Issue #27, opened 2013-11-23
Is it possible to replace *testing.T with an interface?


AlexLuya:
I think the mocking style of testify is more nature than gomock,and hope it can be hooked into ginkgo(a BDD framework:https://github.com/onsi),so posted a function request here:https://github.com/onsi/ginkgo/issues/17,but the developer of ginkgo said:

---

It looks like testify uses (*testing.T) directly instead of an interface (which gomock does). Ginkgo can provide such frameworks with a Fail function that one can wrap in such a way to satisfy most interfaces (this is how we support go mock), but *testing.T refers to a concrete type and passing in the testing.T for the test that Ginkgo runs in would cause the entire suite to fail should one of your mock assertions fail...

---

His suggestion(or needing) is:replaces *testing.T with an interface first,so question:is it possible?.

PR #32, opened 2014-01-14
receive *testing.T via an interface

onsi:
this is related to:
https://github.com/stretchr/testify/issues/27

Receiving testing.T via an interface would allow [Ginkgo](https://github.com/onsi/ginkgo) to provide an implementation of the interface that works with Ginkgo's test runner.

I defined an interface in a new package called `interfaces`... if you don't like that there are other options (e.g. have an interface for this defined in each package)?  I'd be happy to iterate on this.

If this gets pulled in I'll add an adapter to Ginkgo (and some examples) to make using testify's assertion and mock libraries trivial.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -5,9 +5,13 @@ import (
 	"reflect"
 	"runtime"
 	"strings"
-	"testing"
 	"time"
 )
+
+// TestingT is an interface wrapper around *testing.T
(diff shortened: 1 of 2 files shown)


Fernandosilvas35 (changes requested):
> - - [ ] `**__**`

PR #70, opened 2014-07-17
 Add Len() assertion

doloopwhile:
Added `assert.Len(t, obj, l)` which assert `len(obj) == l`.

Of course, `assert.Equal(t, len(obj), l)` do same assertion.

However, I use assertion of `len(obj) == l` often (especially `len(obj) == 1`).
I think `assert.Len(t, obj, l)` very useful to restrict loss of gopher's life
and is not so ad hoc because `len()` is built-in.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -324,6 +324,36 @@ func NotEmpty(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 
 }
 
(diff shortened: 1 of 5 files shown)


pjvds:
Why does Len has the actual as second argument followed by expectation, while the other assertion methods have expectation followed by actual?

PR #92, opened 2014-10-19
Using gospew to print out complex data structures when Equals fails

oxtoacart:
When comparing structs, maps, arrays and slices, this makes the failure message easier to read and actually shows details of the nested data structures.  Here's an example output:

```
Location:       yamlconf_test.go:119
	Error:		Not equal:
			---- Expected ----
			(*yamlconf.TestCfg)(0xc2080387c0)({
			 Version: (int) 1,
			 N: (*yamlconf.Nested)(0xc20800e0c0)({
			  S: (string) (len=1) "3",
			  I: (int) 3
			 })
			})
			
			--------------
			---- Actual ------
			(*yamlconf.TestCfg)(0xc208038e90)({
			 Version: (int) 1,
			 N: (*yamlconf.Nested)(0xc20800e780)({
			  S: (string) "",
			  I: (int) 55
			 })
			})
			
			--------------
	Messages:	Config from updated file should contain correct data
```

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -9,6 +9,8 @@ import (
 	"runtime"
 	"strings"
 	"time"
+
+	"github.com/davecgh/go-spew/spew"
 )
 
 // TestingT is an interface wrapper around *testing.T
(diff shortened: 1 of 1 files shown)


ernesto-jimenez:
Hi @oxtoacart, I'm going through all the open PRs from before I joined the project and got to this one.

I'm going to close since it no longer merges, but we've got an issue filed to implement this: #229 

Beyond pretty printing it would be good to have a diff, since pretty printing will still be messy for big structs.

Thanks!
Ernesto

PR #93, opened 2014-10-20
Fixed regexp error messages to include input values to aide debugging

dlclark:
Looks like a fmt.Sprintf was left off in the regexp tests.


advanderveer:
Would it be possible to have this merged?

tylerb:
LGTM. Thanks for the PR!

PR #96, opened 2014-11-03
Wrap NoError message with quote marks

ac7:
Previous behavior was to print the error with no delimiters:

	No error is expected but got unexpected character in stream

New behavior is to print the error surrounded by quotes:

	Recieved unexpected error "unexpected character in stream"


ernesto-jimenez:
@ac7 I'm going through all the open pull requests and finally got to yours. Thanks!

PR #105, opened 2014-11-26
Add suite test name to test method names

tchajed:
This makes it easier to keep track of which method a test is coming from.

We're using suites to implement parametrized tests (for example, we have in-memory and sqlite implementations of our storage system), with code that looks like this:

```
func TestMemstore(t *testing.T) {
    s := &ProviderSuite{provider.NewSuite().Use("memstore")}
    suite.Run(t, s)
}

func TestSqlite(t *testing.T) {
    s := &ProviderSuite{provider.NewSuite().Use("sqlite")}
    suite.Run(t, s)
}
```

If other people have test names that aren't meaningful, then they could name the suite-calling test method simply "Test" and I could change the patch to handle that case properly.

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"reflect"
 	"regexp"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -31,6 +32,12 @@ func (suite *Suite) SetT(t *testing.T) {
 	suite.Assertions = assert.New(t)
 }
 
+func testName(suiteMethod string, testMethod string) string {
+	methodRegexp := regexp.MustCompile(`([a-zA-Z/_]+\.)?Test([a-zA-Z_]*)`)
+	parts := methodRegexp.FindStringSubmatch(suiteMethod)
+	return fmt.Sprintf("%s.%s", parts[2], testMethod)
+}
+
 // Run takes a testing suite and runs all of the tests attached
 // to it.
 func Run(t *testing.T, suite TestingSuite) {
@@ -43,6 +50,11 @@ func Run(t *testing.T, suite TestingSuite) {
 	methodFinder := reflect.TypeOf(suite)
 	tests := []testing.InternalTest{}
 	for index := 0; index < methodFinder.NumMethod(); index++ {
+		suitePc, _, _, ok := runtime.Caller(1)
+		var suiteMethod string
+		if ok {
+			suiteMethod = runtime.FuncForPC(suitePc).Name()
+		}
 		method := methodFinder.Method(index)
 		ok, err := methodFilter(method.Name)
 		if err != nil {
@@ -51,7 +63,7 @@ func Run(t *testing.T, suite TestingSuite) {
 		}
 		if ok {
 			test := testing.InternalTest{
-				Name: method.Name,
+				Name: testName(suiteMethod, method.Name),
 				F: func(t *testing.T) {
 					parentT := suite.T()
 					suite.SetT(t)


nelsam:
I really like the idea, but I think the implementation is a little too rigid for generic use.  Could you check #115 to see if it will work for you?  You should be able to use `go test -nameformat '{function}.{method}'` to get something similar to what you have as your output here.

The "^Test" portion of the original function name won't be stripped out, though - I can think of ways to do it, but I'm trying to keep the test name format syntax very, very simple.

tchajed:
The implementation in #115 looks good, too. We would probably just use suite.NamedRun, since this is mostly for use with [GoConvey](https://github.com/smartystreets/goconvey), where we would otherwise have to manually configure it to pass `-nameformat`. Even better if `{function}.{method}` is the default, which I think is sane.

You could provide `{testname}` to refer to the test running suite.Run with "Test" stripped. Note that this is subtly different from `{function}` - if suite.Run isn't called directly, `{function}` need not be the TestFoo method.

nelsam:
Yeah, that PR (in its current state) is more of a proof-of-concept than something to actually be merged as is.  Since that will work for you, I'll put some more effort into it over the next few days.  I have [similar code to your example](https://github.com/nelsam/gorq/blob/master/plans/query_plans_test.go#L142-L180) that will benefit from more verbose test names, too - that's the project I'll use as the basis for testing the test names.

My reason for defaulting to `{method}` was for backward compatibility (in case anyone was relying on it); but I'm doubtful anyone is actually relying on that right now.  Assuming I implement `{testname}`, would `{testname}: {suite}.{method}` be acceptable for your use case?  I know I've run in to situations where I want the name of the suite's type in there, too, and `{testname}.{method}` seems slightly misleading in the general case (in the output, I kept on assuming that `{testname}` was the name of my suite type, not the name of the calling function).

If I do implement `{testname}` (I think it's a good idea), I'll ensure that the difference between that and `{function}` is handled as accurately as seems feasible.  Searching through the stack for a function name matching `^Test.*` seems like a good start; checking its signature against `func(*testing.T)` would be even better if I can figure that out.  To go along with that, I'll update it so that it only generates those names if they're requested in the format string, so that we aren't looping through the call stack without reason.

ernesto-jimenez:
Hey @nelsam @tchajed, I'm triaging all the pull requests that were still open.

@tchajed did you have any more comments?

Personally, I'm more inclined for the approach in #115 :)

devnev:
I prefer this PR to #115. Tests shouldn't need configuration to work properly. Not including a suite name of some kind means `go test` output does not uniquely identify tests. I consider that a bug, and it is currently breaking test analysis tools. Using the calling function sticks to the overall functions-as-tests style. Even better would be to use the outer test's name, but that is not exposed.

stevenroose:
This PR is how it should be implemented. You should be able to run `go test -run TestWholeSuite/TestInsideSuite`, just like when using native Go subtests.

stevenroose:
@ernesto-jimenez Are you still looking at PRs?

stevenroose:
#389 is a better solution for that problem than this PR.

ernesto-jimenez:
@stevenroose, I am. Sorry, but I've been travelling a lot the past few months and started a new job. I have to find a big uninterrupted chunk of time to review things closely :)

Testify has thousands of packages depending on it. Including some high profile ones, so I need to review things much more carefully.

I'm hoping to get a day of quality time to review everything either this weekend or the next.

stevenroose:
I'm still using a patch for this problem that I have to manually apply on every glide install. 

Any idea when this can land?

boyan-soubachov:
In my opinion, this is a minor, cosmetic PR that would cause a large change for all of our users. I am closing this due to age but I am happy to review if the community deems this change important.

PR #113, opened 2015-01-06
Add require.Assertions like assert.Assertions

andreas:
The motivation for this PR is to get stop-on-assertion-failure semantics for assert methods in a test suite. Example:

```go
type MySuite struct {
    suite.Suite
    *require.Assertions // override the default assertions
}

func (s *MySuite) TestAddition() {
    s.Equal(3, 1+1) // should fail and exit
    s.Equal(5, 2+2) // this failure should not be part of the output
}
```

This PR contains the following changes:

- A new struct `require.Assertions` which is identical to `assert.Assertions` + tests.
- Update the `require` package to match the `assert` package (`Len`, `Regexp`, `NotRegexp` was missing, `Contains` and `NotContains` had different signatures).
- Correct documentation for `require`, which currently states that `bool`s are returned when in fact they aren't.

### DIFF
--- a/require/forward_requirements.go
+++ b/require/forward_requirements.go
@@ -1,0 +1,211 @@
+package require
+
+import (
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type Assertions struct {
+	t TestingT
+}
+
+func New(t TestingT) *Assertions {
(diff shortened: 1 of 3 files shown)


nelsam:
Looks good to me.

PR #115, opened 2015-01-10
Allow custom test name formats

nelsam:
When using suites, test names were just the raw suite method name, but that's often not
detailed enough to know what the problem is.  This commit adds the ability to pass a custom
name format using certain formatting marks (sort of similar to templating strings) to
include the name of the function that called Run or NamedRun, the name of the suite's type,
and/or the name of the method.

Custom test names can be used one of two ways:
1. When running tests, use `go test -nameformat someFormat` to use "someFormat" as the global format for test names.
2. In your test function, call `suite.NamedRun("someFormat", t, suite)` to use "someFormat" as the format for test names for suite's testing methods.  This is overridden by the `-nameformat` flag.

Valid format strings (so far) are:
- `"{function}"`: The name of the function that called `suite.Run` or `suite.NamedRun`.
- `"{suite}"`: The name of the suite's type.  If the suite is a pointer, the indirect's type name will be used instead.
- `"{method}"`: The name of the test method.

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -6,12 +6,38 @@ import (
 	"os"
 	"reflect"
 	"regexp"
+	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
-var matchMethod = flag.String("m", "", "regular expression to select tests of the suite to run")
+const (
+	testPattern   = "^Test"
+	defaultFormat = "{method}"
+)
+
+var (
+	matchMethod = flag.String("m", "", "Run only those test methods matching the regular expression.")
+	format      = flag.String("nameformat", "",
+		"The format to use for test name output.  Use {function} for the test function name, "+
+			"{suite} for the suite type name, and {method} for the test method name.  Overrides "+
+			"formatting set by NamedRun.")
+)
+
+type testIdentifier struct {
+	function string
+	suite    string
+	method   string
+}
+
+func (t testIdentifier) Name(format string) string {
+	name := strings.Replace(format, "{function}", t.function, -1)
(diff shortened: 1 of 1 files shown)


ernesto-jimenez:
@nelsam I'm going through open PRs and go to the ones before I was contributing to the project :)

I'm going to add a couple of comments I've got in the source code.

suite/suite.go
```diff
@@ -32,8 +58,44 @@ func (suite *Suite) SetT(t *testing.T) {
...
+// nameFormat.  The following patterns will be replaced by details
+// about the current test:
+//
+// * "{function}": The name of the function that called NamedRun.
+// * "{suite}": The name of the testing suite's underlying type.
+// * "{method}": The name of the test method.
+//
+// Be aware that the test flag "nameformat" will override nameFormat
+// globally.
+func NamedRun(nameFormat string, t *testing.T, suite TestingSuite) {
```
ernesto-jimenez: Assertions' signatures start with `*testing.T`, so I would move `t` to the first argument for consistency.

Talking about consistency, I would consider moving the name to be the final argument since it's more consistent with assertions allowing customising the output message.

As a final point, without reading the docs I would expect NamedRun to prefix the method names rather than providing a new format. I would suggest `NamedRun` to call `run(name + ".{method}", t, suite)` and have a different method to pass on a format.

xordspar0:
If this is going to get merged, someone needs to step up and put a little work into it. Besides that, we should decide if this is even needed. Does anyone feel that this would be valuable?

boyan-soubachov:
Due to the age of this PR, I'm closing this. If we feel we want to re-look into this functionality, I'd be happy to review.

Issue #112, opened 2015-01-06
uint32 and int don't compare correctly on go1.4


colinmarc:
Since upgrading to go1.4, I'm seeing assertions like the following fail:

```go
assert.Equal(t, 10, uint32(10))
```

The assertion failure prints the following:

```
--- FAIL: TestFoo (0.00s)
        Location:       foo_test.go:21
	Error:		Not equal: 10 (expected)
			        != 0xa (actual)
```
Thanks!

matryer:
Use `EqualValues`

PR #124, opened 2015-01-24
When testing the equality after conversion, test the underlying values

colinmarc:
In 1.4, comparisons between `int`s and `uint`s seem not to work. This may be
because in 1.4, interface values are always stored as pointers:

> The implementation of interface values has been modified. In earlier
> releases, the interface contained a word that was either a pointer or a
> one-word scalar value, depending on the type of the concrete object stored.
> This implementation was problematical for the garbage collector, so as of
> 1.4 interface values always hold a pointer.

(from https://golang.org/doc/go1.4)

So, when the assertion code obtains a reflect.Value struct, and compares it
with the expected reflect.Value, it may be comparing the pointers held by
that object, and not the scalar value.

Either way, comparing the underlying value of the reflect.Value structs
appears to fix the tests, and by extension #112.

Thanks for the awesome library!

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -43,8 +43,15 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
 	}
 
 	// Attempt comparison after type conversion
-	if actualValue.Type().ConvertibleTo(expectedValue.Type()) && expectedValue == actualValue.Convert(expectedValue.Type()) {
-		return true
+	if actualValue.Type().ConvertibleTo(expectedValue.Type()) {
+		convertedValue := actualValue.Convert(expectedValue.Type())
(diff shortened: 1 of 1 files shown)


colinmarc:
Well, that's embarrassing - I just noticed that the version I was PRing is out of date, and this has already been fixed. :disappointed:

Sorry about that!

matryer:
Thanks Colin.

leolara:
@colinmarc @matryer I am experiencing this issue and I am on the latest `master`

matryer:
Could you submit a PR with a failing test?

leolara:
@matryer here #145

leolara:
@matryer could you run the test?

PR #128, opened 2015-02-07
Exclude `ConvertibleTo` from equality test

icecrime:
`ObjectsAreEqual` using `ConvertibleTo` causes the `ObjectsAreEqual` function to be asymmetrical and producing incorrect assertions.

To give a few invalid examples that this patch fixes:

* `ObjectsAreEqual(2.2, 2) == true` but `ObjectsAreEqual(2, 2.2) == false`
* `ObjectsAreEqual('x', "x") == true` but `ObjectsAreEqual("x", 'x') == false`

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -36,15 +36,6 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
 		return true
 	}
 
-	actualType := reflect.TypeOf(actual)
-	expectedValue := reflect.ValueOf(expected)
-	if expectedValue.Type().ConvertibleTo(actualType) {
-		// Attempt comparison after type conversion
-		if reflect.DeepEqual(actual, expectedValue.Convert(actualType).Interface()) {
-			return true
-		}
-	}
-
 	// Last ditch effort
 	if fmt.Sprintf("%#v", expected) == fmt.Sprintf("%#v", actual) {
 		return true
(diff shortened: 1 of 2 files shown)


vieux:
I don't know about `ObjectsAreEqual` but the `Equal` and `NotEqual` functions (the one you should use) are perfectly fine to me:


```
        assert.Equal(t, 2, 2)
        assert.Equal(t, 2.2, 2.2)
        assert.NotEqual(t, 2.2, 2)
        assert.NotEqual(t, 2, 2.2)

        assert.Equal(t, 'x', 'x')
        assert.Equal(t, "x", "x")
        assert.NotEqual(t, 'x', "x")
        assert.NotEqual(t, "x", 'x')
```

This works as expected

icecrime:
This was introduced in #54 as far as I can tell.

vieux:
@icecrime my bad, I wasn't using master, but a vendored version.

Master it completely broken. Our vendored version is d6577e08ec30538639ac0ea38b562b6f250e9055  and it works as expected

vieux:
@tylerb can you take a look, this is pretty bad

vieux:
@icecrime even better: https://github.com/stretchr/testify/commit/49c5c6cdb11776deb678dd3bb48e1e79e7e33943

tiborvass:
Ping @matryer @tylerb

PR #130, opened 2015-02-11
Add assert.EqualValues, which attempts to convert types to test equality

### LINKED ISSUE #129: Add "ConvertableTo" Equals
Related to #128 -- [ffjson](https://github.com/pquerna/ffjson) https://github.com/pquerna/ffjson/issues/85 were *incorrectly* relying upon the automatic type conversion in `assert.Equal`.

Is it reasonable to add a `ConvertableTo` or `SemiKindaEqual` to `assert` and `require`?

pquerna:
Fixes #129

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -45,6 +45,23 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
 
 }
 
+func ObjectsAreEqualValues(expected, actual interface{}) bool {
+	if ObjectsAreEqual(expected, actual) {
+		return true
+	}
+
+	actualType := reflect.TypeOf(actual)
+	expectedValue := reflect.ValueOf(expected)
+	if expectedValue.Type().ConvertibleTo(actualType) {
+		// Attempt comparison after type conversion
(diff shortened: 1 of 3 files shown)


assert/assertions.go
```diff
@@ -189,6 +206,23 @@ func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{})
 
 }
 
+// Equal asserts that two objects are equal or convertable to the same types
```
tylerb: The name of the function should be `EqualValues`, not `Equal` in the comments.

require/requirements.go
```diff
@@ -42,6 +42,15 @@ func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{})
 	}
 }
 
+// Equal asserts that two objects are equal or convertable to each other.
```
tylerb: The name of the function should be `EqualValues`, not `Equal` in the comments.

tylerb:
Thanks for the PR! Looking good. Please fix those two minor issues and I'll get it merged.

PR #132, opened 2015-02-17
Fix reversed vars in EqualError plus some golint stuff

andybons:
This rolls in the change made in #126 plus some golint fixes.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -405,11 +405,11 @@ func getLen(x interface{}) (ok bool, length int) {
 func Len(t TestingT, object interface{}, length int, msgAndArgs ...interface{}) bool {
 	ok, l := getLen(object)
 	if !ok {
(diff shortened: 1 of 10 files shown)


ernesto-jimenez:
Hey @andybons,

I'm reviewing all the PRs that were still opened and got to this one before I was contributing to the package.

I've added an issue to pass go lint, which I'll take a look at once we've gone through the remaining pull requests so we avoid having to rebase all current PRs: https://github.com/stretchr/testify/issues/230

Also, regarding the underlying issue with `EqualError`, the current message is correct, the issue is with the signature for `EqualError` being inconsistent with most of the library.

There's another issue about making the signatures consistent so it's always `(t, expected, actual)`. Unfortunatelly, it's a breaking change and we need to figure out how to deal with it.

I'm closing the PR. If you have any comments just let me know.

Thanks!
Ernesto

PR #138, opened 2015-02-18
Fixed isEmpty for pointer types, and simplified it.

willfaught:
Fixed the isEmpty bug with nil pointers.

Changed to adopt behavior consistent with reflection. Note that this is a breaking change, e.g. non-nil slices with length 0 were empty before, and now aren't.


matryer:
Hmm, I think an empty slice should be considered empty, don’t you?

willfaught:
Sorry, I mistakenly didn't "go get" the tests, so this hasn't been tested, please ignore.

willfaught:
@matryer I guess it depends on the meaning of 'empty'. I can now see what they were going for, so maybe this wasn't a good idea. I was thinking of it working more like a Zero assertion. I'd love to have that.

PR #145, opened 2015-03-17
Test equality between integer literal and uint32

leolara:
This fails on `go version go1.4.2 darwin/amd64`


leolara:
Hi @DAddYE , Is this working? what are we going to do about this?

PR #147, opened 2015-03-25
Fix race condition with ExpectedCalls

leepa:
There is race condition caused when a method being tested calls a mocked
method within a go routine. For example, caching might be done a go
routine and that caching might be mocked.

There is already a mutex protecting the lists on the Mock object -
however Return() appends to ExpectedCalls and findExpectedCall could
run at the same time.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -112,6 +112,9 @@ func (m *Mock) On(methodName string, arguments ...interface{}) *Mock {
 //
 //     Mock.On("MyMethod", arg1, arg2).Return(returnArg1, returnArg2)
 func (m *Mock) Return(returnArguments ...interface{}) *Mock {
+	m.mutex.Lock()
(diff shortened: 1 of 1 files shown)


DAddYE:
Can you rebase?

sttts:
This problem really complicates tests. It would be great if the PR could be merged.

ernesto-jimenez:
@leepa, can you rebase? otherwise, we'll just make the change and close the PR. Whatever you prefer  :)

jdef:
:+1: for this change, this is killing our tests

leepa:
rebased as clearly it's a big problem ;)

ernesto-jimenez:
@leepa @jdef @sttts merged :)

sttts:
:+1:

DAddYE:
:+1:

PR #148, opened 2015-04-07
Mock can block returns with .After and .WaitUntil

ernesto-jimenez:
You sometimes want to test concurrent behaviour like timeouts, but it was not currently possible with our mocks.

This PR adds the following functions to Mock:

`After` will block m.Called for the given duration before returning.

```go
func (m *Mock) After(d time.Duration) *Mock

Mock.On("MyMethod", arg1).After(1 * time.Millisecond)
```

`WaitUntil` is the primitive under `After`, and it's exposed in case you want somebody wants more control over the returns. `Called` blocks until the `w` channel is closed or receives a message.

```go
func (m *Mock) WaitUntil(w <-chan time.Time) *Mock

Mock.On("MyMethod", arg1).WaitUntil(time.After(1 * time.Millisecond))
```

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -8,6 +8,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"time"
 )
 
 // TestingT is an interface wrapper around *testing.T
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
@matryer how does this PR look?

Would you like me to change anything?

ernesto-jimenez:
@DAddYE thanks! :)

PR #151, opened 2015-04-11
Suite: add support for test name prefixes

allgeek:
When running parameterized test suites, output from go test (and other tooling) isn't as intuitive as it could be, with the name for each test in the suite duplicated for each variation of the suite being run.  I propose adding support for optional test name prefixes that can be used to help distinguish results in those scenarios.

This commit actually has two approaches implemented; a simple RunWithPrefix method, and a TestNamePrefixer interface that suites can implement to provide a prefix.  I'd like feedback on which route is preferred before finalizing the tests.

The RunWithPrefix should be sufficient - the caller of Run() likely created the suite and should have full context on all the options given to a parameterized suite in order to provide meaningful prefixes.

The TestNamePrefixer interface is 'safer' with regard to backward compatibility; if someone in a group (that does not vendor dependencies) uses a new version of testify and adds prefix support, other collaborators will not be forced to upgrade.  It also fits in with other suite functionality, though is a bit less concise in usage.  (FYI the interface name doesn't match method name since the implementation shouldn't be detected as a test, but I wanted to avoid stutter with the package name).

Any other feedback is welcome!

### DIFF
--- a/suite/doc.go
+++ b/suite/doc.go
@@ -18,6 +18,10 @@
 // with "Test" will not be run by testify, and can safely be used as
 // helper methods.
 //
+// For parameterized suites (which may run multiple times for different
+// scenarios), the TestNamePrefixer interface can be implemented to
+// provide meaningful prefixes to test names for each variation.
+//
 // Once you've built your testing suite, you need to run the suite
 // (using suite.Run from testify) inside any function that matches the
 // identity that "go test" is already looking for (i.e.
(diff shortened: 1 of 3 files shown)


allgeek:
Oops, I somehow missed prior PRs #115 and #105 which appear related - however those have been around for quite some time as well.  If this PR, or any others that achieve similar goals, could be considered that would be much appreciated!

ernesto-jimenez:
hey @allgeek, sorry you did not get any replies to this. Since I'm missed this back in they day because I had just started contributing to the project.

I'm going to go back to the other two PRs since they work on similar bases. Thanks!

PR #156, opened 2015-04-24
allow customization of not equal printout

agileknight:
this is useful to print better errors for large mismatching structures
better formatting and diff tools (like go-diff and go-spew) can be used

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -19,6 +19,22 @@ type TestingT interface {
 // Comparison a custom function that returns true on success and false on failure
 type Comparison func() (success bool)
 
+type ErrorFormatter interface {
+	FormatNotEqual(expected, actual interface{}) string
+}
+
+var errorFormatter ErrorFormatter = new(DefaultErrorFormatter)
+
+func SetErrorFormatter(formatter ErrorFormatter) {
+	errorFormatter = formatter
+}
+
+type DefaultErrorFormatter struct {}
+
+func (m *DefaultErrorFormatter) FormatNotEqual(expected, actual interface{}) string {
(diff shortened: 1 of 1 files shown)


ernesto-jimenez:
@agileknight sorry this went unanswered, I'm going trough all the open pull requests and noticed this was still opened.

I agree that comparing large mismatching structures/maps/etc is painful.

Rather than having a pluggable formatter I think it would be better to fix testify's output to show the diff. Opened an issue to address the underlying problem: https://github.com/stretchr/testify/issues/229

Did you have any reason to implement formatter rather than implementing the diff within the library?

ernesto-jimenez:
We've implemented showing a diff when `Equal` fails: https://github.com/stretchr/testify/pull/233

agileknight:
@ernesto-jimenez top! we ended up externally formatting a more readable error text and pass that as an additional argument to assert.Equals but having it out of the box is even better!

ernesto-jimenez:
Awesome, if you have any feedback let me know ;)

agileknight:
@ernesto-jimenez looks good. almost as readable as our (more platform-dependant) implementation highlighting the diff with red/green colors in the console

there was so far just one issue I encountered: map[string]string does not get sorted before generating diff so on some invocations a diff is reported even though the map is filled with equal keys and values. our solution based on go-spew and diffmatchpatch was able to handle this as far as I can tell.

mabye it is of help how we did it:
https://gist.github.com/agileknight/fbf00a42950a8d846bda

ernesto-jimenez:
Thanks @agileknight, I just pushed a change to sort keys too :)

agileknight:
@ernesto-jimenez great, thanks!

PR #158, opened 2015-05-04
Allow mocking methods updating referenced structs

ernesto-jimenez:
Some methods in Go take a pointer to a struct or a map and are supposed
to set some values in the referenced argument. Using Return is not
enough to mock such methods.

We are introducing a Run method that allows setting a Call handler to
make your mock update such referenced values.

Example usage mocking a service that finds a user in a data store and
fills the values in the provided struct:

```go
m.On("Find", 1, mock.AnythingOfType(*User))
  .Return(nil)
  .Run(func(args mock.Arguments) {
  u := args.Get(0).(*User)
  u.ID = 1
  u.Email = "mail@example.com"
})
```

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -42,6 +42,11 @@ type Call struct {
 	// Holds a channel that will be used to block the Return until it either
 	// recieves a message or is closed. nil means it returns immediately.
 	WaitFor <-chan time.Time
+
+	// Holds a handler used to manipulate arguments content that are passed by
+	// reference. It's useful when mocking methods such as unmarshalers or
+	// decoders.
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
@DAddYE, do you mind reviewing this?

Ideally it would be `m.On(...).Run(...).Return(...)`, but it cannot be done right now due to the package internals, so we had to settle for doing `m.On(...).Return(...).Run(...)`

Thanks!

/cc @rnaveiras

DAddYE:
Looks good to me! :+1:

ernesto-jimenez:
Merged, thanks!

Issue #162, opened 2015-05-11
InDelta assert passes if actual is NaN


HawaiianSpork:
If the actual value of the assert is NaN then the assert passes.  I believe this is unexpected.  To fix just add an additional check if the actual isNaN

Example:
```go
assert.InDelta(t, 1.0, math.NaN(), 0.001)
```

ernesto-jimenez:
@HawaiianSpork, makes sense to fix that. Do you want to make a PR?

Otherwise I'll check it out later :)

PR #164, opened 2015-05-12
Address #162: InDelta should fail when actual value is not a number.

HawaiianSpork:
Address #162: InDelta should fail when actual value is not a number.

Comparting a float with NaN is always false so the assertion would always pass.
Added a check that either the actual or expected values are NaN.

InDelta will now fail if either the actual or expected value are NaN.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"math"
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
Looks good, will merge once Travis reports a green build.

Thanks @HawaiianSpork!

ernesto-jimenez:
Done :)

PR #165, opened 2015-05-13
Support map keys for Contains/NotContains assertion.

### LINKED ISSUE #118: assert/require.Contains should work for maps too
`assert/require.Contains` should work for maps too.

For example:

    k := 1
    v := 2
    m := map[int]int{k: v}
    assert.Contains(t, m, k) // true

`assert/require.Contains` should check for map keys and not values because (aside from that obviously making the most sense) Go itself provides a way to test for key membership:

    v, ok := m[k]
    assert.True(ok)

tuvistavie:
Makes `assert.Contains` and `assert.NotContains` work with `map` keys.

```
assert.Contains(t, &map[interface{}]interface{}{"Foo": "Bar"}, "Foo") // passes
assert.Contains(t, &map[interface{}]interface{}{"Bar": "Foo"}, "Foo") // fails
```

This fixes #118 .

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -476,6 +476,16 @@ func includeElement(list interface{}, element interface{}) (ok, found bool) {
 		return true, strings.Contains(listValue.String(), elementValue.String())
 	}
 
+	if reflect.TypeOf(list).Kind() == reflect.Map {
+		mapKeys := listValue.MapKeys()
+		for i := 0; i < len(mapKeys); i++ {
+			if ObjectsAreEqual(mapKeys[i].Interface(), element) {
+				return true, true
+			}
+		}
+		return true, false
+	}
+
 	for i := 0; i < listValue.Len(); i++ {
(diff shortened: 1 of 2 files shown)


willfaught:
+1

DAddYE:
@stretchr/testify-contributors +1 from me

ernesto-jimenez:
LGTM :+1:

ernesto-jimenez:
@tuvistavie sorry! I just realised we +1ed this and did not merge it.

I'm going to merge it now.

arvenil:
How then I can check if map contains value, not key? Shouldn't be there HasKey and Contains methods? First for the key and the second for value?

ernesto-jimenez:
@arvenil making `Contains` look for `map` values would be a breaking change, I'm afraid.

I wouldn't be opposed to a PR including a `ContainsValue` assertion that checks values within a `map`.

You could even modify `Contains` so, when it is checking a `map` where the given value exists but not the key, the error includes something like "X was found in a value, did you mean to use `assert.ContainsValue`?"

Issue #143, opened 2015-03-11
assert.Equal randomly fails testing maps


phemmer:
`assert.Equal` (and other functions) will randomly fail testing equality of maps with the same values but different type.

For example:

```
package main

import (
	"fmt"
	"github.com/stretchr/testify/assert"
	"testing"
)

func TestExactly(t *testing.T) {
	map1 := map[string]interface{}{"foo": 1234, "bar": int16(4321)}
	map2 := map[string]interface{}{"foo": 1234, "bar": int32(4321)}

	for i := 0; i < 3; i++ {
		if assert.Equal(t, map1, map2) {
			fmt.Printf("Everything is ok\n")
		} else {
			fmt.Printf("Everything is NOT OK\n")
		}
	}
}
```

Results in:

```
=== RUN TestExactly
Everything is NOT OK
Everything is ok
Everything is NOT OK
--- FAIL: TestExactly (0.00s)
	Location:	main_test.go:14
	Error:		Not equal: map[string]interface {}{"foo":1234, "bar":4321} (expected)
			        != map[string]interface {}{"foo":1234, "bar":4321} (actual)
		
	Location:	main_test.go:14
	Error:		Not equal: map[string]interface {}{"foo":1234, "bar":4321} (expected)
			        != map[string]interface {}{"bar":4321, "foo":1234} (actual)
		
FAIL
exit status 1
FAIL	_/tmp/test6	0.012s
```

The issue is because the type on `bar` is different between the maps (`int16` in one, `int32` in the other).
`ObjectsAreEqual` uses `reflect.DeepEqual` which properly detects this, but then it continues to check equality of a `fmt.Sprintf("%#v",...)` on the values, and this randomly fails depending on the order it prints the map values in.

The solution seems simple: remove the `fmt.Sprintf()` check. But I wasn't sure why that was in there to begin with.

benma:
I agree, it sounds like a bad idea to compare string representations. This is not robust enough.

Why not simply `return reflect.DeepEqual(expected, actual)` [here](https://github.com/stretchr/testify/blob/e4ec8152c15fc46bd5056ce65997a07c7d415325/assert/assertions.go#L30)?

OttoAllmendinger:
:+1: remove `fmt.Sprintf()` check. The comparison makes it difficult to reason about the `(Not)Equal` methods.

[...]

PR #168, opened 2015-05-19
Added assert.Equals check that was removed

TechnotronicOz:
The recent change to assert.Equals broke all of our tests that tested interface vales. We would really like to see this added back so we don't have to maintain a fork of testify.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -37,6 +37,12 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
 		return true
 	}
 
+	// This was removed in https://github.com/stretchr/testify/commit/e22aedd37671fb115be6c0c25129c405cb575cfd which
+	// breaks nearly all of our tests that test interface values
+	if fmt.Sprintf("%#v", expected) == fmt.Sprintf("%#v", actual) {
+		return true
+	}
+
 	return false
 
 }


DAddYE:
We need versioning before this. https://github.com/stretchr/testify/issues/169

TechnotronicOz:
@DAddYE Now that we are using gopkg.in for versioning, can we create a branch or tag for this to get it in the base and usable?

DAddYE:
@TechnotronicOz yes I'm gonna make it soon 

//cc @stretchr/testify-contributors

ernesto-jimenez:
I'm triaging all open pull requests and saw this is still open. What's the status of this?

@DAddYE, @TechnotronicOz is this still an issue?

TechnotronicOz:
I would still like to see a version of Testify have the old sprintf check so you can still use it if your tests relied on it. We finally ended up just forking Testify and updating it to what we need.

ernesto-jimenez:
@TechnotronicOz how did your tests break?

The previous behaviour was removed because tests would fail randomly https://github.com/stretchr/testify/issues/143

TechnotronicOz:
@ernesto-jimenez We use uint32 for record ids and often do something like

```
assert.True(t, myStruct.Id, 5)
```

which produces 0x5 != 5. Doing the casting in the assertion statement sucks.

ernesto-jimenez:
@TechnotronicOz rather than reverting to comparing using string serialisation I think it would be more appropriate to have `Equal` pass when `ObjectsAreEqual` or `ObjectsAreEqualValues`.

However, we'll have to do some research about the implications of such change. We've got to make sure it would never pass a test that should fail.

I've created a different issue (https://github.com/stretchr/testify/issues/235) and will close this PR.

TechnotronicOz:
@ernesto-jimenez I ended up doing just that, I made our `Equal` a bit more generic so its a pass through to EqualValues and if that fails then runs what Equal did.

ernesto-jimenez:
cool, way better than comparing strings :)

ernesto-jimenez:
btw, @TechnotronicOz, now that we are talking about `Equal`.

Would you find this useful: https://github.com/stretchr/testify/pull/233 ? :)

TechnotronicOz:
@ernesto-jimenez better than that or having to cast in the Equal expression. Having Equal ObjectsAreEqual and ObjectsAreEqualValues makes life way much easier. 

RE: https://github.com/stretchr/testify/pull/233 - that is a slick feature, I dig it.

GeorgeErickson:
Whats the status of this? I still get:

```
	Error:      	Not equal:
	            	expected: int64(1496959500)
	            	actual: int32(1496959500)
```

DAddYE:
Use the right int or use EqualValues

Issue #167, opened 2015-05-19
Unable to test concurrent objects


ANPez:
Because of the changed described in #55, my tests fail randomly when making parallel calls to the same object. My object supports that, but the lock is serializing the calls.

ernesto-jimenez:
@ANPez are you up for doing a PR on this?

I ran into the same issues mocking a gorilla websocket which had two goroutines one calling WriteJSON and the other oner calling ReadJSON. It was quicker for me to do a fast workaround, but it would be gret to get this fixed :)

ANPez:
I was in the exact same situation as you, just fixed it :)

georgelesica-wf:
Hello, my name is George. I've recently taken over maintenance of Testify and
I'm excited to help the project move forward.

In order to improve responsiveness and promote contributions to the Testify
project, I have decided to close pull requests and issues that haven't had any
activity in several months and do not appear to address critical bugs.

If you would still like to see your pull request or issue addressed, I am happy
to help! All you have to do is re-open and update it to reflect the current
state of the master branch. Once that is done, I will triage and recruit
reviewers as-needed.

If you have questions, please feel free to reach out. Thank you for your
interest in improving the Testify project!

PR #171, opened 2015-05-19
Release lock before .WaitUntil

ANPez:
Release the lock before .WaitUntil prevents deadlocks when testing parallel objects.

Address #167: Unable to test concurrent objects


ernesto-jimenez:
@ANPez could you add a test so we avoid regressions in the future?

DAddYE:
> @ANPez could you add a test so we avoid regressions in the future?

Agree

ernesto-jimenez:
@ANPez small nudge, the tests from this PR are not passing :)

ANPez:
Fixed it! Looks like since I wrote it, some code around it changed :)

mock/mock_test.go
```diff
@@ -987,3 +987,29 @@ func Test_Arguments_Bool(t *testing.T) {
...
+	// Lock both goroutines on the .WaitUntil method
+	go func() {
+		mockedService.TheExampleMethod2(false)
+	}()
+	go func() {
+		mockedService.TheExampleMethod2(true)
+	}()
+
+	// Allow the first call to execute, so the second one executes afterwards
+	ch2 <- time.Now()
```
ernesto-jimenez: Maybe it would be good to have each `go func()` write something and assert the order is correct at the end of the test?

PR #176, opened 2015-06-04
Add a basic stacktrace to the assertion output

### LINKED ISSUE #116: Stack trace printed for errors, not just file and line
I commonly write test wrapper functions that assert preconditions and postconditions when calling non-test functions. These wrappers are called in test code in multiple places, and the exact scenario in which one of these wrapper assertions fails is usually not clear and non-trivial to investigate. It would be *incredibly* useful for Testify to print a stack trace for every reported error, or at least to have the option for it to do so.

Sorry if this is a duplicate; I thought I'd seen a similar issue in the past, but couldn't find it when I searched just now.

paulbellamy:
Will fix #116.

Stacktrace stops when it reaches a function matching `TestXxx`, `BenchmarkXxx`, or `ExampleXxx`.

Example output is:
```
--- FAIL: TestFailures-4 (0.00s)
        Location:       failure_test.go:18
                        failure_test.go:9
                        failure_test.go:13
        Error:          Not equal: 1 (expected)
                                != 2 (actual)

FAIL
```

For the following test file:
```Go
package testify

import (
	"testing"
)

func testHelper1(t *testing.T) {
  testHelper2(t)
}

func testHelper2(t *testing.T) {
  assert.Equal(t, 1, 2)
}


func TestFailures(t *testing.T) {
  testHelper1(t)
}
```

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -10,6 +10,8 @@ import (
 	"runtime"
 	"strings"
 	"time"
+	"unicode"
+	"unicode/utf8"
 )
 
 // TestingT is an interface wrapper around *testing.T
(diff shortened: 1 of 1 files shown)


DAddYE:
:+1: 

@stretchr/testify-contributors ?

assert/assertions.go
```diff
@@ -144,19 +180,20 @@ func Fail(t TestingT, failureMessage string, msgAndArgs ...interface{}) bool {
 
 	message := messageFromMsgAndArgs(msgAndArgs...)
 
+	locationInfo := strings.Join(CallerInfo(), "\n\r\t\t\t")
 	if len(message) > 0 {
 		t.Errorf("\r%s\r\tLocation:\t%s\n"+
```
ernesto-jimenez: It's no longer a Location, but an error trade.

Maybe we should change the name?

ernesto-jimenez:
In my opinion, it might be good to reverse the order of the stack.

That way, the assertion call would be at the top, since it's the one causing the error. That would be consistent with the order of Go's stack traces.

@paulbellamy what do you think?

paulbellamy:
Agreed on both counts.

I was debating which order the traces were more useful in. Matching the order of `panic` is a good reason.

ernesto-jimenez:
@paulbellamy looking great.

Just one last question. Should we change the output too so it says something other than 'Location'?

I'm wondering whether everybody will be clear that's an stack trace. What do you think?

paulbellamy:
Am happy to, but I couldn't think of a better name for it.

ernesto-jimenez:
How about "Error Trace"?

paulbellamy:
Cool. Output from that example now looks like:
```
--- FAIL: TestFailures-4 (0.00s)
        Error Trace:    failure_test.go:13
                        failure_test.go:9
                        failure_test.go:18
        Error:          Not equal: 1 (expected)
                                != 2 (actual)

FAIL
```

ernesto-jimenez:
:+1: merging :)

ernesto-jimenez:
thanks!

willfaught:
Looks great! Thanks.

PR #181, opened 2015-06-11
Handle `<autogenerated>` files in `CallerInfo()`

### LINKED ISSUE #180: assert.CallerInfo() panics if file is autogenerated
I honestly have no idea __why__ my file is suddenly `<autogenerated>`; I just invoked a subdirectory:

    go test -v ./subdir

This causes `CallerInfo()` to panic when [extracting the file/path info](https://github.com/stretchr/testify/blob/7e4a149930b09fe4c2b134c50ce637457ba6e966/assert/assertions.go#L88).

I have been attempting to add a test case for this into testify itself and I can not get it to reproduce :/

However I will be submitting a PR shortly that actually fixes the issue, and should be low impact w/o a test case.

ErebusBat:
Fixes #180

This is a huge edge case, but it will panic if this is the case.


DAddYE:
:+1: @stretchr/testify-contributors ?

ernesto-jimenez:
@ErebusBat any chance to write a test for this?

Otherwise we'll likely have a regression in the future.

ErebusBat:
@ernesto-jimenez Not as the code is currently written. I spent about 45 minutes attempting to do just that; however I am not sure why the file  was showing as autogenerated in the first place (it was only this one test).  I attempted to reproduce a generic test to recreate the circumstances without success.  I fear that if I were to magically reproduce the circumstances it would be a false test as there is no guarantee it would be stable.  

However if we refactored the code for getting the current file name out of `CallerInfo()` then a test could easily be written.  Also we would the. Be testing the actual problem (file name being `<autogenerated>`), rather than getting go to reproduce that behavior.  

I would be willing to take on this refactor if you wish, would like like a separate PR just for that?

ernesto-jimenez:
@ErebusBat I did a quick search.

"<autogenerated>" was added to go in this CL: https://codereview.appspot.com/92250044

It was added to address the following issue: https://github.com/golang/go/issues/4388

Here's some code from Go that tests "<autogenerated>" is in the stack: http://bcl.golang.org/test/fixedbugs/issue4388.go

I believe you should be able to write a test based on that :)

Otherwise I'll have a go at it once I have some time.

Thanks!
Ernesto

ernesto-jimenez:
@ErebusBat if you are interested, here's the test: abc938a12b76ff35e817db5d65b8d509c2efc8d8 :)

PR #186, opened 2015-06-20
Adjust NotNil error message

connor4312:
Minor fix. The error message on the NotNil assertion was a little odd to read previously.


DAddYE:
Thanks!

PR #196, opened 2015-07-22
remove overview sections from documentation

### LINKED ISSUE #195: Remove "overview" sections from documentation
The documentation contains a lot of redundant information, specifically the "overview" sections. For example, see https://godoc.org/github.com/stretchr/testify/assert

Notice the 2 huge blocks at the top of the doc giving an overview of the available functions.
1. This information is repeated twice. There's a second identical block right below the first.
2. This information is unnecessary as godoc already generates the overview which can be found in the "index" section.
3. The information is out of date. It tends to not get updated when functions are modified.

Therefore I propose these blocks be removed.

phemmer:
Closes #195


DAddYE:
Thanks!

PR #199, opened 2015-08-06
Use "method" rather than "mode" for http tests

glittershark:
Use `method` (this is what http.NewRequest and the HTTP standard in
general calls this) rather than `mode` for `"GET"`, `"POST"`, etc.


ernesto-jimenez:
@DAddYE looks good to me.

DAddYE:
@glittershark thanks for this!

Issue #185, opened 2015-06-18
EqualError error output swaps the actual and expected values


willfaught:
In my test:

    assert.EqualError(t, err, "missing or invalid currency")

In the output:

    Error Trace:    my_test.go:58
    Error:  	    Not equal: "missing or invalid amount" (expected)
                            != "missing or invalid currency" (actual)
    Messages:	    An error with value "missing or invalid currency" is expected but got "missing or invalid amount".

Note the

    "missing or invalid amount" (expected)

and the

    "missing or invalid currency" is expected

Same for the actual values.

sivagollapalli:
Could we close this issue as PR got merged?

matryer:
Sorry - you were asking the opener :)

willfaught:
Thanks for fixing.

PR #205, opened 2015-08-25
FIX# EqualError error output swaps the actual and expected values 

sivagollapalli:
Issue #185


DAddYE:
Thanks! Good catch!

danielchatfield:
This patch was wrong.

Issue #140, opened 2015-02-23
NotEqual does not print expected and actual values


willfaught:
Unlike Equal. Printing at least one of them would be very useful, although printing both would be best, as I believe some reflection rejiggering is going on in ObjectsAreEqual.

sivagollapalli:
Could we close this issue as PR got merged?

PR #206, opened 2015-08-26
Display actual and expected values on NotEqual

sivagollapalli:
issue #140


ernesto-jimenez:
Since both values are equal, wouldn't be enough to show the actual value?

matryer:
What about `Should not be: VALUE`?

sivagollapalli:
@matryer Modified error message to `Should not be: VALUE` .

PR #209, opened 2015-08-27
Fix order of params in Equal

danielchatfield:
This reverts the change introduced in https://github.com/stretchr/testify/pull/205 which changed it from being correct to being wrong.


ernesto-jimenez:
thanks @danielchatfield

PR #210, opened 2015-08-28
Fix param order in ObjectsAreEqualValues after type conversion

jcamenisch:
one more param order switch :)


ernesto-jimenez:
@jcamenisch this is just for consistency, isn't it?

I'm OK with that :)

jcamenisch:
Yes, exactly.

I was only looking at this line of code because we're using an older version that needs to be updated. But since I noticed the inconsistency, I thought I'd submit this detail tweak.

Anyway, thanks for the great tool, @ernesto-jimenez.

DAddYE:
:+1:

PR #211, opened 2015-08-28
Remove unnecessary branches.

jaguilar:
reflect.DeepEqual already returns exactly what we want in both of these cases.

Tests all seem to pass except testify/require, which fails with:

```
~/gocode/src/github.com/jaguilar/testify$ go test -v ./...
# github.com/jaguilar/testify/require
require/requirements.go:275: undefined: assert.Zero
require/requirements.go:282: undefined: assert.NotZero
```

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -35,11 +35,7 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
 		return expected == actual
 	}
 
-	if reflect.DeepEqual(expected, actual) {
-		return true
-	}
-
-	return false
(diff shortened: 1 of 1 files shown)


ernesto-jimenez:
Thanks @jaguilar, do you mind rebasing the PR?

jaguilar:
Will do.

jaguilar:
OK, that was hairy. I've never done a rebase before but I *think* I figured it out. PTAL.

ernesto-jimenez:
@jaguilar sorry for the delay. Merging now, thanks!

Issue #135, opened 2015-02-18
EqualValues panics when second arg is nil


willfaught:
`assert.EqualValues(t, b, nil)` (b not nil) panics:

    panic: reflect: nil type passed to Type.ConvertibleTo [recovered]
	    panic: reflect: nil type passed to Type.ConvertibleTo

I'm guessing TypeOf(nil) returns nil?

sivagollapalli:
It panics when we do `assert.EqualValues(t, nil, b)`. I have handled both conditions on my PR.

PR #212, opened 2015-08-31
Rescued from panic when user passes nil

sivagollapalli:
issue #135


ernesto-jimenez:
@sivagollapalli sorry it took so long to reply.

Should it return true when `expected == actual == nil`?

Also, could you please add some tests to avoid regressions in the future?

Thanks!

sivagollapalli:
@ernesto-jimenez Handled when both actual and expected value is nil.

ernesto-jimenez:
Hey @sivagollapalli

I just realised this had already been addressed after my last comment.

Thanks!

PR #208, opened 2015-08-27
Allows mock call expectations to be defined in arbitrary order

tcsc:
Originally the expctations required that the return values be specified immediately after the method name and arguments, otherwise the call setup will either panic (best case) or silently modify the *previous*
call specification (worst case).

This change moves the `Return()`, `Run()`, `Once()`, etc methods onto the `Call` struct, and changes the chaining behaviour so that they modify the `Call` data directly rather than referencing the last item in the `ExpectedCalls` array.

The major externally visible change is that the methods all now return a pointer to the `Call` they're operating on, rather than a pointer to the containing `Mock` instance. This would break any code that chains multiple call setups together, so I have added an `On` method to `Call` that passes the call setup through to the `Mock`. 

I've updated the tests to match, and added a few new ones to cover the new functionality.

PR #213, opened 2015-09-06
Start expecting calls as soon as we .On

connor4312:
Previously, it was required to use .Return() after .On(), even if you don't want to return anything. I had originally assumed (based on experience with testing frameworks in other language) that if I didn't want to return anything I could simply omit the .Return(). This lead to some really whacky behaviour, as I had more mocks after it; I was finally able to track down what was going on by reading the testify source :stuck_out_tongue: 

This PR modifies it so that, as soon as .On() is called the call is added to the list of expected calls. Calling .Return() modifies the last call we got, setting the ReturnArguments. This fixes the case I ran into above, and also simplifies/normalizes some logic within the package. No breaking behaviour is introduced.

On a related note, it might be nice to add an .Expect() alias. It seems more descriptive in some cases (e.g. `mock.On("asdf")` vs `mock.Expect("asdf")` without any return), though it does add ambiguity..

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -54,14 +54,6 @@ type Call struct {
 // For an example of its usage, refer to the "Example Usage" section at the top of this document.
 type Mock struct {
 
-	// The method name that is currently
-	// being referred to by the On method.
-	onMethodName string
-
-	// An array of the arguments that are
-	// currently being referred to by the On method.
-	onMethodArguments Arguments
-
 	// Represents the calls that are expected of
 	// an object.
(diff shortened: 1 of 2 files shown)


tcsc:
Looks like you hit the exact same thing I did (that inspired #208.)

connor4312:
Ah, looks like it. Just skimmed over, but your PR looks like a better/more thorough way.

ernesto-jimenez:
@connor4312 I'm closing this PR since we just merged @tcsc 

Thanks you both! :)

Issue #149, opened 2015-04-08
suite.Require() panics with "panic: test executed panic(nil) or runtime.Goexit"


oryband:
```go
r := suite.Require()
for i, tt := range tests {
	res := httptest.request(...)
	...
	r.Equal(http.StatusOK, res.Code)  // Panics with "panic: test executed panic(nil) or runtime.Goexit"
	...
}
```
Also, I've noticed that if i replace `r.Equal(...)` with `suite.Equal(...)`, it will still panic in the next line I'm using `suite.Require()`.

Also, using `Equal()` or `httptest` is not the issue here. This happens with every `suite.Require().X()` call, whatever it is.

Here is the stack trace:
```go
--- FAIL: TestFunc (0.00s)
panic: test executed panic(nil) or runtime.Goexit

goroutine 10 [running]:
testing.func·006()
        /home/ory/.gvm/gos/go1.4.1/src/testing/testing.go:441 +0x181
runtime.Goexit()
        /home/ory/.gvm/gos/go1.4.1/src/runtime/panic.go:288 +0xf6
testing.(*common).FailNow(0xc208044120)
        /home/ory/.gvm/gos/go1.4.1/src/testing/testing.go:329 +0x3d
github.com/stretchr/testify/require.Equal(0x7fd1d5e6f7b0, 0xc208044120, 0x733000, 0xc2080895c8, 0x733000, 0xc2080895d0, 0x0, 0x0, 0x0)
        /home/ory/.gvm/pkgsets/go1.4.1/global/src/github.com/stretchr/testify/require/requirements.go:41 +0xc6
github.com/stretchr/testify/require.(*Assertions).Equal(0xc20800b680, 0x733000, 0xc2080895c8, 0x733000, 0xc2080895d0, 0x0, 0x0, 0x0)
        /home/ory/.gvm/pkgsets/go1.4.1/global/src/github.com/stretchr/testify/require/forward_requirements.go:39 +0x84
github.com/somepath..../handlers.(*MySuite).TestFunc(0xc20803bf20)
        /home/ory/.gvm/pkgsets/go1.4.1/global/src/github.com/somepath..../handlers/file_test.go:270 +0xb11
reflect.Value.call(0x73a860, 0x859030, 0x13, 0x861770, 0x4, 0xc208101f40, 0x1, 0x1, 0x0, 0x0, ...)
        /home/ory/.gvm/gos/go1.4.1/src/reflect/value.go:419 +0x10e5
[...]

PR #216, opened 2015-09-18
Disable memoized require.Assertions in Suite

codygibb:
A quick fix for https://github.com/stretchr/testify/issues/149.

I wasn't able to figure exactly _why_ this was happening, but now `suite.Require()` is actually usable.

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -18,8 +18,7 @@ var matchMethod = flag.String("m", "", "regular expression to select tests of th
 // retrieving the current *testing.T context.
 type Suite struct {
 	*assert.Assertions
-	require *require.Assertions
-	t       *testing.T
+	t *testing.T
 }
 
 // T retrieves the current *testing.T context.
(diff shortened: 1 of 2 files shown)


oryband:
weird. in any case, please merge!

ernesto-jimenez:
@codygibb could you add a failing test so I can see what triggered the problem and we avoid regressions in the future?

Thanks!

ernesto-jimenez:
Looking at the code, this is changing the behaviour of the exported function and also making it inconsistent with `func (*Suite) Assert`.

It would be good to have a test showing why the code was unusable before and whether `Assert` is unusable right now too.

ernesto-jimenez:
@codygibb @oryband finally had some time to dig into the issue and found what the problem was.

The correct solution is in PR https://github.com/stretchr/testify/pull/228

Closing this PR, thanks!

oryband:
:+1: awesome!

PR #217, opened 2015-09-23
Added assertion/requirement that checks if two JSON strings represent equivalent objects

juliancooper:
While writing an API I found myself writing a lot of copy-pasta in my tests to check whether or two JSON strings represent equivalent objects. I can't be the only person using JSON strings so hopefully others will find this useful.

All this assertion does is unmarshal a string into a slice or map of `interface{}` and leverage the Equal assertion to verify equality.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -3,6 +3,7 @@ package assert
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"math"
 	"reflect"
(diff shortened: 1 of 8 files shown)


juliancooper:
Any thoughts or commentary on adding this assertion?

ernesto-jimenez:
Sorry @juliancooper, very busy month, thanks for your patience :)

I'm going to add some comments now.

assert/assertions.go
```diff
@@ -909,3 +910,37 @@ func NotZero(t TestingT, i interface{}, msgAndArgs ...interface{}) bool {
...
+
+// JSONEq asserts that two JSON strings are equivalent.
+//
+//  assert.JSONEq(t, `{"hello": "world", "foo": "bar"}`, `{"foo": "bar", "hello": "world"}`)
+//
+// Returns whether the assertion was successful (true) or not (false).
+func JSONEq(t TestingT, expected string, actual string, msgAndArgs ...interface{}) bool {
+	expectSlice := false
+	expectedJSONAsMap := make(map[string]interface{})
+	expectedJSONAsSlice := make([]interface{}, 0, 0)
```
ernesto-jimenez: you can `Unmarshal` into an `interface{}` rather than doing the two attempts:

```go
package main

import "fmt"
import "encoding/json"

func main() {
	var x interface{}
	json.Unmarshal([]byte("{\"foo\": \"bar\"}"), &x)
	fmt.Printf("%#v\n", x)
	json.Unmarshal([]byte("[{\"foo\": \"bar\"}]"), &x)
	fmt.Printf("%#v\n", x)
}
```

```
map[string]interface {}{"foo":"bar"}
[]interface {}{map[string]interface {}{"foo":"bar"}}
```

assert/assertions_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 var (
-	i      interface{}
+	i     interface{}
```
ernesto-jimenez: Did you run `go ftm`?

I'm guessing `fmt` would have kept this space.

assert/assertions_test.go
```diff
@@ -903,3 +903,18 @@ func TestNotZero(t *testing.T) {
 		True(t, NotZero(mockT, test, "%#v is not the %v zero value", test, reflect.TypeOf(test)))
 	}
 }
+
+func TestJSONEq(t *testing.T) {
```
ernesto-jimenez: Please, split the test in different test cases so it's clear all the different tests that are being tested and when something fails we can know what's the issue from the failing test.

assert/assertions_test.go
juliancooper: `go fmt` definitely eliminated this space. 
```
jcooper-mac:<snip>/github.com/stretchr/testify/assert git:(master) gofmt -d assertions_test.go 
diff assertions_test.go gofmt/assertions_test.go
--- /var/folders/mp/25h872dn1d764k6fy0cxcc300000gn/T/gofmt729445474     2015-10-13 12:55:01.000000000 -0700
+++ /var/folders/mp/25h872dn1d764k6fy0cxcc300000gn/T/gofmt785403993     2015-10-13 12:55:01.000000000 -0700
@@ -11,7 +11,7 @@
 )
 
 var (
-       i      interface{}
+       i     interface{}
        zeros = []interface{}{
                false,
                byte(0),
```

assert/assertions.go
juliancooper: Thanks for the pointer :+1:

assert/assertions_test.go
juliancooper: I did run `go fmt`, should I have not?

juliancooper:
@ernesto-jimenez, I addressed your recommended changes, think you could merge this?

ernesto-jimenez:
@juliancooper great, merging. Thanks!

abest0:
:+1:  Nice addition @juliancooper

Issue #146, opened 2015-03-19
Implements arg order is backwards


willfaught:
Most non-expected-actual functions, like EqualError and Len, put the object to be examined first and the expected value second, but Implements does it in reverse order: type first, then object. This makes it hard to remember.

matryer:
Please fix this and submit a PR. It would be great if it were consistent.

phemmer:
I would argue that `Implements` is correct, and `Len` is backwards.
In most of the functions, the thing being tested against is the second argument, and the thing being tested is the third (and latter) argument(s). Thus in `Len`, the expected value should be the second argument.

Functions I think need to be fixed: `Contains`, `Len`, `HTTPBodyContains`, `HTTPBodyNotContains`, `NotContains`

willfaught:
Agreed.

phemmer:
Looks like another backwards method has been added since this issue was created: `EqualError`

mvdkleijn:
I believe this issue and related PR can move forward again, but we need to take a look at #399 which makes a fair point.

willfaught:
I agree with https://github.com/stretchr/testify/issues/399, actually. [Go has a strong preference for the actual-then-expected order of comparison values](https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures). This ticket is just about the order—whatever it may be—being consistent among the funcs.

willfaught:
Closing due to no activity.

zzh8829:
request to reopen this for v2, the v1 parameter ordering is very counter intuitive

tonyhb:
+1 - `Len` is inconsistent.  The asserted value comes last, vs EqualValues which comes first.  In general, a bunch of these are inconsistent and it's pretty frustrating.

brackendawson:
As this is seems to be the oldest one, and it's already in the v2.0.0 milestone, sure. If v2 ever happens the expected/actual order will be made consistent. Most likely using the order from `assert.Equal` as it's the most ubiquitous.

HeCorr:
In regards to the naming convention, I vote for renaming the sides to "left" and "right" like they are called in Rust, eliminating the discrepancy between all the functions and making the sides unambiguous no matter which standard you're used to (expected-actual/actual-expected).

Thoughts? @willfaught @brackendawson

willfaught:
That’s what I’ve done in my own test helper. Given the age of this issue, I doubt this will ever happen to testify. Better to make your own, better library.

[...]

PR #219, opened 2015-09-24
Elminating backwards and improving consistentency across APIs

Fiery:
There was backwards API as discussed before, e.g. #146 , and here is the final fix to those mentioned.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -439,10 +439,10 @@ func getLen(x interface{}) (ok bool, length int) {
 // Len asserts that the specified object has specific length.
 // Len also fails if the object has a type that len() not accept.
 //
-//    assert.Len(t, mySlice, 3, "The size of slice is not 3")
+//    assert.Len(t, 3, mySlice, "The size of slice is not 3")
 //
 // Returns whether the assertion was successful (true) or not (false).
-func Len(t TestingT, object interface{}, length int, msgAndArgs ...interface{}) bool {
+func Len(t TestingT, length int, object interface{}, msgAndArgs ...interface{}) bool {
 	ok, l := getLen(object)
 	if !ok {
 		return Fail(t, fmt.Sprintf("\"%s\" could not be applied builtin len()", object), msgAndArgs...)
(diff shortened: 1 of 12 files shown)


DAddYE:
Will probably break some compatibility but can't wait to have this merged :+1:

Fiery:
Was wondering why `go test` didn't cover all cases and then realized that I need to test sub packages separately...

ernesto-jimenez:
@Flery, @DAddYE 

Do you have any knowledge about other big packages introducing breaking changes like these without having to rely on versioning through versioning in package URLs?

The core team used to use [gofix](https://blog.golang.org/introducing-gofix) when making breaking changes to the stdlib pre Go 1.0.

This kind of change could easily be fixed automatically in the code, but there would need to be a way to let people know they need to run the gofix/gofix-like tool once they have upgraded the package, rather than just having code not compiling.

Fiery:
@ernesto-jimenez you're right, may be we can consider create a migrating script using gofix and highlight it at very top of README.md to make it noticeable enough? [Kubernetes](https://github.com/kubernetes/kubernetes/blob/master/docs/getting-started-guides/README.md) does a good job to notify users about document discrepancies between different releases.

jzila:
This is great; I was getting frustrated with the inconsistent ordering of expected vs actual parameters, I hope this PR makes it! :+1:

ernesto-jimenez:
@Fiery sorry it took some time to reply, I've been crazy busy the past few weeks.

I've been thinking about this and my concern would be: how will developers know they have a script they can run to fix their code?

In most cases people would just fetch the new version and their tests will break. Specially in the cases were they will not even get a compilation error because we are swapping arguments that take an empty interface.

oryband:
Any effort and getting this merged?

Perhaps you can just bump the version to v2 and be done with it?

xordspar0:
Dependency management tools have gotten better in the last few years, especially the tools built into Go. People using Go modules have to choose when they want to switch to the v2 series; it won't just switch automatically when they do an update. Maybe now is the time to revisit this.

glesica:
I'm fine with this sort of effort, but this PR is outdated and someone would need to clean it up. I'd also like to address the `go generate` issues first so that a hypothetical v2 could be a little easier to work on... see: https://github.com/stretchr/testify/issues/841.

mvdkleijn:
Since #841 is now closed, from what I can tell this can move forward again. There's issue #399 however, that makes a fair point about the argument order.

boyan-soubachov:
I am closing this PR due to the age of it. I imagine a lot of these changes have been implemented over time? If not and you wish to reopen & rebase onto the latest branch, I'd be happy to review for you :)

PR #220, opened 2015-09-24
Fix NotNil assertion for typed nil values

bmatsuo:
The NotNil assertion had an error in its handling of typed nil values.
This change makes use of the helper function isNil (used by the Nil
assertion).  The helper function has correct handling of typed nil
values and when negated provides the expected semantics for
`assert.NotNil(t, x)`.

    if x == nil {
        assert.Fail(t, "nil", x)
    }

edit: I messed up the code block the first time. negated assertions are tricky :wink:

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -287,24 +287,10 @@ func Exactly(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
 //
 // Returns whether the assertion was successful (true) or not (false).
 func NotNil(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
-
-	success := true
-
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
Thanks @bmatsuo 

:+1: and it actually makes sense given how assert.Nil is implemented :)

For symmetry, would you mind testing for the same case in `TestNil`?

bmatsuo:
Good call, @ernesto-jimenez. I just pushed a new commit with the additional test case.

ernesto-jimenez:
:+1: will merge as soon as tests pass :)

ernesto-jimenez:
Thanks @bmatsuo :)

PR #221, opened 2015-09-28
Due to function name format differences, code fails with GCCGO

brahmaroutu:
Error happens when running Docker unit tests using gccgo. Function names are different in gccgo and hence output from runtime.FuncForPC cannot be parsed correctly with the current code.
GCCGO 
functionPath  github_com_docker_libkv_store_mock.WatchTree.pN39_github_com_docker_libkv_store_mock.Mock

Golang
functionPath  github.com/docker/libkv/store/mock.(*Mock).WatchTree

Signed-off-by: Srini Brahmaroutu <srbrahma@us.ibm.com>

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -3,6 +3,7 @@ package mock
 import (
 	"fmt"
 	"reflect"
+	"regexp"
 	"runtime"
 	"strings"
 	"sync"
@@ -265,6 +266,14 @@ func (m *Mock) Called(arguments ...interface{}) Arguments {
 		panic("Couldn't get the caller information")
 	}
 	functionPath := runtime.FuncForPC(pc).Name()
+	//Next four lines are required to use GCCGO function naming conventions.
+	//For Ex:  github_com_docker_libkv_store_mock.WatchTree.pN39_github_com_docker_libkv_store_mock.Mock
+	//uses inteface information unlike golang github.com/docker/libkv/store/mock.(*Mock).WatchTree
+	//With GCCGO we need to remove interface information starting from pN<dd>.
+	re := regexp.MustCompile("\\.pN\\d+_")
+	if re.MatchString(functionPath) {
+		functionPath = re.Split(functionPath, -1)[0]
+	}
 	parts := strings.Split(functionPath, ".")
 	functionName := parts[len(parts)-1]
 


ernesto-jimenez:
Hi @brahmaroutu,

How can we add a test for this?

brahmaroutu:
@ernesto-jimenez You need to have the code compiled and run with gccgo. I am using gccgo 5.2 branch to test docker on x86. Sorry I am not sure if we can add a unit test without changing the compiler.

ernesto-jimenez:
@brahmaroutu am I right to assume our test suite is failing with gccgo?

I think it would be good to:
 - Add a comment explaining the workaround
 - Modify `.travis.yml` to run the tests with gccgo too (we already run tests against several versions of Go)

DAddYE:
If doesn't produce regressions (which seems the case) I'll would merge it. I think is pretty hard to build a gcc trunk on travis keep it updated

brahmaroutu:
@ernesto-jimenez, @DAddYE I am sorry, I am not trying to make the package compile and run with GCCGO, Docker X86 CI builds  with GCCGO are currently broken when this package is added. I am fixing that issue for now. Please review and merge the code. There may be one or two more issue to address if you are interested in complete GCCGO compatibility.

ernesto-jimenez:
@brahmaroutu ok, although with no tests it's very likely there will be a regression in the future, since most people are not using the package with gccgo.

Please, could you add a comment about the workaround and why it is there? It would be good to have some context in case somebody refactors the code or the workaround stops working :)

Once we've got the comment I will merge it.

ernesto-jimenez:
Also, I haven't used gccgo and haven't looked at the internals, but I'm wondering whether that regexp cover all cases or might have to be updated in the future :)

brahmaroutu:
@ernesto-jimenez Please review my comments if that makes sense. 
I think this is general format for function names in GCCGO and will remain generic fix. It should cover all cases and off course it is quite unlikely a function name contain .pN<dd>_ format to cause trouble on the native Go side. 
This fix will not make this package fully GCCGO compliant.

brahmaroutu:
hmm, formatted my regexp above, please read it as ".pN\<dd\>_"

ernesto-jimenez:
:+1: thanks! merging :)

brahmaroutu:
Thanks for merging this.

ernesto-jimenez:
np, good luck with adding support for gccgo in docker! :)

PR #223, opened 2015-10-07
fix race conditions related to mutation of expectedCall.Repeatability

allgeek:
The refactoring in 49804a38 reintroduced a race condition, which has been flagged by the race detector and is causing random failures of some of my tests.

Writes to an expected call's Repeatability field are still protected by the mutex in *Call.Times(), as well as in *Mock.Called(), but there is no protection for reads.  This commit corrects that with the minimal possible changes, although it's not terribly elegant.

It's unfortunate that the public API exposes mock.Calls, mock.ExpectedCalls, and other fields that are not safe for concurrent access, but I'm assuming the prior changes to fix race conditions in b11fb169 were partially motivated by a desire or need to avoid breaking that public API.  It may be a good idea to at least expose the mutex for consumers to safely access things, but I've kept that unexported for now.

findExpectedCall() has been changed to explicitly lock, and uses the exported ExpectedCalls again.

AssertExpectations() is a bit messier, as methodWasCalled() needs the mutex (without more invasive changes), so it's been refactored slightly to enable locking before the Repeatability read.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -205,7 +205,9 @@ func (self *Mock) On(methodName string, arguments ...interface{}) *Call {
 // */
 
 func (m *Mock) findExpectedCall(method string, arguments ...interface{}) (int, *Call) {
-	for i, call := range m.expectedCalls() {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	for i, call := range m.ExpectedCalls {
 		if call.Method == method && call.Repeatability > -1 {
 
 			_, diffCount := call.Arguments.Diff(arguments)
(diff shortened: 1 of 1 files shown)


ernesto-jimenez:
@allgeek thanks for the contribution.

You are correct that we want to avoid breaking the public API since there are many projects depending on this package.

About exporting the mutex, I think it would be more interesting to run some analysis on the open source packages depending on this one to see whether any package is actually using the `Calls` or `ExpectedCalls` fields. I agree those should be private and if possible I would rather make those private than exporting the mutex :)

allgeek:
Our use case is likely abnormal (and I've already adapted our code to eliminate access of `Calls`), but in an internal project we had previously adopted a different assertion library/convention, and were simply using the Mock package by itself.  `Calls` was being accessed by some of our assertion helpers, since we weren't using the testify asserts.

To eliminate the races that caused, we're now providing a fake `TestingT` to `AssertExpectations()` calls, and just relying on the return value to adapt to our typical assertion convention.  That's likely the route the devs would have taken in the first place if `Calls` weren't exported.

Thanks much for the feedback!

PR #227, opened 2015-10-29
Add custom mock argument matching

jgraettinger:
Add an ArgumentMatcher interface and MatchedBy() helper to the mock
package, enabling custom fine-grained argument matching behaviors.

This is particularly helpful when building mocks over complex argument
types where deep equality is cumbersome or can't be used. Ex, a matcher
might match over just the the URL of an argument *http.Request.

[Review on Reviewable](https://reviewable.io/reviews/stretchr/testify/227)

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -458,6 +458,47 @@ func AnythingOfType(t string) AnythingOfTypeArgument {
 	return AnythingOfTypeArgument(t)
 }
 
+// argumentMatcher performs custom argument matching, returning whether or
+// not the argument is matched by the expectation fixture function.
+type argumentMatcher struct {
+	// fn is a function which accepts one argument, and returns a bool.
+	fn reflect.Value
+}
+
+func (f argumentMatcher) Matches(argument interface{}) bool {
+	expectType := f.fn.Type().In(0)
+
+	if reflect.TypeOf(argument).AssignableTo(expectType) {
+		result := f.fn.Call([]reflect.Value{reflect.ValueOf(argument)})
+		return result[0].Bool()
+	} else {
+		return false
+	}
+}
+
+func (f argumentMatcher) String() string {
+	return fmt.Sprintf("func(%s) bool", f.fn.Type().In(0).Name())
+}
+
+// Builds an argumentMatcher from |fn|, which must be a function accepting a
+// single argument (of any type) which returns a bool. If |fn| doesn't match
+// the required signature, MathedBy() panics.
+func MatchedBy(fn interface{}) argumentMatcher {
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
@jgraettinger I think the core idea is great, but I would make some changes.

Looking at the test, I think it would be nicer to do:

```go
mockedService.On("TheExampleMethod",
  MatchedBy(func(a int) bool {
    return a == 1
})
```

You can use the `reflect` package to look at the signature of the function sent to `MatchedBy` which means that we can check whether the type of the argument to the mock call matches the type of the argument of the function provided to `MatchedBy`. It also gets rid of the need for converting the `interface{}` to whatever type you are expecting.

I think it would make the feature nicer to use. What do you think?

ernesto-jimenez:
@jgraettinger I remembered `gocraft/web` uses the reflect package to check if a function matches certain signature.

You can see the code here: https://github.com/gocraft/web/blob/3fd2ffd35c9233bfe91caa84e9c09d7cf8fae964/router_setup.go#L318-L357

It should help.

Also, what use case do you have in mind for exporting the `ArgumentMatcher` interface?

jgraettinger:
@ernesto-jimenez Think your feedback here is great. I've implemented as described, and added a bit more testing here. Also opted to drop the ArgumentMatcher interface altogether. PTAL

mock/mock.go
```diff
@@ -458,6 +458,47 @@ func AnythingOfType(t string) AnythingOfTypeArgument {
...
+		panic(fmt.Sprintf("assert: arguments: %s is not a func", fn))
+	}
+	if fnType.NumIn() != 1 {
+		panic(fmt.Sprintf("assert: arguments: %s does not take exactly one argument", fn))
+	}
+	if fnType.NumOut() != 1 || fnType.Out(0).Kind() != reflect.Bool {
+		panic(fmt.Sprintf("assert: arguments: %s does not return a bool", fn))
+	}
+
+	return argumentMatcher{fn: reflect.ValueOf(fn)}
```
ernesto-jimenez: maybe we should store `reflect.Value(fn).Type().In(0)`?

jgraettinger: Mind elaborating on the rationale here? As an optimization?

IANAE, but as I understand it once you have a reflect.Value, the heavy lifting of type introspection has already been done.

ernesto-jimenez: nevermind, I missed one line where you were using more than `.Type().Int(0)`

ernesto-jimenez:
@jgraettinger looks good to me. The only thing missing is some documentation. Could you add some?

Thanks!
Ernesto

PR #228, opened 2015-11-01
Fixes issue #149: suite.Require() panics

ernesto-jimenez:


suite/suite.go
```diff
@@ -31,6 +31,7 @@ func (suite *Suite) T() *testing.T {
 func (suite *Suite) SetT(t *testing.T) {
 	suite.t = t
 	suite.Assertions = assert.New(t)
+	suite.require = require.New(t)
```
oryband: is `require` supposed to be private?

ernesto-jimenez: You have `Require()` to get the value already.

`Assertions` is exported because it's embedded in the `Suite` struct so when you call `Equal` on a Suite it actually calls `Assertions.Equal`

Issue #123, opened 2015-01-20
IsEmpty does not check for empty structs of the same type


dkdodson3:
I looked at your IsEmpty function to see if I could find a way to check to see if the struct is of the same type and empty. Figured out how to do it with reflection and thought I would share...

```go
package main

import ( 
    "fmt"
    "reflect"
)

type ContainerData struct {
    ContainerUrl string `json:"container_url"`
    AckUrl       string `json:"ack_url"`
}

func main() {
    blah := ContainerData {
        ContainerUrl : "consul-kv://web/containers/abc123",
        AckUrl : "http://requestb.in/11u0wr51",
    }

structType := reflect.TypeOf(blah)
structPointer := reflect.New(structType)
structValue := structPointer.Elem()
structValueInterface := structPointer.Interface()
intPtr := reflect.New(reflect.TypeOf(blah)).Elem().Interface()


fmt.Println(blah)
fmt.Println(structType)
fmt.Println(structPointer)
fmt.Println(structValue)
fmt.Println(structValueInterface)
fmt.Println(intPtr)
fmt.Println(reflect.TypeOf(intPtr))
}
```

matryer:
That’s very cool

georgelesica-wf:
Hello, my name is George. I've recently taken over maintenance of Testify and
I'm excited to help the project move forward.

In order to improve responsiveness and promote contributions to the Testify
project, I have decided to close pull requests and issues that haven't had any
activity in several months and do not appear to address critical bugs.

If you would still like to see your pull request or issue addressed, I am happy
to help! All you have to do is re-open and update it to reflect the current
state of the master branch. Once that is done, I will triage and recruit
reviewers as-needed.

If you have questions, please feel free to reach out. Thank you for your
interest in improving the Testify project!

PR #232, opened 2015-11-02
Fix #136 - Empty returns true for nil pointers

### LINKED ISSUE #136: Empty doesn't check if a pointer value is nil
--- FAIL: TestMain (0.00s)
            Location:       testingutil.go:43
            Error:          Should be empty, but was <nil>

What's wrong with this picture? ;)

ernesto-jimenez:


ernesto-jimenez:
@DAddYE mind double checking?

It's a fix for #136

assert/assertions.go
```diff
@@ -358,6 +358,9 @@ func isEmpty(object interface{}) bool {
 		}
 	case reflect.Ptr:
 		{
+			if objValue.IsNil() {
+				return true
+			}
 			switch object.(type) {
 			case *time.Time:
```
ernesto-jimenez: We probably should not have a special case for `time.Time`.

ernesto-jimenez: Probably related: https://github.com/stretchr/testify/issues/123

Issue #229, opened 2015-11-01
`Equal` errors should show a diff for structs/maps/etc.


ernesto-jimenez:
Right now, when comparing two big structs or maps it's really painful to find what the differences are that make `Equal` fail.

The returned error should include a diff of both values when posible.

PR #233, opened 2015-11-03
Show diff of values when Equal fails

ernesto-jimenez:
This addresses issue #229.

Example output:
```
--- FAIL: TestSomething (0.00s)
        Error Trace:    main_test.go:60
        Error:          Not equal: struct { foo string }{foo:"foo"} (expected)
                                != struct { foo string }{foo:"bar"} (actual)

                        Diff:
                        --- Expected
                        +++ Actual
                        @@ -1,3 +1,3 @@
                         (struct { foo string }) {
                        - foo: (string) (len=3) "foo"
                        + foo: (string) (len=3) "bar"
                         }
```

This should help when Equal fails for big structs, maps, arrays or slices.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -13,6 +13,9 @@ import (
 	"time"
 	"unicode"
 	"unicode/utf8"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/pmezard/go-difflib/difflib"
 )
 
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
@DAddYE mind reviewing this?

DAddYE:
:+1:

PR #237, opened 2015-11-06
Add race detection concurrency test for shared mocks

brildum:
This branch adds a test for concurrency safety for shared mocks, which I've defined as any 2 or more mocks which are used together.

```mock1.On("DoSomethingWithOtherMocks", mock2, mock3).Return(true)```

would make mock1, mock2, mock3 shared.

To reproduce, you must run your tests with the race detector enabled (tested on Go1.5):

```go test -race ./...```

### DIFF
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"math/rand"
+	"sync"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ernesto-jimenez** [comment]
@brildum the test can be simplified to this:

```go
func Test_Shared_Mocks_Concurrency(t *testing.T) {
	b := &SharedTestMockB{}
	a := &SharedTestMockA{}
	a.On("GetA").Return(true)
	b.On("GetB", a).Return(true)

	go func() {
		b.GetB(a)
	}()
	go func() {
		a.GetA()
	}()
}
```

PR #238, opened 2015-11-06
Fix data races with shared mock concurrency

brildum:
See https://github.com/stretchr/testify/pull/237 if you're only interested in a test which surfaces the issue and not the fix.

Instead of using a mutex per-mock, I've switched to a global shared mutex between all mocks. My intuition is that there won't be significant contention in unit-tests, and this will perform well for the vast majority of tests.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -13,6 +13,10 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+var (
+	mu sync.Mutex
+)
+
 // TestingT is an interface wrapper around *testing.T
 type TestingT interface {
 	Logf(format string, args ...interface{})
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
Hi @brildum,

Was this provoking some test fail or block or is it just to address the `-race` warning?

As I showed in the other issue, the test can be simplified. The problem seems to be the fact that Mock is doing a `fmt.Sprintf` of the arguments, which in case is another mock.

I would like to understand how can a test break due to this corner case about using a mock as an argument for another mock.

Thanks!

brildum:
When running with the race detector enabled -- `go test -race` will exit with a non-0 return value. No explicit tests were failing, but many are not executed once the race detector finds an issue.

ernesto-jimenez:
@brildum my main worry would be to confirm there are no unintended consequences from moving to a global lock. We've got hundreds of projects depending on this package and I would like to be sure.

We can prioritise if higher If you can mention a point where actual behaviour would be broken beyond the data race warnings.

Thanks,
Ernesto

ernesto-jimenez:
@brildum did you get it sorted?

PR #239, opened 2015-11-07
Handle case where one object is nil and the other is not

anacrolix:
I get this an awful lot when doing assert.EqualValues on expected error values.


ernesto-jimenez:
Thanks @anacrolix 

Merging :)

PR #240, opened 2015-11-07
Show type of the object instead just "Object" string

### LINKED ISSUE #234: Misleading error message with Implements assertion
Disclaimers:

- I don't know if this behaviour is by design.
- I understand why is happening but I think that it could be a little bit more verbose with the error.

Using this assertion:

    assert.Implements(t, (*TheInterface)(nil), &TheStruct{})

Everything works fine, but if I forget the `&` the error that I will get is:

    Error:		Object must implement TheInterface

Which in my opinion is a little bit misleading.

What do you think? Feel free to close the issue if this behaviour is expected by design.

agonzalezro:
The `Implements` test was just showing a string instead the name of the
object that was not implementing the interface. Now the type of the
object is shown.

Related with #234


ernesto-jimenez:
Merging, thanks! :)

PR #241, opened 2015-11-08
Autogenerate code for require and forwarded assertions

ernesto-jimenez:
Right now each assertion requires to define 4 methods: the actual assertion and three methods wrapping the assertion.

This PR introduces a Go program to automatically generate the wrappers.

This is still work in progress, since it needs a refactor.


matryer:
Oh this is very clever.

PR #247, opened 2015-12-02
True -> False in false assertion comments

blakewilson-wf:
Noticed a few apparent typos in the comments for false asserts/requires


ernesto-jimenez:
thanks

PR #248, opened 2015-12-02
Glide Support Added

akutz:
This patch adds support for the Glide dependency management system by creating a "glide.yaml" file with dependencies on "github.com/davecgh/go-spew" and "github.com/pmezard/go-difflib".

This change will allow other projects to depend upon testify via Glide without having to explicitly depend upon testify's transitive dependencies.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,0 +1,10 @@
+package: github.com/stretchr/testify
+import:
+  - package: github.com/davecgh/go-spew
+    ref:     master
+    repo:    https://github.com/davecgh/go-spew.git
+    vcs:     git
+  - package: github.com/pmezard/go-difflib
+    ref:     master
+    repo:    https://github.com/pmezard/go-difflib.git
+    vcs:     git


ernesto-jimenez:
@akutz I was actually thinking about vendoring those dependencies with godep and rewriting the imports.

akutz:
Hi @ernesto-jimenez,

That would be fine and dandy with me as Glide automagically detects godep files and will convert them on-demand in order to avoid vendoring errors. Then again, there's no reason *not* to include both if that's your pleasure...

akutz:
Hi @ernesto-jimenez,

By the way, I *love* testify! Thank you so much for it!

ernesto-jimenez:
Thanks @akutz :)

I went ahead and did a quick PR with godep and rewriting the import paths: #249

This way testify is standalone rather than pulling external dependencies that might break in the future.

I think it'll be better than relying on glide and/or godep.

What do you think?

akutz:
Hi @ernesto-jimenez,

Well, I personally dislike storing vendored dependencies in the dependents' repos as it creates cruft to pull and to consider. I usually follow a Maven philosophy of locking down/pinning dependencies for tagged releases of a dependent. For example, with Glide (and Godeps), you are able to specify the revision/hash/commit ID/ref on which to depend. During active development I almost always have this set to `master` (or whatever the project's development branch is), but prior to release of my own project I'll lock/pin this to a specific revision/hash/commit ID/ref. 

The above technique avoids the external breakage which you mentioned. One could argue that if Github were to go away that this procedure still allows things to break due to absence of dependencies, but I'd argue that if Github goes away then languages and frameworks such as Golang and Node are in for much larger issues. 

Does that make sense?

ernesto-jimenez:
In my projects I store the vendored dependencies so I can still deploy when GitHub has some downtime, which really happens :)

In this case the bigger problem is the repository being removed or renamed rather than GitHub disappearing. Just pinning is not enough to protect us from that.

ernesto-jimenez:
Tagged dependencies with `godep` and vendored them using GO15VENDOREXPERIMENT

ernesto-jimenez:
already pinned and vendored dependencies with godep.

PR #250, opened 2015-12-04
Fix missing type in signature

garrypolley:
Not a huge change but it was a little confusing when I read it at first and the type was missing.


ernesto-jimenez:
@garrypolley that's correct in Go. When you have more than one arguments with the same type you can omit it in the initial args and just leave it in the last one. Example: http://play.golang.org/p/xjkR-QdvXp

You'll see you can also omit the arguments in return when you are using named return values.

PR #254, opened 2015-12-23
Print full location in failed test.

benma:
Enable with env variable TESTIFY_PRINT_LOCATION=1.

Normal compilation prints the full location with line and column number, so that the editor/programming tools can automatically jump to it (emacs, vim, ...).

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"os"
 	"reflect"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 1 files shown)


assert/assertions.go
```diff
@@ -183,22 +184,39 @@ func indentMessageLines(message string, tabs int) string {
 
 // Fail reports a failure through
 func Fail(t TestingT, failureMessage string, msgAndArgs ...interface{}) bool {
-
+	location := ""
+	printLocation, _ := os.LookupEnv("TESTIFY_PRINT_LOCATION")
```
ernesto-jimenez: Since you are ignoring the `ok` value, you can simply use `os.Env` which is compatible with older Go releases.

ernesto-jimenez:
@benma what's the desired output you want and why?

Why have it behind and env variable rather than a flag or have it always enabled?

benma:
The desired output is the full path of the file and line number of the failed test, as the PR provides. Note that there is no indentation as with the rest, so that Emacs (and probably other editors) can jump to that location automatically like it does with compilation errors.

I did not enable it always because it is a bit invasive and probably not everyone wants this.

How exactly can one enable this with a flag? This sounds even more reasonable to me, and I would change the code accordingly.

assert/assertions.go
benma: Thanks, I will change it once I get back to my dev environment (I am currently traveling).

ernesto-jimenez:
One important question: how do editors work with the errors from standard `go test`?

I believe it just shows the file name and the line number, rather than the whole path. Which means testify is currently aligned with `go test`

I'm a bit concerned about having this hidden behind and environment variable or a flag:

 - With an env variable you'll probably be the only one using the feature ;)
 - With a flag would be a bit more discoverable, but we'll be adding a flag to `go test` which applies only to testify.

@fatih you are maintaining vim-go, what's your take on this request? Is it something you are missing from the `go test` output?

ernesto-jimenez:
Had a chat with @fatih tonight (thanks!), here's a summary from what we discussed based on his experience with vim-go:

Best thing to do to facilitate integration into tools like vim-go would be to align with the output from `go test`. Stdlib prints a single line per error as `file.go:line: message`.

As an example, vim-go has the following regex: `let tokens = matchlist(line, '^\s*\(.\{-}\):\(\d\+\):\s*\(.*\)')`. They have some logic to figure out the path of the file.

In order to align with `go test` and get our errors picked up by vim-go we need to output a single line of text matching the regex and showing a one-line summary.

e.g:
```
--- FAIL: TestExample (0.00s)
main_test.go:18: values should be equal

        Error Trace:    main_test.go:18
        Error:          Not equal: []string{"a", "b", "x"} (expected)
                                != []string{"a", "b", "y", "d"} (actual)

                        Diff:
                        --- Expected
                        +++ Actual
                        @@ -1,5 +1,6 @@
                        -([]string) (len=3 cap=3) {
                        +([]string) (len=4 cap=4) {
                          (string) (len=1) "a",
                          (string) (len=1) "b",
                        - (string) (len=1) "x"
                        + (string) (len=1) "y",
                        + (string) (len=1) "d"
                         }
```

However this would require reviewing all the error messages from testify and might require a breaking to include the one-liner error argument in `Fail`.

In any case, it's better to avoid adding the path. Plugins would already have logic to figure out the path based on the stdlib output. Adding the path can pollute the output considerably and make it less readable.

Furthermore, if we were to do it, I believe we should have it enabled by default rather than hiding it behind a flag or an env variable. That way it'll be integrated by the tool's default `go test` execution rather than require special configuration to have testify place nicely with editor plugins.

Closing this PR in favour of an issue: https://github.com/stretchr/testify/issues/267

Issue #194, opened 2015-07-21
Order of test runs in suite


pkieltyka:
I noticed the tests within a suite run in alphabetical order... why is that..? seems strange. Why wouldn't it follow the same as standard go test cases which is in the order they have been declared?

pkieltyka:
I see this is an issue with reflect.TypeOf(mySuiteStruct) and it returning the methods in alphabetical order.. am I the only one with this issue? I find with end-to-end states, having separate functions for the states is helpful conceptually, but its breaking in practice

TauZero:
It's nice to explicitly test building block methods first, so tests of methods utilizing those can make certain assumptions about the foundational methods and be better focused. Being able to run the tests in a suite in order of definition, as bare tests would be run, or specify a run order (without having to add garbage to the test names to force alphabetical order) would be great.

ernesto-jimenez:
@pkieltyka @TauZero what's the use case to have tests run in certain order?

In my opinion, tests should pass independently of heir order. Tests should keep passing if you reorder the definition or change their names.

@TauZero for what you suggest, you can simply have two different suites and run them in whatever order you prefer.

leebenson:
@ernesto-jimenez IMO, testing order is VERY important. It's common to build on tests sequentially, especially when previous tests have side-effects.

An example I have right now is home-brew session middleware.

The necessary test order is:

1. create new session (TestNewSession)
2. add data (TestAddData)
3. retrieve data (TestRetrieveData)
4. update session to "expired" (TestExpireSession)
5. test auto-deletion of expired sessions (TestDeleteSession)

Out of order, I get a slew of `runtime error: invalid memory address or nil pointer dereference` errors because the steps no longer make any sense.

Alphabetically, the order becomes:

1. add data (TestAddData)
2. test auto-deletion of expired sessions (TestDeleteSession)
3. update session to "expired" (TestExpireSession)
4. create new session (TestNewSession)
5. retrieve data (TestRetrieveData)

In order words...

* I'm adding data (TestAddData) before there's even a session to add to (TestNewSession)
* I'm trying to delete an expired session (TestDeleteSession), that has not been set to expired (TestExpireSession)
* I'm retrieving data (TestRetrieveData) on a session that has already been deleted (TestDeleteSession)

Of course, I could consolidate this functionality into a single test function, but then what would be the point of a test "suite" at all? Why not just have a single go test block and use the assert library?

[...]

PR #258, opened 2016-01-02
tests run by definition order

leebenson:
- fixes stretchr/testify/issues/194

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"reflect"
 	"regexp"
+	"sort"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
Thanks @leebenson,

I'm afraid I'm still unconvinced about the functionality. We can follow the discussion in the issue #194 and we can reopen the PR if and when we agree this should be implemented and this solution is relying on specified behaviour from Go.

leebenson:
@ernesto-jimenez what's the downside?

Alphabetical is an arbitrary order by default. It assumes a 1:1 correlation between name and run order.

If your argument is that definition order _shouldn't_ be linked to run order, then that's equally arbitrary and therefore ordering by method definition makes no material difference to users that don't rely on a particular sequence order.

It's exactly the same thing either way - except my way, you're now satisfying users like myself, the two other guys that opened the issue in #194, and no doubt hundreds of others who DO care about order.

Again - what's the downside?

PR #259, opened 2016-01-07
Add check for empty time.Time struct (non pointer)

SchumacherFM:
The standard library returns always a non-pointer Time type.

Thanks for merging!


SchumacherFM:
Thanks! my tests will now also pass on travis CI :-)

PR #260, opened 2016-01-07
Add EventuallyTrue assert

boudra:
This asserts that a given function will return true before a given duration, if it times out it will fail the test.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -995,7 +995,7 @@ func EventuallyTrue(t TestingT, test func() bool, timeout time.Duration, tick ti
 				return true
 			}
 		case <-timeoutChannel:
-			Fail(t, fmt.Sprintf("Timeout of %s reached", timeout))
+			Fail(t, fmt.Sprintf("Waiting for expression to eventually be true timed out (%s)", timeout))
 			return false
 		case <-ticker.C:
 			go func() {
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
Hi @boudra,

What's the use case for this?

Thanks,
Ernesto

boudra:
Hi @ernesto-jimenez,

I'm currently using this to test a piece of code that sends and receives messages through Redis asynchronously and calls a function handler. I am mocking the function handler to assert that the calls are correct but I need to `time.Sleep` before asserting since Redis takes a couple of milliseconds to transfer the message.

I know the Redis implementation could be mocked so everything runs synchronously but since it's so easy to just spin up a Redis docker container in a CI environment I thought testing with the real thing could be much better.

Another use case could be doing web driver tests where you usually wait for elements to be visible, a text to exist etc.

Cheers

ernesto-jimenez:
How are you mocking the handler? Can you provide a code sample of the test?

You can use channels or `sync.WaitGroup` for synchronisation rather than polling with a ticker. Something like this:

```go
func TestAsync(t *testing.T) {
  done := make(chan struct{})
  handler := func() {
    close(done)
  }
  yourFunction(handler)
  select {
  case <-done:
    // Success
  case <-time.After(500 * time.Millisecond):
    t.Fatal("Failed to call handler")
  }
}
```

Regarding web driver, the package I've used already takes care of the waiting internally. That will probably be better than doing it in the test.

Issue #261, opened 2016-01-08
require.FailNow() removed from exported interface following codegen merge


mattheath:
As a result of the code generation PR being merged (#241) the `FailNow()` method is no longer exported from the `require` package.

The `FailNow()` method is still present on the `TestingT` interface within the package, but presumably the code generation doesn't have this?

0d5a14c5a477957864f3b747d95255ad4e34bcc0 - `require.FailNow()` exists
efd1b850c1e5df1c539e83f61f7d5e113b6484e9 - first commit on the codegen branch (#241) removes `require.FailNow()`
c92828f29518bc633893affbce12904ba41a7cfa - #241 merged into master

ernesto-jimenez:
Hi @mattheath,

Thanks for the heads up, looking into it right now.

PR #262, opened 2016-01-08
Add FailNow back

ernesto-jimenez:
Fix breaking change where we removed FailNow. Note FailNow is also exported in `assert` to keep symmetry between `assert` and `require`.

It's still a small breaking change since FailNow now returns a bool like the rest of the functions. However that should be compatible with all existing code bases since.

### DIFF
--- a/assert/assertion_forward.go
+++ b/assert/assertion_forward.go
@@ -107,6 +107,12 @@ func (a *Assertions) Fail(failureMessage string, msgAndArgs ...interface{}) bool
 }
 
 
+// FailNow fails test
+func (a *Assertions) FailNow(failureMessage string, msgAndArgs ...interface{}) bool {
+	return FailNow(a.t, failureMessage, msgAndArgs...)
+}
+
+
 // False asserts that the specified value is false.
 // 
 //    a.False(myBool, "myBool should be false")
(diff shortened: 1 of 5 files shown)


ernesto-jimenez:
@mattheath, is this OK?

mattheath:
@ernesto-jimenez yep, this fixes #261

mattheath:
Thanks! :heart:

ernesto-jimenez:
Thank you for the quick report! :)

jasdel:
Hi I just noticed this change, adding FailNow() to assert, broke an upstream dependencies of mine. Is it intended for this change break the `assert.TestingT` interface?

ernesto-jimenez:
Hi @jasdel,

How are you depending on `assert.TestingT`?

jasdel:
@ernesto-jimenez a tool use use [gucumber](https://github.com/lsegal/gucumber) which implemented its own [TestingT](https://github.com/lsegal/gucumber/blob/e8116c9c66e641e9f81fc0a79fac923dfc646378/runner.go#L379) internal logic which satisfied the previous `assert.TestingT` but when `FailNow` was added it no longer satisfies the interface. Causing my downstream project to error out.

I can update my project and work to get a fix pushed upstream to gucumber, but wanted to make sure that breaking `assert.TestingT` was intentional.

ernesto-jimenez:
@jasdel It's just an issue trying to align `assert` with `require`. I mistakenly assumed everybody sent `testing.T`.

I'll get this sorted tomorrow so everything stays backwards compatible. See #263 if you want to subscribe.

PR #264, opened 2016-01-09
Fix Backwards Incompatible change to TestingT

ernesto-jimenez:


jasdel:
Awesome, thanks for fixing this so quickly!

ernesto-jimenez:
My pleasure. Sorry for the breaking change :)

PR #269, opened 2016-01-22
Add Suite Focus

jonnyreeves:
Suite Focus allows developers to only run one (or more) tests in a suite, skipping all others by prefixing your test with an `X`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,34 @@ func TestExampleTestSuite(t *testing.T) {
 }
 ```
 
+During debugging, you can focus (ie: only run) on certain tests in a suite by prefixing them with an `X`, eg:
+
+```go
+import (
+    "testing"
+    "github.com/stretchr/testify/suite"
+)
+
+type ExampleTestSuite struct {
+    suite.Suite
+}
+
+// This test is "focused" and will run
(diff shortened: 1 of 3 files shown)


ernesto-jimenez:
Hey @jonnyreeves,

There's already a way to run only certain methods in a suite using the `-m` argument[1][dash-m] with `go test`.

That is aligned with `go test`'s way of focusing on tests with `-run` and `-bench`.

I'm closing this, but thanks :)

[dash-m]: https://github.com/jonnyreeves/testify/blob/feature/suite-focus/suite/suite.go#L16

jonny-improbable:
Thanks for the tip @ernesto-jimenez; I still have a use-case for this feature: IntelliJ supports running individual tests via the Golang IntelliJ plugin, however it does not provide the ability to run a single test within a TestSuite.

Totally understand if you feel this is an upstream issue that should be fixed in the IntellIJ plugin / not in keeping with the rest of the golang ecosystem :)

ernesto-jimenez:
@jonny-improbable, you are right, this would be better handled by updating the IntelliJ plugin to support focusing on tests within a suite :)

PR #274, opened 2016-02-13
Allows specifying .Times(0) to indicate a mock should not be called

vladlosev:
This change will allow one to specify `.Times(0)` to set an expectation of a method _not_ being called with a given set of arguments. It is handy for testing failure scenarios. For example, the example below will test that if `fileServiceMock` fails to open a file, it will try to neither read nor close it:
```go
func TestFileOpenFailure(t *testing.T) {
    fileServiceMock := new(MockFIleService)
    fileServiceMock.on("Open", "filename.txt", "r").Times(1).Return(nil, fmt.Errorf("File not found"))
    fileServiceMock.on("Read", mock.Anything).Times(0)
    fileServiceMock.on("Close", mock.Anything).Times(0)
    defer fileServiceMock.AssertExpectations(t)
    fileProcessor:= NewFileProcessor(fileServiceMock);
    assert.Error(t, fileProcessor.ProcessFile("filename.txt"))
}
```
There is currently a way to verify that a call has not been made using `Mock.AssertNotCalled` but using `.Times(0)` has a couple of advantages. First, its syntax is more consistent with the rest of the expectations and second, it allows to report the unexpected call at the moment it was made, with stack trace available, rather than just report the fact afterwards, without a stack trace.

The change is implemented by changing the meaning of `Call.Repeatability` field. Before, a value `0` had a special meaning of 'any positive amount'. Now, that meaning is assumed by a special constant `Unlimited`, and `0` will mean just that - zero calls expected.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -41,7 +41,7 @@ type Call struct {
 	ReturnArguments Arguments
 
 	// The number of times to return the return arguments when setting
-	// expectations. 0 means to always return the value.
+	// expectations. Unlimited means to always return the value.
 	Repeatability int
 
 	// Holds a channel that will be used to block the Return until it either
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -2,6 +2,7 @@ package mock
 
 import (
 	"fmt"
+	"math"
 	"reflect"
 	"regexp"
 	"runtime"
@@ -59,7 +60,7 @@ func newCall(parent *Mock, methodName string, methodArguments ...interface{}) *C
 		Method:          methodName,
 		Arguments:       methodArguments,
 		ReturnArguments: make([]interface{}, 0),
-		Repeatability:   0,
+		Repeatability:   Unlimited,
 		WaitFor:         nil,
 		RunFn:           nil,
 	}
@@ -100,7 +101,8 @@ func (c *Call) Twice() *Call {
 }
 
 // Times indicates that that the mock should only return the indicated number
-// of times.
+// of times.  The argument can be zero to indicate that the mock should not be
+// called at all.
 //
 //    Mock.On("MyMethod", arg1, arg2).Return(returnArg1, returnArg2).Times(5)
 func (c *Call) Times(i int) *Call {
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
Hi @vladlosev,

Thanks for the PR. Unfortunately, this change will break backwards compatibility since it changes the behaviour of `Call.Repeatibility`, which will break the code from any user's relying on the current meaning of `0`.

There's a big list of packages depending on `testify`, so we must keep changes backwards compatible.

As you saw making the changes. Setting `Repeatability` to `-1` currently means no calls are expected, which would make a test fail when called. Also, if a method is called that you didn't expect (had not set the `On`), the test fail due to an unexpected call.

vladlosev:
Hi @ernesto-jimenez,

I don't believe the change is breaking because specifying `.Times(0)` is always failing in the current code. If you specify it and make some matching calls, it panics when a call is received, as it should. But if you make no matching calls, it still fails with the following message:
```
--- FAIL: TestTimesZeroNoCalls (0.00s)
	mock.go:361: ❌	TestMethod()
	mock.go:375: FAIL: 0 out of 1 expectation(s) were met.
			at: []
FAIL
FAIL	_/Users/vlad/vsco/testify/mock	0.026s
```
(you can check out the test code [here](https://gist.github.com/vladlosev/fad989f51776c2061d72))

So this change should not be breaking any correct existing tests.

ernesto-jimenez:
@vladlosev it does break backwards compatibility.

The following test passes in current testify and fails with your changes:

```go
func TestTimesZeroNoCalls(t *testing.T) {
	var mockedService = new(TestImplementation)
	c := mockedService.On("TestMethod", 5, 6, 7).Return(0, nil).Times(0)
	mockedService.TestMethod(5, 6, 7)
	mockedService.AssertExpectations(t)
}
```

I agree your behaviour is more reasonable, but it unfortunately breaks away from the current exported and documented behaviour. You could argue that those tests are incorrect, but testify has been documented to work this way since 2013.

If and when we make a v2 of testify including breaking changes, we would definitely implement it the way you mention.

vladlosev:
Well, till v2, then.

serejja:
Hey maintainers, do you have some kind of a road map for v2? More than a year passed since this issue and it'd be great to know when this feature is available.

/cc @ernesto-jimenez

ernesto-jimenez:
@serejja we have to come up with a plan to release it. We have been hoping for usage of `dep` to become more prevalent to be able to do a smooth transition to `v2`.

PR #287, opened 2016-03-18
update require.Contains/NotContains comment

mhamrle:


ernesto-jimenez:
@mhamrle the PR is against an old version of testify.

PR #291, opened 2016-04-07
Update MatchedBy example to include actual use of MatchedBy!

willogden:
Small change to comments to include correct usage of MatchedBy


ernesto-jimenez:
@willogden thanks! :)

PR #298, opened 2016-04-17
assert: Add Is{,Not}AssignableTo

anacrolix:


ernesto-jimenez:
@anacrolix quick question: how would you be using the assertion?

The examples in the test could be checked by the compiler. Why would you do it during runtime using a test?

Thanks!
Ernesto

anacrolix:
My initial reason was when I don't have a concrete type to assert the assigned value to. It's an arbitrary type that is required to be assignable. For example, the result of reflect.StructOf.

I also expect that some codegen command needs to be run to dupe the new functions to the require package?

ernesto-jimenez:
> My initial reason was when I don't have a concrete type to assert the assigned value to. It's an arbitrary type that is required to be assignable. For example, the result of reflect.StructOf.

Do you have an example? The example looks a bit weird when saying one value is assignable to another value.

I tend to do more codegen than reflection, so usually work with concrete types :)

> I also expect that some codegen command needs to be run to dupe the new functions to the require package?

yes, `go generate ./...` will generate all the extra functions and methods.

Thanks!
Ernesto

anacrolix:
Whatever use case I had for this has gone. I think because I've had to revert to using code that doesn't make use of Go 1.7's StructOf until it's officially released.

PR #299, opened 2016-04-18
Added Same() and NotSame() assertions.

Monnoroch:
They are similar to Exactly() and NotEqual(), but test equality by reference instead of by value.

### DIFF
--- a/assert/assertion_forward.go
+++ b/assert/assertion_forward.go
@@ -332,7 +332,7 @@ func (a *Assertions) NotRegexp(rx interface{}, str interface{}, msgAndArgs ...in
 }
 
 
-// Equal asserts that two objects are actually the same one.
+// NotSame asserts that two objects are not the same one.
 // 
 //    a.NotSame([]int{123}, []int{123}, "should be different objects")
 // 
@@ -371,7 +371,7 @@ func (a *Assertions) Regexp(rx interface{}, str interface{}, msgAndArgs ...inter
 }
 
 
-// Equal asserts that two objects are actually the same one.
+// Same asserts that two objects are actually the same one.
 // 
 //    obj := []int{123}
 //    a.Same(obj, obj, "should be the same object")
--- a/assert/assertion_forward.go
+++ b/assert/assertion_forward.go
@@ -332,6 +332,16 @@ func (a *Assertions) NotRegexp(rx interface{}, str interface{}, msgAndArgs ...in
 }
 
 
+// Equal asserts that two objects are actually the same one.
+// 
+//    a.NotSame([]int{123}, []int{123}, "should be different objects")
+// 
+// Returns whether the assertion was successful (true) or not (false).
+func (a *Assertions) NotSame(expected interface{}, actual interface{}, msgAndArgs ...interface{}) bool {
+	return NotSame(a.t, expected, actual, msgAndArgs...)
+}
+
+
 // NotZero asserts that i is not the zero value for its type and returns the truth.
 func (a *Assertions) NotZero(i interface{}, msgAndArgs ...interface{}) bool {
 	return NotZero(a.t, i, msgAndArgs...)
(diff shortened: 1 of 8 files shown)


assert/assertions.go
```diff
@@ -253,6 +253,51 @@ func IsType(t TestingT, expectedType interface{}, object interface{}, msgAndArgs
 	return true
 }
 
+// Equal asserts that two objects are actually the same one.
```
exarkun: Should say "Same asserts ..."

assert/assertions.go
```diff
@@ -253,6 +253,51 @@ func IsType(t TestingT, expectedType interface{}, object interface{}, msgAndArgs
...
+
+	if expected != actual {
+		return Fail(t, fmt.Sprintf("Not the same: %#v (expected)\n"+
+			"        != %#v (actual)%s", expected, actual), msgAndArgs...)
+	}
+
+	return true
+}
+
+// Equal asserts that two objects are actually the same one.
```
exarkun: Similar comment here.

assert/assertions_test.go
```diff
@@ -94,6 +94,11 @@ func (a *AssertionTesterConformingObject) TestMethod() {
 type AssertionTesterNonConformingObject struct {
 }
 
+// AssertionTesterConformingObject is an object that conforms to the AssertionTesterInterface interface
```
exarkun: "samenessTestHelper is ..."

require/require.go
```diff
@@ -397,6 +397,18 @@ func NotRegexp(t TestingT, rx interface{}, str interface{}, msgAndArgs ...interf
 }
 
 
+// Equal asserts that two objects are actually the same one.
```
exarkun: "NotSame ..."

require/require.go
```diff
@@ -432,6 +444,19 @@ func Regexp(t TestingT, rx interface{}, str interface{}, msgAndArgs ...interface
...
```
exarkun: "Same ..."

require/require_forward.go
```diff
@@ -333,6 +333,16 @@ func (a *Assertions) NotRegexp(rx interface{}, str interface{}, msgAndArgs ...in
 }
 
 
+// Equal asserts that two objects are actually the same one.
```
exarkun: "NotSame ..."

require/require_forward.go
```diff
@@ -362,6 +372,17 @@ func (a *Assertions) Regexp(rx interface{}, str interface{}, msgAndArgs ...inter
...
```
exarkun: "Same ..."

assert/assertion_forward.go
```diff
@@ -332,6 +332,16 @@ func (a *Assertions) NotRegexp(rx interface{}, str interface{}, msgAndArgs ...in
 }
 
 
+// Equal asserts that two objects are actually the same one.
```
exarkun: "NotSame ..."

assert/assertion_forward.go
```diff
@@ -361,6 +371,17 @@ func (a *Assertions) Regexp(rx interface{}, str interface{}, msgAndArgs ...inter
...
```
exarkun: "Same ..."

assert/assertions.go
Monnoroch: Fixed.

assert/assertions.go
Monnoroch: Fixed.

assert/assertions_test.go
Monnoroch: Fixed.

require/require.go
Monnoroch: Fixed.

require/require.go
Monnoroch: Fixed.

require/require_forward.go
Monnoroch: Fixed.

require/require_forward.go
Monnoroch: Fixed.

assert/assertion_forward.go
Monnoroch: Fixed.

assert/assertion_forward.go
Monnoroch: Fixed.

Monnoroch:
Thanks!
Fixed all the comments.

ernesto-jimenez:
@Monnoroch sorry it took so long to reply!

Would it be better to do the following?

```go
assert.True(t, expected == actual)
// or
assert.True(t, expected == actual, "%#v should be %#v", expected, actual)
```

`assert.True` would have the added benefit that a type mismatch would be detected by the compiler instead of on runtime using reflection.

Monnoroch:
Fair enough. The only downside is writing the error message yourself.

PR #302, opened 2016-04-23
Add Before and After assertions to Testify Assert

andyhaskell:
One possible concern with these assertions is the order of the target and actual times. In assertions like `Equal`, the target value comes first. I chose to put the actual value first in the `Before` and `After` assertions because it feels like a more natural syntax for comparisons

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -666,6 +666,34 @@ func NotPanics(t TestingT, f PanicTestFunc, msgAndArgs ...interface{}) bool {
 	return true
 }
 
+// Before asserts that the actual time is before the expected time.
+//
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
@AndyHaskell sorry it took a while to review this. I've been crazy busy.

`expected` should be before `actual` to be consistent with most of the other assertions.

```go
assert.Before(t, time.Now(), inThePast) // <- testify's pattern
```

Based on the conventions in the package, reads as "assert inThePast is before now.

Your implementation:

```go
assert.Before(t, inThePast, time.Now())
```

For anybody using `testify` would read as "assert that now is before the past".

PR #320, opened 2016-06-30
fix assertion message formatting

mhamrle:
Fail and HTTPBodyNotContains expected that Fail formats messages.


ernesto-jimenez:
@mhamrle those methods are in need of a good refactor :)

I'm merging because of the change to `Error` :)

ernesto-jimenez:
thanks!

PR #324, opened 2016-07-05
Include type information in output

### LINKED ISSUE #322: assert.Equal and similar output should be explicit about type inequality
In `assert.Equal(t, a, b)` if `a` and `b` do not have the same type, the test will fail, but the output will be the same as a value equality test failure.

It is confusing, because on complex equality assertion, ones might not realize the problem is different type, not unequal values. 

Consider this example:

```
type struct Foo { V int }
type Foos []Foo
```

```
assert.Equal(t, Foos{{1}, {2}}, Foos{{1}, {2})
```

compared to:

```
assert.Equal(t, Foos{{1}, {2}}, []Foo{{1}, {2})
```

The second case will fail with a "not equal" message.

Instead, `assert.Equal` and similar should output an explicit message about type inequality.

cc @abhinav @prashantv @breely

jredville:
Modifies the output of (assert|require).Equal to include the type of the variables being compared

Fixes #322

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -262,8 +262,8 @@ func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) 
 
 	if !ObjectsAreEqual(expected, actual) {
 		diff := diff(expected, actual)
-		return Fail(t, fmt.Sprintf("Not equal: %#v (%[1]T (expected)\n"+
-			"        != %#v %[2]T (actual)%s", expected, actual, diff), msgAndArgs...)
+		return Fail(t, fmt.Sprintf("Not equal: %#v (%[1]T) (expected)\n"+
+			"        != %#v (%[2]T) (actual)%s", expected, actual, diff), msgAndArgs...)
 	}
 
 	return true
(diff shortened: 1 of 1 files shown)


assert/assertions.go
```diff
@@ -262,8 +262,8 @@ func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{})
 
 	if !ObjectsAreEqual(expected, actual) {
 		diff := diff(expected, actual)
-		return Fail(t, fmt.Sprintf("Not equal: %#v (expected)\n"+
-			"        != %#v (actual)%s", expected, actual, diff), msgAndArgs...)
+		return Fail(t, fmt.Sprintf("Not equal: %#v (%[1]T (expected)\n"+
```
mhamrle: There is one extra parenthesis. Should be
```
"Not equal: %#v %[1]T (expected)\n"
```

jredville: Fixed in 924cf70. I actually wanted the type info in parentheses, so I added the missing ones

ernesto-jimenez:
Thanks @jredville,

This is a duplicate of what #327 is solving, so closing this one :)

PR #326, opened 2016-07-06
Fix CallerInfo in go-routine

mhamrle:
CallerIfo returned empty array if it is called from go-routine.


ernesto-jimenez:
@mhamrle thanks, this has been fixed in #343.

PR #327, opened 2016-07-07
Clarify assert.Equal failure message given mismatched numeric types

### LINKED ISSUE #155: Return a proper error message when we have type mismatching
It took a lot of time to understand this broken test
```
--- FAIL: TestClickEvent (0.00s)
	Location:	click_test.go:70
	Error:		Not equal: 2 (expected)
			        != 2 (actual)
```

and the reason was that those values had different types: `int32` and `int`.

Would be really good to return a proper error message.

jveski:
Resolves #155.

This change causes type names to be printed when the failure message would otherwise be ambiguous. Specifically, it addresses equality assertions on numeric values of unlike types.

### Example
_Before:_
```go
assert.Equal(t, int64(123), int32(123))
//
// Not equal: 123 (expected)
//       != 123 (actual)
```

_After:_
```go
assert.Equal(t, int64(123), int32(123))
//
// Not equal: int64(123) (expected)
//      != int32(123) (actual)
```

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -262,12 +262,51 @@ func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) 
 
 	if !ObjectsAreEqual(expected, actual) {
 		diff := diff(expected, actual)
-		return Fail(t, fmt.Sprintf("Not equal: %#v (expected)\n"+
-			"        != %#v (actual)%s", expected, actual, diff), msgAndArgs...)
+		expected, actual = formatUnequalValues(expected, actual)
+		return Fail(t, fmt.Sprintf("Not equal: %v (expected)\n"+
+			"        != %v (actual)%s", expected, actual, diff), msgAndArgs...)
 	}
 
 	return true
 
+}
+
+// formatUnequalValues takes two values of arbitrary types and returns string
+// representations appropriate to be presented to the user.
+//
+// If the values are not of like type, the returned strings will be prefixed
+// with the type name, and the value will be enclosed in parenthesis similar
+// to a type conversion in the Go grammar.
+func formatUnequalValues(expected, actual interface{}) (e string, a string) {
+	aType := reflect.TypeOf(expected)
+	bType := reflect.TypeOf(actual)
+
+	if aType != bType && isNumericType(aType) && isNumericType(bType) {
+		return fmt.Sprintf("%v(%#v)", aType, expected),
+			fmt.Sprintf("%v(%#v)", bType, actual)
+	}
+
+	return fmt.Sprintf("%#v", expected),
+		fmt.Sprintf("%#v", actual)
+}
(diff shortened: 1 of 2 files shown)


assert/assertions.go
```diff
@@ -262,14 +262,53 @@ func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{})
...
+	"uint":    true,
+	"uint8":   true,
+	"uint16":  true,
+	"uint32":  true,
+	"uint64":  true,
+}
+
+func isNumericType(t reflect.Type) bool {
+	return numericTypes[t.String()]
+}
```
ernesto-jimenez: Instead of this, you can do:

```go
func isNumericType(t reflect.Type) bool {
	switch t.Kind() {
	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
		return true
	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
		return true
	case reflect.Float32, reflect.Float64:
		return true 
	}
	return false
}
```

assert/assertions_test.go
```diff
@@ -195,6 +195,20 @@ func TestEqual(t *testing.T) {
...
+	Equal(t, `"bar"`, actual, "value should not include type")
+
+	expected, actual = formatUnequalValues(123, 123)
+	Equal(t, `123`, expected, "value should not include type")
+	Equal(t, `123`, actual, "value should not include type")
+
+	expected, actual = formatUnequalValues(int64(123), int32(123))
+	Equal(t, `int64(123)`, expected, "value should include type")
+	Equal(t, `int32(123)`, actual, "value should include type")
+}
```
ernesto-jimenez: Include a check for a struct.

assert/assertions.go
```diff
@@ -262,14 +262,53 @@ func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{})
 
 	if !ObjectsAreEqual(expected, actual) {
 		diff := diff(expected, actual)
-		return Fail(t, fmt.Sprintf("Not equal: %#v (expected)\n"+
-			"        != %#v (actual)%s", expected, actual, diff), msgAndArgs...)
+		expected, actual = formatUnequalValues(expected, actual)
+		return Fail(t, fmt.Sprintf("Not equal: %v (expected)\n"+
+			"        != %v (actual)%s", expected, actual, diff), msgAndArgs...)
```
ernesto-jimenez: since `expected` and `actual` are going to be strings, we can use `%s` in Sprintf

jveski:
@ernesto-jimenez no worries on the delay and I just pushed three commits with the changes you requested.

btw thanks for maintaining this lib. It's really useful, and I know being a maintainer of projects like this can be rough, so thanks. 👍

PR #332, opened 2016-07-21
add >, <, >= and <= assertions

pytlesk4:
Title says it all.

### DIFF
--- a/assert/assertion_forward.go
+++ b/assert/assertion_forward.go
@@ -123,6 +123,18 @@ func (a *Assertions) False(value bool, msgAndArgs ...interface{}) bool {
 }
 
 
+// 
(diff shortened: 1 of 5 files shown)


pytlesk4:
What are the chances of this being accepted?

ernesto-jimenez:
@pytlesk4 sorry for the wait, I've had a crazy busy summer.

I'm a bit conflicted with these assertions :)

I would normally do:

```go
assert.True(t, v <= 12)
assert.True(t, v <= 12, "%d should be <= 12", v)
```

The big upside being that the compiler will already be doing all the type checking for us.

```go
assert.LessThanOrEqual(t, 12, v)
```

Issue #230, opened 2015-11-01
Pass `golint`


ernesto-jimenez:
There's tons of lint warnings to be addressed.

ernesto-jimenez:
PR #266 covers all but two lint warnings:

  - `assert.AnError` is kept unrenamed to keep changes backwards compatible.
  - `mock.argumentMatcher` is unexported for the time being.

PR #333, opened 2016-07-27
Export mock.argumentMatcher

connor4312:
This allows the creation of predicate-generating functions, which can make code a lot cleaner. To give a very simple example, this is not possible with the current non-exported argumentMatcher:

``` go
type Counter struct{ x int }

func counterArg(expectedX int) mock.ArgumentMatcher {
    return mock.MatchedBy(func(c *Counter) bool {
        return c.x == expectedX
    })
}

mock.On("Method", counterArg(1))
mock.On("Method", counterArg(2))
mock.On("Method", counterArg(3))
```

Also fixes the outstanding lint issues from #230

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -472,14 +472,14 @@ func AnythingOfType(t string) AnythingOfTypeArgument {
 	return AnythingOfTypeArgument(t)
 }
 
-// argumentMatcher performs custom argument matching, returning whether or
+// ArgumentMatcher performs custom argument matching, returning whether or
 // not the argument is matched by the expectation fixture function.
-type argumentMatcher struct {
+type ArgumentMatcher struct {
 	// fn is a function which accepts one argument, and returns a bool.
 	fn reflect.Value
 }
 
-func (f argumentMatcher) Matches(argument interface{}) bool {
+func (f ArgumentMatcher) Matches(argument interface{}) bool {
 	expectType := f.fn.Type().In(0)
 
 	if reflect.TypeOf(argument).AssignableTo(expectType) {
@@ -489,7 +489,7 @@ func (f argumentMatcher) Matches(argument interface{}) bool {
 	return false
 }
 
-func (f argumentMatcher) String() string {
+func (f ArgumentMatcher) String() string {
 	return fmt.Sprintf("func(%s) bool", f.fn.Type().In(0).Name())
 }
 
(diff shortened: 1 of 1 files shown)


ernesto-jimenez:
@connor4312 sorry for the late reply. It has been a crazy summer :)

I would rather avoid exporting an struct that cannot be properly initialised.

Instead, it would be better to define an `ArgumentMatcher` interface.

The tricky part is that we must ensure the interface is never satisfied by a struct from some other package, otherwise we might have weird collisions and we would not be able to add anything to the interface without breaking backwards compatibility.

The idea would be to export

```go
type ArgumentMatcher interface {
    Matches(interface{}) bool
    String()
    // This method ensures only `mock` package can implement ArgmentMatchers
    private()
}
```

Then, we include:

```go
func (argumentMatcher) private() { }
```

`MatchedBy` would return `ArgumentMatcher`, and line 570 will assert for `ArgumentMatcher` instead of `argumentMatcher`.

With these changes, you could define functions like yours and the package would be much safer to use.

What do you think?

Do you want to make those changes?

connor4312:
Hm, I see the issue, but to me it seems like that might be a bit of an abuse of the type system. Perhaps a struct or function type which encapsulates a predicate.

```go
// An ArgumentMatcher takes a function argument and returns whether it
// matches the Call's desired signature
type ArgumentMatcher func(v interface{}) bool
```

To be a bit more ambitious, have `On` convert its `args ...interface{}` into ArgumentMatchers (for those which are not already matchers) when the expectation is initially created. This isolates the loosely-typed code into a smaller surface area, which is always a win in my book.

Additionally, some existing code like `AnythingOfTypeArgument` and the magic `Anything` string could be expressed via this generic matchers.

ernesto-jimenez:
My point is that I would rather lock-in this feature so only `mock` can create custom argument matchers for the time being.

Adding a private method to an exported interface is not an abuse of the type system, it is a safeguard to be able to export an interface only your package can satisfy so you are free to modify it without breaking backwards compatibility. `testing.TB` is an example of this pattern from the standard library.

Thinking more about this, I would probably do:

```go
type ArgumentMatcher interface {
    ArgumentMatches(interface{}) error
    // This method ensures only `mock` package can implement ArgmentMatchers
    private()
}
```

And refactor the current `argumentMatcher` to fit the new interface. Which we can do without breaking backwards compatibility due the small surface.

With that change, not only we can definitely `AnythingOfTypeArgument` and `Anything` to return private structs that implement `ArgumentMatcher`.

PR #335, opened 2016-08-12
Mock: Add diff to output when no exact match found

dominicbarnes:
I use this library a lot, but even still I find the output generated when the expected arguments don't match to be incredibly difficult to use when you are dealing with non-trivial values like structs.

This PR adds one more section to the output, it will find the first argument that has any differences and print a diff. (logic lifted directly from the `assert` package) This goes a long way in making it easier to quickly identify differences and iterate.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
+	"github.com/pmezard/go-difflib/difflib"
 	"github.com/stretchr/objx"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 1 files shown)


ernesto-jimenez:
Hey @dominicbarnes, sorry it took me so long to get to your PR. I've been crazy busy this summer.

Can you post the output before and after the changes?

dominicbarnes:
Here's a really simple example, so it probably doesn't show off the usefulness as well as a more complicated struct. (but I think it demonstrates the change pretty clearly)

Before:

```
mock: Unexpected Method Call
-----------------------------

Test_Mock_AssertExpectations_With_Repeatability(map[string]string)
		0: map[string]string{"b":"B"}

The closest call I have is: 

Test_Mock_AssertExpectations_With_Repeatability(map[string]string)
		0: map[string]string{"a":"A"}
```

After:

```
mock: Unexpected Method Call
-----------------------------

Test_Mock_AssertExpectations_With_Repeatability(map[string]string)
		0: map[string]string{"b":"B"}

The closest call I have is: 

Test_Mock_AssertExpectations_With_Repeatability(map[string]string)
		0: map[string]string{"a":"A"}

Difference found in argument 0:

--- Expected
+++ Actual
@@ -1,3 +1,3 @@
 (map[string]string) (len=1) {
- (string) (len=1) "b": (string) (len=1) "B"
+ (string) (len=1) "a": (string) (len=1) "A"
 }
```

PR #338, opened 2016-08-19
Changing format of NoError output to make use of errors

nyarly:
While %+v shouldn't significantly change the formatting of most errors, it prints a stacktrace for errors that have been created or wrapped with pkgs/errors, which is extremely helpful for debugging.


nyarly:
@ernesto-jimenez Could I get some feedback?

ernesto-jimenez:
@nyarly sorry, I've been travelling and very busy :)

Sounds good. Merging.

PR #343, opened 2016-09-06
Fix #341

### LINKED ISSUE #341: No stack trace for failures in subtest
When using the new [subtest](https://tip.golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks) feature in Go 1.7, the stack trace is not properly included in the output when there's a failure. Here's test output with failure in a normal test (first) and the same failure as a subtest (second). At the same time, `debug.PrintStack()` does correctly print the stack.

```Text
goroutine 18 [running]:
runtime/debug.Stack(0x0, 0x0, 0x0)
        C:/Go/src/runtime/debug/stack.go:24 +0x80
runtime/debug.PrintStack()
        C:/Go/src/runtime/debug/stack.go:16 +0x29
github.com/dcormier/subtests.subtest(0xc04207e180)
        C:/Dev/Go/src/github.com/dcormier/subtests/main_test.go:19 +0x29
github.com/dcormier/subtests.TestWithoutSubtest(0xc04207e180)
        C:/Dev/Go/src/github.com/dcormier/subtests/main_test.go:11 +0x32
testing.tRunner(0xc04207e180, 0x68ded0)
        C:/Go/src/testing/testing.go:610 +0x88
created by testing.(*T).Run
        C:/Go/src/testing/testing.go:646 +0x2f3
--- FAIL: TestWithoutSubtest (0.00s)
        Error Trace:    main_test.go:21
                        main_test.go:11
        Error:          This is a failure

goroutine 20 [running]:
runtime/debug.Stack(0x0, 0x0, 0x0)
        C:/Go/src/runtime/debug/stack.go:24 +0x80
runtime/debug.PrintStack()
        C:/Go/src/runtime/debug/stack.go:16 +0x29
github.com/dcormier/subtests.subtest(0xc04207e3c0)
        C:/Dev/Go/src/github.com/dcormier/subtests/main_test.go:19 +0x29
testing.tRunner(0xc04207e3c0, 0x68ded8)
        C:/Go/src/testing/testing.go:610 +0x88
created by testing.(*T).Run
        C:/Go/src/testing/testing.go:646 +0x2f3
--- FAIL: TestWithSubtest (0.00s)
    --- FAIL: TestWithSubtest/This_is_a_subtest (0.00s)
        Error Trace:    5:
        Error:          This is a failure

FAIL
```

Notice that in the failure for "TestWithoutSubtest" the "Error Trace" is what one might expect, but in the "TestWithSubtest" failure the value for "Error Trace" is simply "5:".

[Here](https://github.com/stretchr/testify/files/448171/main_test.go.txt) is the very simple example showing this.

adamwg:
Go 1.7 subtests are called directly from testing.tRunner, not from the
test that contains the t.Run call. Because of this, the call stack
doesn't contain a function starting with Test, Benchmark, or Example,
causing assert.CallerInfo() to run off the end of the call stack and
return nil.

Look for testing.tRunner explicitly to solve this problem, and return
the list of callers if we do run off the call stack so that if there's a
similar problem in the future we still get an inflated stack trace
rather than no trace at all.


ernesto-jimenez:
@adamwg thanks for the patience, I'm back online and merging now :)

PR #349, opened 2016-09-14
Fix Error Trace with Subtest

yuroyoro:
In subtest supported from go1.7, `CallerInfo()` could not detect
correct funcname, flie and line.
Then no source location shown in 'Error Trace" field.

This is caused by `runtime.FuncForPC(pc)` returns funcname with
clolure indexis like "pkgname.FuncName.func1.2.3".

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -63,6 +63,8 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
 	return false
 }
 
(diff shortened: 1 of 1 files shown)


pkieltyka:
Yes, for some reason I can't see the source lines to assert errors either. I hope its possible. I will try to solve this one day too if someone doesn't do it before me.

ernesto-jimenez:
This should be addressed with #343, which was just merged.

If it's not solved, open an issue with an simple example and I'll take a look and fix it :)

Thanks!

PR #353, opened 2016-09-22
Run gofmt on assertion_forward.go

ChimeraCoder:
Fixes whitespace issues in `assertion_forward.go`. (All other files were already formatted).


ernesto-jimenez:
@ChimeraCoder those files are automatically generated by `_codegen`. These changes would be overridden next time we run `go generate` and the code is re-generated.

We can format the code within the `_codegen` program using `go/format`, if you want to do a PR for that :)

Issue #339, opened 2016-08-21
go1.7 vet flags "possible formatting directive" when Error is used


prashantv:
When the `assert.Error` method is used with formatting arguments, go vet complains as of 1.7:

```
flag_test.go:61: possible formatting directive in Error call
```

For more information and a repro, see:
https://github.com/uber-go/zap/pull/135

The specific check that is run is here:
https://golang.org/src/cmd/vet/print.go#L587

I'm not sure if there's a good fix for this, but I wanted to flag this, as I typically:
- recommend adding context to the `assert.Error` calls
- suggest usage of `assert.Error` instead of `assert.NotNil`

However, this "correct" usage of the library leading to a vet failure may mean users will either avoid one or the other.

ernesto-jimenez:
@prashantv this will require creating methods like `Errorf` and `Equalf` as counterparts of `Error` and `Equal` with formatted messages.

Let me look into it :)

ernesto-jimenez:
@prashantv what does https://github.com/stretchr/testify/pull/356 look like?

PR #356, opened 2016-09-24
Add `assertion`f assertions like Errorf and Equalf

ernesto-jimenez:
Related to #339

In Go 1.7, the govet will complain with the following code:

``` go
assert.Error(t, err, "test %d", i)
```

The complain is:

```
no formatting directive in Error call
```

The cause being the fact that `Error`'s signature ends with
`msgAndArgs ...interface{}` instead of `msg string, args ...interface{}`.

In this PR we add format assertions like `Equalf` and `Errorf`.

In a v2 of testify we would remove `msgAndArgs` from all the assertions and
have people use the format assertions if they want to customise the error.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -18,6 +18,8 @@ import (
 	"github.com/pmezard/go-difflib/difflib"
 )
 
+//go:generate go run ../_codegen/main.go -output-package=assert -template=assertion_format.go.tmpl
+
 // TestingT is an interface wrapper around *testing.T
 type TestingT interface {
 	Errorf(format string, args ...interface{})
@@ -268,7 +270,7 @@ func labeledOutput(content ...labeledContent) string {
 
 // Implements asserts that an object is implemented by the specified interface.
 //
-//    assert.Implements(t, (*MyInterface)(nil), new(MyObject), "MyObject")
+//    assert.Implements(t, (*MyInterface)(nil), new(MyObject))
 func Implements(t TestingT, interfaceObject interface{}, object interface{}, msgAndArgs ...interface{}) bool {
 
 	interfaceType := reflect.TypeOf(interfaceObject).Elem()
@@ -293,7 +295,7 @@ func IsType(t TestingT, expectedType interface{}, object interface{}, msgAndArgs
 
 // Equal asserts that two objects are equal.
 //
-//    assert.Equal(t, 123, 123, "123 and 123 should be equal")
+//    assert.Equal(t, 123, 123)
 //
 // Returns whether the assertion was successful (true) or not (false).
 //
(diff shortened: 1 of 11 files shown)


assert/assertions.go
```diff
@@ -246,7 +248,7 @@ func Fail(t TestingT, failureMessage string, msgAndArgs ...interface{}) bool {
 
 // Implements asserts that an object is implemented by the specified interface.
 //
-//    assert.Implements(t, (*MyInterface)(nil), new(MyObject), "MyObject")
```
prashantv: Why are the messages removed in the examples? I'd expect the methods without the `f` suffix to still accept arguments which are appended similar to how `fmt.Println` does.

assert/assertions_test.go
```diff
@@ -555,7 +555,7 @@ func TestNoError(t *testing.T) {
 	}()
 
 	if err == nil { // err is not nil here!
-		t.Errorf("Error should be nil due to empty interface", err)
+		t.Errorf("Error should be nil due to empty interface: %s", err)
```
prashantv: nit: should use `%v` for types that aren't bytes or strings
same below at line 595

require/require.go
```diff
@@ -33,6 +40,20 @@ func Contains(t TestingT, s interface{}, contains interface{}, msgAndArgs ...int
 	}
 }
 
+// Containsf asserts that the specified string, list(array, slice...) or map contains the
+// specified substring or element.
+//
+//    assert.Containsf(t, "Hello World", "World", "extra error message")
```
prashantv: The examples for `Containsf` should contain a formatting string + arguments. Otherwise `Contains` is fine.

assert/assertions_test.go
ernesto-jimenez: This is part of our internal tests, so this should be fine.

assert/assertions.go
ernesto-jimenez: We cannot do it for this version, since it would be a breaking change, but I'm thinking that testify's assertions provide plenty of context around the failure.

I a v2 of testify, we would probably have assertions not take any message at all, and offer the "format" assertions like `Errorf` to those who want to provide some context.

This has some extra benefits, since people right now make mistakes like this:

```go
assert.Error(t, err, "expected error message")
```

Because they believe that would check that `err.Error() == "expected error message"` and the code compiles and runs.

require/require.go
ernesto-jimenez: This is part of the code-generated comments.

`, "extra error message"` is added to the end of usage examples on "formatted assertions".

Would need to figure out what we can add instead that would be a relevant non-confusing message+variable that works for any assertion.

assert/assertions.go
prashantv: Sorry, I don't mean change the interface, just the examples should suggest formatting.

assert/assertions_test.go
prashantv: Yep, this will work but the `%s` format specified is not intended to be used for arbitrary types:

From https://golang.org/pkg/fmt/:
```
%s	the uninterpreted bytes of the string or slice
```

Whereas we really want `%v`:
```
%v	the value in a default format
```

ains:
Any update on getting this merged in @ernesto-jimenez?

mohit:
Any updates. Seems like this is still required.

PR #361, opened 2016-10-22
Add GreaterThan and SmallerThan assertions

lccezinha:
Add this to make easier to assert over numerical values and when is needed to verify which value is greater/smaller than other one.

For some time I had to use assertions like this:

``` go
package main

import "testing"
import "github.com/stretchr/testify/assert"

func TestGreaterThan(t *testing.T) {
  firstVallue, secondVallue := 1, 10

  assert.True(t, firstValue > secondValue)
  assert.False(t, firstValue > secondValue)

  // OR 

  assert.True(t, firstValue < secondValue)
  assert.False(t, firstValue < secondValue)
} 
```

So my idea is to make this easier with something like that:

``` go
package main

import "testing"
import "github.com/stretchr/testify/assert"

func TestGreaterThan(t *testing.T) {
  firstVallue, secondVallue := 1, 10

  assert.GreaterThan(t, firstValue, secondValue)
  assert.GreaterThan(t, firstValue, secondValue)

  // OR 

  assert.SmallerThan(t, firstValue, secondValue)
  assert.SmallerThan(t, firstValue, secondValue)
}
```

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1058,3 +1058,51 @@ func diff(expected interface{}, actual interface{}) string {
 
 	return "\n\nDiff:\n" + diff
 }
+
+// GreaterThan asserts that first numerical value is greater than second numerical value.
+//
+//    assert.GreaterThan(t, firstNumericalValue, secondNumericalValue, "First value must be greater than second value")
+//
+// Returns whether the assertion was successful (true) or not (false).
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
@lccezinha thanks.

We already had a PR for this back on #332. In my opinion what you have been doing is better since it allows the compiler to take care of all the type checking on build time.

lccezinha:
@ernesto-jimenez Thanks for you feedback.

PR #364, opened 2016-10-29
Ensure that assert.Fail properly align its output

### LINKED ISSUE #83: Multilines don't play nicely with vim-go
I also opened the issue over there, but figured I'd additionally post here.

https://github.com/fatih/vim-go/issues/168

I attached two screenshots. One shows how `go test` renders in a terminal, and the other when using `:GoTest` within vim. I'm using testify mostly for assert functionality.

I chatted with fatih a bit in #go-nuts, and it was said that vim-go has pretty basic support for multilines. 

screenshot 2014-09-09 19 46 45
screenshot 2014-09-09 19 47 40

nmiyake:
Previous implementation depended on tab alignment. This worked
when the output was not prepended with anything, but would break
if the output was prepended with further spaces (which can occur
in environments like IntelliJ test runners). This commit fixes it
so that the output is always aligned logically.

Fixes #83

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -157,7 +157,7 @@ func getWhitespaceString() string {
 	parts := strings.Split(file, "/")
 	file = parts[len(parts)-1]
 
-	return strings.Repeat(" ", len(fmt.Sprintf("%s:%d:      ", file, line)))
+	return strings.Repeat(" ", len(fmt.Sprintf("%s:%d:        ", file, line)))
 
 }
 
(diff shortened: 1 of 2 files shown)


nmiyake:
Example output before -- first output is running in IntelliJ, second directly in terminal:

image

Example output after -- first output is running in IntelliJ, second directly in terminal:

image

nmiyake:
Note that this PR also makes it such that the "Messages" output is aligned, which is a behavior change. I believe this behavior is more conceptually correct and looks nicer, but that can be done as a separate discussion/PR if there's contention around that (whereas making sure the alignment is consistent between terminal and runner output is indisputably a bug that should be fixed).

@ernesto-jimenez (or any other project maintainers), please let me know if you have any feedback.

nmiyake:
@ernesto-jimenez any feedback? This PR is getting fairly stale and would like to know whether it can move forward (or if it needs to be modified).

ernesto-jimenez:
@nmiyake finally had some time to review it.

Looks good, merging it now :)

PR #367, opened 2016-11-15
Remove isNumericType check

### LINKED ISSUE #366: Assert panics when comparing 1 to nil
I have an example [here](https://gist.github.com/packrat386/4f276d5b47ad4a3f23d0e4cf333f0cce)

It looks like calling [Kind()](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L311) on the type returned by nil doesn't work.

packrat386:
I'm not sure why this check is needed. It seems worthwhile to print the
type any time the types aren't equal. Closes #366.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -295,29 +295,13 @@ func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) 
 // with the type name, and the value will be enclosed in parenthesis similar
 // to a type conversion in the Go grammar.
 func formatUnequalValues(expected, actual interface{}) (e string, a string) {
-	aType := reflect.TypeOf(expected)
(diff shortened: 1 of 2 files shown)


packrat386:
I made this PR as what I saw to the be the simplest way to address #366, but I'm open to other suggestions if we want to keep the old behaviour.

ernesto-jimenez:
thanks @packrat386

You can also remove `aType` and `bType`. Then we can change Sprintf to be something like: `fmt.Sprintf("%T(%+v)", expected, expected)`

packrat386:
@ernesto-jimenez is the current version what you were thinking?

ernesto-jimenez:
@packrat386 looks, good, thanks!

And sorry for the delay reviewing. I've been traveling a lot.

Issue #187, opened 2015-06-23
Running Parallel Tests in Suite


eyphka:
When running multiple tests in a suite which have been set to suite.t().Parallel(), tests will pass even if they should fail.

dwlnetnl:
That's true, but as it's currently structured it isn't possible to run suite tests in parallel. This is because before and after each suite test ```SetupTest``` and ```TearDownTest``` is run.

ernesto-jimenez:
@eyphka do you want to make a PR to address it? :)

ernesto-jimenez:
Related: https://github.com/stretchr/testify/issues/52

eyphka:
Sure

oryband:
Reviving this issue. Presumably it should be possible to run suite tests in parallel.

The problem with parallelism here is that the suite members used in tests have a shared single instance for all. Tests running in parallel could potentially access these members at the same time, causing them to "collide" with each other via these members.

However, my suggestion is that testify will add a special `Parallel()` flag to the suite. This new flag will instantiate new instance members for every new `Parallel()` test, calling `Setup/TearDownTest()` as well.

This will of course require extra work from the test developer, to make sure `Setup/TearDownTest()` are suitable for being called with/without parallel (As some tests will be run in parallel, and some don't).

Thoughts?

andrew-plunk:
I have a test suite runner here that supports testing.T.Parallel().
https://github.com/andrew-plunk/testify/pull/1/commits/d3e6d555c9d4be38fccab470b9bb0ee4b8537ead

The problem with supporting parallel execution on Suite is that the TestingSuite's public interface T() returns a *testing.T. It is very hard to make this goroutine safe. Rather than wrapping access to the returned *testing.T, I went down the path of removing the dependencies on the Suite struct's internal state, and simply pass the internal tests' *testing.T in via function arguments.

This removes the ability to call Assert() on the testing suite, but is a minior inconvenience IMO because the assert package supports passing a *testing.T object into it's functions.

Let me know how you all feel about this strategy. I will work on cleaning up that branch and adding docs/tests for a pr if it works for you.

dansimau:
I just whipped up a slightly different approach to running tests in parallel:
https://gist.github.com/dansimau/128826e692d7834eb594bb7fd41d2926

Similar to #369, it introduces a `RunParallel`, except that it uses reflection to create a new instance of the suite for each test run.

[...]

PR #369, opened 2016-11-18
Add RunParallel for running suites with t.Parallel

andrew-plunk:
Hello. My name is Andrew and I love using testify.

The purpose of this pr is to add a test suite runner that enables running suites in parallel (addressing #187). 

I am making this pr for feedback on my general methodologies here before writing more tests.

The general idea is to limit the amount of state exposed by the TestingSuite interface and to synchronize access to shared data via the Setup* and TearDown* interfaces.

RunParallel differs from Run in a few ways. It does not operate on the TestingSuite interface because it is very difficult to provide goroutine saftey when TestingSuite#T() returns a pointer to a testing object. Instead we take a functional approach here by passing in the *testing.T object created in the InternalTest closure into each test method.

Internal data can be safely set here via the Setup* and TearDown* interface methods. We serialize access to the SetupTest and TearDownTest methods because they can be called in parallel. Mutating test state can be done via pointer receivers on these methods.

All test methods should accept value receivers so that they don't cause data races. Any shared state those tests need can be safely setup via the above interface methods.

All feedback is welcomed and if anyone can point me to docs on how to best contribute to this project, I would appreciate it.

### DIFF
--- a/suite/parallel_suite.go
+++ b/suite/parallel_suite.go
@@ -4,8 +4,9 @@ import (
 	"fmt"
 	"os"
 	"reflect"
-	"sync"
 	"testing"
+
+	"github.com/mitchellh/copystructure"
 )
 
 // RunParallel takes a testing suite and runs all of the tests attached
@@ -17,8 +18,6 @@ import (
 //    internalState int
 // }
 //
-// Setup and Teardown functions which need to modify test state can accept pointer receivers safely.
-// The calling of these functions is synchronized between the tests.
 // func (s *ParallelTestSuite) SetupTest() {
 //     s.internalState++
 // }
@@ -28,9 +27,8 @@ import (
 // }
 //
 // All test methods in the suite will be run in parallel.
-// All tests should accept value receivers to prevent data races in the test suite.
-// All tests must accept t *testing.T as their only argument.
-// func (s ParallelTestSuite) Test(t *testing.T) {
+// All test methods must accept t *testing.T as their only argument.
+// func (s *ParallelTestSuite) Test(t *testing.T) {
 //     // do something
 // }
 //
@@ -38,8 +36,7 @@ import (
 // func TestParallelSuite(t *testing.T) {
 // 	RunParallel(t, new(ParallelTestSuite))
 // }
-// RunParallel synchronizes calls to SetupTestSuite and TearDownTestSuite. Any data accessed outside
-// of those functions must be manually synchronized to be goroutine safe.
+// RunParallel creates a copy of the test suite for each test method to prevent sharing data.
 func RunParallel(t *testing.T, suite interface{}) {
 	if setupAllSuite, ok := suite.(SetupAllSuite); ok {
 		setupAllSuite.SetupSuite()
@@ -49,9 +46,6 @@ func RunParallel(t *testing.T, suite interface{}) {
 			tearDownAllSuite.TearDownSuite()
 		}
 	}()
-
-	// Locks for synchronizing access to SetupTest() and TearDownTest().
-	setupMtx, teardownMtx := sync.Mutex{}, sync.Mutex{}
 
 	methodFinder := reflect.TypeOf(suite)
 	tests := []testing.InternalTest{}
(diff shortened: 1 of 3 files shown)


suite/parallel_suite.go
```diff
@@ -0,0 +1,100 @@
...
+					defer func() {
+						if tearDownTestSuite, ok := suite.(TearDownTestSuite); ok {
+							teardownMtx.Lock()
+							tearDownTestSuite.TearDownTest()
+							// This will deadlock if TearDownTest() panics.
+							teardownMtx.Unlock()
+						}
+					}()
+
+					method.Func.Call([]reflect.Value{reflect.ValueOf(suite), reflect.ValueOf(t)})
```
dustinlindquist: I believe there is still a race condition here. Because you release the setup Lock above and then kick this test off. Where another Setup test method could then grab that lock and modify the state this test depended on before it kicks off. See comment above.

suite/parallel_suite.go
```diff
@@ -0,0 +1,100 @@
...
+				Name: method.Name,
+				F: func(t *testing.T) {
+					t.Parallel()
+					setupTestSuite, ok := suite.(SetupTestSuite)
+					if ok {
+						setupMtx.Lock()
+						setupTestSuite.SetupTest()
+						// This will deadlock if SetupTest() panics.
+						// We cannot defer unlocking here because that would seralize running the tests.
+						setupMtx.Unlock()
```
dustinlindquist: Before you release the lock I think you should begin the test with the valueOf(suite). That way the copy of the state you just set up is passed to the test before the lock is released, allowing that state to change.

andrew-plunk: :+1:

andrew-plunk:
This pr was updated to use a deep copy of the suite (rather than locks) when passing the suite a a receiver to test methods. This should solve any concurrency issues.

I would like feedback on the difference in the suite interface accepted in Run vs RunParallel.

I like having RunParallel test methods accept *testing.T as their only arg because, follows the pattern setup in the standard lib's [testing package](https://golang.org/pkg/testing/). If required, I have no problem making RunParallel accept a TestingSuite and calling SetT/T like Run().

dansimau:
> I would like feedback on the difference in the suite interface accepted in Run vs RunParallel.
> I like having RunParallel test methods accept *testing.T as their only arg because, follows the pattern setup in the standard lib's testing package

I also prefer this pattern, however, I don't know if it justifies making `suite.RunParallel` backwards incompatible with `suite.Run`.

If you change the signature then anyone who wants to use `suite.RunParallel` has to change the method signature of every single test. Whereas it is very easy to make this a one line change for people by not changing the method signature.

My vote would be to make this backwards compatible.

(Just had a thought though: could `suite.Run` and `suite.RunParallel` be modified to actually accept both types of method signatures? We're already using reflection here; is it possible to figure out which way to call or set `testing.T` using reflection?)

andrew-plunk:
> My vote would be to make this backwards compatible.

Agreed. I will make RunParallel compatible with Run.

> (Just had a thought though: could suite.Run and suite.RunParallel be modified to actually accept both types of method signatures? We're already using reflection here; is it possible to figure out which way to call or set testing.T using reflection?)

I can work on a pr today to make this happen.

andrew-plunk:
@dansimau It is backwards compatible now.

andrew-plunk:
@ernesto-jimenez if you get a chance I would appreciate some feedback on this pr.

arvenil:
Any news on this?

arvenil:
Hah, I just discovered that go 1.7 allows subtests, which basically makes `Suite` obsolete. Instead of creating complex mechanism for creating tests that share same `Suite` it is now possible to create subtests. https://tip.golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks
Parallel testing is also extremely easy with this method. Maybe this will be useful for someone.

varunrau:
are there any plans to get this across the finish line? 🙂

aplunk:
Feel free to close this out. I don't have access to the account that I used to open this pr anymore or I would.

PR #371, opened 2016-11-20
Add comments for Equal and NotEqual to clarify pointer comparison

### LINKED ISSUE #358: More clear docs for `Equal` and `NotEqual`
There's currently some confusion when people realise that `Equal` and `NotEqual` compare values. 

See https://github.com/stretchr/testify/issues/296

We need better documentation for those two assertions.

nmiyake:
Clarify that pointer equality is determined by equality of values
rather than memory address.

Fixes #358


ernesto-jimenez:
@nmiyake back reviewing after a few months of travelling.

Changes look good. Thanks!

PR #375, opened 2016-12-01
Improved error reporting of unexpected method calls

### LINKED ISSUE #373: Not specifying arguments to mock causes index out of range error
When mocking a function that takes arguments, not specifying those arguments causes the test to panic with following error: "runtime error: index out of range". [Example](https://gist.github.com/seppestas/a56882d8eb6ad3c5aca815b45b376ab7)

A message stating the mock was called with `closestCall.Arguments` instead of no arguments would be a lot nicer.

Not specifying the function arguments is an obvious mistake when the developer intends to write a test where the arguments can be anything. In this case `mock.Anything` should be used since not specifying the arguments means the mock is expected to be called without arguments. IMO this is not very clear though, the `Mock.On` method is way less descriptive than E.g [gomock's MockMatcher](https://godoc.org/github.com/golang/mock/gomock/mock_matcher), which requires to explicitly state the excepted arguments using ['MockMatcher.EXCEPT'](https://godoc.org/github.com/golang/mock/gomock/mock_matcher#MockMatcher.EXPECT).

Personally, I think setting the return value of a mocked function and checking it's arguments should be split, allowing to separate test set up and assertions more clearly.

seppestas:
Fixes #373. Panic now tells us the method was called with [] instead
of an obscure index out of range error.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -252,18 +252,16 @@ func (m *Mock) findClosestCall(method string, arguments ...interface{}) (bool, *
 	return true, closestCall
 }
 
-func callString(method string, arguments Arguments, includeArgumentValues bool) string {
-
-	var argValsString string
-	if includeArgumentValues {
-		var argVals []string
-		for argIndex, arg := range arguments {
-			argVals = append(argVals, fmt.Sprintf("%d: %#v", argIndex, arg))
+func callString(method string, arguments Arguments) string {
+	var argVals []string
+	for i, arg := range arguments {
+		if i > 0 {
+			argVals = append(argVals, ",")
 		}
-		argValsString = fmt.Sprintf("\n\t\t%s", strings.Join(argVals, "\n\t\t"))
+		argVals = append(argVals, fmt.Sprintf("%#v", arg))
 	}
 
-	return fmt.Sprintf("%s(%s)%s", method, arguments.String(), argValsString)
+	return fmt.Sprintf("%s(%s) with arguments %s", method, arguments.String(), argVals)
 }
 
 // Called tells the mock object that a method has been called, and gets an array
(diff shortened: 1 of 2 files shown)


mock/mock.go
```diff
@@ -252,18 +252,16 @@ func (m *Mock) findClosestCall(method string, arguments ...interface{}) (bool, *
...
-		for argIndex, arg := range arguments {
-			argVals = append(argVals, fmt.Sprintf("%d: %#v", argIndex, arg))
+func callString(method string, arguments Arguments) string {
+	var argVals []string
+	for i, arg := range arguments {
+		if i > 0 {
+			argVals = append(argVals, ",")
 		}
-		argValsString = fmt.Sprintf("\n\t\t%s", strings.Join(argVals, "\n\t\t"))
+		argVals = append(argVals, fmt.Sprintf("%#v", arg))
```
ernesto-jimenez: We should:
 - **Use `%s` instead of `%#v`**: it is pretty common methods to receive `[]byte` and `%#v` would print output like `[]byte{0x6f, 0x6b}`
 - **Split arguments in different lines**: it is common for people to receive big structs or long strings. Having all the arguments in a single line will affect readability.

ernesto-jimenez (changes requested):
@seppestas thank you very much for the contribution, the refactoring and changes to the messages are good, but we should make a couple of tweaks to have good readability when methods accept `[]byte` or are receiving big arguments.

mock/mock.go
seppestas: Mmh, I don't really agree. Depending on what the []byte contains, I'd actually preter it to be printed the way `%#v` prints it over printing it as a string. IMO, if you're using a []byte to hold text Go has to manipulate or print, you're doing something wrong. Relevant: https://blog.golang.org/strings. Also, what if the argument is a string nor []byte?

Splitting the arguments makes sense though.

seppestas: For me that argument doesn't make a lot of sense. I could argue most numbers used by developers are smaller than 1000, but that doesn't mean making something that can only handle numbers up to 999 is a good idea, even if it makes things "simpler" for "most" people.

I can come up with plenty of functions returning a []byte that do not contain plain text like:
- [`x509.CreateCertificate`](CreateCertificate)
- [`elf.section.Data`](https://golang.org/pkg/debug/elf/#Section.Data)
- Good ol' `io.Reader.read`, arguably the most versatile interace in golang, used for everything from [bzip2](https://golang.org/pkg/compress/bzip2/#NewReader) to [png](https://golang.org/pkg/image/png/#Decode), and yes, also for text.

Also, when I'd pass `[]byte{0x62, 0x61, 0x72}` to a function, and some error tells me I passed a string containing `bar`, I don't consider that expected or something I'd want. Let alone an error spewing out a bunch of random crapp when I passed binary data to a function expecting, and using []byte for this is perfectly valid for this, binary data.

As a last argument: the `callString` function show arguments to mocked functions, not return types. Any sane function doing something with text should take a `string` (or even better, a [`text.Scanner`](https://golang.org/pkg/text/scanner/)), not a []byte. So while `json.Marshal` might return a []byte, if you want to make a function that does something with the marshalled text, it should take a string, and the output of `json.Marshal` should be wrapped with a string().

Trust me, as someone having to deal with `char *` and `uint8_t *` on a daily basis, be glad Go has a special type for text.

stevenroose: I also use `[]byte` mostly for binary data over text. (There is `string` after all..)

However, the `%#v` notation is quite verbose. `"0x%x"` for hexadecimal might be a good compromise between the two.

seppestas: But the nice feature of `%v` is that it works for pretty much all types, not only `[]byte` or `string`.

I also think that using Go's default formatting for types is more consistent than trying to come up with new formats, and it makes for a more "expected" result, even if it's a bit verbose.

A good alternative might be to shorten the `argVals` so a big `[]byte` is not completely printed, but limited to e.g 60 or so characters. This would mean seeing something like:
```
Expected: []byte{0x66, 0x6f, 0x6f, 0x66, 0x6f, 0x6f...
Received: []byte{0x62, 0x61, 0x72, 0x62, 0x61, 0x72...
```

rcorre:
Any chance this will still be merged? I just spent a while trying to debug an `IndexOutOfRange` panic that happened because I forgot one argument in my `On` call.

dolmen:
Closing because more than 9 years old.

PR #376, opened 2016-12-02
descriptive assertion error when comparing string aliases

Charlesetc:
Hi

I think the current error message when comparing string aliases is a bit confusing:

```go
Error:          Not equal: 
expected: "hi there"
received: "hi there"
``` 
With this diff, it becomes:

```go
Error:          Not equal: 
expected: string("hi there")
received: main.MyType("hi there")
``` 

:pear: thanks for taking a look!


ernesto-jimenez:
@Charlesetc this has been addressed with another PR that was pending and we just merged.

If you update testify, the output should be the one you were specifying.

Thank you!

PR #382, opened 2016-12-24
add Subset and NotSubset assertions

GlenDC:
Currently it isn't possible to test if a list is a (perfect) subset of another list. On top of that it is also not possible to see if a list is equal to another list when the order is not known.

This PR makes all of this possible.


GlenDC:
@ernesto-jimenez rebased master into this PR's branch as requested :)

ernesto-jimenez:
@GlenDC merged, thanks!

PR #384, opened 2017-01-05
Add `Mock#Reset()`

jonnyreeves:
Provide a methods for clearing all programed expectations from a mock; useful during a TestSuite tear-down method to ensure you have a clean slate between test-runs.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -457,6 +457,13 @@ func (m *Mock) calls() []Call {
 	return append([]Call{}, m.Calls...)
 }
 
+// Reset removes all programmed expectations from a mock
+func (m *Mock) Reset() {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	m.ExpectedCalls = nil
+}
+
 /*
 	Arguments
 */
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -1156,3 +1156,17 @@ func Test_WaitUntil_Parallel(t *testing.T) {
 	// Allow the first call to execute, so the second one executes afterwards
 	ch2 <- time.Now()
 }
+
+func Test_Reset(t *testing.T) {
+	// make a test impl object
+	var mockedService = new(TestExampleImplementation)
+
+	mockedService.
+		On("TheExampleMethod2", 1, 2, 3).
+		Return(0)
+
+	assert.Equal(t, 1, len(mockedService.ExpectedCalls))
+
+	mockedService.Reset()
+	assert.Equal(t, 0, len(mockedService.ExpectedCalls))	
+}


codeactual:
Should all `Calls` and `testData` also be reset to fully clear the slate?

ernesto-jimenez:
@jonnyreeves could you provide an example on how you would use it?

I would like to understand the use case better :)

jimmiebtlr:
I'll hop in with my use case.

I start a router with a passed in DB, and potentially some others.  I'd like to reset the db mock between tests without needing to stop and start the server.

So pretty much any case where something can be resource or time intensive to start and stop, that you want to test with a mock.

ernesto-jimenez:
@jimmiebtlr you can already do that as follows:

```go
// mock
type myMock struct {
    mock.Mock
}
// initialise
m := &myMock{}
// [...]
// Reset
m.Mock = mock.Mock{}
```

Have you ever used this?

jimmiebtlr:
Oh, sweet, didn't realize you could do that.

Monnoroch:
I have another use case. Here's a component:
```
type StuffDoer struct {
    HelpfulDependency HelpfulInterface
}

func (self *StuffDoer) DoStuff() string {
    ...
}
```

Of course I want to test it. I have test suit with one test per feature for the `DoStuff` method. Most tests want `self.HelpfulDependency.On("HelpfulMethod").Return(result1)`, but I have a single test that wants `self.HelpfulDependency.On("HelpfulMethod").Return(result2)`. Ideally I set up the default case in the `SetupTest()` and override it once in the single test that requires a different response. Right now, doing `.On()` for the second time does not override, so I either have to `.Reset()` first (which is absent) or recreate the `StuffDoer` object with the newly created mock. Ideally I would want testify to do what Mockito does: override previous calls with `.On()`, but `.Reset()` is an okay alternative.

daniellowtw:
@ernesto-jimenez I also think abstracting the resetting of mocks can improve readability, and also standardize the ways of doing that. It is also more aligned with the intuition when using the mocks, as suggested in @Monnoroch example.

Using the suite package, it would be nicer to read and easier to understand
```
func(s *MyTestSuite) SetupTest() {
	s.myMock.Reset()
}
```
Instead of other possibilities such as 
```
func(s *MyTestSuite) SetupTest() {
	// option 1
	s.myMock.ExpectedCalls = nil
	s.myMock.Calls = nil
}
```
or 
```
func(s *MyTestSuite) SetupTest() {
	// option 2
	s.myMock.Mock = mock.Mock{}
}
```

jvitoroc:
May I suggest a simpler solution that would require a simple modification? 
https://github.com/jvitoroc/testify/blob/master/mock/mock.go#L225
```
func (m *Mock) Reset() {
	*m = Mock{}
}
```

PR #394, opened 2017-02-01
Add using 'Equal' method of structure to test equality

bgaifullin:
If struct implements Equal method it should be used instead of
recursively compare fields, because in some cases recursively
comparing fields returns false result. for example type decimal.Decimal
can have different internal representation for the same values.

### DIFF
--- a/assert/equals.go
+++ b/assert/equals.go
@@ -1,0 +1,174 @@
+package assert
+
+import (
+	"reflect"
+	"unsafe"
+)
+
+const (
+	// if struct implements this method, it will be used to check equality
+	equalMethodName = "Equal"
+)
+
+type visit struct {
+	a1  unsafe.Pointer
+	a2  unsafe.Pointer
+	typ reflect.Type
+}
+
+// check equality of two interfaces, in case if interface has method 'Equal', uses it
+func deepEqual(x, y interface{}) bool {
+	if x == nil || y == nil {
+		return x == y
+	}
+	v1 := reflect.ValueOf(x)
+	v2 := reflect.ValueOf(y)
+	if v1.Type() != v2.Type() {
+		return false
+	}
+	return deepValueEqual(v1, v2, make(map[visit]bool), 0)
+}
+
+func deepValueEqual(v1, v2 reflect.Value, visited map[visit]bool, depth int) bool {
+	if !v1.IsValid() || !v2.IsValid() {
+		return v1.IsValid() == v2.IsValid()
+	}
+	if v1.Type() != v2.Type() {
+		return false
+	}
+
+	// if depth > 10 { panic("deepValueEqual") }	// for debugging
+	hard := func(k reflect.Kind) bool {
+		switch k {
+		case reflect.Array, reflect.Map, reflect.Slice, reflect.Struct:
+			return true
+		}
+		return false
+	}
+
+	if v1.CanAddr() && v2.CanAddr() && hard(v1.Kind()) {
+		addr1 := unsafe.Pointer(v1.UnsafeAddr())
+		addr2 := unsafe.Pointer(v2.UnsafeAddr())
+		if uintptr(addr1) > uintptr(addr2) {
+			// Canonicalize order to reduce number of entries in visited.
+			// Assumes non-moving garbage collector.
+			addr1, addr2 = addr2, addr1
+		}
+
+		// Short circuit if references are already seen.
+		typ := v1.Type()
+		v := visit{addr1, addr2, typ}
+		if visited[v] {
+			return true
+		}
+
+		// Remember for later.
+		visited[v] = true
+	}
+
+	switch v1.Kind() {
+	case reflect.Bool:
+		return v1.Bool() == v2.Bool()
+	case reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64, reflect.Int:
+		return v1.Int() == v2.Int()
+	case reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uint, reflect.Uintptr:
+		return v1.Uint() == v2.Uint()
+	case reflect.Float32, reflect.Float64:
+		// TODO add precision
+		return v1.Float() == v2.Float()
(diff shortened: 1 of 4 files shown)


bgaifullin:
up

bgaifullin:
up

bgaifullin:
Folks, please review the patch. I cannot integrate package decimal to my project without this PR.

bgaifullin:
ping

bgaifullin:
Is there any chance that this PR will be merged?

bgaifullin:
up

.idea/libraries/GOPATH__testify_.xml
```diff
@@ -0,0 +1,17 @@
...
+    <SOURCES>
+      <root url="file://$USER_HOME$/go/src/github.com" />
+      <root url="file://$USER_HOME$/go/src/golang.org" />
+      <root url="file://$USER_HOME$/go/src/gopkg.in" />
+    </SOURCES>
+    <excluded>
+      <root url="file://$PROJECT_DIR$" />
+    </excluded>
+  </library>
+</component>
```
ernesto-jimenez: Remove IDE files

.idea/libraries/Go_SDK.xml
```diff
@@ -0,0 +1,10 @@
+<component name="libraryTable">
+  <library name="Go SDK">
+    <CLASSES>
+      <root url="file:///usr/local/go/src" />
+    </CLASSES>
+    <SOURCES>
+      <root url="file:///usr/local/go/src" />
+    </SOURCES>
+  </library>
+</component>
```
ernesto-jimenez: this one too

assert/equals.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+		return deepValueEqual(v1.Elem(), v2.Elem(), visited, depth+1)
+	case reflect.Struct:
+
+		equal := v1.MethodByName(equalMethodName)
+		// if struct has equal method, use it
+		if equal.IsValid() {
+			result := equal.Call([]reflect.Value{v2})
+			return result[0].Bool()
+		}
```
ernesto-jimenez: What happens if the `Equal` method has a signature you don't expect?

bgaifullin:
@ernesto-jimenez thanks, I have addressed your comments.
Removed ide files.
It will panic if Equal method signature does not fit required, so added check for Equal method signature. Also covered this case by tests.

ernesto-jimenez:
@bgaifullin I just realised most of the code is copied and pasted from the standard's library source code.

I'm a bit uncomfortable with forking code from the standard library:
- The license requires keeping the copyright notices, which are missing
- We are copyng all the source but not the tests
- Testify could get out of sync with the stdlib behaviour (e.g: Go 1.8 included a fix on reflect.DeepEqual)

What is the issue aligning your expectation with the result?

Would it be possible to address it without forking `reflect.DeepEqual`?

bgaifullin:
@ernesto-jimenez you are right, most of code of function deepEqual is forked from standard library.
My first goal is solving problem with comparing type Decimal (the appropriate PR to shopspring/decimal  is mentioned above), but this PR also actual for any type that has different behaviour between member-wise comparison and comparison via Equal method. for example the default type time. you can have the same time in different Locations and for it the standard DeepEqual returns false, but method time.Equal returns true.
I do not know how to implement this without forking the standard DeepEqual. Comparison vis using method Equal is not related to function DeepEqual. 
Also I cannot implement the separate method for comparison via method Equal, because there can be a complex structures like slices, arrays, map other structures, which contains members which implements Equal method. for the instance: 
```
t1 =: time.Now()
assert.Equal([]time.Time{t1}, []time.Time{t1.In(time.UTC())})
```
This assert should returns true, even local time zone is not UTC 
I also do not like forking functional, but I do not see other ways to implement this.

Any suggestions are welcome.

ernesto-jimenez:
@bgaifullin have you considered creating an issue to support this in the standard library?

It seems like it would be the best place to contribute such a change. If they have any objections, I would be interested in learning what those objections are.

bgaifullin:
@ernesto-jimenez there is proposal to implement such think in go standard library: https://github.com/golang/go/issues/20444
Unfortunately it was rejected. :(

ernesto-jimenez:
@bgaifullin as they point out on that issue, the change would be a breaking change. We would like to avoid that too since we want to keep testify backwards compatible.

Looking at the issue for the stdlib, the objection was to change the behaviour of `DeepEqual`, but they might be willing to consider a proposal to add a new function with that feature.

It might be worth raising a proposal that is backwards compatible.

bgaifullin:
@ernesto-jimenez ok, I will try to raise such proposal.

bgaifullin:
@ernesto-jimenez you can vote for [proposal ](https://github.com/golang/go/issues/20519)

bgaifullin:
@ernesto-jimenez the proposal has been declined. Can you accept this PR, or I should close it ?

bgaifullin:
@ernesto-jimenez  What do you think about using third party library to check equality.
for example https://github.com/pierrre/compare or https://github.com/google/go-cmp

jeromefroe:
+1 for investigating using `go-cmp` to replace `reflect.DeepEqual`.

This seems particularly relevant given Go 1.9's changes to `time.Time` will likely break many people's tests. Furthermore, I think most people would agree that if the values have an `Equal` method they should be compared using that method before falling back on basic types. 

In regards to backwards compatibility, a major version bump would probably satisfy most people's needs since they can vendor older versions if they so please. In fact, it seems likely to me that there will be more people whose tests are broken because of the changes to `time.Time` in Go 1.9 than people whose tests would be broken by using an equality function which checks for an `Equal` method first.

An alternative option, as well, is to introduce a new equality function for this purpose. Then users can choose when they want to use `Equal` and when they want to use the new function.

ernesto-jimenez:
Moving to `go-cmp` would be great, yes.

I've created a specific issue and labeled as `help wanted` https://github.com/stretchr/testify/issues/535

Closing this in favour of that implementation.

PR #404, opened 2017-02-23
Adds a helper to quickly get the calls for a given method

aubm:
Hello, let me know if you like the idea, I will write some tests for it.
This is something that I found myself doing repeatedly, then I figured it would be useful to have it directly implemented in the package.


ernesto-jimenez:
@aubm why were you doing this?

PR #412, opened 2017-03-17
Tighten language by increasing overall consistency in wording in text…

nesvold:
…s and argument names: use 'actual' instead of 'received'


nesvold:
@ernesto-jimenez Something you could consider?

PR #417, opened 2017-03-29
diffArguments: remove unnecessary range-for

### LINKED ISSUE #393: diffArguments panics with index out of range
If `expected` is longer than `actual`, `diffArguments` panics with `index out of range`:

https://github.com/stretchr/testify/blob/211175bc2881649c78872c5853521ad390060a80/mock/mock.go#L717

This happened to me when I added an argument to a function, but the stubbed function calls were still specified with the old, shorter argument list.

josephholsten:
When `len(actual) > len(expected)`, this for loop would cause a panic.

Alternative to this implementation, there could be a check for
`if len(actual) != len(expected)`, but generating a meaningful message
describing that is already provided by `diff()`. So we defer to diff in
every case.

Fixes #393

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -715,10 +715,8 @@ func typeAndKind(v interface{}) (reflect.Type, reflect.Kind) {
 }
 
 func diffArguments(expected Arguments, actual Arguments) string {
-	for x := range expected {
-		if diffString := diff(expected[x], actual[x]); diffString != "" {
-			return fmt.Sprintf("Difference found in argument %v:\n\n%s", x, diffString)
-		}
+	if diffString := diff(expected, actual); diffString != "" {
+		return fmt.Sprintf("Difference found in arguments:\n\n%s", diffString)
 	}
 
 	return ""


josephholsten:
Fixes #393, needs better commit msg, tests.

josephholsten:
test appears only to be failing on go 1.1, go 1.2 & go 1.3; all with error:
```
The command "eval godep restore " failed. Retrying, 2 of 3.
```

nathanboktae:
@josephholsten could you trigger a new build? it was probably an intermittent error. I'd love this PR to be accepted - just ran into this.

josephholsten:
@nathanboktae triggered now! 👉

josephholsten:
ok, looks like we're dying with:
```
$ godep restore

godep: Dep (github.com/davecgh/go-spew/spew) restored, but was unable to load it with error:

	Package (bytes) not found

godep: Dep (github.com/pmezard/go-difflib/difflib) restored, but was unable to load it with error:

	Package (bufio) not found

godep: Dep (github.com/stretchr/objx) restored, but was unable to load it with error:

	Package (bytes) not found

godep: Error checking some deps.
```

josephholsten:
@nathanboktae I've created #438 to verify that these tests are broken in master.

DAddYE:
Already done

DAddYE:
Should be already done now.

Issue #420, opened 2017-04-01
Add default struct with mock.Mock (syntactic sugar)


alexbt:
When creating mocks, I usually need to define one or more structs as follow:

I was trying to embedded mock.Mock in a repeatable manner, but I can very well do:
type (
	mockWriter mock.Mock
	mockUsername mock.Mock
	elasticSearchMockNoResult mock.Mock
	elasticSearchMockOneResult mock.Mock
)
```
type (
	mockWriter struct {mock.Mock}
	mockUsername struct {mock.Mock}
	elasticSearchMockNoResult struct {mock.Mock}
	elasticSearchMockOneResult struct {mock.Mock}
)
```

What I like to do, is define a Mock struct that I can just reuse (defined in another go file):
```
type Mock struct {mock.Mock}
```

Now, my previous struct declaration looks like this (I just think it looks nicer):
```
type (
	mockWriter test.Mock
	mockUsername test.Mock
	elasticSearchMockNoResult test.Mock
	elasticSearchMockOneResult test.Mock
)
```
 
**It would be nice if this was available in testify's mock package, so I would just do:**

```
type (
	mockWriter mock.MockStruct
	mockUsername mock.MockStruct
	elasticSearchMockNoResult mock.MockStruct
	elasticSearchMockOneResult mock.MockStruct
)
```

alexbt:
PR: https://github.com/stretchr/testify/pull/421

alexbt:
beginner's misuse of go..

alexbt:
PR: https://github.com/stretchr/testify/pull/421

ernesto-jimenez:
The best syntactic sugar is to use `go generate` to automatically generate the whole mock based on an interface

https://blog.gopheracademy.com/advent-2016/how-to-write-good-tests-for-solid-code/

PR #421, opened 2017-04-02
420: Add a default Mock struct

alexbt:
I'm not sure what are your thoughts on this addition.

With this, I'd be able to define my mock structs as follow:

```
type (
	mockWriter mock.MockStruct
	mockUsername mock.MockStruct
	elasticSearchMockNoResult mock.MockStruct
	elasticSearchMockOneResult mock.MockStruct
)
```

It's really just syntactic sugar, but it reads better in my opinion.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -176,6 +176,11 @@ type Mock struct {
 	mutex sync.Mutex
 }
 
+// Default Struct definition using Mock
+type MockStruct struct {
+	Mock
+}
+
 // TestData holds any data that might be useful for testing.  Testify ignores
 // this data completely allowing you to do whatever you like with it.
 func (m *Mock) TestData() objx.Map {


alexbt:
associated issue: https://github.com/stretchr/testify/issues/420

alexbt:
beginner's misuse of go

alexbt:
associated Issue: https://github.com/stretchr/testify/issues/420

DAddYE:
Mmm I don't get why you can't

```go
myMock {mock.Mock}
```

alexbt:
is this valid `myMock {mock.Mock}` ? Shouldn't it be `myMock struct {mock.Mock}` ?
Please correct me If I'm wrong here, I'm fairly new to Go.

yes, (as detailed in the "issue" https://github.com/stretchr/testify/issues/420) I can do the following, which I think is what you're saying: 
```
type (
	mockWriter struct {mock.Mock}
	mockUsername struct {mock.Mock}
	elasticSearchMockNoResult struct {mock.Mock}
	elasticSearchMockOneResult struct {mock.Mock}
)
```

I just thought it would be nicer if I could avoid the repetitive ```struct { ``` :
```
type (
	mockWriter mock.MockStruct
	mockUsername mock.MockStruct
	elasticSearchMockNoResult mock.MockStruct
	elasticSearchMockOneResult mock.MockStruct
)
```

I totally get if you think this is a useless addition :)

For me, I end up creating my own wrapper in each of my projects: ```type Mock struct {mock.Mock}``` just so I can use it without inlining a struct definition every time.

ernesto-jimenez:
The best syntactic sugar is to use `go generate` to automatically generate the whole mock based on an interface

https://blog.gopheracademy.com/advent-2016/how-to-write-good-tests-for-solid-code/

PR #431, opened 2017-04-17
Compare bytes with bytes.Equal instead of reflect.DeepEqual

dennwc:
In some tests we compare ~1M byte slices with testify library, but since it uses `reflect.DeepEqual` this operation is very slow.

This PR adds a `[]byte` as a special case for `ObjectsAreEqual` and uses `bytes.Equal` to make comparison faster.

Benchmark results:
```
benchmark                 old ns/op     new ns/op     delta
BenchmarkBytesEqual-4     695795        261           -99.96%
```


bgaifullin:
I this this is covered by PR #394

PR #439, opened 2017-05-09
add mock.MethodCalled(methodName, args...)

yiminc:
Current Mock.Called(args...) records a method call, but the method name is the calling method. This limits a very useful use case. We are working on a library, where user could register function with a name. Then those registered functions could be invoked by name from some workflow code. The library is using reflect to get the function and passing in arguments to those registered functions. We don't know the function signature at our library's compile time. In code, it is conceptually like:
lib.Register("foo", func(...) {...})
lib.Register("bar", func(...) {...})
Then, the workflow code is something conceptually like:
result1 := lib.Invoke("foo", args...)
result2 := lib.Invoke("bar", args...)
processResult(result1, result2).
Now, we want our test suite to be able to test those workflow code,and allow user to mock any registered function. User's test code would look like:
m.On("foo", args...).Return(some_static_result)
m.On("bar", args...).Return(func(...) {...})
Then user assign this mock object m to our test suite, and in our library test framework, before we pass the arguments via reflect to the actual registered function, we want to check if a mock is available by:
found, _ := m.findExpectedCall(registeredFnName, arguments)
if found {
retArgs := m.MethodCalled(functionName, args...)
// deal with the mock call...
} else {
// no mock call found, call the actual registered function.
}

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -213,7 +213,7 @@ func (m *Mock) On(methodName string, arguments ...interface{}) *Call {
 // 	Recording and responding to activity
 // */
 
-func (m *Mock) findExpectedCall(method string, arguments ...interface{}) (int, *Call) {
+func (m *Mock) FindExpectedCall(method string, arguments ...interface{}) (int, *Call) {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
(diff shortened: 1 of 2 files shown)


yiminc:
poke @ernesto-jimenez :)

PR #430, opened 2017-04-17
Added goconvey style assertions to the mock package

kingofsouth:
Added assertions that work with the [goconvey](https://github.com/smartystreets/goconvey) testing framework.

See example test [here](https://github.com/kingofsouth/mockery/blob/convey-tests-for-testify/convey/convey_test.go#L34-L71)

Issue #437, opened 2017-05-09
allow mock to record specified method


yiminc:
Current `Mock.Called(args...)` records a method call, but the method name is the calling method. This limits a very useful use case. We are working on a library, where user could register function with a name. Then those registered functions could be invoked by name from some workflow code. The library is using reflect to get the function and passing in arguments to those registered functions.  We don't know the function signature at our library's compile time. In code, it is conceptually like:
```
lib.Register("foo", func(...) {...})
lib.Register("bar", func(...) {...})
```
Then, the workflow code is something conceptually like:
```
result1 := lib.Invoke("foo", args...)
result2 := lib.Invoke("bar", args...)
processResult(result1, result2).
```
Now, we want our test suite to be able to test those workflow code,and allow user to mock any registered function. User's test code would look like:
```
m.On("foo", args...).Return(some_static_result)
m.On("bar", args...).Return(func(...) {...})
```
Then user assign this mock object m to our test suite, and in our library test framework, before we pass the arguments via reflect to the actual registered function, we want to check if a mock is available by:
```
found, _ := m.findExpectedCall(registeredFnName, arguments)
if found {
  retArgs := m.MethodCalled(functionName, args...)
 // deal with the mock call...
} else {
  // no mock call found, call the actual registered function.
}
```

mvdkleijn:
I believe this was already addressed by PR #444

PR #441, opened 2017-05-26
Revert PRs with

ernesto-jimenez:
Reverts:
- #417: bug was already fixed in another PR and code was unformatted
- #430: added exported functions had confusing names and code was unformatted
- #439: exported an internal method

@DAddYE FYI

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,37 +180,6 @@ For more information on how to write mock code, check out the [API documentation
 
 You can use the [mockery tool](http://github.com/vektra/mockery) to autogenerate the mock code against an interface as well, making using mocks much quicker.
 
-### Plays well with [`goconvey`](https://github.com/smartystreets/goconvey)
-`goconvey` is a BDD style testing framework for gophers. The `mock` package defines several assertion functions intended to 
-work with that framework
-
-Here is an example test function testing a piece of code that operates on a mock called `Example`.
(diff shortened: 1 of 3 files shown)


ernesto-jimenez:
@DAddYE addressing #430 and #439 needs some analysis

We must be very careful with what we export.

josephholsten:
@ernesto-jimenez: looks like my patch was redundant, revert is the right move for it. Thanks for the follow up!

yiminc:
@ernesto-jimenez, #439 is to support a very valid use case, see issue  https://github.com/stretchr/testify/issues/437 Please help on what else we could do to make that use case possible. Thanks.

kingofsouth:
@ernesto-jimenez reopened my pull request [here](https://github.com/stretchr/testify/pull/443). Ran go fmt sorry for missing that before. Let me know if any explanations are needed or if you would like changes to function naming. I mostly tried to keep the naming convention that was already in place in the various assert functions only dropping the assert verb such that it would read better in the so clauses.

yiminc:
@ernesto-jimenez reopened my PR here https://github.com/stretchr/testify/pull/444. This new PR avoid exposing internal method, also added documentation to the new methods. Please put comments to the PR if you would like anything to be changed. Thanks.

ernesto-jimenez:
Replied to both PRs. Thank you all!

PR #443, opened 2017-05-27
Added goconvey style assertions

kingofsouth:
Added assertions that work with the [goconvey](https://github.com/smartystreets/goconvey) testing framework. This would expose all of the various assert functions on the mock struct as exported package functions that could be passed to [convey.So](https://godoc.org/github.com/smartystreets/goconvey/convey#So) for verifying mock interactions.

Made sure that the various assertion functions would be able to directly operate on mocks generated by [mockery](https://github.com/vektra/mockery)

See example test [here](https://github.com/kingofsouth/mockery/blob/convey-tests-for-testify/convey/convey_test.go#L34-L71)

See more on goconvey assertions [here](https://github.com/smartystreets/goconvey/wiki/Assertions)

This reverts commit 2b76a9702ee5548248cded996e8614e3469e1cd2.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -1,6 +1,7 @@
 package mock
 
 import (
+	"bytes"
 	"fmt"
 	"reflect"
 	"regexp"
@@ -13,6 +14,10 @@ import (
 	"github.com/pmezard/go-difflib/difflib"
 	"github.com/stretchr/objx"
 	"github.com/stretchr/testify/assert"
+)
+
+const (
+	convey_test_passed_message = ""
 )
 
 // TestingT is an interface wrapper around *testing.T
@@ -368,6 +373,25 @@ func AssertExpectationsForObjects(t TestingT, testObjects ...interface{}) bool {
 // AssertExpectations asserts that everything specified with On and Return was
 // in fact called as expected.  Calls may have occurred in any order.
 func (m *Mock) AssertExpectations(t TestingT) bool {
+	errorMsg := ExpectationsWereMet(m)
+	if errorMsg != convey_test_passed_message {
+		t.Errorf(errorMsg)
+		return false
+	} else {
+		return true
+	}
+}
+
+// ExpectationsWereMet is a goconvey style assertion see: https://github.com/smartystreets/goconvey/wiki/Assertions
+// It asserts that everything specified with On and Return was
+// in fact called as expected.  Calls may have occurred in any order
+func ExpectationsWereMet(iFace interface{}, args ...interface{}) string {
+	m, err := getMock(iFace)
+	if err != nil {
+		return err.Error()
+	}
+
+	b := new(bytes.Buffer)
 	var somethingMissing bool
 	var failedExpectations int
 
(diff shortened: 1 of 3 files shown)


mock/mock.go
```diff
@@ -377,55 +401,176 @@ func (m *Mock) AssertExpectations(t TestingT) bool {
...
+	}
+
+	return passedTest
+}
+
+func getMock(iFace interface{}) (Mock, error) {
+	var mock Mock
+	if m, ok := iFace.(Mock); ok {
+		mock = m
+	} else {
```
kingofsouth: Probably should refactor this else out of here

ernesto-jimenez:
@kingofsouth I have two concerns about this:
- exporting methods that are called like assetions but return strings is quite confusing to me as somebody who has never used `goconvey`
- It's making significant changes to the package just to make it work with `goconvey`

It would probably be better to have a new package focused on the integraiton of  goconvey with testify?

e.g:
```go
type fakeT struct {
    buf bytes.Buffer
    failed bool
}

func (t *fakeT) Logf(format string, args ...interface{}) {
    fmt.Sprinf(&t.buf, format, args...)
}

func (t *fakeT) Errorf(format string, args ...interface{}) {
    t.Logf(format, args...)
    t.FailNow()
}

func (t *fakeT) FailNow() {
   t.failed = true
}

func (t *fakeT) error() string {
    return b.String()
}

type testifyMock interface {
    AssertCalled(t mock.TestingT, methodName string, arguments ...interface{}) bool
}

func ShouldHaveReceived(mock interface{}, expected ...interface{}) string {
   m, ok := mock.(testifyMock)
   if !ok {
       return fmt.Sprintf("%T is not a testify mock", mock)
   }
   if len(expected) < 1 {
       return "missing expected method call"
   }
   method, ok := expected[0].(string)
   if !ok {
       return fmt.Sprintf("expected method name (string) and got %#v", expected[0])
   }
   t := &fakeT{}
   m.AssertCalled(t, method, expected[1:]...)
   if t.failed {
       return t.error()
   }
   return ""
}
```

You can probably build that package without any changes to testify :)

kingofsouth:
@ernesto-jimenez while it is true that I could build a package to provide this functionality without any modifications to testify, I think that would be significantly less useful to communities currently using the two packages together in a less integrated way.

Are you more comfortable with this change in a less invasive form? 

I would be happy to move any `convey` assertions into a separate package (Possibly under `mock`?) and refactor it to make better use of the [TestingT](https://github.com/stretchr/testify/blob/d77da356e56a7428ad25149ca77381849a6a5232/require/requirements.go#L4-L7) interface to make this work without modifications to the main `mock` package.

From my point of view as a consumer of both: this project provides a very useful tool with its `mock` package and has value irrespective to how assertions are performed,

ernesto-jimenez:
@kingofsouth on the other hand, I never use goconvey. Adding a new package within testify I would never use will be tricky to maintain.

It is probably better for you to create a package that has goconvey style assertions that integrate with testify. Then we can link to your package from our README.

You could even [use go generate](https://github.com/stretchr/testify/blob/master/assert/forward_assertions.go#L16) to automatically generate goconvey assertions for each assertion in `testify/assert`.

tamccall:
Created [package](https://github.com/tamccall/moconvey) with convey style assertions.

Need to get travis ci setup for the build then I'll modify this change to just be a readme update

PR #444, opened 2017-05-27
add MethodCalled to the mock

yiminc:
This is a second try to address issue https://github.com/stretchr/testify/issues/437
The original PR https://github.com/stretchr/testify/pull/439 was merged then reverted.
This PR avoid exposing internal function which is the reason pull #439 was reverted.
Really hope this could be merged as this would support a very useful use case.
Please put comments if you see anything that needs to be changed for this PR to be merged.
Thanks.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -285,9 +285,16 @@ func (m *Mock) Called(arguments ...interface{}) Arguments {
 	}
 	parts := strings.Split(functionPath, ".")
 	functionName := parts[len(parts)-1]
+	return m.MethodCalled(functionName, arguments...)
+}
 
+// MethodCalled tells the mock object that the given method has been called, and gets
+// an array of arguments to return. Panics if the call is unexpected (i.e. not preceded
+// by appropriate .On .Return() calls)
+// If Call.WaitFor is set, blocks until the channel is closed or receives a message.
+func (m *Mock) MethodCalled(methodName string, arguments ...interface{}) Arguments {
 	m.mutex.Lock()
-	found, call := m.findExpectedCall(functionName, arguments...)
+	found, call := m.findExpectedCall(methodName, arguments...)
 
 	if found < 0 {
 		// we have to fail here - because we don't know what to do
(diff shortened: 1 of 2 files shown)


mock/mock.go
```diff
@@ -209,6 +209,12 @@ func (m *Mock) On(methodName string, arguments ...interface{}) *Call {
 	return c
 }
 
+// IsCallExpected checks if a method call is expected.
+func (m *Mock) IsCallExpected(method string, arguments ...interface{}) bool {
+	found, _ := m.findExpectedCall(method, arguments...)
+	return found >= 0
+}
```
ernesto-jimenez: Why add an `IsCallExpected`?

The test would already know which expectations it has setup.

yiminc: In our use case, user could optionally setup mock for some of the registered methods (with their expected arguments). When our test suite runs users's tested methods (which calls registered methods), it will have to decide if it should call a mock (if there is one) or it should call the real registered method if there is no mock.

yiminc: There could be many registered methods, we don't want to force user to mock all those registered methods. For example, user's code might call into 10 registered methods, and only one of them needs to be mocked.

ernesto-jimenez: @yiminc do you have a complete code sample of how it would work? Both your mock and how the users would use it.

It would help get my head around the use case.

Thanks!

ernesto-jimenez: @yiminc you can accomplish what you want without exposing `testify/mock`'s internals.

Something like this:

```go
type TestRuntimeEnvironment struct {
    mock.Mock
    mocked map[string]struct{}
}

func (t *TestRuntimeEnvironment) On(methodName string, arguments ...interface{}) *mock.Call {
    if t.mocked == nil {
        t.mocked = make(map[string]struct{})
    }
    t.mocked[methodName] = struct{}
    return t.Mock.On(methodName, arguments...)
}

func (t *TestRuntimeEnvironment) ExecuteActivity(activityName string, args... interface{}) (result, error) {
    _, hasMock := t.mocked[activityName]
    if hasMock {
        retArgs := t.MethodCalled(activityName, args...)
        // proceed with retArgs
        ...
    } else {
        // no mock found, so proceed with normal code path to call the registered method
        ...
    }
}
```

yiminc: You are assuming that user always use mock.Anything for all arguments, that is not true. It is a big limitation to empose this enforcement.

ernesto-jimenez: @yiminc, that solution assumes nothing about the arguments you send to `On`. It would work with any set of arguments.

yiminc: I'm confused. Assume that the setup is:
`testEnv.On("downloadFile", "test-file-id").Return(FileInfo{Path:"path_to_downloaded_file"}, nil)
`
We would want the hasMock to be true only if the methodName is "downloadFile" and the argument is "test-file-id". With the solution you suggested, the hasMock will be true for methodName="downloadFile" with any argument.

ernesto-jimenez: You could use the arguments if you wanted too, what I shared was just a quick example.

That would be better handled in your code rather than exposing `Mock`'s internals. The use case is quite different from `mock`'s purpose.

yiminc: It will basically have to duplicate the findExpectedCall() logic. But we don't have access to the m.mutex. And it will also break our code if there is any change to the findExpectedCall() in the future.

yiminc: The IsCallExpected does not really expose too much of internal stuff. It should not break anything, or make any harm to the mock library, yet it make the use case much more viable.

ernesto-jimenez: It exposes querying the internal state of the mock, which is unnecessary for the purpose of the package. As soon as we expose it, people will be able to abuse it and we must keep it to maintain backwards compatibility.

yiminc: OK, while I'm not convinced, I will comply. Will update this PR to remove IsCallExpected(). @ernesto-jimenez, thank you for your time spend on this, appreciated!

ernesto-jimenez:
@yiminc thanks! can you rebase and squash the commits?

yiminc:
Rebase and squashed.

ernesto-jimenez:
@yiminc merged, thanks :)

PR #450, opened 2017-05-29
Simple validation of Equal/NotEqual assertion arguments

### LINKED ISSUE #286: Misleading error when trying to compare two functions
While Go doesn't support comparing functions to each other natively, 

```
invalid operation: time.Now == time.Now (func can only be compared to nil)
```

the current failure message does not reflect this, and probably could support it anyway for the purposes of tests, maybe by comparing the pointer addresses or something.

```
assert.Equal(t, time.Now, time.Now)

Not equal: (func() time.Time)(0xa6c20) (expected)
                != (func() time.Time)(0xa6c20) (actual)

```

medzin:
It resolves #286 by providing a simple validation function for Equal/NotEqual assertions. I have no idea how to test error messages produced by go testing framework, so I moved part of the logic to a helper function.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math"
 	"reflect"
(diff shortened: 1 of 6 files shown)


assert/assertions.go
```diff
@@ -14,6 +14,8 @@ import (
 	"unicode"
 	"unicode/utf8"
 
+	"errors"
```
ernesto-jimenez: move errors with the rest of the stdlib imports

assert/assertions.go
```diff
@@ -302,6 +304,10 @@ func IsType(t TestingT, expectedType interface{}, object interface{}, msgAndArgs
 // Pointer variable equality is determined based on the equality of the
 // referenced values (as opposed to the memory addresses).
 func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool {
+	if err := validateEqualArgs(expected, actual); err != nil {
+		return Fail(t, fmt.Sprintf("Invalid operation: %#v == %#v (%s)",
+			expected, actual, err), msgAndArgs...)
+	}
```
ernesto-jimenez: people sometimes mix up expected and actual.

I would say it's better to fail if either `expected` or `actual` are a function.

People can compare functions using `.Nil` and `.NotNil`.

medzin:
Code generated and commits squashed.

ernesto-jimenez:
@medzin merged, thanks!

PR #453, opened 2017-05-30
Add IsCallExpected()

yiminc:
This is a follow up to https://github.com/stretchr/testify/pull/444
It expose a way for user to know if a call is expected before calls to the MethodCalled() function. @ernesto-jimenez, I know that you think that this expose some internal state and dis-like this method. But we need a way to make the use case possible. Another way we could do is to return error from MethodCalled() if the call is not expected, right now it will panic. Let me know what do you think, and I will update the PR if you think returning error from MethodCalled is good.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -209,6 +209,15 @@ func (m *Mock) On(methodName string, arguments ...interface{}) *Call {
 	return c
 }
 
+// IsCallExpected checks whether a method call is expected.
+func (m *Mock) IsCallExpected(method string, arguments ...interface{}) bool {
+	m.mutex.Lock()
+	found, _ := m.findExpectedCall(method, arguments...)
+	m.mutex.Unlock()
+
+	return found >= 0
+}
+
 // /*
 // 	Recording and responding to activity
 // */


yiminc:
I put up another PR https://github.com/stretchr/testify/pull/454
Either of these 2 options would work perfectly for our use case. We only need one of them.

ernesto-jimenez:
Hi @yiminc, as we discussed in the other PR, this would be exposing internal state from the mock.

yiminc:
Just curious, what kind of problem are you worried about having IsMethodCalled() exposed? I don't see any obvious problem that could rise by having this API. It is a legit request from user to know if a call is expected.

ernesto-jimenez:
A user already knows whether a call is expected because they set up the expectations themselves. Regular users should never need this and exporting it only has downsides for the purpose of the package:
- We have to maintain another method that we must keep backwards compatible.
- People can abuse the method to check the state of the mock when they should never do that.

The use case is to allow third parties to build custom mocking libraries on top of our existing mocking logic, which is fine as long as it can work without exposing internal state or compromising on our public API.

yiminc:
Agreed that regular user would never needs those. However, isn't it a legit requirement for 3rd party mocking library use case to know if a mock call is expected?
It is a pretty simple API (require the same input as MethodCalled() API and return bool). I don't see any backwards compatible issue of having this IsMethodCalled() as long as Called and MethodCalled are backwards compatible.
I still don't get how people can abuse this API to do any harm (to the mock library).

ernesto-jimenez:
> Agreed that regular user would never needs those. However, isn't it a legit requirement for 3rd party mocking library use case to know if a mock call is expected?

A third party mocking library using ours in the background can already know what expectations it has set.

> It is a pretty simple API (require the same input as MethodCalled() API and return bool). I don't see any backwards compatible issue of having this IsMethodCalled() as long as Called and MethodCalled are backwards compatible.
> I still don't get how people can abuse this API to do any harm (to the mock library).

They can abuse it to write poor tests. The only thing it allows for most of our users is to write poor tests.

yiminc:
As the pseudocode I showed in PR 444, there is no easy way 3rd party could tell if a mock call is expected. 3rd party would basically have to duplicate the logic in findExpectedCall() to know if the call is expected, but 3rd party code do not have proper way to lock and would have data race issue. Also, if there is any logic update in the  findExpectedCall() method, 3rd party library will fail.

PR #454, opened 2017-05-30
return error from MethodCalled if call is not expected

yiminc:
This is a second option to address the left over issue to https://github.com/stretchr/testify/pull/444
Instead of panic from MethodCalled() if the call is not expected, it will return error.
We need a way to be able to safely call the MethodCalled(). Either via API to check if the call is expected, or the MethodCalled() returns error instead of panic. 

Also open to any suggestion on other options to achieve the goal.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -285,14 +285,18 @@ func (m *Mock) Called(arguments ...interface{}) Arguments {
 	}
 	parts := strings.Split(functionPath, ".")
 	functionName := parts[len(parts)-1]
-	return m.MethodCalled(functionName, arguments...)
+	result, err := m.MethodCalled(functionName, arguments...)
+	if err != nil {
+		panic(err)
+	}
+	return result
 }
 
 // MethodCalled tells the mock object that the given method has been called, and gets
-// an array of arguments to return. Panics if the call is unexpected (i.e. not preceded
+// an array of arguments to return, and error if the call is unexpected (i.e. not preceded
 // by appropriate .On .Return() calls)
 // If Call.WaitFor is set, blocks until the channel is closed or receives a message.
-func (m *Mock) MethodCalled(methodName string, arguments ...interface{}) Arguments {
+func (m *Mock) MethodCalled(methodName string, arguments ...interface{}) (Arguments, error) {
 	m.mutex.Lock()
 	found, call := m.findExpectedCall(methodName, arguments...)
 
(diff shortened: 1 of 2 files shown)


yiminc:
For issue 1), would it be OK if we keep MethodCalled() unchanged, and adding a new version that return error?
For issue 2), i don't understand the issue here. If the expectation is satisfied more than expected times, the MethodCalled won't return error, instead, the AssertExpectations() will fail. That is expected behavior.

We don't really want to control when to raise error, we want to know when the call is not expected. Your suggested solution require user to implement TestingT which they don't really care about.
Also, there are several existing methods on *Mock that require TestingT as input parameter, it would be weird to have SetT(t) as well, because instead of require from input parameter, people would expect mock to use its internal t after SetT() is called. But we cannot remove those from input as that would be backward incompatible.

ernesto-jimenez:
> For issue 1), would it be OK if we keep MethodCalled() unchanged, and adding a new version that return error?

That's not an option, since calling `MethodCalled` modifies the internal state of the mock, reducing the repeatability of the expectation.

> For issue 2), i don't understand the issue here. If the expectation is satisfied more than expected times, the MethodCalled won't return error, instead, the AssertExpectations() will fail. That is expected behavior.

That's not the case, the following code will panic due to an unexpected method call:

```go
func TestFoo(t *testing.T) {
        m := &myMock{}
        m.On("test").Return(nil).Once()
        m.MethodCalled("test")
        m.MethodCalled("test")
}
```

> We don't really want to control when to raise error, we want to know when the call is not expected. Your suggested solution require user to implement TestingT which they don't really care about.

No, you could implement the `TestingT` yourself and handle that within your mocking library.

> Also, there are several existing methods on *Mock that require TestingT as input parameter, it would be weird to have SetT(t) as well, because instead of require from input parameter, people would expect mock to use its internal t after SetT() is called. But we cannot remove those from input as that would be backward incompatible.

Agreed, but the `Assert*` methods have other issues we cannot solve right now due to backwards compatibility.

`SetT` would be helpful to stop panicking in cases like the one mentioned before and start failing the tests instead.

yiminc:
OK, thanks for clarify the issue 2. Could you give a brief idea on how to implement TestingT that could solve our problem. We want to ignore the error if there is no mock call found when MethodCalled is invoked, and we want to know that there was error happen due to either missing mock, or already reaching to maximum expected times. The TestingT does not define Fatelf, I guess you mean FailNow. The FailNow on testing.T needs to be called on the same goroutinue that starts the tests which not necessary the same goroutinue call MethodCalled.

ernesto-jimenez:
You can see an example using a custom `testingT` in the following comment: https://github.com/stretchr/testify/pull/443#issuecomment-304469309

That wouldn't allow you to differentiate whether the error was because there were zero expectations or reachability was reached. However, you could do what I suggested in https://github.com/stretchr/testify/pull/444#discussion_r118918691 to see whether you had setup any expectations for that method at all.

yiminc:
Your custom TestingT's FailNow() or Errorf() won't stop execution of the calling goroutinue, then the code has to return something after FailNow or Errorf is called. But since the call is not expected, there is nothing to return, or it has to return nil as Arguments. That means user has to check if the returned Arguments is nil or not. That is breaking change.

devdinu:
If #572 addresses this entirely, we could close this.

PR #455, opened 2017-06-01
Panics with value

### LINKED ISSUE #445: Consideration for extended panic assertion?
There might be an interesting use case for a `assert.PanicsWith()` assertion, where the caller could provide a function to be called to assert on the error that was recovered:

Imagine you're testing a function that should panic before it reaches some mocked methods:
```go
func Function(value int, resource Resource) {
  if value > 0 {
    resouce.Use()
  }
  panic(fmt.Errorf("illegal value"))  
}

func TestFunction(...) {
  mockResource := ...
  // don't stub-out Resource.Use() for this test
  assert.Panics(t, func() { Function(0, mockResource) })
}
```

This test case can pass either because it reached the actual `panic` line (good!) or because it reached the unmocked `Resource.Use()` line (bad!)

One could argue that for this test, you can workaround by stubbing-out all possible resource calls, then verifying that they were not called using `resource.AssertNotCalled()`, but this approach seems brittle and heavy-handed.

Alternatively, what if there were support for a panic-checking callback, like so:

```go
assert.PanicsWith(t, func() { Function(0, mockResource) }, 
        func(err error) { assert.Equals(err.Error(), "illegal value") } )
```

Opinions?  Thank you for the consideration

tartale:
* Added new `assert.PanicsWithValue()` function, plus unit tests

Fixes #445

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -811,6 +811,25 @@ func Panics(t TestingT, f PanicTestFunc, msgAndArgs ...interface{}) bool {
 	return true
 }
 
+// PanicsWithValue asserts that the code inside the specified PanicTestFunc panics, and that
+// the recovered panic value equals the expected panic value.
+//
+//   assert.PanicsWithValue(t, "crazy error", func(){ GoCrazy() })
+//
+// Returns whether the assertion was successful (true) or not (false).
+func PanicsWithValue(t TestingT, expected interface{}, f PanicTestFunc, msgAndArgs ...interface{}) bool {
+
+	funcDidPanic, panicValue := didPanic(f)
+	if !funcDidPanic {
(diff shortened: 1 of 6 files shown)


assert/assertions.go
```diff
@@ -811,6 +811,27 @@ func Panics(t TestingT, f PanicTestFunc, msgAndArgs ...interface{}) bool {
 	return true
 }
 
+// PanicsWithValue asserts that the code inside the specified PanicTestFunc panics, and that
+// the recovered panic value equals the expected panic value.
+//
+//   assert.PanicsWithValue(t, "crazy error", func(){
+//     GoCrazy()
+//   }, "Calling GoCrazy() should panic with 'crazy error'")
```
ernesto-jimenez: Can you please remove the message? A example error message is added automatically to generated `PanicWithValuef` functions

tartale:
@ernesto-jimenez the changes you requested have been made

PR #459, opened 2017-06-02
add mock clock support for the Mock

yiminc:
people often times use mock clock in unit test because we don't want unit test to wait on clock for long time. unit test should run fast. this PR make it possible for those use cases where they use their own mock clock to coordinate calls. If the clock is not set, it will use the default real clock.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -22,6 +22,25 @@ type TestingT interface {
 	FailNow()
 }
 
+// Clock is an interface around time
+type Clock interface {
+	Now() time.Time
+	After(d time.Duration) <-chan time.Time
+}
+
+type realClock struct{}
+
+// Now returns the current local time.
+func (realClock) Now() time.Time {
+	return time.Now()
+}
+
+// After waits for the duration to elapse and then sends the current time
+// on the returned channel.
+func (realClock) After(d time.Duration) <-chan time.Time {
+	return time.After(d)
+}
+
 /*
 	Call
 */
@@ -130,7 +149,10 @@ func (c *Call) WaitUntil(w <-chan time.Time) *Call {
 //
 //    Mock.On("MyMethod", arg1, arg2).After(time.Second)
 func (c *Call) After(d time.Duration) *Call {
-	return c.WaitUntil(time.After(d))
+	c.lock()
+	defer c.unlock()
+	c.WaitFor = c.Parent.clock.After(d)
+	return c
 }
 
 // Run sets a handler to be called before returning. It can be used when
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
@yiminc sorry, but I'm a bit confused about the use case, could you elaborate on what you want to achieve?

`Mock` already has a `WaitUntil` method so you can have full control over the order in which mocks unblock calls.

yiminc:
The WaitUntil() does make the use case possible but it is not as easy to use. The After() method directly use time.After() and waits on real clock time, that is most likely not what people want in unit test environment.

The use case that I'm trying to solve is following:
We are working on a service and a client library to make writing long running workflow easy. (It will be open source on github very soon. :)
A long running workflow could call into registered methods (which could be service running on other box) , those methods could take long time in real world. In order to make testing the long running workflow possible in unit test environment, we need to have full control of the clock and time. The workflow could have logic around time, for example, if a registered method takes too long to complete, the workflow would take some action, like cancel it or terminate and restart, or doing some other things. If we want to test this workflow handling logic code, we need to simulate the time used by the registered method. We are able to mock the method thanks to https://github.com/stretchr/testify/pull/444. We still need the ability to add delay time to the mock call. 
Right now, our user has to:
testSuite := lib.NewTestSuite()
testEnv := testSuite.NewTestEnvironment()
mockClock := testEnv.GetClock()
waitCh := mockClock.After(time.Hour)
testEnv.On("registeredFoo", ...).WaitUntil(waitCh).Return(...)

We want to make it:
testEnv.On("registered_foo", ...).After(time.Hour).Return(...)
In our testSuite.NewTestEnvironment() we would set up our internal mock clock to the testify.Mock struct. This way, the whole concept of the controlled clocked is hide away from our users.

So, yes, with WaitUntil(), it is possible to make it work. But the problems are:
1) It is much more intuitive and easy to use .After(duration) for this use case.
2) Users are more likely to use .After(duration) and only to realize it does not work, by the time they would waste lots of time. And it is confusing for our users.
3) User needs to understand the concept of the controlled time in our test suite and be more confused about the mock clock that they get and not clear if they need to control the mock clock by themselves or not.
4) I think this is not a specific problem that only our team is facing. I think in unit test, it is more likely that people do not want to wait on real clock because unit test needs to be fast. So I believe this is a problem that others would face too, and address it here in this framework would benefit the open source community.

yiminc:
ping @ernesto-jimenez :)

dineshkumar-cse:
@yiminc If you mean time.After waits on the clock immediately and not during method call, its fixed in commit b3596e92796169eee4a0050bceaa767a48cb0075.

@ernesto-jimenez So we could close this issue.

ernesto-jimenez:
👍 

Thanks @dineshkumar-cse !

Issue #469, opened 2017-07-05
Empty/NotEmpty return confusing results for custom types with underlying type string


bits01:
Consider the following example:
```
type key string

func TestEmpty(t *testing.T) {
	var k key = ""
	assert.Empty(t, k, "confusing to fail")
	assert.NotEmpty(t, k, "this is very error prone when used in tests")
}
```

The `assert.Empty` above fails. While I understand what is going on, it is confusing and very error prone in tests when one uses `assert.NotEmpty` instead.

bits01:
Please take a look at PR #470.

reflect.Value.Len() can cover more types, per godoc:
> Len returns v's length. It panics if v's Kind is not Array, Chan, Map, Slice, or String.
https://golang.org/pkg/reflect/#Value.Len

bits01:
PR #470 merged. Thank you.

PR #470, opened 2017-07-05
Issue #469 fix

bits01:
reflect.Value.Len() can cover more types, per godoc:
> Len returns v's length. It panics if v's Kind is not Array, Chan, Map, Slice, or String.


bits01:
ping @DAddYE @ernesto-jimenez

DAddYE:
LGTM

PR #479, opened 2017-07-18
Adds a helper to quickly get the calls for a given method

aubm:
This the same than https://github.com/stretchr/testify/pull/404

I mishandled something that got the previous PR automatically closed, my bad.

@ernesto-jimenez  something like that could be useful if one ever need a quick way to get all the calls for a specific method name. Suppose you want to make sure that some code called the `Warn` method on some mock logger, but you don't want to deal with all the noisy `Info` calls.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -463,6 +463,17 @@ func (m *Mock) calls() []Call {
 	return append([]Call{}, m.Calls...)
 }
 
+// CallsForMethod returns a list of call for a given method
+func (m *Mock) CallsForMethod(method string) []Call {
+	calls := []Call{}
+	for _, call := range m.Calls {
+		if call.Method == method {
+			calls = append(calls, call)
+		}
+	}
+	return calls
+}
+
 /*
 	Arguments
 */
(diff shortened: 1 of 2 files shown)


philtay:

ernesto-jimenez:
@aubm, why would you need access to the internals of the mock?

For the usecase you mentioned, you can simply do `m.AssertNumberOfCalls(t, "Warn", 1)` or `m.AssertCalled(t, "Warn", "some argument")`

aubm:
@ernesto-jimenez good point, it is probably a good idea to use these methods.
Only thing is, they expect you to use the `t` variable. As for me, I happen to use only the `mocks` subpackage of this repo, leaving the rest appart, and I use this thing called [Gomega](https://onsi.github.io/gomega/) that doesn't let me use the `t` variable directly in my tests.

devdinu:
@aubm you could directly use `m.Calls` instead of exposing a function for the same.
I would still suggest you  not to use these, coz it might be changed in upcoming version. ideally it should be hidden.

Maybe you could share the code when you say you can't use `t`, it would help us understand and help.

PR #486, opened 2017-08-04
Add new assertion ErrorContains

### LINKED ISSUE #497: Feature request: ErrorContains
I'm using this pattern a lot:

```
require.Error(t, err)
require.Contains(t, err.Error(), "some part of the error string")
```

Although `require.ErrorEqual` is pretty useful at times, often I only want to match on part of the error message (maybe because some of it is dynamic for each test run or in different test environments).

Proposal: Add `require.ErrorContains()`

dnephin:
Fixes #497

I have found that in many cases an error message isn't entirely predictable. This may be the case when a function wraps an error using`errors.Wrap()`, and the underlying error changes by platform or for some other reason.

I often us an `ErrorContains()` assertion to test these cases. Would you be interested in adding this assertion to `testify` ?

I've named the function `ErrorContains()` which I think reads better than `ContainsError()` but I realize that is less consistent with `EqualError()`. `ErrorContains()` also has the advantage that in godoc it is sorted next to `Error()` so all the error-related assertions are grouped together.

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -84,6 +84,17 @@ func Errorf(t TestingT, err error, msg string, args ...interface{}) bool {
 	return Error(t, err, append([]interface{}{msg}, args...)...)
 }
 
+// ErrorContainsf asserts that the error is not-nil and that it contains the
+// expected error string.
+//
+//   actualObj, err := SomeFunction()
+//   assert.ErrorContainsf(t, err,  expectedErrorString, "error message %s", "formatted")
+//
+// Returns whether the assertion was successful (true) or not (false).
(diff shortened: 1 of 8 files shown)


dnephin:
Does anyone have an opinion about this PR?

philtay:

ernesto-jimenez:
@dnephin commented on the actual issue: https://github.com/stretchr/testify/issues/497#issuecomment-354553709

I'm a bit wary about nudging people into following anti-patterns.

Closing this for now in favour of the discussion on the issue to centralise the argument. We can always reopen the issue if needed.

PR #488, opened 2017-08-14
Fix unprotected call fields access in MethodCalled()

danielheller:
This change fixes a race condition I discovered when a multithreaded test in a service I work on failed under -race.  The included test case simulates that failure (concurrent mutation of a Call with invocations on the mock).  The test will fail with a data race if run under the race detector; the new locking ensures that call fields are not accessed without the protection of the parent mutex.


DAddYE:
@ernesto-jimenez LGTM

PR #490, opened 2017-08-22
ConditionWait assertion

kulti:
Hi!

Sometimes we need to check some state, which setting up in another goroutine. We can use a `Condition` assertion with infinite loop and probably timeout, but could you please add this primitive to `testify` framework?

Thanks.

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -16,6 +16,11 @@ func Conditionf(t TestingT, comp Comparison, msg string, args ...interface{}) bo
 	return Condition(t, comp, append([]interface{}{msg}, args...)...)
 }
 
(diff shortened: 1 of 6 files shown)


philtay:

ernesto-jimenez:
@kulti I have three concerns with the current solution:
1. When `Comparison` never returns `true`, it will leak a goroutine that is running the condition an infinite loop.
2. The infinite loop will get CPU consumption to 100% and compete for resources with other goroutines.
3. It encourages bad practices, by having racy test that rely on waits rather than using things like `context` cancelation.

Closing the PR since you would be able to implement your own condition waits within your `Comparison` funciton using channels and `context`, which would be better and avoid exhausting resources.

Issue #275, opened 2016-02-16
test whether two slices have equal content


LarsFronius:
I wonder if this inside the scope of testify.
Basically I have two slices of a certain type, which can be of random order and I want to be able to compare, that any of the data of one slice appears on the other slice as well.
A reflect.deepequal does not get the job done, neither does assert.Equal, so is this something to consider as a feature for testify?
e.g.

```go
type foobar struct {
    Key string,
    Value string
}

var foobar_slice []*foobar
foobar_slice = append(settings_slice, &pushservice.Setting{Key: "foobar", Value: "barbaz"})
foobar_slice = append(settings_slice, &pushservice.Setting{Key: "bazfoo", Value: "foobaz"})

var barbaz_slice []*foobar
barbaz_slice = append(settings_slice, &pushservice.Setting{Key: "bazfoo", Value: "foobaz"})
barbaz_slice = append(settings_slice, &pushservice.Setting{Key: "foobar", Value: "barbaz"})

assert.EqualSlice(t, foobar_slice, barbaz_slice)
```

ernesto-jimenez:
Hi @LarsFronius, `assert.EqualSlice` is a bit misleading, since two slices with different order are different.

Before finding the right name for the method there would need to be some clarification on the behaviour:

 - Is it just confirming that all events from the `actual` slice are contained in the `expected` slice and vice versa?
 - What would happen with duplicated elements? e.g: `actual` and `equal` have the same elements but `actual` has one that is duplicated. Would the assertion fail?

Sounds like you might be thinking about `assert.ContainsSameElements`.

ernesto-jimenez:
I think this would be out of scope.

It's probably better for the developer to sort both arrays and then call `assert.Equal`

narg95:
`assert.ContainsSameElements` would be great! 
In my case, I have a integration test and I get some data from the database that it may come in different order every time I run it, and I just want to assert that the elements that I expect are there. This function will make my assertion easier. 
I know it is closed but +1.

DennisMao:
Now it's  the  `assert.ElementsMatch(t, expectedSlice, actualSlice)`

sarathsp06:
@DennisMao is there a way to do the same in `Mock.On`

ilovejs:
...
```
Expected :[]string{"g1", "f3", "ze"}
Actual   :[]string{"f3", "g1", "ze"}
```

fangpinsern:
[...]

PR #491, opened 2017-08-22
ElementsMatch array/slice assertion ignoring order

emou:
An assertion that compares the elements of the slices/arrays disregarding the order,
i.e. it checks whether each element in the first slice/array appears the same number of times in it
as in the second slice/array.

This name seemed like it would be easy to find.

Possible alternatives for the name:
- ContainsSameElements
- IsPermutation (C++: http://en.cppreference.com/w/cpp/algorithm/is_permutation)
- MatchArray (rspec: http://www.rubydoc.info/github/rspec/rspec-expectations/RSpec/Matchers:match_array)
- EqualSorted
- Other ideas?

This implementaiton is O(N^2), while sorting both lists first would be O(nlogn).
However, this one doesn't need to copy both lists and sort them (which would be more complicated code). Also, it uses less memory (N bytes of memory instead of 8 x N).

I realize this was deemed as out of scope
https://github.com/stretchr/testify/issues/275
but I decided to give it a shot as I needed it also.


adamwg:
I would use this if merged. That said, you can emulate it by checking that two lists are subsets of one another:

```go
assert.Subset(t, a, b)
assert.Subset(t, b, a)
```

asko:
Is there progress on this?

philtay:

ernesto-jimenez:
Thanks @emou, merging.

PR #492, opened 2017-08-23
Make NotSubset actually fail the test on nil subset

emou:
The assertion function was simply returning false, which doesn't actually fail a test.
An example test that should have failed but doesn't:

  func TestNotSubset(t *testing.T) {
      assert.NotSubset(t, []string{"x"}, nil)
  }

NOTE: It might be worthwhile refactoring all the tests to make sure to not only check whether the return value of an assertion was false, but also check if the test would actually be failed.


philtay:

ernesto-jimenez:
@emou thanks for including the test too. Merging.

PR #495, opened 2017-08-31
Add suite variant for benchmarks

JelteF:
I wanted setup, teardown and easy access to assert/require for benchmarks i'm working on. I think it might be useful for other people as well.


JelteF:
One thing I'm not sure about if we want to reuse the `SetupTest`, `TearDownTest`, `BeforeTest` and `AfterTest` interfaces or if new ones for `Benchmark` should be created. I kept it like this for now for sake of simplicity, but I don't mind adding the extra interfaces if it is decided that that improves clarity. 
If the interfaces are reused the documentation for them should at least be updated to include that they are run before and after benchmarks as well.

philtay:

suite/suite.go
```diff
@@ -84,21 +121,13 @@ func Run(t *testing.T, suite TestingSuite) {
...
-							afterTestSuite.AfterTest(methodFinder.Elem().Name(), method.Name)
-						}
-						if tearDownTestSuite, ok := suite.(TearDownTestSuite); ok {
-							tearDownTestSuite.TearDownTest()
-						}
+						afterHandlers(suite, suiteName, method.Name)
 						suite.SetT(parentT)
 					}()
+
 					method.Func.Call([]reflect.Value{reflect.ValueOf(suite)})
```
johnny87: to measure the time of the function. You should to add b.RestartTim and then b.startTime and after the calling b.StopTime. Otherwise the result would be wrong

PR #506, opened 2017-09-21
Allow a method call to be optional

shhsu:
Sometimes when we are testing a large and complex method, some of its dependencies on a utility level might or might not be called depends on whether the dependee has encountered an error and bails out early.

I added a Maybe() method on the Call struct so we mark a method call optional so when AssertExpectations is called it wouldn't trigger an error.


philtay:

ernesto-jimenez:
@shhsu created a new PR in order to be able to refactor the test: #506 

You still appear as the commit author.

Closing this in favour of the other one.

Thanks!
Ernesto

PR #513, opened 2017-10-17
correct diffArguments call

yiminc:
The expected and actual arguments are swapped in the call to diffArguments.


philtay:

ernesto-jimenez:
@yiminc why would you swap the arguments?

I believe the current error is OK, since we expected to be able to use the provided arguments and the closest one we found is different.

We cannot say the closest match is what we expected.

If the issue is with how clear the copy of the error is, it would be great to discuss that in an issue.

Could you create one with the reason why you find the current message confusing so we can discuss an alternative format?

Closing the PR since the actual change will probably be different.

Thanks!

PR #523, opened 2017-11-27
Remove "?" after url when query string (values) is not provided

azbshiri:
Why?
To be able test RESTful API endpoints without "?" after. for example
"/appointemnts" instead of "/appointments?" or "/appointments/1?".

### DIFF
--- a/assert/http_assertions.go
+++ b/assert/http_assertions.go
@@ -12,7 +12,10 @@ import (
 // an error if building a new request fails.
 func httpCode(handler http.HandlerFunc, method, url string, values url.Values) (int, error) {
 	w := httptest.NewRecorder()
-	req, err := http.NewRequest(method, url+"?"+values.Encode(), nil)
+	if values != nil {
+		url = fmt.Sprintf("%s?%s", url, values.Encode())
+	}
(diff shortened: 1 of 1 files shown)


philtay:

assert/http_assertions.go
```diff
@@ -12,7 +12,10 @@ import (
 // an error if building a new request fails.
 func httpCode(handler http.HandlerFunc, method, url string, values url.Values) (int, error) {
 	w := httptest.NewRecorder()
-	req, err := http.NewRequest(method, url+"?"+values.Encode(), nil)
+	if values != nil {
+		url = fmt.Sprintf("%s?%s", url, values.Encode())
+	}
+	req, err := http.NewRequest(method, url+values.Encode(), nil)
```
ernesto-jimenez: This change is incorrect since it is duplicating `values.Encode()`

Instead of `/?foo=bar` it would use `/?foo=barfoo=bar`

ernesto-jimenez:
@azbshiri closing for now since the PR has a bug and the change is just in style.

Feel free to issue a new one that is fixed and has the corresponding unit tests.

Thanks!

PR #527, opened 2017-12-10
generalize isEmpty

vincentcr:
I had some issues using the `Empty` assertion with a custom string type: let's say I define `type TString string`, then `assert.Empty(t, TString(""))` would fail. The `Empty` would also not work against empty custom structs. 

Fixed by doing a `reflect.DeepEqual` against the type's `reflect.Zero`.


philtay:

ernesto-jimenez:
@vincentcr the case for `TString("")` worked as expected, but the other two didn't.

Merging. Thanks!

Issue #464, opened 2017-06-13
After() should block _after_ the mocked method is called


skonefal:
Currently the timer in `After(duration)` method is started immediately when mock is defined. This is a problem when user would like to test two mocked methods that are run one after another. 

For example in following snippet:
```go
func doWaiting(w1 Waiter, w2 Waiter) {
  w1.Wait()
  w2.Wait()
}

func TestWaiter() {
  w1 := make(waiterMock)
  w2 := make(waiterMock)
  w1.On("Wait").After(100 * time.Milisecond)
  w2.On("Wait").After(100 * time.Milisecond)
 
  doWaiting(w1, w2)
}
```

[Documentation](http://godoc.org/github.com/stretchr/testify/mock#Call.After) specifies, that the **call** will block so that **expected** execution time of `doWaiting` function is 200ms.

Unfortunately, the **actual** execution time is 100ms, because timer in `After` starts when mock is defined.

This leads to mock code that misleading:

```go
  w1.On("Wait").After(100 * time.Milisecond)
  w2.On("Wait").After(200 * time.Milisecond)
```

dineshkumar-cse:
I've raised a PR https://github.com/stretchr/testify/pull/529 comments are welcome.

ernesto-jimenez:
Fixed in #529

PR #529, opened 2017-12-29
Fixing After(duration) to wait during method call

dineshkumar-cse:
After was using call.WaitUntil(time.After(duration)),
`ch := time.After(duration)` returns a channel with the timer immediately started.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -55,6 +55,8 @@ type Call struct {
 	// receives a message or is closed. nil means it returns immediately.
 	WaitFor <-chan time.Time
 
+	waitTime time.Duration
+
 	// Holds a handler used to manipulate arguments content that are passed by
 	// reference. It's useful when mocking methods such as unmarshalers or
 	// decoders.
@@ -133,7 +135,10 @@ func (c *Call) WaitUntil(w <-chan time.Time) *Call {
 //
 //    Mock.On("MyMethod", arg1, arg2).After(time.Second)
 func (c *Call) After(d time.Duration) *Call {
-	return c.WaitUntil(time.After(d))
+	c.lock()
+	defer c.unlock()
+	c.waitTime = d
+	return c
 }
 
 // Run sets a handler to be called before returning. It can be used when
@@ -326,18 +331,12 @@ func (m *Mock) MethodCalled(methodName string, arguments ...interface{}) Argumen
 		}
 	}
 
-	switch {
-	case call.Repeatability == 1:
+	if call.Repeatability == 1 {
 		call.Repeatability = -1
-		call.totalCalls++
-
-	case call.Repeatability > 1:
+	} else if call.Repeatability > 1 {
 		call.Repeatability--
-		call.totalCalls++
-
-	case call.Repeatability == 0:
-		call.totalCalls++
 	}
+	call.totalCalls++
 
 	// add the call
 	m.Calls = append(m.Calls, *newCall(m, methodName, arguments...))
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -22,3 +22,4 @@ _testmain.go
 *.exe
 
 .DS_Store
+*.swp
```
ernesto-jimenez: Can you remove this from the `.gitignore`?

It is a detail from your editor's configuration. You set a global `.gitignore` on your machine so Git always ignores `.swp` files.

mock/mock.go
```diff
@@ -52,6 +52,8 @@ type Call struct {
 	// receives a message or is closed. nil means it returns immediately.
 	WaitFor <-chan time.Time
 
+	WaitTime time.Duration
```
ernesto-jimenez: Unexport this, so it has to be set with `After`.

mock/mock.go
```diff
@@ -335,6 +332,7 @@ func (m *Mock) MethodCalled(methodName string, arguments ...interface{}) Argumen
 	if call.WaitFor != nil {
 		<-call.WaitFor
 	}
+	time.Sleep(call.WaitTime)
```
ernesto-jimenez: Rather than doing `time.Sleep`, do `call.WaitFor = time.After(call.waitTime)` before the if.

Otherwise it might be possible to set two different waiting times for the same call.

mock/mock_test.go
```diff
@@ -1221,6 +1222,45 @@ func Test_MockReturnAndCalledConcurrent(t *testing.T) {
 	wg.Wait()
 }
 
+type timer struct{ Mock }
+
+func (s *timer) GetTime(i int) string {
+	return s.Called(i).Get(0).(string)
+}
+
+func TestAfterTotalWaitTimeWhileExecution(t *testing.T) {
```
ernesto-jimenez: These tests slow down the test run by ˜3 seconds.

Can you reduce wait times a lot more to keep tests quick?

mock/mock.go
dineshkumar-cse: Then in that case we would be losing the channel which is set via `time.WaitUntil` (which i can control when to send data not just with timer.

mock/mock.go
dineshkumar-cse: Followed thinking its a convention, because other members are exported.

mock/mock.go
ernesto-jimenez: That is not allowed right now anyway, since `After` currently calls `WaitUntil` and overrides the property.

The change I mention would be aligned with how things work right now.

By the way, remember to lock/unlock the mutex in `After`.

dineshkumar-cse: @ernesto-jimenez i've added `time.Sleep(waitTime)` if `waitFor` is nil, adding it before would make the duration override, in this case channel is used. Anyway It depends on the client usage `WaitUntil(ch).After(d) or .After(d).WaitUntil(ch)` and not defined in here. so i hope this is fine.

mock/mock_test.go
```diff
@@ -1221,6 +1222,36 @@ func Test_MockReturnAndCalledConcurrent(t *testing.T) {
...
+	waitDuration := 10
```
ernesto-jimenez: would `1ms` be enough here?

The faster we can make it, the better :)

ernesto-jimenez:
@dineshkumar-cse awesome! do you mind squashing into a single commit?

Thanks

dineshkumar-cse:
Done. @ernesto-jimenez Thanks for reviewing it. We can close https://github.com/stretchr/testify/issues/464

ernesto-jimenez:
Cool, will merge once CI passes.

Thanks!

fixes https://github.com/stretchr/testify/issues/535

PR #546, opened 2018-01-13
[FIX] Cleanup code

im-kulikov:
- `dep prune`
- collides with package name
- remove unused aliases for import
- prelloc


im-kulikov:
fail here:
https://github.com/stretchr/testify/blob/87b1dfb5b2fa649f52695dd9eae19abe404a4308/suite/suite_test.go#L35-L43

im-kulikov:
Ping @ernesto-jimenez

mock/mock_test.go
```diff
@@ -1260,7 +1260,7 @@ func Test_Arguments_Bool(t *testing.T) {
 func Test_WaitUntil_Parallel(t *testing.T) {
 
 	// make a test impl object
-	var mockedService *TestExampleImplementation = new(TestExampleImplementation)
+	var mockedService = new(TestExampleImplementation)
```
dineshkumar-cse: Infact you don't need the `:=`

im-kulikov: `var mockedService = new(TestExampleImplementation)`
and 
`mockedService := new(TestExampleImplementation)`
doesn't matter

dineshkumar-cse: Yeah, you've refactored a lot with convention, so suggested. BTW i'm curious whether we can remove vendor.

im-kulikov: not sure, that I understand correctly this part `and could also share context behind main.go file change`

- vendor - yeah, we can remove it.. or add in gitignore

im-kulikov:
any updates @ernesto-jimenez?

.travis.gogenerate.sh
```diff
@@ -5,7 +5,10 @@ if [[ "$TRAVIS_GO_VERSION" =~ ^1\.[45](\..*)?$ ]]; then
 fi
 
 go get github.com/ernesto-jimenez/gogen/imports
+go get golang.org/x/tools/cmd/goimports
 go generate ./...
+goimports -w ./..
```
ernesto-jimenez: This will actually modify the source code.

The point of this check is to confirm `go generate ./...` is idempotent and all the generated code has been committed, so we should not be running `goimports` here.

If we wanted to run `goimports`, we could possibly do it as part of the code generation, but it is only going to be reordering imports.

ernesto-jimenez:
Thanks!

.travis.gogenerate.sh
im-kulikov: this needs for correct formatting:
https://github.com/im-kulikov/testify/blob/b446e260ab346c645d75436340301f8a4e385088/.travis.gogenerate.sh#L11-L16



this code make changes too - https://github.com/im-kulikov/testify/blob/b446e260ab346c645d75436340301f8a4e385088/.travis.gofmt.sh#L3-L7

im-kulikov: go generate bad formatting and code has incorrect formatting..
thats why I run `go import` to check resort imports like GoLand / VSCode
otherwise it fails at this check - https://github.com/stretchr/testify/blob/12b6f73e6084dad08a7c6e575284b177ecafbc71/.travis.yml#L13

ernesto-jimenez: [We format the code](https://github.com/stretchr/testify/blob/master/_codegen/main.go#L83) already using `[go/format](https://godoc.org/go/format)` before writing it.

The check you pointed out on `.travis.gofmt.sh` doesn't actually modify the code, it checks that the code is properly formatted in CI.

```
usage: gofmt [flags] [path ...]
[...]
  -l    list files whose formatting differs from gofmt's
```

Your changes should not fail the `gofmt` check, since the generated code is already properly formatted.

im-kulikov: oh.. I'll try to explain:

after `go generate`
``` go
 import (
    assert "github.com/stretchr/testify/assert"
    http "net/http"
    url "net/url"
    time "time"
)
```

after VSCode / GoLand editing:
```
import (
	"net/http"
	"net/url"
	"time"

)
```

VSCode / GoLand cleanup unused import-aliases and resort in two groups (native and external)

im-kulikov: after that, we run gofmt and... CI fails

im-kulikov: why? because go generate incorrect sorting (not like go import), and after that make changes in existing code

ernesto-jimenez: `goimports` does some extra formatting on top of `gofmt`. Our code generation formats the code according to `gofmt` and I never use `goimports` for code generation because it can cause issues if the code was missing an import and you have two packages with that same name in your `GOPATH`.

There are three rules I follow for generated code in this regard:
1. Generated code must be formatted with `go/format`
2. Generated code must not rely on `goimports`
3. `go generate ./...` should be idempotent, it's results checked in, and CI should check it ran

This change breaks points 2  and 3.

How the editor handles the generated files is not of much importance because:
* Generated code must never be edited by hand
* If your setup makes any changes to those files, `go generate ./...` should overwrite them anyway.

PR #550, opened 2018-01-25
Prune vendor & update deps

Hanzei:
`dep` got a new release (https://golang.github.io/dep/blog/2018/01/23/announce-v0.4.0.html) which allows [options for prune](https://golang.github.io/dep/docs/Gopkg.toml.html#prune). This removes all non go files and test files and makes `vendor` smaller. This PR also updates the dependencies.


ernesto-jimenez:
Thanks!

PR #555, opened 2018-02-02
Adding logging when mock assertions fails

### LINKED ISSUE #507: AssertExpectations should tell you what mock failed an assertion
If you use `mock.AssertExpectationsForObjects()` to test multiple mocks and one of them fails an assertion, you have no idea which mock actually failed.

Another issue is that you also have no idea what assertions failed.

dineshkumar-cse:
Adding log for assertExpectationsForObjects mock failure.
```
func AssertExpectationsForObjects(t TestingT, testObjects ...interface{}) bool {
```
fixes #507

### DIFF
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -3,6 +3,7 @@ package mock
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


mock/mock_test.go
```diff
@@ -1323,6 +1323,32 @@ func (s *timer) GetTime(i int) string {
 	return s.Called(i).Get(0).(string)
 }
 
+type tCustomLogger struct {
+	*testing.T
+	err string
```
ernesto-jimenez: how about we do
```
type tCustomLogger struct {
    *testing.T
    logs []string
}
```

That way we can use `logs` to have an entry in the array for each `Logf` call.

PR #557, opened 2018-02-11
Display arguments Diff during panic

### LINKED ISSUE #556: Inaccurate MatchedBy error message
This might sorta-maybe be related to #432.

I'm trying to use MatchedBy to verify that I get the proper object. When I have MatchedBy just return true tests work:

``` go
s.foo.On("UpdateFoo", mock.MatchedBy(func(foo *model.Foo) bool {
    return true
})).Return(nil)
```

However, as soon as I'm actually trying to compare values on foo like this:

``` go
s.foo.On("UpdateFoo", mock.MatchedBy(func(foo *model.Foo) bool {
    // LastSeen is a time.Time
    return foo.LastSeen.Equal(timer.Timestamp())
})).Return(nil)
```

Then I get this panic:

```
UpdateFoo(*model.Foo)
		0: &model.Foo{ /* stuff */}

The closest call I have is: 

UpdateFoo(mock.argumentMatcher)
		0: mock.argumentMatcher{fn:reflect.Value{typ:(*reflect.rtype)(0x769b00), ptr:(unsafe.Pointer)(0x8036c0), flag:0x13}}
```

**Update:** After plenty of staring at my test cases and spotting that my test values didn't line up with what I wanted to match it against, I isolated this to Testify returning that error as soon as MatchedBy can't match the object we're testing. Simply having MatchedBy return false will give the same error.

That error message should be changed to give some information about what caused it, in this case MatchedBy() failing to match.

devdinu:
when expected method call wasn't found, but there's a closest method display the diff during panic.
especially this helps for argumentMatcher. closes #556

### DIFF
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -8,9 +8,10 @@ import (
 	"testing"
 	"time"
 
+	"runtime"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"runtime"
 )
 
 /*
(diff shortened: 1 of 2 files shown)


mock/mock_test.go
```diff
@@ -1384,6 +1384,23 @@ func TestAfterTotalWaitTimeWhileExecution(t *testing.T) {
...
+		}
+	}()
+
+	m := new(timer)
+	m.On("GetTime", MatchedBy(func(i int) bool { return false })).Return("SomeTime").Once()
+
+	res := m.GetTime(1)
+	require.Equal(t, "SomeTime", res)
+	m.AssertExpectations(t)
+}
```
ernesto-jimenez: Should we add a test without `MatchedBy` to make sure in the future we avoid messing up the messages in those cases?

devdinu: i've updated the commit with a test for that case.

ernesto-jimenez:
@devdinu there is a bug when selecting which diff to show. Here is an example test that shows the bug and the fix: https://github.com/stretchr/testify/commit/aa1c55e0270653d4d2aa6035ba012b693e1f53a7

ernesto-jimenez:
@devdinu can you fixup the commit into a single one?

Thanks!

PR #563, opened 2018-02-25
Bump go version to 1.10

Hanzei:
Since go 1.10 go versions must be set as strings

[Building a Go Project \- Travis CI](https://docs.travis-ci.com/user/languages/go/#What-This-Guide-Covers)
> Note that, in order to choose Go 1.10, you must use go: "1.10" (a string), not go: 1.10 (a float). Using a float results in the use of Go 1.1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ language: go
 sudo: false
 
 go:
-  - 1.7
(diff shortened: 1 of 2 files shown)


ernesto-jimenez:
Thanks!

PR #569, opened 2018-03-06
Add T object and Test method to Mock

### LINKED ISSUE #489: Should mock really panic instead of t.Fail?
I'm not sure which is best overall, but I'd like to share my experience just to bring this into consideration.

I was testing a component of my software which will recover whenever a panic occurs, so having an unexpected method call panic broke the logic of my test.

If mock was passed the testing.T object and called t.Fail instead of panic, my test would work fine. That said, I can also imagine circumstances where you'd rather have a panic.

It might be good to make this behaviour configurable. The way I would do this is by adding an exported field to the mock.Mock struct `T *testing.T`, then if `T != nil` call `panic`, otherwise `T.Error`

posener:
This makes is possible to fail the test instead of panicing in case
that the method was called with unexpected arguments.

Fixes #489

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -501,7 +501,7 @@ type Arguments []interface{}
 const (
 	// Anything is used in Diff and Assert when the argument being tested
 	// shouldn't be taken into consideration.
-	Anything string = "mock.Anything"
+	Anything = "mock.Anything"
 )
 
 // AnythingOfTypeArgument is a string that contains the type of an argument
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -191,6 +191,10 @@ type Mock struct {
 	// Holds the calls that were made to this mocked object.
 	Calls []Call
 
+	// T is An optional variable that holds the test struct, to be used when an
+	// invalid mock call was made.
+	T TestingT
+
 	// TestData holds any data that might be useful for testing.  Testify ignores
 	// this data completely allowing you to do whatever you like with it.
 	testData objx.Map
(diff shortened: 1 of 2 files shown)


mock/mock.go
```diff
@@ -191,6 +191,10 @@ type Mock struct {
 	// Holds the calls that were made to this mocked object.
 	Calls []Call
 
+	// T is An optional variable that holds the test struct, to be used when an
+	// invalid mock call was made.
+	T TestingT
```
ernesto-jimenez: Can you unexport `TestingT`?

The package already exports too much of it's internals and we should keep new things to a minimum.

mock/mock.go
```diff
@@ -501,7 +526,7 @@ type Arguments []interface{}
 const (
 	// Anything is used in Diff and Assert when the argument being tested
 	// shouldn't be taken into consideration.
-	Anything string = "mock.Anything"
+	Anything = "mock.Anything"
```
ernesto-jimenez: FYI, for future contributions, it would be best to keep this kind of things in a different PR.

mock/mock_test.go
```diff
@@ -92,6 +92,34 @@ func (i *TestExampleImplementation) TheExampleMethodFuncType(fn ExampleFuncType)
 	return args.Error(0)
 }
 
+// MockTestingT mocks a test struct
+type MockTestingT struct {
+	Mock
```
ernesto-jimenez: I would rather avoid using `mock` to mock a dependency for its own test :)

Let's implement this with a simple mock instead to keep the tests for the package depending on the package itself working correctly :)

mock/mock.go
posener: I didn't add the `TestingT` interface you mean unexport `T` I guess?

posener:
Thanks @ernesto-jimenez , pushed the fixes you asked

posener:
I'm closing this (not deleting, so you can still merge)
I think I have a nicer API, I'll open a new PR, you can merge the one you like better..

posener:
Sorry, the new API didn't work out for me.
I think that once this is merged, we could add a `New` function in mockery to make the creation of the mock and assigning the test field in one command.

PR #572, opened 2018-03-11
 Add T object and Test method to Mock

### LINKED ISSUE #489: Should mock really panic instead of t.Fail?

posener:
This makes is possible to fail the test instead of panicing in case that the method was called with unexpected arguments.

Instead of #569
Fixes #489

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -191,6 +191,10 @@ type Mock struct {
 	// Holds the calls that were made to this mocked object.
 	Calls []Call
 
+	// test is An optional variable that holds the test struct, to be used when an
+	// invalid mock call was made.
+	test TestingT
+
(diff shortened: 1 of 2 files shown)


posener:
@ernesto-jimenez thanks, rebased and resolved

PR #576, opened 2018-03-14
add a few function types to assist table driven tests

coryb:
Basically this PR is just adding these function aliases:
```
type ValuesChecker func(TestingT, interface{}, interface{}, ...interface{}) bool
type ValueChecker func(TestingT, interface{}, ...interface{}) bool
type BoolChecker func(TestingT, bool, ...interface{}) bool
type ErrorChecker func(TestingT, error, ...interface{}) bool
```
I have added tests and Examples so there will example usage in the docs.

### Why?
I find myself frequently using table driven tests to test various edge cases of my functions, but I generally want to test both the positive and negative cases.

Lets say I have a trival function `func MyFunc(input string) error`, typically the data tests would look something like this (without `testify`):
```
    tests := []struct {
        name    string
        arg     string
        wantErr bool
    }{
        {"good test", "good", false},
        {"bad test", "bad", true},
    }

    for _, tt := range tests {
            if err := MyFunc(tt.arg); (err != nil) != tt.wantErr {
                t.Errorf("MyFunc() error = %v, wantErr %v", err, tt.wantErr)
            }
        })
    }
```

With `testify` it gets a bit worse with the conditional:
```
    for _, tt := range tests {
            if err := MyFunc(tt.arg); (err != nil) != tt.wantErr {
                assert.Error(t, err, tt.name)
            } else {
                assert.NoError(t, err, tt.name)
            }
        })
    }
```

But with the function aliases in this PR, I can make more clear what I am testing (in my opinion anyway):
```
    tests := []struct {
        name  string
        arg   string
        check assert.ErrorChecker
    }{
        {"good test", "good", assert.NoError},
        {"bad test", "bad", assert.Error},
    }

    for _, tt := range tests {
            tt.check(t, MyFunc(tt.arg), tt.name)
        })
    }
```

I could obviously just add the function aliases to my project tests, but I think it useful enough that others might want to use them as well, and it would be nice just having them directly in testify.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -27,6 +27,22 @@ type TestingT interface {
 	Errorf(format string, args ...interface{})
 }
 
+// ComparisonAssertionFunc is a common function prototype when comparing two values.  Can be useful
+// for table driven tests.
+type ComparisonAssertionFunc func(TestingT, interface{}, interface{}, ...interface{}) bool
+
+// ValueAssertionFunc is a common function prototype when validating a single value.  Can be useful
+// for table driven tests.
+type ValueAssertionFunc func(TestingT, interface{}, ...interface{}) bool
+
+// BoolAssertionFunc is a common function prototype when validating a bool value.  Can be useful
+// for table driven tests.
+type BoolAssertionFunc func(TestingT, bool, ...interface{}) bool
+
+// ValuesAssertionFunc is a common function prototype when validating an error value.  Can be useful
+// for table driven tests.
+type ErrorAssertionFunc func(TestingT, error, ...interface{}) bool
+
 // Comparison a custom function that returns true on success and false on failure
 type Comparison func() (success bool)
 
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -2,6 +2,7 @@ package assert
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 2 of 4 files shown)


ernesto-jimenez:
Thank @coryb,

Sounds like an interesting idea!

Have you thought about the following points:

* Alternative naming to `Checker` suffix? I'm wondering whether it is clear enough or something like `BoolAssertionFunc` would be clearer.
* Are `(Value|Values)Checker` names clear enough? Would `ValueAssertionFunc` and `ComparisonAssertionFunc` be clearer?
* Should we add some examples to the docs of the types themselves?
* Do you think we should add the same types to the `require` package for symmetry?

Other than that, I would recommend avoid adding `tt.name` on the checks `msgAndArgs`. That information would already appear in the thanks to `t.Run`. If we were to do `testify` v2, we would probably remove the `msgAndArgs` from all assertions.

coryb:
Hey, thanks for spending the time to think about this.  

1) Yeah, I was not super happy with the `Checker` name and I didnt want to spend much more time thinking about it I wanted to get your opinions on it :)  `AssertionFunc` prefix sounds reasonable to me.

2) `ValueAssertionFunc` and `ComparisonAssertionFunc` both sound good to me, I was also not fond of the `Value` vs `Values` but didn't think of anything better.

3) Maybe I don't understand this question, the `Example*` funcs in the `assertions_test.go` will automatically show up in the godoc documentation like this:
image
Did you want something more? 
In my first commit the Examples actually used to run as part of the `go test` process, but `govet` didn't like that the Examples were non-niladic (had to pass in the `t *testing.T` object).  With the second commit it just validates they are valid syntax, but does not run them.

4) Yeah, I think it makes sense for `require` also, I was uncertain how `require` was kept in sync with `assert` and I didn't dig too deep there, so i will add it in there.

I will also remove the `tt.name` from the `msgAndArgs` in the examples, I agree it is redundant.

coryb:
I updated the PR per your feedback: changed names, added to `require`, removed `tt.name` from `msgAndArgs`.
Here is an example of the documentation for the `ComparisonAssertionFunc`:
image

coryb:
@ernesto-jimenez, rebased and pushed.  Thanks!

ernesto-jimenez:
@coryb merged. Thanks!

PR #577, opened 2018-03-19
Add placeholder example to docs

shaneHowearth:
This adds an example to the docs that uses a placeholder in the On arguments


ernesto-jimenez:
@shaneHowearth thanks! :)

PR #579, opened 2018-03-21
Use go-cmp.Equal instead of reflect.DeepEqual

### LINKED ISSUE #535: Use go-cmp instead of reflect.DeepEqual
The package `github.com/google/go-cmp` is better suited for comparison in tests than `reflect.DeepEqual`.

We should replace our `DeepEqual` usage with `go-cmp`.

Replacement should be backwards-compatible.

posener:
Closes #535 

This works with https://github.com/posener/go-cmp/tree/detect-cycles2 branch of go-cmp which is in PR here: https://github.com/google/go-cmp/pull/79

* Will give us more control on comparing objects.
* Notice the differences in the diffs, if they are acceptable.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -18,7 +18,6 @@ import (
 	"unicode/utf8"
 
 	"github.com/davecgh/go-spew/spew"
-	"github.com/pmezard/go-difflib/difflib"
 	yaml "gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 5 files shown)


ernesto-jimenez:
@posener 🙌 for the work on go-cmp!

It seems you went with a [new PR](https://github.com/google/go-cmp/pull/85), as soon as that gets merged, I would be keen to rebase this and get it merged.

kishaningithub:
The [new PR](https://github.com/google/go-cmp/pull/85) got merged! Waiting for this to be merged and released

brackendawson:
With this being a breaking change and testify v2 not planned, this PR can't progress so I'm closing it now. If you feel this is not correct then feel free to at me here with the reason.

PR #581, opened 2018-03-24
Moving call and arguments to separate file, and added tests for call

devdinu:
- Moved out call, and arguments to separate file in mock package, since the mock file is too big.
- added test for basic call functionality, and we could more tests later


boyan-soubachov:
This looks like a good separation. I would be happy to review and get this over the line if you rebase the PR and bring it up to date :)

dolmen:
That was a good separation. But this is now outdated by 5 years of changes.

A similar refactor would be good.

PR #582, opened 2018-03-24
Refactor ObjectsAreEqual()

HaraldNordgren:
Found this when looking into implementing a FileExists feature. Then I realized it already is there in later versions than what I was using.

Not sure if you guys want "random" refactors. But submitting this anyway to try to make the code even clearer.


erdnaxeli:
This code is exactly the same. You are just doing "if exp is not []byte" instead of "if exp is []byte". I don't see the point.

HaraldNordgren:
You are correct. Close it if you want to.

PR #584, opened 2018-03-28
Add non nillable type check for NotNil assertion

### LINKED ISSUE #570: assert: NotNil accepts non-nillable types and doesn't fail
https://github.com/stretchr/testify/blob/b89eecf5ca5db6d3ba60b237ffe3df7bafb7662f/assert/assertions.go#L390-L410

`isNil` will always return false, so `NotNil` will always return true for int, string, struct, etc.

I would expect it to error with "invalid type".

medzin:
This commit adds non nillable type check to NotNill assertion. When non nillable type is passed to that assertion - it will fail. It resolves #570.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -409,6 +409,9 @@ func NotNil(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
+	if isNonNillableType(object) {
+		return Fail(t, fmt.Sprintf("Non nillable type %T used in NotNil assertion.", object), msgAndArgs...)
+	}
 	if !isNil(object) {
 		return true
 	}
@@ -428,6 +431,16 @@ func isNil(object interface{}) bool {
 	}
 
 	return false
+}
+
+func isNonNillableType(object interface{}) bool {
+	switch reflect.ValueOf(object).Kind() {
+	case reflect.Array, reflect.Chan, reflect.Func, reflect.Interface, reflect.Map,
+		reflect.Ptr, reflect.Slice, reflect.UnsafePointer:
+		return false
+	default:
+		return true
+	}
 }
 
 // Nil asserts that the specified object is nil.
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -309,7 +309,6 @@ func TestFormatUnequalValues(t *testing.T) {
 }
 
 func TestNotNil(t *testing.T) {
-
 	mockT := new(testing.T)
 
 	if !NotNil(mockT, new(AssertionTesterConformingObject)) {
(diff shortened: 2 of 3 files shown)


assert/assertions.go
```diff
@@ -407,6 +407,9 @@ func NotNil(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
+	if isNonNillableType(object) {
```
devdinu: we should make similar changes to `Nil`, keeping the contract `!Nil(anyType) == NotNil(anyType)` same.
also we could log it, instead of failing the test. i left a comment in #570

dnephin: I don't see any reason this should be a log. If the function is called with a value that is not nil, it should fail the test. That's the entire purpose of the assertion.

medzin: I agree with @dnephin - incorrectly used assertions should not work.

assert/assertions_test.go
```diff
@@ -306,7 +306,38 @@ func TestNotNil(t *testing.T) {
...
+		{func() {}, false},
+		{make(chan struct{}), false},
+		{&someString, false},
+		{[]string{}, false},
+		{map[string]string{}, false},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(fmt.Sprintf("valueType=%T", testCase.value), func(subtest *testing.T) {
+			if !(isNonNillableType(testCase.value) == testCase.expectedResult) {
```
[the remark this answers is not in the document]
medzin: @devdinu changed! 😄

assert/assertions.go
```diff
@@ -407,6 +407,9 @@ func NotNil(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
...
+		return Fail(t, "Non nillable type used in NotNil assertion.", msgAndArgs...)
```
ernesto-jimenez: We should probably include what the type is.

```go
Fail(t, fmt.Sprintf("Non nillable type %T used in NotNil assertion.", object), msgAndArgs...)
```

assert/assertions.go
ernesto-jimenez: I think failing is OK.

medzin:
@ernesto-jimenez but how you want to handle interfaces? Interfaces are nullable - even if the value underneath is not, you can assign `nul` to interface type.

assert/assertions_test.go
```diff
@@ -1677,7 +1705,6 @@ func TestValueAssertionFunc(t *testing.T) {
 		value     interface{}
 		assertion ValueAssertionFunc
 	}{
-		{"notNil", true, NotNil},
```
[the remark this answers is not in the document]
medzin: I added string slice here.

PR #585, opened 2018-03-30
compared expected panic to stringified panic

cbelsole:
## Problem
After writing a few panic tests I noticed that I could not get some tests to pass. This was because the object passed to panic was not a string and therefore not equal to the expected string I was passing to `PanicWithValue`. This was being thrown from a third party library that I did not control. My solution was just to use `Panics` instead, but if the function ever changes and hits a different panic I could get a false positive. 

## Solution
The change here is to stringify the panics before you compare to the expected string. That way, no matter what type the panic is you can compare against it and assert that you are seeing the panic you are expecting.

## Example
```go
require.PanicsWithValue(t, "{this should match the panic}", func(){
  panic(struct{ a string }{"this should match the panic"})
}) // fails
```

## Testing
* The tests pass because they are already using strings.
* Added a test for matching against a struct.

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -426,7 +426,7 @@ func Panicsf(t TestingT, f PanicTestFunc, msg string, args ...interface{}) bool 
 // the recovered panic value equals the expected panic value.
 //
 //   assert.PanicsWithValuef(t, "crazy error", func(){ GoCrazy() }, "error message %s", "formatted")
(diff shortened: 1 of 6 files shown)


ernesto-jimenez:
@cbelsole I'm afraid that this would be a breaking change in testify.

What we can definitely do is tweak the error so we print with `%#v` instead of `%v` so you can see you are getting a struct.

If you really care about handling an specific panic from a thirs party library, I would suggest you capture it within your code and decide how to handle it.

ernesto-jimenez:
Making error messages more clear on https://github.com/stretchr/testify/pull/611

PR #588, opened 2018-04-04
Disallow mock.Anything from satisfying missing function parameter

### LINKED ISSUE #587: mock does not enforce arity if parameter is Anything
```go
package main_test

import (
	"testing"

)

type Foo interface {
	Bar()
}

type MockFoo struct {
	mock.Mock
}

// Bar provides a mock function with given fields:
func (_m *MockFoo) Bar() {
	_m.Called()
}

func TestFoo(t *testing.T) {
	foo := new(MockFoo)
	foo.On("Bar", mock.Anything).Return()

	foo.Bar() // expected error for unmocked call
}

```

The logic in `Arguments.Diff` does not check whether a given argument is required by the function signature before checking if the value is `Anything`. Phrased another way, there is no explicit arity check on the mocked method.

dnathe4th:
Fixes #587

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -658,7 +658,6 @@ func (args Arguments) Is(objects ...interface{}) bool {
 func (args Arguments) Diff(objects []interface{}) (string, int) {
 	//TODO: could return string as error and nil for No difference
 
-	var missing = "(Missing)"
 	var output = "\n"
 	var differences int
 
@@ -671,22 +670,14 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 		var actual, expected interface{}
 		var actualFmt, expectedFmt string
 
-		if len(objects) <= i {
-			actual = missing
-			actualFmt = missing
-		} else {
-			actual = objects[i]
-			actualFmt = fmt.Sprintf("(%[1]T=%[1]v)", actual)
+		if len(objects) <= i || len(args) <= i {
+			differences++
+			output = fmt.Sprintf("%s\t%d: FAIL:  inconsistent function arity, mocked argument count (%d) != call argument count (%d) \n", output, i, len(args), len(objects))
+			return output, differences
 		}
 
-		if len(args) <= i {
-			expected = missing
-			expectedFmt = missing
-		} else {
-			expected = args[i]
-			expectedFmt = fmt.Sprintf("(%[1]T=%[1]v)", expected)
-		}
-
+		actual = objects[i]
+		expected = args[i]
 		if matcher, ok := expected.(argumentMatcher); ok {
 			if matcher.Matches(actual) {
 				output = fmt.Sprintf("%s\t%d: PASS:  %s matched by %s\n", output, i, actualFmt, matcher)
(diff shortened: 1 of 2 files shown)


mock/mock.go
```diff
@@ -701,7 +702,9 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 
 			// normal checking
 
-			if assert.ObjectsAreEqual(expected, Anything) || assert.ObjectsAreEqual(actual, Anything) || assert.ObjectsAreEqual(actual, expected) {
+			expectedAnythingAndRequired := assert.ObjectsAreEqual(expected, Anything) && actual != missing
```
devdinu: One suggestion would be to move away from usage of the string, you could change the missing to `var missingArgument = errors.New("(Missing)")`, so we will be comparing on error values, and its easy to change output if required. like this function could return error (custom type) and based on it, the output should be formatted.

mock/mock_test.go
```diff
@@ -1204,6 +1204,18 @@ func Test_Arguments_Diff_DifferentNumberOfArgs(t *testing.T) {
...
+	var args = Arguments([]interface{}{Anything})
+	var diff string
+	var count int
+	diff, count = args.Diff([]interface{}{})
+
+	assert.Equal(t, 1, count)
+	assert.Contains(t, diff, `(Missing) != mock.Anything`)
+
+}
+
```
devdinu: You could add a test case for `Arguments([]interface{}{}).Diff(Arguments([]interface{}{Anything}))`, expected is missing

devdinu (changes requested):

dnathe4th:
How about removing the missing string sentinel all together and returning a new error outright? I kept this new check inside the loop rather than before it in case there are additional errors the user should be informed about.

mock/mock.go
```diff
@@ -669,17 +669,14 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 	for i := 0; i < maxArgCount; i++ {
 		var actual, expected interface{}
 
-		if len(objects) <= i {
-			actual = "(Missing)"
-		} else {
-			actual = objects[i]
+		if len(objects) <= i || len(args) <= i {
```
devdinu: @dnathe4th We should ideally do this as part of `mock.On` function, and we need to consider `varargs`, and have to be tested extensively, lets create an issue and raise PR separately. I would suggest just with previous commit along with test case would be good to close the PR.

dnathe4th: I replied on #587 I am not sure it is possible to do this check as part of `mock.On` because there is no way for a function on `*mock.Mock` to be aware of what struct has embedded it as far as I could tell. 

Functionally these two should be the same, but if you like the previous commit better with the extra checks in the conditional at the end I can revert back to that. This version seems strictly better to me from a readability perspective since it has an explicit check on arg length.

From my digging through the existing tests and my understanding of the `reflect` package, it doesn't seem like varargs needs any special handling since tests like `Test_Mock_On_WithMixedVariadicFunc` pass on this PR, but maybe I am misunderstanding what is needed here.

devdinu: yep, checked varargs are passed as `slice`, so we're good.

lets just pull this out of loop, or print it only once when required.

dnathe4th: Even though this is inside the loop it should only print once because I return early. I implemented it like this do the `differences` counter would still return the total number of errors against the mock, which seemed more correct than returning before the loop with a return value of `differences=1`

If you feel strongly I can pull it outside the loop.

devdinu: `len(objects), len(args)` wouldn't change inside the loop, so wouldn't this still return `differences=1` in first loop itself.

dnathe4th: Hey @devdinu , where did we end up on this? Is this change only acceptable with the check before the loop?

devdinu: @dnathe4th The change looks good. @ernesto-jimenez could help in merging this.

ernesto-jimenez (changes requested):
Could we simplify the logic here?

If we are going to short-circuit the loop when it goes beyond the length of either of the arrays. Could we:
* Check for inconsistent arity before the loop
* Remove maxArg and iterate through `len(args)`
* Remove the length checks

The main issue is that we would miss wether the other arguments matched.

An alternative might be that we keep the loop as it is, add a couple of bool variables: `actualMissing` and `expectedMissing`, and use them on the normal checking.

dnathe4th:
Hey team, how can I move this forward?

dolmen:
This seems still relevant as of 37a3cb7cc084291e6d3c420f4a897b5102752dc8.

PR #593, opened 2018-04-16
fix docs: http assertions

srenatus:
The doc comments were missing the method in two cases.

Q: I might be missing something, but does the `msgAndArgs` business work for `http_assertions`? I don't see those being used (or passed to some other `assert.*` function) in the function bodies of the `HTTP...` methods. 🤔


ernesto-jimenez:
@srenatus thanks!

Instead of `msgAndArgs` you should use the `...f` functions like `[HTTPBodyContainsf ](https://godoc.org/github.com/stretchr/testify/assert#HTTPBodyContainsf)`.

You can append `f` to any assertion to set an error message and arguments.

We would remove `msgAndArgs` if it weren't a breaking change.

Issue #518, opened 2017-11-13
Support diff for strings


KingCrunch:
When comparing strings

```
assert.Equals(t, "foo\nbar", "foo\nbaz")
```

it would be convenient to get the diff of both strings like one gets when comparing objects.

nicolaiskogheim:
It's not nice, but I get the diff by converting strings to arrays.

```go
assert.Equals(t, strings.Split(expected, "\n"), strings.Split(actual, "\n"))
```

ernesto-jimenez:
What should the output be like?

nicolaiskogheim:
I suppose KingCrunch wants the string to be split on newlines so that the diff shows the "foo" part as equal and "bar" vs "baz" as a difference. That's what you get if you do like I did in my previous comment.

I often have two long strings (like some compact JSON), and to get a readable diff I have to prettify the JSON somehow and split the strings on newlines before I pass it to `assert.Equals`.

ernesto-jimenez:
@nicolaiskogheim do you know about [assert.JSONEq](https://godoc.org/github.com/stretchr/testify/assert#JSONEq) ?

nicolaiskogheim:
Nope! Look nice, though, thanks! 
Great that someone fixed the general issue as well.

Issue #589, opened 2018-04-07
improve output


mh-cbon:
hi, small suggestion about the output i have to deal with right now using your test library, thanks for it.

I have terrible output such as 

```sh
Not equal: 
		
	            	expected: "[\n{\n  \"Assignee\": \"1\",\n  \"Filename\": \"filename-1\",\n  \"Line\": 1,\n  \"Message\": [\n    \"multi\",\n    \"line\"\n  ]\n},\n{\n  \"Assignee\": \"\",\n  \"Filename\": \"filename-1\",\n  \"Line\": 2,\n  \"Message\": [\n    \"no-assignee\"\n  ]\n},\n{\n  \"Assignee\": \"2\",\n  \"Filename\": \"filename-2\",\n  \"Line\": 3,\n  \"Message\": [\n    \"message-1\"\n  ]\n},\n{\n  \"Assignee\": \"asticode\",\n  \"Filename\": \"some-file\",\n  \"Line\": 4,\n  \"Message\": [\n    \"I should be false\"\n  ]\n},\n{\n  \"Assignee\": \"astitodo\",\n  \"Filename\": \"testdata/level1.go\",\n  \"Line\": 10,\n  \"Message\": [\n    \"Something else comes here\"\n  ]\n}\n]"
		
	            	actual: "[\n  {\n    \"Assignee\": \"1\",\n    \"Filename\": \"filename-1\",\n    \"Line\": 1,\n    \"Message\": [\n      \"multi\",\n      \"line\"\n    ]\n  },\n  {\n    \"Assignee\": \"\",\n    \"Filename\": \"filename-1\",\n    \"Line\": 2,\n    \"Message\": [\n      \"no-assignee\"\n    ]\n  },\n  {\n    \"Assignee\": \"2\",\n    \"Filename\": \"filename-2\",\n    \"Line\": 3,\n    \"Message\": [\n      \"message-1\"\n    ]\n  },\n  {\n    \"Assignee\": \"asticode\",\n    \"Filename\": \"some-file\",\n    \"Line\": 4,\n    \"Message\": [\n      \"I should be false\"\n    ]\n  },\n  {\n    \"Assignee\": \"astitodo\",\n    \"Filename\": \"testdata/level1.go\",\n    \"Line\": 10,\n    \"Message\": [\n      \"Something else comes here\"\n    ]\n  }\n]\n"
FAIL
```

But, in my atom terminal because there is no forced line break, it appears such as,

```sh
Not equal: 
		
[...]

PR #594, opened 2018-04-27
RFC: support diff display for string variables

timfeirg:
This PR brings diff support to strings, which looks something like this:

image

Which without diff support, would look like this: 

image

I think this could close https://github.com/stretchr/testify/issues/518 https://github.com/stretchr/testify/issues/589

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1338,12 +1338,18 @@ func diff(expected interface{}, actual interface{}) string {
 		return ""
 	}
 
-	if ek != reflect.Struct && ek != reflect.Map && ek != reflect.Slice && ek != reflect.Array {
+	if ek != reflect.Struct && ek != reflect.Map && ek != reflect.Slice && ek != reflect.Array && ek != reflect.String {
 		return ""
 	}
 
-	e := spewConfig.Sdump(expected)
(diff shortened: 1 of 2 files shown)


timfeirg:
request for comment @ernesto-jimenez

devdinu:
@timfeirg You could add a test to to verify `string` with newlines, comparison failure output to match our expectation `Equal(t, "some\nmissing\nbla", "some\nmissing")`

we could also try to split string with whitespace and we could do a diff on it. so it would be helpful for a long single line string.

timfeirg:
I've already add such tests if I'm understanding you correctly, see https://github.com/stretchr/testify/pull/594/files#diff-8592448e051dac4c311b0cfee177add7R262
@devdinu

devdinu:
I would say that test is little cryptic, it uses `*bufferT` and some implementation. Its good to have more tests, i was mentioning we could add `TestStringEqual` and add different test cases for string comparison with spaces, newlines, without whitespace success case etc, also it makes it easier to change the code later.

timfeirg:
making contributions to testify is my biggest dream since I was a child.

need your input on this PR @ernesto-jimenez

ernesto-jimenez:
Thanks! 
🙌

PR #595, opened 2018-05-05
Implement assert.NotEmptyThenClear

HaraldNordgren:
This a helper to simplify a pattern that I use often in my everyday work with `testify`.

Generally when an API endpoint returns a struct, I want to assert that it's equal to an expected model, but a some fields have auto-generated values like a timestamps or UUIDs that cannot be predicted.

The options then are
1) Assert fields separately instead of asserting the whole struct, which makes for very verbose code.
2) Check that the unpredictable fields are non-empty and then clear, which is cleaner but also a bit verbose. The PR is trying to achieve this option but in an cleaner way.

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -349,6 +349,17 @@ func NotEmptyf(t TestingT, object interface{}, msg string, args ...interface{}) 
 	return NotEmpty(t, object, append([]interface{}{msg}, args...)...)
 }
 
+// NotEmptyThenClearf asserts that the specified object is NOT empty.  I.e. not nil, "", false, 0
+// or either a slice or a channel with len == 0. And then set it to empty.
+//
+//  assert.NotEmptyThenClearf(t, &obj, "error message %s", "formatted")
+func NotEmptyThenClearf(t TestingT, object *interface{}, msg string, args ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	return NotEmptyThenClear(t, object, append([]interface{}{msg}, args...)...)
+}
+
 // NotEqualf asserts that the specified values are NOT equal.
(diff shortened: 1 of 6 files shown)


HaraldNordgren:
Two clear problems for me here

**Number 1:**

I don't know how to generate the `require` counter-part of the assertion function. I get the same result on my native macOS as on Docker:

```
$ go run _codegen/main.go
2018/05/05 23:03:12 8:1: expected 'IDENT', found 'import'
exit status 1
```

**Number 2:**

Tests are failing for Go 1.8 and 1.9 (it works for 1.10 and tip). The only failing part is

```
./.travis.gofmt.sh
```

It appears also from https://github.com/stretchr/testify/commits/master that the test have not be run since 2c9035a78fc7b52c6acd9750c42d70cffb5fde81 which doesn't seem ideal for a test library. So they could have been broken anytime between then and now.

Please correct me if I'm wrong.

HaraldNordgren:
Ping @ernesto-jimenez

HaraldNordgren:
Ping @ernesto-jimenez

HaraldNordgren:
Ping @ernesto-jimenez. I figured out how to generate files to unbreak tests.

I think this would be a very nice feature to have in the library if you would please take a look.

HaraldNordgren:
Ping @ernesto-jimenez. I think this would be a very nice feature to have in the library if you would please take a look.

HaraldNordgren:
@ernesto-jimenez Merge this please?

HaraldNordgren:
@ernesto-jimenez Merge this?

ernesto-jimenez:
@HaraldNordgren we shouldn't be modifying the data we are comparing. 

In those cases, I would suggest you create a helper method in your code to compare those types.

PR #596, opened 2018-05-06
Format matched arguments with %v to avoid panic

gz-c:
Some values will panic when fmt.Sprintf'd with `%s`, e.g. https://godoc.org/github.com/boltdb/bolt#Tx.  

When mocking a method that has such a value in its arguments, the program will panic due to Argument.Diff attempting to print the interface{} type with `%s`.

Normally the compiler will prevent using such a value in a formatter but the code bypasses this by operating on `interface{}`.

This PR changes the formatter to use `(%T=%v)`. `%T` is used to mimic the previous behaviour which generated outputs like `%!s(int=123)`.


ernesto-jimenez:
@gz-c thanks!

PR #601, opened 2018-05-10
Added suport for HTTPHeadersContains method

### LINKED ISSUE #598: Assert for HTTP Headers
Assert package has 2 useful method to verify HTTP status - https://godoc.org/github.com/stretchr/testify/assert#HTTPSuccess and body - https://godoc.org/github.com/stretchr/testify/assert#HTTPBodyContains . It will be nice also to add method to verify HTTP headers.

falkerson:
Current commit fixes #598 and provide support for additional helper testing
method which can verify that response returns proper headers.

### DIFF
--- a/assert/http_assertions.go
+++ b/assert/http_assertions.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"reflect"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


assert/assertion_format.go
```diff
@@ -195,6 +195,21 @@ func HTTPErrorf(t TestingT, handler http.HandlerFunc, method string, url string,
 	return HTTPError(t, handler, method, url, values, append([]interface{}{msg}, args...)...)
 }
```
devdinu: @falkerson This is auto generated code, which shouldn't be edited.

brackendawson:
Closed for the reasons in #598

PR #603, opened 2018-05-14
add constructor to assertions object to smooth out subtests

ichiban:
Currently, you can't redeclare `assert := assert.New(t)` in subtests.

This PR enables it by defining constructor method.

```go
package main

import (
	"testing"

)

func TestAssertions_New(t *testing.T) {
	assert := assert.New(t)
	assert.True(true)
	t.Run("subtest", func(t *testing.T) {
		assert := assert.New(t) // this doesn't compile without `func (a *Assertions) New(t TestingT) *Assertions`
		assert.True(true)
	})
}
```

### DIFF
--- a/assert/forward_assertions.go
+++ b/assert/forward_assertions.go
@@ -13,4 +13,9 @@ func New(t TestingT) *Assertions {
 	}
 }
 
+// New makes a new Assertions object for the subtest TestingT.
+func (a *Assertions) New(t TestingT) *Assertions {
+	return New(t)
+}
+
 //go:generate go run ../_codegen/main.go -output-package=assert -template=assertion_forward.go.tmpl -include-format-funcs


ichiban:
Thank you for your reply, @devdinu!

I know that `assert := assert.New(t)` is an idiom that shadows the package name with the variable name. Though, I didn't want to rename the variable to other than `assert` because it required more changes than it should.

My original problem was applying `t.Run()` to tests predating subtests (`Test_original`). To do so, there're 2 options: rename the outer `assert` so that it doesn't shadow package (`Test_rename`), or don't declare the outer `assert` (`Test_package`). Both require changes to `// lots of assertions` and it would be great if I had another solution without breaking them. Adding a constructor method can keep lots of assertions intact (`Test_new`).

```go
package main

import (
	"testing"

)

func Test_original(t *testing.T) {
	assert := assert.New(t)

	// lots of assertions
	assert.True(true)
	assert.True(true)
	assert.True(true)

	{ // subtest A
		assert.True(true)
	}

	{ // subtest B
		assert.True(true)
	}
}

func Test_rename(t *testing.T) {
	asser := assert.New(t)

	// lots of assertions
	asser.True(true)
	asser.True(true)
	asser.True(true)

	t.Run("subtest A", func(t *testing.T) {
		assert := assert.New(t)
		assert.True(true)
	})

	t.Run("subtest A", func(t *testing.T) {
		assert := assert.New(t)
		assert.True(true)
	})
}

func Test_package(t *testing.T) {
	// lots of assertions
	assert.True(t, true)
	assert.True(t, true)
	assert.True(t, true)

	t.Run("subtest A", func(t *testing.T) {
		assert := assert.New(t)
		assert.True(true)
	})

	t.Run("subtest A", func(t *testing.T) {
		assert := assert.New(t)
		assert.True(true)
	})
}

func Test_new(t *testing.T) {
	assert := assert.New(t)

	// lots of assertions
	assert.True(true)
	assert.True(true)
	assert.True(true)

	t.Run("subtest A", func(t *testing.T) {
		assert := assert.New(t)
		assert.True(true)
	})

	t.Run("subtest A", func(t *testing.T) {
		assert := assert.New(t)
		assert.True(true)
	})
}
```

devdinu:
I'm curious why do you need to do `assert.New` instead of using those method on packages directly.

ichiban:
@devdinu Good point! I'd do so if I wrote tests from scratch but I have existing assertions in the form of `assert.True(true)` to be wrapped in subtests. I need it to wrap them in subtests without changing the outer assertions and also the assertions in subtests. I don't want to waste others' time reviewing those nonessential changes.

assert/forward_assertions.go
```diff
@@ -13,4 +13,9 @@ func New(t TestingT) *Assertions {
 	}
 }
 
+// New makes a new Assertions object for the subtest TestingT.
+func (a *Assertions) New(t TestingT) *Assertions {
```
ernesto-jimenez: This seems confusing to me.

how about:

```go
func (a *Assertions) Run(name string, fn func(*Assertions)) {
 	a.t.Run(name, func(t *testing.T) {
 	 	a := New(t)
 	 	fn(a)
 	})
}
```

ichiban: Hi @ernesto-jimenez, thank you for the great idea! It looks compelling and I can easily imagine myself writing like:

```go
func Test_run(t *testing.T) {
	assert := assert.New(t)

	assert.Run("subtest", func(assert *assert.Assertions) { // Wait, package `assert` is shadowed
		assert.True(true)
	})
}
```

Oh wait, `*assert.Assertions` will be shadowed if I keep declaring `assert := assert.New(t)`. Also, `a.t` is a `TestingT` instead of `*testing.T` which doesn't have `Run()` method.

dolmen:
:+1: for `func (a *Assertions) Run(name string, fn func(*Assertions))` as suggested by @ernesto-jimenez .

I'm closing this PR and I have opened #1446.

PR #606, opened 2018-05-26
Optimise Template

Mungrel:
Fixes #602 

Returns early without calling:
```
	if h, ok := t.(tHelper); ok {
		h.Helper()
	}
```
Need to run the code generator on the new template, but couldn't get it to run, just got:
```
2018/05/26 21:53:23 8:1: expected 'IDENT', found 'import'
exit status 1
```

Issue #602, opened 2018-05-11
Conditions are too expensive on the successful path.


jcferretti:
Conditions call `testing.T.Helper()` in the non-failing path, which is expensive (uses mutexes and `runtime.Callers`: https://golang.org/src/testing/testing.go).
We have some tests for numerical code that make heavy use of `require` to validate results; these tests check many possible values of the input range. The tests as written were working fine with go 1.8; when trying go 1.10 the time this tests take skyrocketed; this was due to the calls to Helper (which were not happening in 1.8; require code checks for the implementation of an interface to call only if it is available, it wasn't prior to 1.9).  See transcript of test times at the end of this writeup.

Require implementations do:

```
func Condition(t TestingT, comp assert.Comparison, msgAndArgs ...interface{}) {
	if h, ok := t.(tHelper); ok {
		h.Helper()
	}
	if !assert.Condition(t, comp, msgAndArgs...) {
		t.FailNow()
	}
}
```

To only incur `Helper()`'s cost on the fail path, perhaps they could do instead:

```
func Condition(t TestingT, comp assert.Comparison, msgAndArgs ...interface{}) {
	if assert.Condition(t, comp, msgAndArgs...) {
		return
        }
	if h, ok := t.(tHelper); ok {
		h.Helper()
	}
	t.FailNow()
}
```

Test times:

```
// Go 1.10
=== RUN   TestNew
Test finished.
--- PASS: TestNew (0.00s)
=== RUN   TestIsSet
Test finished.
--- PASS: TestIsSet (0.01s)
=== RUN   TestOrEquals
Test finished.
--- PASS: TestOrEquals (0.01s)
=== RUN   TestSetClear
Test finished.
--- PASS: TestSetClear (5.03s)
=== RUN   TestIsAnySet
Test finished.
--- PASS: TestIsAnySet (1.82s)
=== RUN   TestIsAnyClear
Test finished.
--- PASS: TestIsAnyClear (1.82s)
=== RUN   TestSetRange
Test finished.
--- PASS: TestSetRange (68.92s)
=== RUN   TestClearRange
Test finished.
--- PASS: TestClearRange (68.59s)
PASS
00:02:11.683470 testrunner.go:230: Max RSS memory used: 16281600 bytes
=== RUN   TestNew
Test finished.

// Go 1.8
--- PASS: TestNew (0.00s)
=== RUN   TestIsSet
Test finished.
--- PASS: TestIsSet (0.00s)
=== RUN   TestOrEquals
[...]

PR #612, opened 2018-06-09
Optimise other templates

Mungrel:
In relation to issue: #602 and PR: #606 

The calls to helper (afaik), are called further down the line anyway, so shouldn't need to be called earlier.

I guess the next step would be to go through `assertions.go` and shift the calls to helper to the failing path


ernesto-jimenez:
@Mungrel, unfortunately `t.Helper` works per function: https://godoc.org/testing#T.Helper

That means that every testify function in our stack must be marked as a helper or it will appear as the source of the error.

PR #614, opened 2018-06-13
EqualError: Accept error as expected value as well as string

jinnovation:
This modification is useful in cases where we want to assert equality against
some exported error value, e.g. `assert.EqualError(t, err, pkg.ErrSomeError)`,
letting us avoid the unnecessary error "dereference" to string.

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -82,11 +82,11 @@ func Equalf(t TestingT, expected interface{}, actual interface{}, msg string, ar
 //
 //   actualObj, err := SomeFunction()
 //   assert.EqualErrorf(t, err,  expectedErrorString, "error message %s", "formatted")
-func EqualErrorf(t TestingT, theError error, errString string, msg string, args ...interface{}) bool {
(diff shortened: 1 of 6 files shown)


slizco:
I'd also like to see this PR included. As an addition, it'd be nice for the assertion to return `true` if both the expected and actual errors passed equate to `nil` (in the case where an `error` is passed for expected value instead of `string`). This should still be a non-breaking change.

georgelesica-wf:
Are there any tests that can be updated just to exercise the new code path?

glesica:
Is this still a desirable change? I'm going to move it to the release after the next release since it won't be a breaking change and then we can push out the next release sooner.

jinnovation:
> I'm going to move it to the release after the next release since it won't be a breaking change and then we can push out the next release sooner.

Sounds good.

jinnovation:
`ErrorIs` was added in v1.7.0 that supersedes this horridly stale PR of mine. 😇 Closing.

PR #616, opened 2018-06-14
Optimise Assertions

Mungrel:
Related issue: #602 

Moves the calls to `Helper()` to the failing path for assertions.


Mungrel:
Getting this error in CI only for `Go:tip` in a seemingly unrelated file. Can't seem to find anything online that works to fix it. Unless it's something else entirely.

screen shot 2018-06-14 at 9 06 45 pm

PR #620, opened 2018-06-22
Fixing issue of passing varargs and string to Sprintf. Closes #619

### LINKED ISSUE #619: Tests are failing with go tip versions
From go version `6b3e343408` tests started failing 
https://travis-ci.org/stretchr/testify/jobs/391186749 for all PRs.

```
./mock.go:506: missing ... in args forwarded to printf-like function
./mock.go:523: missing ... in args forwarded to printf-like function
```

devdinu:
Build fails with recent go version in mock package.
```
./mock.go:506: missing ... in args forwarded to printf-like function
./mock.go:523: missing ... in args forwarded to printf-like function
```

This fixes #619.

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"regexp"
 	"runtime/debug"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


georgelesica-wf (dismissed):
Looks good to me. I ran into this and while this seems like a strange fix, it works.

devdinu:
The tests were failing for all PRs, and recent PR's tests succeeds. So this must be specific to some `tip` version of go. We could drop this PR if its non mandated by go changes in release.

georgelesica-wf:
Good call, let's leave it open but put it on hold for now until we're sure we don't need it. Thanks.

PR #622, opened 2018-06-25
assert.Equal: fix a panic when comparing types where the underlying type is string

autarch:
The code was checking if the type's _kind_ was string and then attempting to
do a type assertion of `var.(string)`. However, this assertion only works if
the type actually _is_ a string. If it's some other type (like `type foo
string`) then this ends up giving us a runtime panic.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1345,7 +1345,7 @@ func diff(expected interface{}, actual interface{}) string {
 	}
 
 	var e, a string
-	if ek != reflect.String {
+	if et.Name() != "string" {
 		e = spewConfig.Sdump(expected)
 		a = spewConfig.Sdump(actual)
 	} else {
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -1791,3 +1791,18 @@ func TestErrorAssertionFunc(t *testing.T) {
 		})
 	}
 }
+
+type notstring string
+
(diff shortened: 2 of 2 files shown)


autarch:
Travis failure is because of #619

jwilder:
I'm hitting this issue w/ 1.2.2.  Any chance we can get this merged?  I verified the change fixes it for me as well.

jwilder (approved):

georgelesica-wf:
@autarch if you don't mind, could you confirm that https://github.com/stretchr/testify/pull/661 also fixes this issue?

autarch:
@georgelesica-wf - the test in that other PR looks to be functionally the same as the one I added, so I'm sure it fixes the issue as best as I could test it. I don't have any other code that causes this issue to run the PR branch against.

georgelesica-wf:
Cool, thanks! Would you have any objection if we merged the one I linked? I won't close this one until that's done, however.

autarch:
Yeah, I'm fine with anything that fixes this bug.

vitalyisaev2:
Perhaps I'm little bit late to the party, tried to fix it too in #674

autarch:
This was fixed a while back in a different commit.

dolmen:
Fix for #644.
Fixed by #661.

PR #623, opened 2018-06-26
Allow tests to fail on master

Hanzei:
When testing against go tip we can allow tests to fail. This PR also ensures that we run test against the newest go version e.g. `1.10.3` instead of `1.10`

Ref #619

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,11 @@ go:
   - "1.10"
   - tip
 
+matrix:
+  allow_failures:
+    - go: master
+fast_finish: true
+
 script:
   - ./.travis.gogenerate.sh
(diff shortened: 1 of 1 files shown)


devdinu:
PR #620 fixes this issue.

Hanzei:
I would still like to see this merged, as it prevents future problems with go:master

Hanzei:
Any update?

devdinu:
@georgelesica-wf We can close this PR.

devdinu:
@Hanzei Thought we could close this, since the latest PR's are passing in CI, and we don't need allow failures, if you still think there's reasons to do so, please share it, and we can reopen it and discuss.

PR #627, opened 2018-07-07
assert.Equal: fix panic on user type from string

jsidew:
To repro the bug, pull only assert/issue_test.go, then run unit tests and watch the failing output.
If you also want the fix, pull also assert/assertions.go, run the tests again, which all should pass ;)


jsidew:
... I just realized this issue duplicates #622

dolmen:
The issue is #644.

As #622 mentioned above is closed as a duplicate of #661 (merged) which fixed #644, I expect this one is fixed too.

So closing.

PR #631, opened 2018-07-11
mock: move regexp compilation outside of `Called`

aud10slave:
This diff moves the regular expression compilation for GCCGO functions outside the `mock.Called`. Its doing unnecessary work on every function call.


mock/mock.go
```diff
@@ -18,6 +18,9 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// regex for GCCGO functions
+var gccgoRE = regexp.MustCompile("\\.pN\\d+_")
```
dolmen, line 22: Use backquote to reduce escaping.

```suggestion
var gccgoRE = regexp.MustCompile(`\.pN\d+_`)
```

dolmen (changes requested):

mock/mock.go
```diff
@@ -455,8 +458,7 @@ func (m *Mock) Called(arguments ...interface{}) Arguments {
 	// For Ex:  github_com_docker_libkv_store_mock.WatchTree.pN39_github_com_docker_libkv_store_mock.Mock
 	// uses interface information unlike golang github.com/docker/libkv/store/mock.(*Mock).WatchTree
 	// With GCCGO we need to remove interface information starting from pN<dd>.
-	re := regexp.MustCompile("\\.pN\\d+_")
-	if re.MatchString(functionPath) {
+	if gccgoRE.MatchString(functionPath) {
 		functionPath = re.Split(functionPath, -1)[0]
```
dolmen, line 462:
```suggestion
		functionPath = gccgoRE.Split(functionPath, -1)[0]
```

dolmen (changes requested):
We must squash commits.

dolmen (approved):

PR #638, opened 2018-07-16
Add panic stack to output from NotPanics assertion failures

### LINKED ISSUE #636: NotPanics failures should include the panicking stack
If a `NotPanics` assertion fails, the test output should include the stack from the panic. This would speed up debugging. For example,

```go
func makeNew(c *http.Client) (*http.Response, error) {
	req, err := http.NewRequest(http.MethodPost, "https://example.com/api/v1/things/new", nil)
	if err != nil {
		return nil, err
	}

	return c.Do(req)
}

func TestPanic(t *testing.T) {
	var c *http.Client

	doReq := func() {
		resp, err := makeNew(c)
		require.NoError(t, err)

		resp.Body.Close()
	}

	require.NotPanics(t, doReq)
}
```

That gives you output like this:

```text
--- FAIL: TestNotPanics (0.00s)
	require.go:945:
			Error Trace:	notpanics_test.go:29
			Error:      	func (assert.PanicTestFunc)(0x171bf90) should not panic
			            		Panic value:	runtime error: invalid memory address or nil pointer dereference
			Test:       	TestNotPanics
```

Including the stack of the panicked goroutine would help in quickly spotting what the mistake was.

dcormier:
Fixes #636.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -11,6 +11,7 @@ import (
 	"reflect"
 	"regexp"
 	"runtime"
+	"runtime/debug"
 	"strings"
 	"time"
 	"unicode"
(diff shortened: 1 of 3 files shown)


dcormier:
OK, @georgelesica-wf. It's been updated.

PR #648, opened 2018-08-16
Update the LICENSE to match the 'standard' MIT license file

omarkohl:
The reason is that since the name of the license is not stated
explicitly and the license text differed very slightly I had to perform
a diff to ensure it actually is MIT licensed. Also GitHub does not
detect the license for this project currently (probably for the same
reason).

Changes:

* Remove the line 'Please consider promoting this project...' that is
  not part of the MIT license
* Use the same formatting as choosealicense.com
* Explicitly state 'MIT License'
* Update the copyright year

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,22 +1,21 @@
-Copyright (c) 2012 - 2013 Mat Ryer and Tyler Bunnell
+MIT License
 
-Please consider promoting this project if you find it useful.
+Copyright (c) 2012-2018 Mat Ryer and Tyler Bunnell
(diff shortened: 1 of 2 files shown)


georgelesica-wf:
It looks to me like this should be fine, but I think we should review it carefully nonetheless. Assuming there are no problems here, I think it's a good idea to make the license more obvious.

omarkohl:
OK, I'll update it

lookuptable:
Friendly ping...

PR #651, opened 2018-08-22
fix #644

### LINKED ISSUE #644: Panic: Equal on type aliases causes panic
Current behaviour:

With go 1.9.3 and testify 1.2.2 or latest master (f35b8ab0b5a2cef36673838d662e249dd9c94686)  if I require/assert.Equal two aliased types, it causes testify to panic:

```
goroutine 5 [running]:
testing.tRunner.func1(0xc42010a0f0)
	/Users/brendan.fosberry/.gvm/gos/go1.9.3/src/testing/testing.go:711 +0x2d2
panic(0x1289000, 0xc420066780)
	/Users/brendan.fosberry/.gvm/gos/go1.9.3/src/runtime/panic.go:491 +0x283
github.com/stretchr/testify/assert.diff(0x126fae0, 0xc420010d10, 0x126fae0, 0x130a0b0, 0x0, 0x0)
	/Users/brendan.fosberry/go/src/github.com/stretchr/testify/assert/assertions.go:1352 +0x494
github.com/stretchr/testify/assert.Equal(0x141c480, 0xc42010a0f0, 0x126fae0, 0xc420010d10, 0x126fae0, 0x130a0b0, 0x0, 0x0, 0x0, 0xc420039f01)
	/Users/brendan.fosberry/go/src/github.com/stretchr/testify/assert/assertions.go:338 +0x271
_/Users/brendan.fosberry/workspace/derp.TestMyTypeRequire(0xc42010a0f0)
	/Users/brendan.fosberry/workspace/derp/derp_test.go:19 +0x11f
testing.tRunner(0xc42010a0f0, 0x12e9c88)
	/Users/brendan.fosberry/.gvm/gos/go1.9.3/src/testing/testing.go:746 +0xd0
created by testing.(*T).Run
	/Users/brendan.fosberry/.gvm/gos/go1.9.3/src/testing/testing.go:789 +0x2de
FAIL	_/Users/brendan.fosberry/workspace/derp	0.038s
```

Expected behaviour: testify should determine if the two types are equal, or fail gracefully.

Here is a test case to replicate the issue

```
package main

import (
        "testing"

)

type MyType string

const (
        MyTypeA MyType = "a"
        MyTypeB MyType = "b"
)

func TestMyTypeRequire(t *testing.T) {
        mt := MyTypeA
        assert.Equal(t, mt, MyTypeA)
        assert.Equal(t, mt, MyTypeB)
}
```

I don't know much about reflect, but I think it comes down to this:

```
	var e, a string
	if ek != reflect.String {
		e = spewConfig.Sdump(expected)
		a = spewConfig.Sdump(actual)
	} else {
		e = expected.(string)
		a = actual.(string)
	}
```

ek == String comes back as true for string type aliases, however a type assertion causes a panic since they are of type string.

My workaround for now is not to use Equal with type aliases, however this is something that should be handled internally.

yahelmanor:
Fix by change the condition in the if statement from the problematic reflect based condition to safer one.
I try before this solution, other reflect based solution but something in the reflect package seams to not work like i was expected.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1345,7 +1345,7 @@ func diff(expected interface{}, actual interface{}) string {
 	}
 
 	var e, a string
-	if ek != reflect.String {
+	if _, ok := expected.(string); !ok {
 		e = spewConfig.Sdump(expected)
 		a = spewConfig.Sdump(actual)
 	} else {


devdinu:
@yahelmanor Could you share what wasn't working with reflect check, please share code or test which 's breaking and that would be assisting the need for change as well.

yahelmanor:
the test that panics (from the #644 issue) is
``` golang
package main

import (
        "testing"

)

type MyType string

const (
        MyTypeA MyType = "a"
        MyTypeB MyType = "b"
)

func TestMyTypeRequire(t *testing.T) {
        mt := MyTypeA
        assert.Equal(t, mt, MyTypeA)
        assert.Equal(t, mt, MyTypeB)
}
```
i try to replace the original condition of `ek != reflect.String` that is checking that `exacted`  variable have the same kind as string, this check not cover the case of kind is equal to string but you still can`t convert it.

a small demonstration to this problem is in this program
```golang
package main

import (
	"fmt"
	"reflect"
)

type s string

func main() {
	var sA s = "a"
	i := interface{}(sA)
	fmt.Println("sA,i kinds =", reflect.TypeOf(sA).Kind(), ",", reflect.TypeOf(i).Kind())
//print: sA,i kinds = string , string
	fmt.Println("can we convert to string by reflect?", reflect.TypeOf(i).ConvertibleTo(reflect.TypeOf("")))
//print: true
	fmt.Println("just to chack the type of <\"\"> is:", reflect.TypeOf(""))
//print: string
	_, ok := i.(string)
//here it panics in the test
	fmt.Println("can we actualy convert?", ok)
//print: false
	s, ok := reflect.ValueOf(i).Convert(reflect.TypeOf("")).Interface().(string)
	fmt.Println("and via reflect?", ok, ",s =", s)
//print: and via reflect? true ,s = a
}
```
in the demonstration the kind of `sA` is string but still it can`t be convert.


so it can be possible to change the assert statement into the reflect way but i change the condition so in case that make the original problem it will go throw the  `spewConfig.Sdump` function that use reflect so i think there is no need to use the reflect conversion in that case but it open to discussion.

georgelesica-wf:
I believe this issue has been addressed. Feel free to re-open and explain if that is not the case. Thanks!

Issue #466, opened 2017-06-21
bug in parallel tests


yiminc:
When enable parallel testing, we would expect the TearDownSuite() to be called after all the tests completed. However, after this commit, https://github.com/stretchr/testify/commit/3104bf5483d3fba84f24933949f0533f4467880b
the parallel testing is broken as the TearDownSuite() is called before tests are done.
Here is a simple repro code to repro:
```
type MyTestSuite struct {
	suite.Suite
}
func TestMySuite(t *testing.T) {
	suite.Run(t, new(MyTestSuite))
}
func (s *MyTestSuite) SetupSuite() {
	fmt.Println("SetupSuite")
}
func (s *MyTestSuite) TearDownSuite() {
	fmt.Println("TearDownSuite")
}
func (s *MyTestSuite) SetupTest() {
	s.T().Parallel()
	fmt.Println("SetupTest")
}
func (s *MyTestSuite) TearDownTest() {
	fmt.Println("TearDownTest")
}
func (s *MyTestSuite) Test_A() {
	time.Sleep(time.Second)
}
func (s *MyTestSuite) Test_B() {
	time.Sleep(time.Second * 2)
}

```
The output of above tests is:
```
SetupSuite
TearDownSuite
SetupTest
SetupTest
TearDownTest
TearDownTest
```

thaJeztah:
relates to https://github.com/stretchr/testify/pull/389, https://github.com/stretchr/testify/issues/346

/cc @bayesianmind @ernesto-jimenez PTAL

piepmatz:
The same problem applies to parallel subtests: `TearDownTest()` is run before the test is completed.

PR #655, opened 2018-09-01
Add support for subtests (go 1.7)

itzamna314:
Adds a `.Run(..)` method onto test suites, mimicking the semantics of [testing.Run](https://godoc.org/testing#T.Run).  This will make writing table-driven tests in test suites cleaner and simpler, by handling the management of `*testing.T` instances transparently so that sub-tests can use the same `suite.Assert()` and `suite.Require()` methods, and have all output and test logic properly fall under the correct subtest.

The major benefits of managing the `*testing.T` instance are:
- Subtest output will be properly printed under each subtest, rather than all grouped under the test method
- Subtests can be correctly executed using `go test -run MyTestSuite/MyTestMethod/MySubTest`

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -55,6 +55,20 @@ func (suite *Suite) Assert() *assert.Assertions {
 	return suite.Assertions
 }
 
+// Run provides suite functionality around golang subtests.  It should be
+// called in place of t.Run(name, func(t *testing.T)) in test suite code
+// to expose setup and teardown functionality for each subtest.  The passed-in
+// func will be executed as a subtest with a fresh instance of t.  Provides
+// compatibility with go test pkg -run TestSuite/TestName/SubTestName.
+func (suite *Suite) Run(name string, subtest func()) {
+	oldT := suite.T()
+	oldT.Run(name, func(t *testing.T) {
+		suite.SetT(t)
+		subtest()
+		suite.SetT(oldT)
+	})
+}
+
 // Run takes a testing suite and runs all of the tests attached
 // to it.
 func Run(t *testing.T, suite TestingSuite) {
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -59,6 +59,7 @@ type SuiteTester struct {
 	TearDownTestRunCount  int
 	TestOneRunCount       int
 	TestTwoRunCount       int
+	TestSubtestRunCount   int
 	NonTestMethodRunCount int
 
 	SuiteNameBefore []string
@@ -153,6 +154,30 @@ func (suite *SuiteTester) NonTestMethod() {
 	suite.NonTestMethodRunCount++
 }
 
+func (suite *SuiteTester) TestSubtest() {
+	beforeCount := suite.TestSubtestRunCount
+	suite.TestSubtestRunCount++
+
+	for _, t := range []struct {
+		testName string
+	}{
+		{"first"},
+		{"second"},
+	} {
+		suiteT := suite.T()
+		suite.Run(t.testName, func() {
+			// We should get a different *testing.T for subtests, so that
+			// go test recognizes them as proper subtests for output formatting
+			// and running individual subtests
+			subTestT := suite.T()
+			suite.NotEqual(subTestT, suiteT)
+		})
+	}
+
+	assert.Equal(suite.T(), suite.TestOneRunCount, beforeCount+1)
+	suite.Equal(suite.TestOneRunCount, beforeCount+1)
+}
+
 // TestRunSuite will be run by the 'go test' command, so within it, we
 // can run our suite using the Run(*testing.T, TestingSuite) function.
 func TestRunSuite(t *testing.T) {
@@ -168,18 +193,20 @@ func TestRunSuite(t *testing.T) {
 	assert.Equal(t, suiteTester.SetupSuiteRunCount, 1)
 	assert.Equal(t, suiteTester.TearDownSuiteRunCount, 1)
 
-	assert.Equal(t, len(suiteTester.SuiteNameAfter), 3)
-	assert.Equal(t, len(suiteTester.SuiteNameBefore), 3)
-	assert.Equal(t, len(suiteTester.TestNameAfter), 3)
-	assert.Equal(t, len(suiteTester.TestNameBefore), 3)
+	assert.Equal(t, len(suiteTester.SuiteNameAfter), 4)
+	assert.Equal(t, len(suiteTester.SuiteNameBefore), 4)
+	assert.Equal(t, len(suiteTester.TestNameAfter), 4)
+	assert.Equal(t, len(suiteTester.TestNameBefore), 4)
 
 	assert.Contains(t, suiteTester.TestNameAfter, "TestOne")
 	assert.Contains(t, suiteTester.TestNameAfter, "TestTwo")
 	assert.Contains(t, suiteTester.TestNameAfter, "TestSkip")
+	assert.Contains(t, suiteTester.TestNameAfter, "TestSubtest")
 
 	assert.Contains(t, suiteTester.TestNameBefore, "TestOne")
 	assert.Contains(t, suiteTester.TestNameBefore, "TestTwo")
 	assert.Contains(t, suiteTester.TestNameBefore, "TestSkip")
+	assert.Contains(t, suiteTester.TestNameBefore, "TestSubtest")
 
 	for _, suiteName := range suiteTester.SuiteNameAfter {
 		assert.Equal(t, "SuiteTester", suiteName)
@@ -197,15 +224,16 @@ func TestRunSuite(t *testing.T) {
 		assert.False(t, when.IsZero())
 	}
 
-	// There are three test methods (TestOne, TestTwo, and TestSkip), so
+	// There are four test methods (TestOne, TestTwo, TestSkip, and TestSubtest), so
 	// the SetupTest and TearDownTest methods (which should be run once for
-	// each test) should have been run three times.
-	assert.Equal(t, suiteTester.SetupTestRunCount, 3)
-	assert.Equal(t, suiteTester.TearDownTestRunCount, 3)
+	// each test) should have been run four times.
+	assert.Equal(t, suiteTester.SetupTestRunCount, 4)
+	assert.Equal(t, suiteTester.TearDownTestRunCount, 4)
 
 	// Each test should have been run once.
 	assert.Equal(t, suiteTester.TestOneRunCount, 1)
 	assert.Equal(t, suiteTester.TestTwoRunCount, 1)
+	assert.Equal(t, suiteTester.TestSubtestRunCount, 1)
 
 	// Methods that don't match the test method identifier shouldn't
 	// have been run at all.


suite/suite_test.go
```diff
@@ -153,6 +154,30 @@ func (suite *SuiteTester) NonTestMethod() {
...
+		{"first"},
+		{"second"},
+	} {
+		suiteT := suite.T()
+		suite.Run(t.testName, func() {
+			// We should get a different *testing.T for subtests, so that
+			// go test recognizes them as proper subtests for output formatting
+			// and running individual subtests
+			subTestT := suite.T()
+			suite.NotEqual(subTestT, suiteT)
```
kwk: Can you add a similar test that the old `*testing.T` is restored when `Run` was exited?

itzamna314: :+1: added

kwk:
> @kwk Is there anything else we need to get this approved and merged?

@itzamna314 I'm not a maintainer and not a contributor, so I cannot tell.

itzamna314:
Still looking for a review from a contributor/maintainer.

Looks like I'll keep copy-pasting this code into every project that uses `testify/suite` till then 😿

suite/suite.go
```diff
@@ -55,6 +55,20 @@ func (suite *Suite) Assert() *assert.Assertions {
...
 
+// Run provides suite functionality around golang subtests.  It should be
+// called in place of t.Run(name, func(t *testing.T)) in test suite code
+// to expose setup and teardown functionality for each subtest.  The passed-in
+// func will be executed as a subtest with a fresh instance of t.  Provides
+// compatibility with go test pkg -run TestSuite/TestName/SubTestName.
+func (suite *Suite) Run(name string, subtest func()) {
+	oldT := suite.T()
+	oldT.Run(name, func(t *testing.T) {
+		suite.SetT(t)
```
kwk: I suggest to use a defer here to ensure the old `t` is always reset.
```go
func (suite *Suite) Run(name string, subtest func()) {
	suite.T().Run(name, func(t *testing.T) {
		defer func(oldT *testing.T) {
			suite.SetT(oldT)
		}(suite.T())
		suite.SetT(t)
		subtest()
	})
}
```

letientai299:
This PR is there for 3 months already. Cc @viswajithiii and @georgelesica-wf since you are the most recent contributor on master branch. Anything we can do to have this get reviewed and merged?

georgelesica-wf:
@letientai299 looking at it right now. I just came on to help a couple months back, and I haven't had time to look much at the backlog of PRs, I've just been dealing with them as they are brought up.

suite/suite_test.go
```diff
@@ -153,6 +154,31 @@ func (suite *SuiteTester) NonTestMethod() {
...
+		})
+		suite.Equal(suiteT, suite.T())
+	}
+
+	assert.Equal(suite.T(), suite.TestOneRunCount, beforeCount+1)
+	suite.Equal(suite.TestOneRunCount, beforeCount+1)
```
georgelesica-wf: These would seem to pass trivially since you capture and increment on lines 158 and 159. What am I missing?

georgelesica-wf:
@itzamna314 hey would you mind resolving merge conflicts?

suite/suite_test.go
itzamna314: Its comparing `TestOneRunCount`, and the thing that got incremented was `TestSubtestRunCount`.

But as far as I can tell, that's an arbitrary test that was passing coincidentally.  Its been a few months since I wrote that, and maybe I had a good reason then, but I can't figure it out now at all.  So I pulled those lines of the test out, to avoid future confusion.  Good catch!

itzamna314:
I got the merge conflicts resolved, but ci is failing.  It built on all of the targets except for tip, and for some reason tip took 4 minutes while the other targets took about 1 each.  I couldn't figure out how to re-run.

georgelesica-wf:
The failure on tip is a known issue. I'm hoping to fix it soon, certainly before the next release. I'm going to have one other person take a look at this tomorrow and then I'll merge it. Sorry again for the delay.

itzamna314:
Thanks!

suite/suite.go
```diff
@@ -63,6 +63,20 @@ func failOnPanic(t *testing.T) {
 	}
 }
 
+// Run provides suite functionality around golang subtests.  It should be
+// called in place of t.Run(name, func(t *testing.T)) in test suite code
+// to expose setup and teardown functionality for each subtest.  The passed-in
```
brandonbodnar-wk: > to expose setup and teardown functionality for each subtest.

This comment is confusing to me. Based on its wording, I would expect some type of before and after method is ran by the suite runner before and after each subtest (similar to the `SetupSuite` and `TearDownSuite` ran before and after the suite itself, or the `SetupTest`, `BeforeTest`, `AfterTest`, and `TearDownTest` ran before and after each test). I don't believe that is the intent though based on the code, and that the intent is to allow people to write subtests in which they may themselves execute setup and teardown code within their passed in function.

To address, I suggest altering this statement to something along the lines of

>  It may be called in place of t.Run(name, func(t *testing.T)) within tests inside a test suite.

suite/suite.go
```diff
@@ -63,6 +63,20 @@ func failOnPanic(t *testing.T) {
...
+// func will be executed as a subtest with a fresh instance of t.  Provides
```
brandonbodnar-wk: > will be executed as a subtest with a fresh instance of t.

I think this can create some confusion that a `*testing.T` instance will be passed to the passed-in function as a parameter. While obviously not the case based on the function signature of subtest for `Suite.Run`, we could probably clean up this documentation to make clear that it is the suite's testing context (as returned by `Suite.T()`) that is switched prior to execution of the given function.

suite/suite.go
```diff
@@ -63,6 +63,20 @@ func failOnPanic(t *testing.T) {
...
+// compatibility with go test pkg -run TestSuite/TestName/SubTestName.
+func (suite *Suite) Run(name string, subtest func()) {
```
brandonbodnar-wk: Any reason not to return a `bool` similar to the return value of `T.Run`?

suite/suite.go
```diff
@@ -63,6 +63,20 @@ func failOnPanic(t *testing.T) {
 	}
 }
 
+// Run provides suite functionality around golang subtests.  It should be
```
brandonbodnar-wk: I think we need to add some additional documentation concerning the potential impact of calling [T.Parallel](https://golang.org/pkg/testing/#T.Parallel), especially as it can impact subsequent calls to functions dependent on the state of `Suite`, such as `Suite.Assert`, `Suite.Require`, and `Suite.T`. If someone calls `suite.T().Parallel()` within the subtest, then the `T.Run` call within `Suite.Run` will return and so will `Suite.Run`. This could lead to another test or subtest running that modifies the suite's testing context, modifying state for the still running first now parallel subtest. Users will need to be careful regarding when they call `T.Parallel()` and that they already have any testing context (`Suite.T()`), assertion context (`Suite.Assert()`), or require context (`Suite.Require()`) pulled out that they might need prior to calling `T.Parallel`

suite/suite.go
```diff
@@ -63,6 +63,20 @@ func failOnPanic(t *testing.T) {
...
+// called in place of t.Run(name, func(t *testing.T)) in test suite code
+// to expose setup and teardown functionality for each subtest.  The passed-in
+// func will be executed as a subtest with a fresh instance of t.  Provides
+// compatibility with go test pkg -run TestSuite/TestName/SubTestName.
+func (suite *Suite) Run(name string, subtest func()) {
+	oldT := suite.T()
+	oldT.Run(name, func(t *testing.T) {
+		suite.SetT(t)
+		defer suite.SetT(oldT)
```
brandonbodnar-wk: Should this be moved to before the `oldT.Run` call and outside the inner function here?

My primary concern is `oldT.Run` returning prior to the subtest finishing. According to [T.Run](https://golang.org/pkg/testing/#T.Run), this can occur if `t.Parallel` is called. If two subtests are running in parallel on this suite, that may lead to `oldT := suite.T()` in the second subtest as setting `oldT` as the testing context of the first subtest (since it is still executing  and thus does not had its defer called). Then when the second subtest finishes executing, it will set the suite testing context to the first subtest testing context instead of to the parent context as expected.

brandonbodnar-wk:
Reviewing some other existing areas of code with relation to using `T.Parallel` within a suite, it looks like we have a number of other problematic areas within Suite that prevent `T.Parallel` from working as expected (for example as pointed out in #466, the suite teardown method runs before executing the parallel tests). While I still feel that the `defer suite.SetT(oldT)` issue I noted above should be addressed in this PR as related to its code addition, it looks like we have a similar issue within the actual test runner for suite that already exists. @georgelesica-wf I assume we should create a new issue for this.

georgelesica-wf:
@itzamna314 Hey can you look at addressing Brandon's comments? Ideally I think we want to try not to break parallel runs any more than we already do.

suite/suite.go
itzamna314: Nope, I didn't realize `t.Run` returned a bool.  Updated to expose that return value.

suite/suite.go
itzamna314: :+1:

georgelesica-wf:
@itzamna314 thank you for your contribution and patience!

xcile:
Shouldn't SetupTest() and other interface methods be run for each golang subtest? It's a bit confusing right now that the Suite has subtests, but the suite is common for each test.

itzamna314:
I don't think so. The desired behavior on each subtest may be different than the desired behavior on each top-level test. 

It's pretty simple to run some setup/teardown for each subtest. Just add a `.Run()` method with your suite as the receiver, and call the embedded `TestSuite.Run()` from your method. This shadowing approach is similar to overriding parent methods in languages with inheritance.

You could make that behavior more explicit, by adding an interface that does subtest setup and teardown and then using type assertions to run that interface's setup/teardown methods. It would work similar to test-level and suite-level setup/teardown. That behavior would be fully compatible with this implementation.

xcile:
I still think it's confusing that the golang subtests does not work seamless with testify/suite. If they would, you could use the flexibility of table testing together with the power of testify/suite. However, I also agree with you that its quite easy to accomplish what I'm asking for. Thanks for the overriding tip, that might come useful sometime.

I've solved it now by using a separate suite instance for each subtest which becomes quite powerful:
```golang
func (s *MySuite) TestSomeCases() {

   subTests := map[string]MySuite {
      "test1": func(suite Suite) Suite {
         suite.SUT = mySutFunc
         suite.inputData = "some data"
         suite.expectedOutput = "ok"
         return suite
      }(*s),
   }

   for name, test : subTests {
      test.Run(name, func(){
         result := test.SUT(test.inputData)
         ...
      })
   }
}
```

BenSayers:
I arrived here after looking for an approach to writing table driven tests using testify/suite. I agree with @xcile it's unexpected that `suite.Run` does not invoke the setup and teardown methods. Having something that automatically calls these methods is the primary driver for using this feature in the first place.

mattolenik:
I also came here looking to write table-driven tests with Testify suites. As I understand it, this PR may be solving a slightly different problem than what some of us are looking to solve when we think "table-driven tests with Testify". I used the following pattern to run a single suite many times with different parameters. It will give output separated by `TestName` in the way you'd expect. Posting it here as it may be useful to other folks who also stumble upon this thread 😃 

```Go
type TestParams struct {
    // Name to be given for each run of the suite
    TestName String
    // Various test parameters etc
    Param1 string
    Param2 int
}

// Your standard Testify suite struct
type MySuite struct {
    suite.Suite
    params *TestParams
}

// This is your test table
var params = []TestParams{
    {"FirstTest", "abc", 123},
    {"SecondTest", "def", 456},
    {"ThirdTest", "xyz", 789},
    // etc...
}

// Top-level function invoked by `go test`
func TestMySuite(t *testing.T) {
    for _, param := range params {
        t.Run(param.TestName, func(tt *testing.T) {
            s := new(MySuite)
            s.params = &param
            suite.Run(tt, s)
        })
    }
}

// Parameters can be utilized in setup/teardown or individual test cases
func (s *MySuite) SetupTest() {
    foo(s.params.Param1)
    bar(s.params.Param2)
}

func (s *MySuite) TestMyFeature() {
    // test code here...
    DoSomethingTesty(s.params.Param1)
}
```

Thanks to everyone in this thread for the contribution and discussion, I think we all appreciate the work that goes into this fantastic library! 👍

Krisa:
@mattolenik thanks that was helpful. It also works well in combination with `BeforeTest` and `AfterTest`. I'd wish this library to have documentation and such examples

PR #658, opened 2018-09-04
call testing.T.Helper when calls are unexpected

bhcleek:


bhcleek:
Can one of the maintainers review this small PR? It really does help elide testify's call locations when a mock's expectations are not satisfied.

brackendawson:
I don't know what you intend this to do, because you didn't attach any context to the PR. I think it does nothing.

PR #607, opened 2018-05-30
Add go.mod and go.modverify

leighmcculloch:
What
===
Add `go.mod` and `go.sum`, using `Go 1.11`.

Why
===
Now that golang/go#24301 has been accepted, lets start using go.mod files
alongside Godep, and keep the two in sync.

Notes
===
There are no changes required to testify to support go.mod. The files were
generated by running `go build` and `go mod tidy.

Merging
===
This PR is intended to be merged after #659 which adds Go1.11 to the list of
supported builds.

PR #659, opened 2018-09-09
Add Go1.11 to CI

leighmcculloch:
What
===
Add Go1.11 to CI.

Why
===
Go1.11 has been released and should be tested against and supported.


leighmcculloch:
@ernesto-jimenez Should I remove Go1.8 too?

leighmcculloch:
I updated the PR removing 1.8.

devdinu:
It should be backward compatible, as long as we don't face any issues we could keep the 1.8

leighmcculloch:
@Hanzei @devdinu I've reverted the changes regarding Go1.8. The PR is limited to only adding Go1.11 again now.

README.md
```diff
@@ -319,7 +319,7 @@ To update Testify to the latest version, use `go get -u github.com/stretchr/test
 Supported go versions
 ==================
 
-We support the three major Go versions, which are 1.8, 1.9 and 1.10 at the moment.
+We support the four major Go versions, which are 1.8, 1.9, 1.10, and 1.11 at the moment.
```
johnlockwood-wf: Please change leave the official support at three versions, but do shift to 1.11.

georgelesica-wf: Agreed, let's not expand our commitment going forward, but as long as we still do support 1.8 we can leave it listed.

leighmcculloch: @georgelesica-wf I want to make sure I understand your comment correctly. You said agreed to only support three versions, but you also stated to keep 1.8 listed, and those two things seem at odds. Can you confirm, do you want this to read that 1.8 to 1.11 is officially supported, or 1.9 to 1.11?

leighmcculloch: After re-reading a few times, I think you meant to keep the statement at supporting the three major versions (1.9 to 1.11), but that 1.8 could stay listed in the build tests. I've updated the PR to say that. If that isn't correct please let me know.

georgelesica-wf: Yep, that is exactly what I meant. Sorry, I should have phrased that better. I just didn't want to us to change the commitment to three versions even though we may technically support four or more at a given point in time. Thanks!

georgelesica-wf:
@leighmcculloch we're waiting for someone to grant me merge permissions, sorry so slow!

leighmcculloch:
@ernesto-jimenez @matryer Do either of you have merge access, or able to provide @georgelesica-wf with merge access? ❤️

leighmcculloch:
Thanks for merging @georgelesica-wf !

leighmcculloch:
Now that this has been merged I've updated #607 (Add go.mod) based on this.

PR #670, opened 2018-09-18
Fix typo ("PASS" -> "FAIL")

### LINKED ISSUE #669: Output should say "FAIL" instead of "PASS" when a MatchedBy isn't satisfied

xtonyjiang:
Fixes #669

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -691,7 +691,7 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 				output = fmt.Sprintf("%s\t%d: PASS:  %s matched by %s\n", output, i, actualFmt, matcher)
 			} else {
 				differences++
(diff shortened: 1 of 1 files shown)


devdinu:
`func Test_Arguments_Diff_WithArgMatcher(t *testing.T)`  have the case to check assert with argumentMatcher and Diff is called for `Assert`, so when it's mismatched, output should've `FAIL`.


Can you share the code & test case to reproduce and discuss.

bodnarbm:
@devdinu, I believe the test cited ([Test_Arguments_Diff_WithArgMatcher](https://github.com/stretchr/testify/blob/f35b8ab0b5a2cef36673838d662e249dd9c94686/mock/mock_test.go#L1249-L1269)) only asserts the portion of the message after the PASS or FAIL statements. Thus, it was not caught here where the message had the proper "not matched by" output, but had the erroneous  "PASS".

georgelesica-wf:
Re-running CI, I'll merge once it finishes.

PR #675, opened 2018-09-28
fix: InDeltaf function comment error

lsytj0413:


boyan-soubachov:
Given that we haven't had any progress since September 2018, I will clone these changes into a PR

boyan-soubachov:
Having investigated this in further detail, it's actually an issue with the `codegen` utility. I am busy on a related PR which will fix this permanently. This one will get overwritten on the next `go generate` run.

Closing this PR.

PR #679, opened 2018-10-02
Fixed type assertion bug

kuntao:


pwfcurry:
Can I suggest adding the following test cases -

```go
	if Equal(mockT, pointerOf("a"), pointerOf("b")) {
		t.Error("Equal should return false (pointer equality is based on equality of underlying value)")
	}
	if Equal(mockT, "a", pointerOf("b")) {
		t.Error("Equal should return false")
	}
```

pwfcurry:
On second thoughts, perhaps this isn't the best approach. When comparing string pointers, the diff is unhelpful -

```
	Error:      	Not equal: 
	            	expected: (*string)(0xc42052a6d0)
	            	actual  : (*string)(0xc42052a6e0)
```

I've submitted https://github.com/stretchr/testify/pull/680 which gives a more useful message.

PR #680, opened 2018-10-05
fixed Equals on string pointers

pwfcurry:


pwfcurry:
E.g.,

```
	Error:      	Not equal: 
	            	expected: (*string)(0xc42025e120)
	            	actual  : (*string)(0xc42025e130)
	            	
	            	Diff:
	            	--- Expected
	            	+++ Actual
	            	@@ -1,2 +1,2 @@
	            	-(*string)((len=3) "abc")
	            	+(*string)((len=4) "abcd")
```

assert/assertions.go
```diff
@@ -1333,8 +1335,8 @@ func diff(expected interface{}, actual interface{}) string {
 		return ""
 	}
 
-	et, ek := typeAndKind(expected)
-	at, _ := typeAndKind(actual)
+	et, ek, expectedIsPointer := typeAndKind(expected)
+	at, _, actualIsPointer := typeAndKind(actual)
```
devdinu: @pwfcurry Instead of adding `bool` and changing `typeAndKind` func, we could check `ek` and `ak` is reflect.Ptr.

Could you add the testcase to verify the output, when we deal with pointers.

devdinu:
As a suggestion, we could inverse the type check in diff.
```
		e = expected.(string)
		a = actual.(string)
``` 
when `e` and `a` is of type string, otherwise we could do `spewConfig.Sdump` that should ensure the type is string, and wouldn't panic.

You could add a test to verify there's no panic in case of different types (so we wouldn't face this again)

assert/assertions.go
pwfcurry: ek/ak are strings at this point, as we're returning `reflect.TypeOf(*str).Kind()`.

I've added tests in assert/assertions_test.go

Melaninneal:
can we get this merged ASAP? this basic functionality that should have been in place from start

assert/assertions_test.go
```diff
@@ -208,6 +208,12 @@ func TestEqual(t *testing.T) {
 	if !Equal(mockT, &struct{}{}, &struct{}{}) {
 		t.Error("Equal should return true (pointer equality is based on equality of underlying value)")
 	}
+	if Equal(mockT, pointerOf("a"), pointerOf("b")) {
+		t.Error("Equal should return false (pointer equality is based on equality of underlying value)")
+	}
+	if Equal(mockT, "a", pointerOf("b")) {
+		t.Error("Equal should return false")
+	}
```
georgelesica-wf: Would these tests have failed before your changes, though? It doesn't seem that way. Could you verify?

georgelesica-wf:
Would you mind adding a description to the PR that outlines the problem and your solution? Thanks!

georgelesica-wf:
Also, please consider the comment above from @devdinu.

PR #686, opened 2018-10-27
Time duration diff format

### LINKED ISSUE #626: formatUnequalValues with %v instead of %#v?
Is there a way to get the default format when assert.Equal fails? Specifically, I am comparing time.Duration values, and getting an int64 of microseconds is not as friendly as Duration.String. ([example](https://play.golang.org/p/vA3xeX5-OEr))

I understand that the current behavior is generally more useful. Rather than changing the default, I am hoping there is some option I can invoke to make my own test output more readable.

Sample output:
```
        assertions.go:256:
                        Error Trace:    integration_test.go:125
                        Error:          Not equal:
                                        expected: 180000000000
                                        actual  : 0
                        Test:           TestDurationFormat
```

Preferred output:
```
        assertions.go:256:
                        Error Trace:    integration_test.go:125
                        Error:          Not equal:
                                        expected: 3m0s
                                        actual  : 0s
                        Test:           TestDurationFormat
```

devdinu:
Time Duration mismatch from nanoseconds (int64) to readable format.

Fixes #626.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -357,9 +357,11 @@ func formatUnequalValues(expected, actual interface{}) (e string, a string) {
 		return fmt.Sprintf("%T(%#v)", expected, expected),
 			fmt.Sprintf("%T(%#v)", actual, actual)
 	}
-
-	return fmt.Sprintf("%#v", expected),
-		fmt.Sprintf("%#v", actual)
+	switch expected.(type) {
+	case time.Duration:
+		return fmt.Sprintf("%v", expected), fmt.Sprintf("%v", actual)
+	}
+	return fmt.Sprintf("%#v", expected), fmt.Sprintf("%#v", actual)
 }
 
 // EqualValues asserts that two objects are equal or convertable to the same types
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -1499,6 +1499,15 @@ Diff:
 	Equal(t, expected, actual)
 }
 
+func TestTimeEqualityErrorFormatting(t *testing.T) {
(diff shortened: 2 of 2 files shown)


devdinu:
@boyan-soubachov could you review this and merge

assert/assertions.go
```diff
@@ -357,9 +357,11 @@ func formatUnequalValues(expected, actual interface{}) (e string, a string) {
 		return fmt.Sprintf("%T(%#v)", expected, expected),
 			fmt.Sprintf("%T(%#v)", actual, actual)
 	}
-
-	return fmt.Sprintf("%#v", expected),
-		fmt.Sprintf("%#v", actual)
+	switch expected.(type) {
```
boyan-soubachov: Any reason for preferring a `switch` over a plain `if`? Would it make sense to have it as an:
```
if expected.(type) == time.Duration {
    return ...
}
return ...
```
and if we need to add more cases in the future, we can convert to a `switch`?

assert/assertions_test.go
```diff
@@ -1499,6 +1499,15 @@ Diff:
 	Equal(t, expected, actual)
 }
 
+func TestTimeEquality(t *testing.T) {
```
boyan-soubachov: Would it make sense to have a more explicit test name which very apparently describes what we're testing? A suggestion would be: `TestTimeEqualityFormatting`.

devdinu: Updated.

assert/assertions.go
devdinu: type check needs switch

Issue #514, opened 2017-10-21
Very confusing output if call.Repeantability exceeded


decibel:
If too many calls with the same exact inputs happen, the error report is extremely confusing:

```
mock: Unexpected Method Call
---------------------------

MyFunc(...)
   0: 'a'

The closest call I have is:

MyFunc(...)
   0: 'a'
```

I think the best way to handle this would be for Mock.MethodCalled() to include `closestCall.totalCalls` in it's output somehow. Is there a better way?

It would also be nice to report how many calls have already occured, vs how many were originally allocated, but that's hard to do since Repeatability gets modified. It seems that Repeatability could just be left alone and totalCalls compared to it, no?

ernesto-jimenez:
The whole setup to keep track of calls is quite messy because the original package exposed too much of its internals like `Repeatability`, which means we cannot change that behaviour without introducing a breaking change.

We can definitely improve on the error message and specify that calls have been exhausted on the closest call.

Something like:
```
mock: Unexpected Method Call
---------------------------

MyFunc(...)
   0: 'a'

The closest call I have is:

MyFunc(...) - Total calls received: 3 (maximum reached)
   0: 'a'
```

PR #687, opened 2018-10-28
Improve output when method called too many times

zkry:
Allow the MethodCalled logic to be able to differentiate between a mismatched call and a call that exceeded its limit.

fixes: #514 

In order to be able to distinguish between a mismatching call that exceeded its limit vs the correct call that exceeded its limit, I changed the signature of the findClosestCall method to return this information. 

I also noticed a bug related to this in the findClosestCall method where a diff count of 0 meant both this-is-the-first-iteration, as well as this-is-a-perfect-match. I updated this so that a diffCount value of -1 meant that this is our fist iteration while a zero value meant a perfect match. Not doing this will make subsequent comparisons after a perfect match overwrite that match.

The final result looks like this:
screenshot 2018-10-28 at 12 24 47


st3penta:
Duplicate of https://github.com/stretchr/testify/pull/767

brackendawson:
Closing as duplicate of #767

PR #689, opened 2018-11-16
Add a WithinTimes assertion

goodspark:
I've been testing some code that generates timestamps and found that I'm often thinking in terms of 'time should be between a and b'. So I think this might be helpful.

Example of how I'd like to write my tests:

```go
before := time.Now()
data := DoSomething()
after := time.Now()

assert.WithinTimes(t, before, data.timestamp, after)
```

Without this, I have to use `WithinDuration`, which I find _slightly_ cumbersome to setup and read.

```go
before := time.Now()
data := DoSomething()
span := time.Since(before) /2
midpoint := before.Add(span)

assert.WithinDuration(t, midpoint, data.timestamp, span)
```

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -933,6 +933,24 @@ func WithinDuration(t TestingT, expected, actual time.Time, delta time.Duration,
 	return true
 }
 
+// WithinTimes asserts that the actual time is between the two given times.
+//
+//   before := time.Now()
+//   now := time.Now()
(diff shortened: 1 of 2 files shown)


goodspark:
I have no idea how to get `go generate` working. I'm on go 1.11.1 and the `go generate ./...` command yields the same problems as https://travis-ci.org/stretchr/testify/jobs/456142014#L477

PR #691, opened 2018-11-24
Add NaN function

### LINKED ISSUE #624: assert two NaN values
when i try to assert two NaN values the assert result was
```
$ go test
--- FAIL: TestNaN (0.00s)
	main_test.go:10: 
			Error Trace:	main_test.go:10
			Error:      	Not equal: 
			            	expected: NaN
			            	actual  : NaN
			Test:       	TestNaN
FAIL
FAIL	command-line-arguments	0.230s
```
the code that genrate this error is
```golang
package test

import (
        "math"
        "testing"
)

func TestNaN(t *testing.T) {
        assert.Equal(t, math.NaN(), math.NaN())
}
```

because the `assert.Equal` is based on `reflerct.DeepEqual` in it documentations written that two floats a,b will be equal if a == b is true.
the golang the act according IEEE 754 is defined NaN == NaN as false.
so this may need discussion about adding assert checking for NaN values.

yahelmanor:
Adding NaN to assert.
Adding tests for the NaN function.

resolve #624 .

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"math/cmplx"
 	"os"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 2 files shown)


georgelesica-wf:
I'm hesitant to expand the API for this since it is a pretty niche use-case that is already covered pretty cleanly, as noted in the issue discussion: `assert.True(t, math.IsNaN(i))`.

The way I see it, there are two benefits to having a special assertion for a given type or situation:

  1. Brevity / conciseness / readability
  2. Useful failure messages

In this case, the alternative is nearly as concise, so we need to think about (2). In this case, the failure message would include the value that was actually found, which might be helpful. But if you're expecting NaN, how likely is the actual value to matter? I would think that the operands used to produce the value would be more useful, but we can't automatically include those in the message since we don't know them. Therefore, users will still, likely, end up writing custom failure messages.

So I'm just not sure that this is helpful enough to warrant its own assertion, particularly given that I wouldn't expect this assertion to be used frequently.

I'm happy to discuss it further, however.

kylebrandt:
@georgelesica-wf I think the case that gets messy is when you have NaNs inside a data structure. `go-cmp` takes some options on how to compare things which is pretty nice, see (https://godoc.org/github.com/google/go-cmp/cmp#example-Option--EqualNaNs). 

This allows for basic Object in / Object out table tests when working with any data structure that has NaN values in it.

PR #694, opened 2018-12-04
Add String method to Mock to fix #423

### LINKED ISSUE #423: Using mock struct as arg to mocked method causes data race
If a mock struct with a method is passed as an argument to a method on another mock struct, and both methods are invoked simultaneously in separate goroutines, a data race occurs. In one goroutine, the first mock instance is written to update call counts or the list of calls on its method, while in another goroutine the second mock instance is logging its arguments, which includes a walk of all the data on the first instance.

For example, test the following files with `go test -race`:

main.go
```
package main

import "sync"

type IInterface1 interface {
	SomeFunc1() error
}

type IInterface2 interface {
	SomeFunc2(IInterface1) error
}

type Implementation1 struct {
}

func (i Implementation1) SomeFunc1() error {
	return nil
}

type Implementation2 struct {
}

func (i Implementation2) SomeFunc2(i1 IInterface1) error {
	return nil
}

func DoSomeStuff(i1 IInterface1, i2 IInterface2) error {
	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		for j := 0; j < 10000; j++ {
			i2.SomeFunc2(i1)
		}
		wg.Done()
	}()
	for i := 0; i < 10000; i++ {
		i1.SomeFunc1()
	}
	wg.Wait()
	return nil
}

func main() {
	var i1 Implementation1
	var i2 Implementation2
	_ = DoSomeStuff(i1, i2)
}
```

main_test.go
```
package main

import (
	"testing"

)

type (
	MockBugTestSuite struct {
		suite.Suite
	}
)

type IInterface1Mock struct {
	mock.Mock
}

func (_m IInterface1Mock) SomeFunc1() error {
	ret := _m.Called()

	var r0 error
	if rf, ok := ret.Get(0).(func() error); ok {
		r0 = rf()
	} else {
		r0 = ret.Error(0)
	}

	return r0
}

type IInterface2Mock struct {
	mock.Mock
}

func (_m IInterface2Mock) SomeFunc2(i1 IInterface1) error {
	ret := _m.Called(i1)

	var r0 error
	if rf, ok := ret.Get(0).(func(IInterface1) error); ok {
		r0 = rf(i1)
	} else {
		r0 = ret.Error(0)
	}

	return r0
}

func (s *MockBugTestSuite) TestDoSomeStuff() {
	var i1 IInterface1Mock
	var i2 IInterface2Mock

	i1.On("SomeFunc1").Return(nil)
	i2.On("SomeFunc2", mock.Anything).Return(nil)
	err := DoSomeStuff(i1, i2)
	s.Nil(err)
}

func TestPermTestSuite(t *testing.T) {
	suite.Run(t, new(MockBugTestSuite))
}
```

### LINKED ISSUE #625: race condition on master in mock
Upgraded to master and tests that previously ran without issues now fail the race detector:

```
$ go test ./model/modelhelpers/ -race -run TestUpdateRefreshScheduleTimerSuccess
==================
WARNING: DATA RACE
Read at 0x00c4203a34e8 by goroutine 21:
  reflect.typedmemmove()
      /usr/local/Cellar/go/1.10.2/libexec/src/runtime/mbarrier.go:265 +0x0
  reflect.packEface()
      /usr/local/Cellar/go/1.10.2/libexec/src/reflect/value.go:119 +0x101
  reflect.valueInterface()
      /usr/local/Cellar/go/1.10.2/libexec/src/reflect/value.go:978 +0x16a
  reflect.Value.Interface()
      /usr/local/Cellar/go/1.10.2/libexec/src/reflect/value.go:948 +0x51
  fmt.(*pp).printValue()
      /usr/local/Cellar/go/1.10.2/libexec/src/fmt/print.go:699 +0x3700
  fmt.(*pp).printValue()
      /usr/local/Cellar/go/1.10.2/libexec/src/fmt/print.go:853 +0x2550
  fmt.(*pp).printArg()
      /usr/local/Cellar/go/1.10.2/libexec/src/fmt/print.go:689 +0x192
  fmt.(*pp).doPrintf()
      /usr/local/Cellar/go/1.10.2/libexec/src/fmt/print.go:1099 +0x912
  fmt.Sprintf()
      /usr/local/Cellar/go/1.10.2/libexec/src/fmt/print.go:203 +0x73
  some.company.com/some/go-repo/vendor/github.com/stretchr/testify/mock.Arguments.Diff()
      /Users/andrewdeandrade/go/src/some.company.com/some/go-repo/vendor/github.com/stretchr/testify/mock/mock.go:678 +0x1416
  some.company.com/some/go-repo/vendor/github.com/stretchr/testify/mock.(*Mock).findExpectedCall()
      /Users/andrewdeandrade/go/src/some.company.com/some/go-repo/vendor/github.com/stretchr/testify/mock/mock.go:267 +0x175
  some.company.com/some/go-repo/vendor/github.com/stretchr/testify/mock.(*Mock).MethodCalled()
      /Users/andrewdeandrade/go/src/some.company.com/some/go-repo/vendor/github.com/stretchr/testify/mock/mock.go:343 +0xb3
  some.company.com/some/go-repo/vendor/github.com/stretchr/testify/mock.(*Mock).Called()
      /Users/andrewdeandrade/go/src/some.company.com/some/go-repo/vendor/github.com/stretchr/testify/mock/mock.go:333 +0x18a
  some.company.com/some/go-repo/mocks.(*ProvidesTimers).Remove()
      /Users/andrewdeandrade/go/src/some.company.com/some/go-repo/mocks/provides_timers.go:40 +0x2a1

Previous write at 0x00c4203a34e8 by goroutine 20:
  sync/atomic.CompareAndSwapInt32()
      /usr/local/Cellar/go/1.10.2/libexec/src/runtime/race_amd64.s:293 +0xb
  sync.(*Mutex).Lock()
      /usr/local/Cellar/go/1.10.2/libexec/src/sync/mutex.go:74 +0x4d
  some.company.com/some/go-repo/vendor/github.com/stretchr/testify/mock.(*Mock).AssertCalled()
      /Users/andrewdeandrade/go/src/some.company.com/some/go-repo/vendor/github.com/stretchr/testify/mock/mock.go:489 +0xc8
  some.company.com/some/go-repo/model/modelhelpers.TestUpdateRefreshScheduleTimerSuccess()
      /Users/andrewdeandrade/go/src/some.company.com/some/go-repo/model/modelhelpers/helpers_test.go:267 +0xafe
  testing.tRunner()
      /usr/local/Cellar/go/1.10.2/libexec/src/testing/testing.go:777 +0x16d

Goroutine 21 (running) created at:
  some.company.com/some/go-repo/model/modelhelpers.removeTimerAsync()
      /Users/andrewdeandrade/go/src/some.company.com/some/go-repo/model/modelhelpers/helpers.go:94 +0x10f
  some.company.com/some/go-repo/model/modelhelpers.UpdateRefreshScheduleTimer()
      /Users/andrewdeandrade/go/src/some.company.com/some/go-repo/model/modelhelpers/helpers.go:152 +0x2c2
  some.company.com/some/go-repo/model/modelhelpers.TestUpdateRefreshScheduleTimerSuccess()
      /Users/andrewdeandrade/go/src/some.company.com/some/go-repo/model/modelhelpers/helpers_test.go:241 +0x5e0
  testing.tRunner()
      /usr/local/Cellar/go/1.10.2/libexec/src/testing/testing.go:777 +0x16d

Goroutine 20 (running) created at:
  testing.(*T).Run()
      /usr/local/Cellar/go/1.10.2/libexec/src/testing/testing.go:824 +0x564
  testing.runTests.func1()
      /usr/local/Cellar/go/1.10.2/libexec/src/testing/testing.go:1063 +0xa4
  testing.tRunner()
      /usr/local/Cellar/go/1.10.2/libexec/src/testing/testing.go:777 +0x16d
  testing.runTests()
      /usr/local/Cellar/go/1.10.2/libexec/src/testing/testing.go:1061 +0x4e1
  testing.(*M).Run()
      /usr/local/Cellar/go/1.10.2/libexec/src/testing/testing.go:978 +0x2cd
  main.main()
      _testmain.go:52 +0x22a
```

Has anyone else encountered this?

neilisaac:
Proposed fix for #423 and possibly #625 if it's the same issue.

See comment on Mock.String for explanation.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -202,6 +202,14 @@ type Mock struct {
 	mutex sync.Mutex
 }
 
+// String provides a %v format string for Mock.
+// Note: this is used implicitly by Arguments.Diff if a Mock is passed.
+// It exists because go's default %v formatting traverses the struct
+// without acquiring the mutex, which is detected by go test -race.
+func (m *Mock) String() string {
+	return fmt.Sprintf("%[1]T<%[1]p>", m)
+}
+
 // TestData holds any data that might be useful for testing.  Testify ignores
 // this data completely allowing you to do whatever you like with it.
 func (m *Mock) TestData() objx.Map {
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -1495,3 +1495,40 @@ func unexpectedCallRegex(method, calledArg, expectedArg, diff string) string {
 func ConcurrencyTestMethod(m *Mock) {
 	m.Called()
 }
+
+func TestConcurrentArgumentRead(t *testing.T) {
+	methodUnderTest := func(c caller, u user) {
+		go u.Use(c)
+		c.Call()
+	}
+
+	c := &mockCaller{}
+	defer c.AssertExpectations(t)
+
+	u := &mockUser{}
+	defer u.AssertExpectations(t)
+
+	done := make(chan struct{})
+
+	c.On("Call").Return().Once()
+	u.On("Use", c).Return().Once().Run(func(args Arguments) { close(done) })
+
+	methodUnderTest(c, u)
+	<-done // wait until Use is called or assertions will fail
+}
+
+type caller interface {
+	Call()
+}
+
+type mockCaller struct{ Mock }
+
+func (m *mockCaller) Call() { m.Called() }
+
+type user interface {
+	Use(caller)
+}
+
+type mockUser struct{ Mock }
+
+func (m *mockUser) Use(c caller) { m.Called(c) }


glesica:
OK, this looks pretty harmless to me. I'd like to get some confirmation that it actually fixes the problems various people seem to be having before we merge it, though. I'll ping some of the people who have brought up this issue and get feedback here.

bhcleek:
The test I created for #625, https://github.com/stretchr/testify/compare/master...bhcleek:race-test, still fails.

In any case, I don't see how this PR could solve the problem caused by a mutex not being held since the mutex is not locked in the new `String` method.

neilisaac:
This avoids the race in #423 by avoiding traversing the mock.Mock data structure.  mock.Arguments.Diff uses fmt.Sprintf, which will use reflection to traverse the arguments, triggering the race detector.  This PR simply avoids that traversal by providing a trivial String implementation.

This should fix a common special case of a more general issue in testify/mock.  This come up in some of my code where a mocked interface takes an interface type as an argument which is also mocked in the same test.  The test case I added in this PR fails `go test -race ./...` in master.

@bhcleek's test case here https://github.com/stretchr/testify/compare/master...bhcleek:race-test demonstrates the general case of this bug: doing recursive string formatting in mock.Arguments.Diff can cause races that the user can't easily avoid (since it still does the string formatting first regardless of what kind of argument matcher is being used.)

neilisaac:
I thought about this some more.  I think a holistic fix is tricky if you want to be able to print the actual value passed when it doesn't match an expected call.

In general, there is a data race in tests that have concurrent access to a data structure passed as an argument to a mock that's traversed in its own String implementation.  context.WithValue violates this as in @bhcleek's example.

Here's the simplest test case for the general problem:
```go
func TestConcurrentSyncArgument(t *testing.T) {
	m := &Mock{}
	m.On("TestConcurrentSyncArgument", Anything).Return()

	s := struct{ x int }{1}
	go func() { s.x++ }()

	m.Called(&s)

	m.AssertExpectations(t)
}
```

This test case will still fail with this PR, but I think it's worth merging since it fixes the more special case where passing mocks to mocks always breaks.

neilisaac:
Is there any objection to merging this?  As noted, it doesn't fix the general case, but does fix a special case that crops up when you pass a mock to a mock in tests with concurrency (as demonstrated by the test case in this PR.)

mock/mock.go
```diff
@@ -203,6 +203,14 @@ type Mock struct {
 	mutex sync.Mutex
 }
 
+// String provides a %v format string for Mock.
+// Note: this is used implicitly by Arguments.Diff if a Mock is passed.
+// It exists because go's default %v formatting traverses the struct
+// without acquiring the mutex, which is detected by go test -race.
+func (m *Mock) String() string {
```
xorcare: It is very strange to implement the stringer method through a pointer and not through a value. most often this method is implemented by value.

neilisaac: @xorcare the Mock holds a mutex, so we can't pass by value or the race detector will detect the copy

neilisaac:
Despite not fixing the general case, this PR fixes a real issue demonstrated in this unit test: https://github.com/stretchr/testify/pull/694/files#diff-d4376f2727e58f552d2da8aa2c4deb7fR1501.  

This library does string formatting on arguments to Mock.On, and Mock contains a Mutex, which the race detector detects.  It seems wrong that you can't use a mock as an argument to another mock, so can we merge this fix so I can remove workarounds in my tests?

danielbh:
Any updates on this merge please? It would be very helpful for us. Thank you.

7fold:
We are experiencing issues with this as well.

boyan-soubachov (approved):

Issue #695, opened 2018-12-04
Recover in suite


YaroslavDev:
IMHO, panic in tests is not a good idea. Not only because it's hard to understand what test panicked, but panic also crashes entire test run and some tests don't even start. I think it's a good idea to isolate each test run by putting a defer/recover so that you know exactly what test caused panic and give possibility for other tests to run. 

Github PR implementing this feature: https://github.com/stretchr/testify/pull/696

iaroslav-ciupin:
P.S.
If you want to test that some of your code panics as expected, you must explicitly put a recover in your test, catch actual panic and compare it with expected one. 
IMHO,
Panic **SHOULD NOT** propagate outside scope of a test case, because in a context of testing *panic is just another type of error* that should be caught and checked.

itzamna314:
+1 to this.  With test suites that have some environmental setup/teardown, a panic will exit tests before the teardown can happen.  Then you've got resources you have to clean up manually.

Real-world example: I have a test suite that sets up a docker database container, then runs tests against it, and removes the container plus all volumes when everything is complete.  If I have a panic in my code, I have to manually go through and clear out the container and its volumes.

YaroslavDev:
Merged

PR #696, opened 2018-12-04
Recover panic in suite

YaroslavDev:
## Synopsis

PR implements feature suggested in https://github.com/stretchr/testify/issues/695 (don't panic in tests, but gracefully recover, mark test as failed and run other tests)

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -55,10 +55,19 @@ func (suite *Suite) Assert() *assert.Assertions {
 	return suite.Assertions
 }
 
+func failOnPanic(t *testing.T) {
+	r := recover()
+	if r != nil {
+		t.Errorf("test panicked: %v", r)
+		t.FailNow()
+	}
+}
+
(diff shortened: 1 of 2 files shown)


iaroslav-ciupin:
@ernesto-jimenez @matryer can you please take a look ?

iaroslav-ciupin:
@georgelesica-wf , can you please take a look ?

georgelesica-wf:
Since this is a pretty significant change I'm going to get a few people to look at it and discuss, but I'll keep you posted.

benechols-wf:
+1 This is roughly the behavior of go tests already (panics are caught by the test framework and reported asa failure) so i think it makes sense for a custom test sit suite to provide the same feature.  Also, being able to execute cleanup steps even if there isa panic makes sense.

georgelesica-wf:
Sorry, I've been pretty busy. I'll merge it now, thank you for your patience! Sorry it took awhile.

PR #699, opened 2018-12-05
Support arbitrary objects as the sole element in msgAndArgs

viswajithiii:
Currently, if `msgAndArgs` contains exactly one element, it's expected to be a `string` -- and if it's not, the framework panics, which makes it hard to debug the actual test failure. So, if the object passed is not a string, then print it using the `%+v` modifier.

This will make life slightly more convenient for clients. It doesn't break existing programs in any way.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -179,7 +179,11 @@ func messageFromMsgAndArgs(msgAndArgs ...interface{}) string {
 		return ""
 	}
 	if len(msgAndArgs) == 1 {
-		return msgAndArgs[0].(string)
+		msg := msgAndArgs[0]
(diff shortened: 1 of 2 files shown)


viswajithiii:
@georgelesica-wf Can you review this?

(Test failure is only on `tip`, and appears to be unrelated.)

viswajithiii:
Thanks @georgelesica-wf!

viswajithiii:
Thanks for merging! Do you know when you'll cut a release next?

georgelesica-wf:
I need to figure out the process the project follows for that, but I'll try to do it next week.

PR #712, opened 2019-01-03
Preserve stack frame for mock parent method call

### LINKED ISSUE #705: Build is broken on Go tip
As of https://github.com/stretchr/testify/pull/699 the CI build on Go tip appears to be broken. We should consider fixing this immediately so as to avoid blocking future pull requests or causing breakage when the next release comes out.

letientai299:
Fixes #705 

Go tip contains following commmit, that inlines function with single
call bodies.

https://github.com/golang/go/commit/13baf4b2cd34dfb41c570e35b48ec287713f4d7f

`(*Call).On()` is the exact target for the improvement in Go repo. Due to
the inlining, assert.CallerInfo() can't not detect the file and line
number of the call to `(*Mock).On()` from `(*Call).On()`. Thus, the test
fails.

Adding the compiler directive `go:noinline` prevent this effect and make
mock package works with go tip as before.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -176,6 +176,7 @@ func (c *Call) Maybe() *Call {
 //    Mock.
 //       On("MyMethod", 1).Return(nil).
 //       On("MyOtherMethod", 'a', 'b', 'c').Return(errors.New("Some Error"))
+//go:noinline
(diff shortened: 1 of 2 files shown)


letientai299:
This would fix #705

georgelesica-wf:
Fixes #705

Issue #597, opened 2018-05-08
How to assert the address of 2 pointers are (not) equal?


littledot:
According to the docs, pointers are considered equals as long as their ref'd values are equal. 

```
// NotEqual asserts that the specified values are NOT equal.
//
//    assert.NotEqual(t, obj1, obj2)
//
// Pointer variable equality is determined based on the equality of the
// referenced values (as opposed to the memory addresses).
```

However, this can be difficult when testing clone functions because in addition to testing the referenced values, I also need to test the memory address are new and not the same. 

```go

type A struct {
	i int
}

func (a *A) Clone() *A {
	clone := *a
	return &clone
}

func TestClone(t *testing.T) {
	orig := &A{1}
	copy := orig.Clone()

	assert.Equal(t, *orig, *copy)
	assert.NotEqual(t, orig, copy) // Fail, how to compare memory address values?
}
```

Are there any functions that compares memory addresses values of pointers?

devdinu:
@littledot As its a pointer, You could just use `==` operator to verify its different memory.

`orig==copy` is `false` and `orig==orig` is `true`.

littledot:
Yes I can do == with assert.True(), but the default error log for assert.True() is vague due to a lack of context. Hence I'd like to know if there is a dedicated assert function for comparing memory addresses? Which will also pretty print the pointers on assertion failures.

littledot:
What are your thoughts on a new assertion function `assert.Same(*testing.T, interface{}, interface{})` that simply wraps the `==` check? 

https://play.golang.org/p/ZEzrtMb7oQN

devdinu:
looks good, we also have to check the types of the two pointers, and report properly if its different, and not just print the different address.

drdgvhbh:
Any updates on this?

gavincabbage:
Was looking for this today so I took a stab at it based on @littledot's suggestion

juliaogris:
`NotSame` would also be great.

piepmatz:
Can be closed, `Same` and `NotSame` are available.

ilya-korotya:
@devdinu This issue was resolved. Could you close the issue, please?

antsupovsa:
Maybe you know how recursive doing Same or NotSame?

PR #714, opened 2019-01-10
add ptr comparison func `Same` to assert pkg

gavincabbage:
Adds pointer comparison function `Same` to `assert` package, based on discussion in #597.

Both arguments must be pointer variables. Pointer variable sameness is determined based on the equality of both type and value.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -350,6 +350,37 @@ func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) 
 
 }
 
+// Same asserts that two pointers reference the same object.
+//
+//    assert.Same(t, ptr1, ptr2)
+//
+// Both arguments must be pointer variables. Pointer variable sameness is
+// determined based on the equality of both type and value.
+func Same(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+
+	expectedPtr, actualPtr := reflect.ValueOf(expected), reflect.ValueOf(actual)
+	if expectedPtr.Kind() != reflect.Ptr || actualPtr.Kind() != reflect.Ptr {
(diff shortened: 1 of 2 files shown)


gavincabbage:
Please review when you have a moment @georgelesica-wf. Thanks!

georgelesica-wf:
Two pointers that point at the same piece of memory should have the same value, why not just assert that the pointers themselves are equal? For example... https://play.golang.com/p/1nnLRb2nPA-

gavincabbage:
Thanks for taking a look @georgelesica-wf 

Can you clarify your question? Are you wondering why I bothered with the pointer type checking?

```
expectedType, actualType := reflect.TypeOf(expected), reflect.TypeOf(actual)
	if expectedType != actualType {
		return Fail(t, fmt.Sprintf("Pointer expected to be of type %v, but was %v",
			expectedType, actualType), msgAndArgs...)
	}
```

I added that per @devdinu's comment in #597 "we also have to check the types of the two pointers, and report properly if its different" - however, given Go has no pointer arithmetic, I'm unable to write a test to exercise that condition, so it may be unnecessary. I would be interested to hear more about what they meant, perhaps I misunderstood. Happy to remove that part if we agree it is unnecessary.

To give more background, the purpose of this PR is best described by @littledot's original issue: to provide more meaningful output than `assert.True(t, ptr1==ptr2)` when checking two pointers reference the same object (per the docs, `assert.Equal` wouldn't be appropriate).

Thanks again for the review. Looking forward to incorporating your feedback.

syndbg:
It would be great if is merged upstream.

It's very useful and saves writing for yourself the proposed `assert.Same` functionality

gavincabbage:
@georgelesica-wf Just wanted to reach out in case you had any more thoughts on this PR and my response to your question. If you're not interested, please feel free to close this PR, though it does seem there is appetite for this feature given the discussion around it here and in #597.

Thanks for the time!

georgelesica-wf:
OK, let's move forward with it since there's demand. I'm going to get a couple other sets of eyes on the PR and then I'll merge.

PR #717, opened 2019-01-21
Fix issue with GoLand debugger using testify

eldanelias:
GoLand's debugger hangs given *goquery.Document as the value to the Sprintf function
This makes it impossible to debug tests in GoLand which is too bad
%v --> %#v make it so the debugger won't hang

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -676,7 +676,7 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			actualFmt = "(Missing)"
 		} else {
 			actual = objects[i]
-			actualFmt = fmt.Sprintf("(%[1]T=%[1]v)", actual)
+			actualFmt = fmt.Sprintf("(%[1]T=%#[1]v)", actual)
 		}
 
 		if len(args) <= i {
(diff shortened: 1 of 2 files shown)


eschwartz:
@glesica any chance we could get a look at this one?

@eldanelias -- did you ever figure out what was causing this? It seems to be certain data types that get stuck here...

glesica:
Is this still reproducible? Because it kinda sounds like a problem with the debugger. Not that we shouldn't mitigate it here, I'm just wondering if JetBrains has perhaps fixed it on their end since the pull request was created?

eschwartz:
@glesica -- here's an example of it happening in an open source repo,
if I try to put a breakpoint on line 143.

It doesn't happen every time, so I'm not sure exactly what scenarios it fails in. But the changes in this PR fix it. It sure would be nice to get this in, if it's not a big deal..

mvdkleijn:
@eschwartz - Unless I'm blind, you did not link to anything? :smile:

boyan-soubachov:
I would be wary of modifying the existing API to work-around specific tooling. In my opinion this is an issue that should be fixed with GoLand (an IDE should never hang) rather than worked-around in the package.

glesica:
I'm not sure what to think here, but I agree that it wouldn't be great to alter the test output for everyone to work around what appears to be a bug in GoLand.

boyan-soubachov:
I am closing this PR as we've had no feedback and the issue sounds like it's with GoLand's debugger rather than testify. Happy to re-open it if anyone disagrees.

PR #721, opened 2019-01-25
Prevent misleading test result output

letientai299:
Some tests intentionally use and let golang `InternalTest` to fail. Those `InternalTest` print "FAIL" messages to stdout, with is confusing readers. This commit redirect `os.Stdout` during those test to suppress the messages and revert original `Stdout` back one done.

---

Before this commit, we got following lines after running `go test -v ./....`

```
...
PASS
ok      github.com/stretchr/testify/require     (cached)
=== RUN   TestSuiteRequireTwice
=== RUN   TestSuiteRequireTwice
=== RUN   TestSuiteRequireTwice/TestRequireOne
=== RUN   TestSuiteRequireTwice/TestRequireTwo
--- FAIL: TestSuiteRequireTwice (0.00s)
    --- FAIL: TestSuiteRequireTwice/TestRequireOne (0.00s)
        require.go:157:
                Error Trace:    suite_test.go:37
                Error:          Not equal:
                                expected: 1
                                actual  : 2
                Test:           TestSuiteRequireTwice/TestRequireOne
    --- FAIL: TestSuiteRequireTwice/TestRequireTwo (0.00s)
        require.go:157:
                Error Trace:    suite_test.go:42
                Error:          Not equal:
                                expected: 1
                                actual  : 2
                Test:           TestSuiteRequireTwice/TestRequireTwo
...
```
I had to debug the source code to know why exit code (`$?`) is still 0.

### DIFF
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -19,6 +19,11 @@ type SuiteRequireTwice struct{ Suite }
 // suite.requirements was not initialised in suite.SetT()
 // A regression would result on these tests panicking rather than failing.
 func TestSuiteRequireTwice(t *testing.T) {
+	stdout := os.Stdout
+	os.Stdout, _ = ioutil.TempFile("", "")
+	defer func() {
+		os.Stdout = stdout
+	}()
 	ok := testing.RunTests(
 		allTestsFilter,
 		[]testing.InternalTest{{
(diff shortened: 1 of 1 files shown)


boyan-soubachov:
These are just the unit tests for testify itself. These are not used or consumed by our end-users. I think it makes more sense to keep the code as simple as possible given the fringe (at the moment, none) benefit of this.

If the community feels strongly about looking into this, I'd be happy to re-open the PR :)

PR #724, opened 2019-01-31
Introduce Eventually assertion.

gmiejski:
Introducing `assert.Eventually`, to verify if given condition will be met in given time.

Such method is useful for testing asynchronous actions, for example in integration tests we can use `assert.Eventually` to verify that condition is met in given time, instead of using `time.Sleep()` with predefined and always too long wait duration - this way tests complete faster.

Example usage? Send an event to some event bus, then wait for condition to be satisfied in given time (for example wait that HTTP endpoint will return something based on that event consumption)

Other testing tools contain similar methods, for example: http://spockframework.org/spock/javadoc/1.2/spock/util/concurrent/PollingConditions.html

Please share if you find it as valuable as I do.

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -113,6 +113,17 @@ func Errorf(t TestingT, err error, msg string, args ...interface{}) bool {
 	return Error(t, err, append([]interface{}{msg}, args...)...)
 }
 
+// Eventuallyf asserts that given condition will be met in waitFor time,
+// periodically checking target function each tick.
+//
+//    assert.Eventuallyf(t, func() bool { return true; }, time.Second, 10*time.Millisecond, "error message %s", "formatted")
+func Eventuallyf(t TestingT, condition func() bool, waitFor time.Duration, tick time.Duration, msg string, args ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	return Eventually(t, condition, waitFor, tick, append([]interface{}{msg}, args...)...)
+}
+
 // Exactlyf asserts that two objects are equal in value and type.
 //
 //    assert.Exactlyf(t, int32(123, "error message %s", "formatted"), int64(123))
--- a/assert/assertion_forward.go
+++ b/assert/assertion_forward.go
@@ -215,6 +215,28 @@ func (a *Assertions) Errorf(err error, msg string, args ...interface{}) bool {
 	return Errorf(a.t, err, msg, args...)
 }
 
+// Eventually asserts that given condition will be met in waitFor time,
+// periodically checking target function each tick.
+//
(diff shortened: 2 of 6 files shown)


gmiejski:
Bump? If testify maintainers are not interested in such feature, than we can close this PR.

kandros:
Hello @gmiejski, I tried your branch but I'm unsure of how it should behave when checking values available via a closure

using this it fails
```
func Test(t *testing.T) {
	b := false
	assert.Eventually(t, func() bool {
		t.Log("ticking")
		t.Log(b)
		return b
	}, 3*time.Second, 1*time.Second)
	b = true
}
```

Result
```
--- FAIL: Test (3.00s)
    main_test.go:13: ticking
    main_test.go:14: false
    main_test.go:13: ticking
    main_test.go:14: false
    main_test.go:12: 
                Error Trace:    main_test.go:12
                Error:          Condition never satisfied
                Test:           Test
FAIL
FAIL    command-line-arguments  3.019s
```

gmiejski:
@kandros
Here are my clarifications to when this assertion comes in handy:
- this assertion is blocking the same way as any other assertion - so state changes (`b = true`) will not be executed if it's places in the main "thread" - it will be executed after assertion already failed
- this assertion is useful only when state is being changed in `another goroutine` and we're waiting for something to be True `in the nearest future`. Without this approach (which as I mentioned is commonly adopted in other libraries for other languages) you would have to put artificial `time.sleep(...)` which is nonoptimal, as you have to put bigger sleep time than is necessary for the condition to become true, to avoid too early false negatives test results

Proper minimal example usage, similar to yours would be this:
```
func Test(t *testing.T) {
	var counter uint64

	go func() {
		time.Sleep(2*time.Second)
		atomic.AddUint64(&counter, 1)
	}()

	assert.Eventually(t, func() bool {
		t.Log("ticking")
		currentValue := atomic.LoadUint64(&counter)
		t.Log("currentValue: ", currentValue)
		return currentValue == 1
	}, 3*time.Second, 1*time.Second)
}
```

Result:
```
=== RUN   Test
--- PASS: Test (2.00s)
	main_test.go:42: ticking
	main_test.go:44: currentValue:  0
	main_test.go:42: ticking
	main_test.go:44: currentValue:  1
PASS
```

Also remember that reading values modified by different goroutines might lead to stale reads - in your example I've used atomic taken from here: https://gobyexample.com/atomic-counters 

reach me out if you have any other questions.

georgelesica-wf:
This is intriguing. I'm going to recruit a couple reviewers and we'll try to get it moving.

benechols-wf:
I feel like this can be boiled down to just a for/select with a timer and another result channel and so feels like its more complicated than is needed (especially the waitfor vs tick).

However, I understand that this does mean you don't have to write that bit yourself and so saves a little code in some cases.

+1

PR #725, opened 2019-02-08
Don't setup suite if no methods satisfy filter

Saluev:
When suite setup is long, the necessity to wait for all suite setups for testing one single method of one of suites bothers a lot. Here's a little fix of that behavior.

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -82,7 +82,7 @@ func Run(t *testing.T, suite TestingSuite) {
 	suite.SetT(t)
 	defer failOnPanic(t)
 
-	setupDone := false
+	suiteSetupDone := false
 	
 	methodFinder := reflect.TypeOf(suite)
 	tests := []testing.InternalTest{}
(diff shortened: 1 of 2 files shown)


suite/suite.go
```diff
@@ -82,15 +82,8 @@ func Run(t *testing.T, suite TestingSuite) {
...
-	if setupAllSuite, ok := suite.(SetupAllSuite); ok {
-		setupAllSuite.SetupSuite()
-	}
-	defer func() {
-		if tearDownAllSuite, ok := suite.(TearDownAllSuite); ok {
-			tearDownAllSuite.TearDownSuite()
-		}
-	}()
-
+	setupDone := false
```
georgelesica-wf: Can we rename this to something like `suiteSetupDone`?

georgelesica-wf:
Looks good to me, just a couple small issues. I'm going to have one or two others look at it as well, but I expect we'll have it merged by the end of the week, if not sooner.

suite/suite.go
Saluev: Sure!

Saluev:
Thanks a lot! I've added the test you requested.

Saluev:
Hi @georgelesica-wf, are there any other issues with this PR I can fix?

Saluev:
That's great, thank you!

PR #727, opened 2019-02-11
Show stack trace on panic

g7r:


snigle:
Please merge it, it's difficult to debug without stack trace :s

g7r:
@georgelesica-wf :pray:

georgelesica-wf:
Looks reasonable to me. I'm going to ask a couple other people to take a look just for completeness. My only concern is if people are depending on the output to look a certain way for scripts and such, but hopefully that won't be an issue.

PR #732, opened 2019-02-22
assert.Len: print value last

christopinka:
https://github.com/stretchr/testify/issues/207#issuecomment-464956602 - Adjust assert.Len() fail message making the object last so that large object output doesn't obscure the length difference.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -553,6 +553,7 @@ func Len(t TestingT, object interface{}, length int, msgAndArgs ...interface{}) 
 	}
 	ok, l := getLen(object)
 	if !ok {
+		//TODO: should it be - \"%s\" could not be applied <to> builtin len() or maybe "len() could not be applied to \"%s\" 
 		return Fail(t, fmt.Sprintf("\"%s\" could not be applied builtin len()", object), msgAndArgs...)
 	}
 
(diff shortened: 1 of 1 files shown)


assert/assertions.go
```diff
@@ -553,11 +553,12 @@ func Len(t TestingT, object interface{}, length int, msgAndArgs ...interface{})
 	}
 	ok, l := getLen(object)
 	if !ok {
+		//TODO: should it be - \"%s\" could not be applied <to> builtin len() or maybe "len() could not be applied to \"%s\" 
 		return Fail(t, fmt.Sprintf("\"%s\" could not be applied builtin len()", object), msgAndArgs...)
```
christopinka: Isn't it - the application of the function ƒ to its argument x? len() applied to object? There could be something I'm missing...

srmaxwell3: Yes, I think it should say: "Could not apply len() to \"%s\"".

viktorasm:
This would be a very welcome change - having difference at the end of message makes it hard to read

dolmen (changes requested):
Must be rebased on master.

brackendawson:
Related to #1801

dolmen:
Has #1646 (truncate long messages) solved the readability issue in another way?

At least length is nowadays displayed last, as requested.

PR #733, opened 2019-02-22
refactoring of assertion_order

alexpantyukhin:
Refactoring of the assertion order functionality:
 - Using constants `greater`, `equal`, `less` instead of `-1`, `0`, `1`. Now the code become more readable.
 - Unify working with comparing results. There was almost the same copy-pasted code for `Greater`, `GreaterOrEqual`, `Less`, `LessOrEqual` functions. Now these call one function with all common logic. It could help to change behavior for all of them in future.

Linked issue https://github.com/stretchr/testify/issues/903 .

### DIFF
--- a/assert/assertion_order.go
+++ b/assert/assertion_order.go
@@ -5,6 +5,12 @@ import (
 	"reflect"
 )
 
+const(
+	greater = 1
+	equal = 0
+	less = -1
+)
+
 func compare(obj1, obj2 interface{}, kind reflect.Kind) (int, bool) {
 	switch kind {
 	case reflect.Int:
@@ -12,13 +18,13 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (int, bool) {
 			intobj1 := obj1.(int)
 			intobj2 := obj2.(int)
 			if intobj1 > intobj2 {
-				return -1, true
+				return greater, true
 			}
 			if intobj1 == intobj2 {
-				return 0, true
+				return equal, true
 			}
 			if intobj1 < intobj2 {
-				return 1, true
+				return less, true
 			}
 		}
 	case reflect.Int8:
@@ -26,13 +32,13 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (int, bool) {
 			int8obj1 := obj1.(int8)
 			int8obj2 := obj2.(int8)
 			if int8obj1 > int8obj2 {
-				return -1, true
+				return greater, true
 			}
 			if int8obj1 == int8obj2 {
-				return 0, true
+				return equal, true
 			}
 			if int8obj1 < int8obj2 {
-				return 1, true
+				return less, true
 			}
 		}
 	case reflect.Int16:
@@ -40,13 +46,13 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (int, bool) {
 			int16obj1 := obj1.(int16)
 			int16obj2 := obj2.(int16)
 			if int16obj1 > int16obj2 {
-				return -1, true
+				return greater, true
 			}
 			if int16obj1 == int16obj2 {
-				return 0, true
+				return equal, true
 			}
 			if int16obj1 < int16obj2 {
-				return 1, true
+				return less, true
 			}
 		}
 	case reflect.Int32:
@@ -54,13 +60,13 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (int, bool) {
 			int32obj1 := obj1.(int32)
 			int32obj2 := obj2.(int32)
 			if int32obj1 > int32obj2 {
-				return -1, true
+				return greater, true
 			}
 			if int32obj1 == int32obj2 {
-				return 0, true
+				return equal, true
 			}
 			if int32obj1 < int32obj2 {
-				return 1, true
+				return less, true
 			}
 		}
 	case reflect.Int64:
@@ -68,13 +74,13 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (int, bool) {
 			int64obj1 := obj1.(int64)
 			int64obj2 := obj2.(int64)
 			if int64obj1 > int64obj2 {
-				return -1, true
+				return greater, true
 			}
 			if int64obj1 == int64obj2 {
-				return 0, true
+				return equal, true
 			}
 			if int64obj1 < int64obj2 {
-				return 1, true
+				return less, true
 			}
 		}
 	case reflect.Uint:
@@ -82,13 +88,13 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (int, bool) {
 			uintobj1 := obj1.(uint)
 			uintobj2 := obj2.(uint)
 			if uintobj1 > uintobj2 {
-				return -1, true
+				return greater, true
 			}
 			if uintobj1 == uintobj2 {
-				return 0, true
+				return equal, true
 			}
 			if uintobj1 < uintobj2 {
-				return 1, true
+				return less, true
 			}
 		}
 	case reflect.Uint8:
@@ -96,13 +102,13 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (int, bool) {
 			uint8obj1 := obj1.(uint8)
 			uint8obj2 := obj2.(uint8)
 			if uint8obj1 > uint8obj2 {
-				return -1, true
+				return greater, true
 			}
 			if uint8obj1 == uint8obj2 {
-				return 0, true
+				return equal, true
 			}
 			if uint8obj1 < uint8obj2 {
-				return 1, true
+				return less, true
 			}
 		}
 	case reflect.Uint16:
@@ -110,13 +116,13 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (int, bool) {
 			uint16obj1 := obj1.(uint16)
 			uint16obj2 := obj2.(uint16)
 			if uint16obj1 > uint16obj2 {
-				return -1, true
+				return greater, true
 			}
 			if uint16obj1 == uint16obj2 {
-				return 0, true
+				return equal, true
 			}
 			if uint16obj1 < uint16obj2 {
-				return 1, true
+				return less, true
 			}
 		}
 	case reflect.Uint32:
@@ -124,13 +130,13 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (int, bool) {
 			uint32obj1 := obj1.(uint32)
 			uint32obj2 := obj2.(uint32)
 			if uint32obj1 > uint32obj2 {
-				return -1, true
+				return greater, true
 			}
 			if uint32obj1 == uint32obj2 {
-				return 0, true
+				return equal, true
 			}
 			if uint32obj1 < uint32obj2 {
-				return 1, true
+				return less, true
 			}
 		}
 	case reflect.Uint64:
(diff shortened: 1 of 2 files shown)


alexpantyukhin:
@georgelesica-wf Could you please review this PR?

assert/assertion_order.go
```diff
@@ -296,14 +250,25 @@ func LessOrEqual(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...inter
...
 	return true
 }
+
+
+func containsValue(values []int, value int) bool{
+    for _, v := range values {
+        if (v == value) {
+            return true
+        }
+	}
```
nineinchnick: Wrong indentation

assert/assertion_order.go
```diff
@@ -286,6 +235,11 @@ func Less(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...interface{})
 //    assert.LessOrEqual(t, "a", "b")
 //    assert.LessOrEqual(t, "b", "b")
 func LessOrEqual(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...interface{}) bool {
+	return compareTwoValues(t, e1, e2, []int {less, equal}, "\"%s\" is not less or equal than \"%s\"", msgAndArgs);
+}
+
+
+func compareTwoValues(t TestingT, e1 interface{}, e2 interface{}, allowedComparesResults []int, failMessage string, msgAndArgs ...interface{}) bool {
```
nineinchnick: How about using a struct with three boolean fields for each comparsion result, instead of a slice? I guess this needs a benchmark, but iterating over a slice of 1 to 2 elements seems like an overkill.

georgelesica-wf:
Please update the description so that it contains a high-level overview of your changes. You should include the problem you're attempting to solve, an overview of the solution, and how your changes might impact users. Also, if this PR is tied to any open issues, please link to them.

assert/assertion_order.go
alexpantyukhin: @nineinchnick . I thought about that. The compare function returns constants `greater`, `equal`, `less`. Do you think iterating over the slice takes much time there? If we have a struct we should anyway interpret what the value we got. It will make the code more difficult.
I guess now it's readable and simple. What do you think?

alexpantyukhin:
@georgelesica-wf Sorry to be late. I have added a detail description into the head of the PR. This is not linked with some issue, Just some thoughts and implementations of refactoring.

assert/assertion_order.go
alexpantyukhin: solver, thx!

assert/assertion_order.go
boyan-soubachov: I think it's a great idea to extract the common logic for comparing two values. Do you think however that it would make more sense if it only focused on the comparison and the calling functions did the 'failMessage' formatting?

alexpantyukhin: @boyan-soubachov Thank you for your review! Oh, I guess I see what you mean: extracting raising final `Fail`  into upper level methods. Good input! I also thought about this idea. But I choose  the current one because:
 - The current approach minimize duplication of code (in the suggested solution we would need to copy paste code for call `Fail`. It's not much, but anyway...) 
 - We already have other checking for values `Kinds` and that values are comparable before. And I think it's another one plus to put calling `Fail` with `failMessage` inside this method.

boyan-soubachov: Fair enough, that makes sense

assert/assertion_order.go
```diff
@@ -296,14 +249,24 @@ func LessOrEqual(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...inter
...
-	if res != 1 && res != 0 {
-		return Fail(t, fmt.Sprintf("\"%v\" is not less than or equal to \"%v\"", e1, e2), msgAndArgs...)
+	if !containsValue(allowedComparesResults, compareResult) {
+		return Fail(t, fmt.Sprintf(failMessage, e1, e2), msgAndArgs...)
 	}
 
 	return true
 }
+
+func containsValue(values []int, value int) bool {
```
boyan-soubachov: Could you please add unit tests for this function, even if it is very basic and simple in its definition. We would like to prevent future regressions and codify the exact specifications & intention of it.

assert/assertion_order.go
```diff
@@ -286,6 +235,10 @@ func Less(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...interface{})
 //    assert.LessOrEqual(t, "a", "b")
 //    assert.LessOrEqual(t, "b", "b")
 func LessOrEqual(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...interface{}) bool {
+	return compareTwoValues(t, e1, e2, []int{less, equal}, "\"%s\" is not less or equal than \"%s\"", msgAndArgs)
+}
+
+func compareTwoValues(t TestingT, e1 interface{}, e2 interface{}, allowedComparesResults []int, failMessage string, msgAndArgs ...interface{}) bool {
```
boyan-soubachov: Could you please add unit tests to this function? They don't seem to have been added and we would definitely want to lock-down the behaviour and interface to prevent regressions.

assert/assertion_order.go
```diff
@@ -5,193 +5,199 @@ import (
 	"reflect"
 )
 
+const (
+	greater = 1
```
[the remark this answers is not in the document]
alexpantyukhin: Good catch! fixed!)

assert/assertion_order.go
```diff
@@ -5,193 +5,199 @@ import (
...
+	equal   = 0
+	less    = -1
+)
+
 func compare(obj1, obj2 interface{}, kind reflect.Kind) (int, bool) {
```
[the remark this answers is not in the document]
alexpantyukhin: Yesm right, Fixed

assert/assertion_order.go
alexpantyukhin: Added.

assert/assertion_order.go
alexpantyukhin: @boyan-soubachov  Added.

assert/assertion_order_test.go
```diff
@@ -116,3 +116,84 @@ func TestLessOrEqual(t *testing.T) {
...
+		value  CompareType
+		result bool
+	}{
+		{values: []CompareType{compareGreater}, value: compareGreater, result: true},
+		{values: []CompareType{compareGreater, compareLess}, value: compareGreater, result: true},
+		{values: []CompareType{compareGreater, compareLess}, value: compareLess, result: true},
+		{values: []CompareType{compareGreater, compareLess}, value: compareEqual, result: false},
+	} {
+		compareResult := containsValue(currCase.values, currCase.value)
+		if compareResult != currCase.result {
```
boyan-soubachov: We can dogfood our own `assert.Equal` function here :)

assert/assertion_order_test.go
```diff
@@ -116,3 +116,84 @@ func TestLessOrEqual(t *testing.T) {
...
+		{v1: make([]int, 5, 5), v2: make([]int, 5, 5)},
+	} {
+		compareResult := compareTwoValues(mockT, currCase.v1, currCase.v2, []CompareType{compareLess, compareEqual, compareGreater}, "testFailMessage")
+		if compareResult {
+			t.Errorf("Values %s and %s should be not comparable", currCase.v1, currCase.v2)
+		}
+	}
+}
+
+func Test_compareTwoValuesCrrectCompareResult(t *testing.T) {
```
boyan-soubachov: Typo here ('Crrect' => 'Correct').

assert/assertion_order_test.go
```diff
@@ -116,3 +116,84 @@ func TestLessOrEqual(t *testing.T) {
...
+	}{
+		{v1: 1, v2: 2, compareTypes: []CompareType{compareLess}},
+		{v1: 1, v2: 2, compareTypes: []CompareType{compareLess, compareEqual}},
+		{v1: 2, v2: 2, compareTypes: []CompareType{compareGreater, compareEqual}},
+		{v1: 2, v2: 2, compareTypes: []CompareType{compareEqual}},
+		{v1: 2, v2: 1, compareTypes: []CompareType{compareEqual, compareGreater}},
+		{v1: 2, v2: 1, compareTypes: []CompareType{compareGreater}},
+	} {
+		compareResult := compareTwoValues(mockT, currCase.v1, currCase.v2, currCase.compareTypes, "testFailMessage")
+		if !compareResult {
```
boyan-soubachov: We could dogfood our own `assert.True` here.

assert/assertion_order_test.go
```diff
@@ -116,3 +116,84 @@ func TestLessOrEqual(t *testing.T) {
...
+	for _, currCase := range []struct {
+		v1 interface{}
+		v2 interface{}
+	}{
+		{v1: CompareStruct{}, v2: CompareStruct{}},
+		{v1: map[string]int{}, v2: map[string]int{}},
+		{v1: make([]int, 5, 5), v2: make([]int, 5, 5)},
+	} {
+		compareResult := compareTwoValues(mockT, currCase.v1, currCase.v2, []CompareType{compareLess, compareEqual, compareGreater}, "testFailMessage")
+		if compareResult {
```
boyan-soubachov: As below, we could use our own `assert.False()`

assert/assertion_order_test.go
alexpantyukhin: @boyan-soubachov  yes, it's tricky. Do you think is it OK to use our own `assert.Equal` ? which is potentially should be tested inside `testify`?

boyan-soubachov: Oh yes, we actively encourage it :)

2nd paragraph of https://github.com/stretchr/testify/blob/master/README.md#contributing

`Equal` is well tested and if we use it ourselves, we're more likely to find any issues with it _before_ any of our external users do. This means we can also fix it before it becomes a common problem for our external users.

alexpantyukhin: @boyan-soubachov Yes, I see.. Fixed :)

assert/assertion_order_test.go
alexpantyukhin: fixed :)

assert/assertion_order.go
```diff
@@ -229,26 +218,7 @@ func Greater(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...interface
...
-	if !isComparable {
-		return Fail(t, fmt.Sprintf("Can not compare type \"%s\"", reflect.TypeOf(e1)), msgAndArgs...)
-	}
-
-	if res != -1 && res != 0 {
-		return Fail(t, fmt.Sprintf("\"%v\" is not greater than or equal to \"%v\"", e1, e2), msgAndArgs...)
-	}
-
-	return true
+	return compareTwoValues(t, e1, e2, []CompareType{compareGreater, compareEqual}, "\"%s\" is not greater or equal than \"%s\"", msgAndArgs)
```
boyan-soubachov: Excuse me, one last thing I missed: The language here has changed and seems incorrect compared to what it used to be previously:
> is not greater than or equal to

Do you mind making sure it's consistent with what it used to be before?

assert/assertion_order.go
```diff
@@ -286,6 +237,10 @@ func Less(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...interface{})
 //    assert.LessOrEqual(t, "a", "b")
 //    assert.LessOrEqual(t, "b", "b")
 func LessOrEqual(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...interface{}) bool {
+	return compareTwoValues(t, e1, e2, []CompareType{compareLess, compareEqual}, "\"%s\" is not less or equal than \"%s\"", msgAndArgs)
```
boyan-soubachov: As above, this should be:
> is not less than or equal to

assert/assertion_order.go
alexpantyukhin: @boyan-soubachov  fixed)

assert/assertion_order.go
alexpantyukhin: fixed

boyan-soubachov:
@alexpantyukhin , please rebase the branch and resolve the conflicts before I can merge it :)

alexpantyukhin:
@boyan-soubachov thamk you very much for your review) It is rebased.

PR #753, opened 2019-04-19
codegen: add flag for additional imports

nomis52:
Go 1.5 introduced the 'source' compiler. See
https://github.com/golang/go/issues/11415 for details.

This avoids the problem of using out of date information when generating
code. It also means we don't have to have to run `go install` prior to
generating.

I also added the ability to specify additional imports since we use
codegen to generate files that depend on types outside testify.

### DIFF
--- a/_codegen/main.go
+++ b/_codegen/main.go
@@ -28,11 +28,12 @@ import (
 )
 
 var (
-	pkg       = flag.String("assert-path", "github.com/stretchr/testify/assert", "Path to the assert package")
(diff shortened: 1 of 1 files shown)


nomis52:
I can change this to use imported.ForCompiler() if you like, but that was only introduced in 1.12

dolmen:
@nomis52 This PR doesn't seem relevant anymore after rebase. Could you either confirm (and close) or explain how it is still useful?

dolmen:
The remaining changes are very light. Better close and reinvent the code when needed than keeping it withe expectation it will be useful one day.

PR #760, opened 2019-05-02
Hack a fix for code generation

georgelesica-wf:
I'm really not sure if this is a good idea or not, but it seems to fix the code generation issues we've seen it a few PRs.


boyan-soubachov:
@georgelesica-wf some of the changes in the PR are not relevant anymore due to us dropping Go 1.9 support. Would you like to rebase and I'd be happy to review?

PR #765, opened 2019-05-15
Add a DynamicReturn method to Mock

dgsb:
The purpose is to easily be able to return different values upon
consecutive call on the mock.


dgsb:
Hello, we can forget this pull request.
There are more clever way to implement this behaviour, like [vektra](github.com/vektra/mockery) does at the mock implementation level.

PR #481, opened 2017-07-20
Detect plurality for suffix in Len

mlsteele:
Change this:
`... should have 2 item(s), but has 3`

To this:
`... should have 1 item, but has 3`
`... should have 2 items, but has 3`

PR #768, opened 2019-05-20
Detect plurality for suffix in Len

mlsteele:
Change this:
`... should have 2 item(s), but has 3`

To this:
`... should have 1 item, but has 3`
`... should have 2 items, but has 3`

Previously killed by forkers https://github.com/stretchr/testify/pull/481

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -588,7 +588,11 @@ func Len(t TestingT, object interface{}, length int, msgAndArgs ...interface{}) 
 	}
 
 	if l != length {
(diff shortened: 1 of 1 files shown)


boyan-soubachov:
I don't think this issue provides much added value to the suite and would prefer to keep the code complexity as low as possible.

If the community feels strongly about this, I'd be happy to reconsider and review the PR in more detail but I find that using the brackets notation (i.e. "item(s)") is easily readable, from my perspective anyways.

Again, if the community really wants this, I'd be happy to review and look into it. In the mean time, I will close the issue.

PR #769, opened 2019-05-20
Fix docs and output for LessOrEqual and friends

mlsteele:
In documentation change this and other similar entries:
```
- LessOrEqual asserts that the first element in greater or equal than the second
+ LessOrEqual asserts that the first element is less than or equal to the second 
```

In test output for `LessOrEqual(t, 2, 1)` change this and similar:
```
-                Error:          "%!s(int=2)" is not less or equal than "%!s(int=1)"
+                Error:          "2" is not less than or equal to "1"
```


Jumanjii:
I was just beginning these modifications when I tell me "Wait, maybe this pr already exists" :-D

Issue #743, opened 2019-03-22
Assertion for matching YAML


rndstr:
Same as `JSONEq()`, a `YAMLEq()` would be useful to have.

PR #775, opened 2019-06-13
Add YAMLEq to assert that two YAML strings are equivalent

nmi:
Like JSONEq, I added YAMLEq to check whether YAML strings are equivalent.

Fixes https://github.com/stretchr/testify/issues/743


nmi:
@georgelesica-wf If you have time, could you review this PR?

georgelesica-wf:
This looks fine to me, I'm going to get another set of eyes to take a look and then I'll merge it. Thanks!

Issue #51, opened 2014-04-23
Not a bug,but a suggestion:mocking should be overrideable


AlexLuya:
I am using a BDD framework that supports BeforeEach(),so I want to set some default mocking,then override needed mocking in concrete method,but I found mocking same method with same parameters but different returns,only  first mock will work.So hoping mocking will be overrideable,and this will make test more concise for "big" method with many dependencies.

ernesto-jimenez:
@AlexLuya you just need to specify that each of those method calls is expected only once and then it'll work as @matryer described. See: https://github.com/stretchr/testify/blob/master/mock/mock_test.go#L590

If that's not what you need please reopen the issue or open a new one. thanks!

yangshun:
I think the example has since been changed to: https://github.com/stretchr/testify/blob/master/mock/mock_test.go#L651-L657

rockuw:
@ernesto-jimenez Specifying `Once()` may not be enough. See the following case:

```go
m := NewMockModule()
s := NewModule(m)
m.On("Foo").Return(nil).Once()
s.DoSth() // which may call m.Foo() multiple times, even variable times

m.On("Foo").Return(err) // expecting overrideing
s.DoSth()
```

So overriding is still desirable.

PR #783, opened 2019-07-12
Mock.Get returns existing expectation #51

temoto:
Hello. Huge thanks for testify, it removed a lot of waste time from my work.

I believe this patch mostly satisfies #51 for those who wanted to define default calls and maybe override them later.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -257,6 +257,32 @@ func (m *Mock) On(methodName string, arguments ...interface{}) *Call {
 	return c
 }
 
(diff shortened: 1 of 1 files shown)


azharulacquia (approved):

brackendawson:
Sorry, but this is definitely a no. Unset was added (quite poorly) to attempt to do the same thing and it has caused problems with our API.

The correct way to remove or change an expectation is to; a) not share a mock between multiple tests and b) not set expectations you don't expect. It is always possible to know before a test what calls are expected.

PR #785, opened 2019-07-19
Format code with goimports

muesli:
Fixed formatting and import order.

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -84,7 +84,7 @@ func Run(t *testing.T, suite TestingSuite) {
 	defer failOnPanic(t)
 
 	suiteSetupDone := false
-	
+
 	methodFinder := reflect.TypeOf(suite)
 	tests := []testing.InternalTest{}
 	for index := 0; index < methodFinder.NumMethod(); index++ {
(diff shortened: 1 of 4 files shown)


suite/suite_test.go
```diff
@@ -350,7 +350,7 @@ func TestRunSuite(t *testing.T) {
 type SuiteSetupSkipTester struct {
 	Suite
 
-	setUp bool
+	setUp    bool
```
xorcare: these changes are not related to imports

suite/suite.go
```diff
@@ -84,7 +84,7 @@ func Run(t *testing.T, suite TestingSuite) {
 	defer failOnPanic(t)
 
 	suiteSetupDone := false
-	
+
```
xorcare: these changes are not related to imports

suite/suite_test.go
muesli: goimports does both import ordering as well as calling gofmt.

bencampbell-wf: Hmm. This change would lead me to believe the go formatting check (defined during CI)[https://github.com/stretchr/testify/blob/master/.travis.gofmt.sh#L5) is not being run.

FYI: @georgelesica-wf

georgelesica-wf: Agreed, weird...

georgelesica-wf: Fixed, hopefully... https://github.com/stretchr/testify/pull/804

PR #788, opened 2019-07-19
Handle ineffectual assignments

muesli:
Removed ineffectual assignments.


mock/mock_test.go
```diff
@@ -1271,7 +1271,7 @@ func Test_Arguments_Diff_WithArgMatcher(t *testing.T) {
 	assert.Equal(t, 1, count)
 	assert.Contains(t, diff, `(bool=false) not matched by func(int) bool`)
 
-	diff, count = args.Diff([]interface{}{"string", 123, false})
+	diff, _ = args.Diff([]interface{}{"string", 123, false})
```
boyan-soubachov: Shouldn't we check the value of `count` instead of ignoring it?

muesli: You're right, will fix!

muesli:
Added the missing count assertion. Good for another review!

PR #792, opened 2019-07-24
Add NotSame

### LINKED ISSUE #784: add "NotSame"
We need a negative for the `Same` assertion

boyan-soubachov:
* Added NotSame test for the assert package
* Added NotSame test for the require package
* Included formatted variants of NotSame for both assert and require

Closes #784

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ Contributing
 
 Please feel free to submit issues, fork the repository and send pull requests!
 
-When submitting an issue, we ask that you please include a complete test function that demonstrates the issue.  Extra credit for those using Testify to write the test code that demonstrates it.
+When submitting an issue, we ask that you please include a complete test function that demonstrates the issue. Extra credit for those using Testify to write the test code that demonstrates it.
 
 Code generation is used. Look for `CODE GENERATED AUTOMATICALLY` at the top of some files. Run `go generate ./...` to update generated files.
 
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -526,19 +526,6 @@ func Samef(t TestingT, expected interface{}, actual interface{}, msg string, arg
 	return Same(t, expected, actual, append([]interface{}{msg}, args...)...)
 }
 
-// NotSamef asserts that two pointers do not reference the same object.
-//
-//    assert.NotSamef(t, ptr1, ptr2, "error message %s", "formatted")
-//
-// Both arguments must be pointer variables. Pointer variable sameness is
-// determined based on the equality of both type and value.
-func NotSamef(t TestingT, expected interface{}, actual interface{}, msg string, args ...interface{}) bool {
-	if h, ok := t.(tHelper); ok {
-		h.Helper()
-	}
-	return NotSame(t, expected, actual, append([]interface{}{msg}, args...)...)
-}
-
 // Subsetf asserts that the specified list(array, slice...) contains all
 // elements given in the specified subset(array, slice...).
 //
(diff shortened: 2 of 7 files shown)


boyan-soubachov:
Comments and feedback greatly appreciated :pray:

boyan-soubachov:
Excuse me, I realised I missed the code-generation part. Will redo the proper way.

ceven:
Hi Boyan, what do you think of refactoring your code to better re-use existing implementation of `Same`? You might have to slightly refactor the `Same` function for that purpose.

boyan-soubachov:
Hi @ceven , I think you've got a very good point there. I initially looked at that approach and came to the conclusion that the code could become complex to follow. Having had another look at it again though, I think you're right and we probably can pull some common logic out into function/s. 

I am a bit hesitant to have a toggle flag in the (common) function specify whether we're looking for same or not same since it seems to violate the single-responsibility principle. Thoughts?

I think a possible way to go about it is to have a common 'same' check function that covers the type and values of the pointers while returning only a `boolean`. We can then return a `Fail()` depending on the result and calling function (`Same()`/`NotSame()`). Thoughts?

ceven:
2nd option sounds good. It's also the way other methods have been written in `testify` (e.g. [Empty](https://github.com/stretchr/testify/blob/af88917d7dc5a6a253e7809e41bf046b25090a97/assert/assertions.go#L530-L594)) so doing it that way would keep consistency across the library.

boyan-soubachov:
@ceven , refactored. Please note that we do lose some information in this case as we don't throw a specific message if the types are different when doing a `Same()` test but it's still very clear given that the fail message shows both types and values.

Feedback would be much appreciated as always :smiley: :pray:

ceven:
@boyan-soubachov code looks good, but is there any reason why you didn't extract more of the common logic into the common comparator? AFAIS everything above (pointer comparison...) is duplicated.
Again referring to the implementation of Empty: https://github.com/stretchr/testify/blob/af88917d7dc5a6a253e7809e41bf046b25090a97/assert/assertions.go#L530-L594 
No big deal but interested to hear your thoughts

boyan-soubachov:
@ceven , it's a good point you make. I had considered it and my logic for leaving them separate is purely one of readability & simplicity for the `sameComparator` function. Idea being that input validation is handled by the functions calling `sameComparator` whereas the actual comparison (and comparison only) is handled by the `sameComparator`.

That way `sameComparator` only compares (it would be valid for anything, not just pointers).

Happy to change/incorporate the pointer checking logic if you think it makes more sense or is definitely worth it :)

ceven:
You should definitely write reusable code, but also refactor where some is duplicated. In this case I'd argue that `sameComparator` could include more common logic, and yes you may have to rename it if it does to keep the name aligned with the logic it performs. 
It's minor anyways, I'll leave the decision up to you.

boyan-soubachov:
Fair enough, will do the full refactor :)

PureWhiteWu:
http://www.stretchr.com/ is for sale now, so I think this repo is no longer maintained.
I've created a new organization testify-go, and forked this repo.
https://github.com/testify-go/testify
I will be maintaining the fork of this repo.
Would you please create the pr again in the new repo?

RmbRT:
This PR is broken, it removed Assertions.NotSame(), even though it was initially included.

boyan-soubachov:
@RmbRT, I actually manually added the generated code without noticing. I then removed it and in the latest PR I’ve rerun ‘go generate’ to do it properly as per the readme

RmbRT:
ah okay

PR #793, opened 2019-07-26
Fixed didPanic to now detect panic(nil).

### LINKED ISSUE #794: didPanic does not detect panic(nil)
This is because you use `recover() != nil` to detect whether a panic  occurred.
I have opened PR  #793 to fix this.

RmbRT:
Previously, the function would not detect panic(nil) calls.
In didPanic, removed the anonymous function call, instead,
added named return values. Added extra test cases for the
panic(nil) call.

Closes #794.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1004,27 +1004,21 @@ func Condition(t TestingT, comp Comparison, msgAndArgs ...interface{}) bool {
 type PanicTestFunc func()
 
 // didPanic returns true if the function passed to it panics. Otherwise, it returns false.
-func didPanic(f PanicTestFunc) (bool, interface{}, string) {
+func didPanic(f PanicTestFunc) (didPanic bool, message interface{}, stack string) {
+	didPanic = true
 
-	didPanic := false
-	var message interface{}
-	var stack string
-	func() {
-
-		defer func() {
-			if message = recover(); message != nil {
-				didPanic = true
-				stack = string(debug.Stack())
-			}
-		}()
-
-		// call the target function
-		f()
-
+	defer func() {
+		message = recover()
+		if didPanic {
+			stack = string(debug.Stack())
+		}
 	}()
 
-	return didPanic, message, stack
+	// call the target function
+	f()
+	didPanic = false
 
+	return
 }
 
 // Panics asserts that the code inside the specified PanicTestFunc panics.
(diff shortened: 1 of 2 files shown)


MariusVanDerWijden:
Nice!

MariusVanDerWijden (dismissed):

sebastianst:
ACK good catch

sebastianst (dismissed):

ggwpez:
Hes my brother :100:

assert/assertions.go
```diff
@@ -901,25 +901,15 @@ func Condition(t TestingT, comp Comparison, msgAndArgs ...interface{}) bool {
...
-
-	}()
-
-	return didPanic, message
+	defer func() { message = recover() }()
+	// call the target function
+	f()
+	didPanic = false
 
+	return
```
boyan-soubachov, line 912: Is this line not redundant/pointless now?

assert/assertions_test.go
```diff
@@ -730,10 +730,18 @@ func TestCondition(t *testing.T) {
 
 func TestDidPanic(t *testing.T) {
 
-	if funcDidPanic, _ := didPanic(func() {
-		panic("Panic!")
-	}); !funcDidPanic {
-		t.Error("didPanic should return true")
+	const panicMsg = "Panic!"
+
+	if funcDidPanic, msg := didPanic(func() {
```
boyan-soubachov, line 735: Would it be a bit clearer if we separated these `if` statements into two lines, assignment and `if` check?

e.g.
```
funcDidPanic, msg := didPanic(func() {...})
if !funcDidPanic || msg != panicMsg {
    ...
}
```

Antonboom:
Any updates? :)

Antonboom:
Any updates? :)

boyan-soubachov (dismissed):
Thanks for your contribution :)

boyan-soubachov:
This PR needs to be updated and conflicts resolved before I can merge it.

RmbRT:
Rebased, resolved conflicts.

assert/assertions.go
RmbRT: I think it's required when returning named return values.

assert/assertions_test.go
RmbRT: I think it's fine as it is. Reads as naturally as any other go code to me.

PR #796, opened 2019-07-29
panic: add .Panic() to mock.Call

sarathsp06:
Panic mocks a function call that panics with the specified message

Usage:
   
       c := mockedService.
		On("TheExampleMethod", "A", "B", true).
		Panic("panic message for example method")

The above mocked service would panic with message `panic message for example method`


sarathsp06:
Can someone reviw this

ben-bourdin451:
This PR fixes #797 

(mentioning the issue here to link the PR to it)

PR #799, opened 2019-08-02
Fixed suite TearDown ordering 

### LINKED ISSUE #466: bug in parallel tests

devdinu:
Using WaitGroup to synchronise between tests so TearDown runs at end.

Having `suite.T().Parallel()` in `TestA` causes it to hang. goroutine wait in `<-t.signal` at [testing.go](https://github.com/golang/go/blob/master/src/testing/testing.go#L961) and suite.go Run to wait on `testsSync` `:108` 

Closes #466


devdinu:
@georgelesica-wf could we merge this.

georgelesica-wf:
CI is broken right now and I haven't had time to fix it. It's not running our checks, just the tests. If someone can figure out how to get the code gen to run and can finish making [this PR](https://github.com/stretchr/testify/pull/804) pass then we can start merging again. Until then I'd rather hold off.

devdinu:
@boyan-soubachov Could you help reviewing this and merge.

suite/suite_order_test.go
```diff
@@ -0,0 +1,47 @@
+package suite
```
boyan-soubachov: What is your logic for putting this in a separate file? Does it not make sense to just have this as another test in `suite_test.go`?

suite/suite.go
```diff
@@ -111,6 +114,9 @@ func Run(t *testing.T, suite TestingSuite) {
 		test := testing.InternalTest{
 			Name: method.Name,
 			F: func(t *testing.T) {
+				defer func() {
```
boyan-soubachov: Can we not just `defer testSync.Done()` instead of wrapping it in an anonymous function?

devdinu: Had some logic which i've removed later. good catch, have updated.

suite/suite_order_test.go
devdinu: The other file becoming huge, thought could separate the tests based on what we integration vs small func tests.

boyan-soubachov: I see. I do agree that it's becoming a bit large but I don't think navigating it is still a huge problem. 

I'd be happy to discuss any ideas you might have about improving its readability & maintainability, whether it involves splitting it up or some other way.

PR #804, opened 2019-08-15
Fix CI script

georgelesica-wf:
The script isn't running properly in Travis. This appears to be because the `script` block was indented incorrectly.


georgelesica-wf:
Yep, there we go, now it's running the scripts again.

boyan-soubachov:
I think this isn't an issue anymore :) closing

PR #807, opened 2019-08-28
Use Repeatability as tie-breaker for closest match

uberbo:


mock/mock.go
```diff
@@ -281,15 +281,17 @@ func (m *Mock) findExpectedCall(method string, arguments ...interface{}) (int, *
 
 func (m *Mock) findClosestCall(method string, arguments ...interface{}) (*Call, string) {
 	var diffCount int
+	var repeatability int
 	var closestCall *Call
 	var err string
 
 	for _, call := range m.expectedCalls() {
 		if call.Method == method {
```
boyan-soubachov: Would it make the code clearer if we refactored this `if` clause into a guard clause? i.e. 
```
if call.Method != method {
    continue
}
...
```

mock/mock.go
```diff
@@ -281,15 +281,17 @@ func (m *Mock) findExpectedCall(method string, arguments ...interface{}) (int, *
...
 
 			errInfo, tempDiffCount := call.Arguments.Diff(arguments)
-			if tempDiffCount < diffCount || diffCount == 0 {
+			if tempDiffCount < diffCount || diffCount == 0 || (tempDiffCount == diffCount && call.Repeatability > repeatability) {
```
boyan-soubachov: Would it make sense to extract this logic into a separate function? It seems to be getting relatively complex and we definitely need to add unit tests for this new logic to make sure there are no regressions regarding previous behaviour.

uberbo: I was going for a minimal change. I have now added a more extensive rewrite of the method along with more testing.

mock/mock.go
uberbo: I think it is clearer without the continue, especially now that the inner content contains less logic, but let me know if you still prefer the continue.

boyan-soubachov: Thank you. We prefer a simpler approach to the code, even if it means adding a few more lines sometimes 🙂

mock/mock_test.go
```diff
@@ -1497,9 +1533,33 @@ func TestClosestCallMismatchedArgumentValueInformation(t *testing.T) {
...
+}
+
+func Test_isBetterMatchThanReturnsTrueIfRepeatabilityIsGreaterThanOther(t *testing.T) {
+	assert.True(t, matchCandidate{call: &Call{Repeatability: 1}, diffCount: 1}.isBetterMatchThan(matchCandidate{call: &Call{Repeatability: -1}, diffCount: 1}))
+}
+
+func Test_isBetterMatchThanReturnsFalseIfRepeatabilityIsLessThanOrEqualToOther(t *testing.T) {
+	assert.False(t, matchCandidate{call: &Call{Repeatability: 1}, diffCount: 1}.isBetterMatchThan(matchCandidate{call: &Call{Repeatability: 1}, diffCount: 1}))
+}
+
```
boyan-soubachov: One last thing, IMO there's no need to have individual test functions for each case. The convention we use is one `Test_*` function per function being tested with the individual cases called inside it :)

PR #808, opened 2019-08-29
Fix panic for Eventually functions

### LINKED ISSUE #805: assert.Eventually with low ticks cause panic: send on closed channel
# Setup
If you put an:
`assert.Eventually(t, func() bool { return true }, time.Millisecond, 50 * time.Microsecond)` inside your test, you will get flaky results and more often then not, you will get a panic:
`panic: send on closed channel`

# Workaround
Increasing the tick to around `100 * time.Microsecond` removes the flakyness

# Solution
Add a waitgroup potentially so that [this channel close](https://github.com/stretchr/testify/blob/221dbe5ed46703ee255b1da0dec05086f5035f62/assert/assertions.go#L1483) waits for the gofunc writing to the channel completes.

### LINKED ISSUE #835: race condition in Eventually
I've used a tightly timed 'Eventually' in my test:
```assert.Eventually(t, func() bool { ... }, time.Millisecond*100, time.Millisecond)```
ran it with `-race -count=1000`  and i got:
```
==================
WARNING: DATA RACE
Write at 0x00c0000ba5b0 by goroutine 34:
  runtime.closechan()
      /.../go/go1.12.9/src/runtime/chan.go:334 +0x0
  github.com/stretchr/testify/assert.Eventually()
      /.../go/go1.12.9/pkg/mod/github.com/stretchr/testify@v1.4.0/assert/assertions.go:1490 +0x29e
  github.com/stretchr/testify/require.Eventually()
      /.../go/go1.12.9/pkg/mod/github.com/stretchr/testify@v1.4.0/require/require.go:278 +0xaf
  command-line-arguments_test.TestPeriodicalTriggerStopWorksWhenContextIsCancelled.func1()
    ...

Previous read at 0x00c0000ba5b0 by goroutine 62:
  runtime.chansend()
      /.../go/go1.12.9/src/runtime/chan.go:142 +0x0
  github.com/stretchr/testify/assert.Eventually.func1()
      /.../go/go1.12.9/pkg/mod/github.com/stretchr/testify@v1.4.0/assert/assertions.go:1494 +0x54
```

### PR DESCRIPTION
**Pull request by jszwec**
Fixes #805
Fixes #835 

/cc @gmiejski @georgelesica-wf

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1475,24 +1475,26 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 		h.Helper()
 	}
 
+	ch := make(chan bool, 1)
+
 	timer := time.NewTimer(waitFor)
-	ticker := time.NewTicker(tick)
-	checkPassed := make(chan bool)
 	defer timer.Stop()
+
+	ticker := time.NewTicker(tick)
 	defer ticker.Stop()
-	defer close(checkPassed)
-	for {
+
+	for tick := ticker.C; ; {
 		select {
 		case <-timer.C:
-			return Fail(t, "Condition never satisfied", msgAndArgs...)
-		case result := <-checkPassed:
-			if result {
+			return false
+		case <-tick:
+			tick = nil
+			go func() { ch <- condition() }()
+		case v := <-ch:
+			if v {
 				return true
 			}
-		case <-ticker.C:
-			go func() {
-				checkPassed <- condition()
-			}()
+			tick = ticker.C
 		}
 	}
 }
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -1921,3 +1921,12 @@ func TestEventuallyTrue(t *testing.T) {
 
 	True(t, Eventually(t, condition, 100*time.Millisecond, 20*time.Millisecond))
 }
+
+func TestEventuallyIssue805(t *testing.T) {
+	mockT := new(testing.T)
+
+	NotPanics(t, func() {
+		condition := func() bool { <-time.After(time.Millisecond); return true }
+		False(t, Eventually(mockT, condition, time.Millisecond, time.Microsecond))
+	})
+}

### CODE REVIEW

**Comment by jszwec** [comment]
I believe we should also consider limiting the number of `condition` function calls to one at the time. It is possible to have a few of these running in parallel. The caller may not be aware that under certain circumstances the condition function should be thread safe or idempotent. However, arguably this would break the current description: `periodically checking target function each tick`.

Let me know what you think.

**Comment by bencampbell-wf** [comment]
I am wondering if it is worth re-implementing this using a pipeline pattern as described on the [Go Blog](https://blog.golang.org/pipelines). I figure it gives us the concurrency safety we are missing in the master implementation as well as defines some components for other concurrent matchers that may come along.

I spiked an example: https://play.golang.org/p/qJcXUfgnmgA

**Comment by georgelesica-wf** [comment]
Can we consider the above suggestion?

**Comment by jszwec** [comment]
I am not against using pipeline pattern, because I use it a lot in my daily work. However, in my personal opinion, we should settle for something smaller and tailored specifically for the job.

We could achieve the same by just disabling the ticker channel like this: https://play.golang.org/p/GhfafyX9epo

**Comment by j0z3f** [comment]
I think there might be a bug in this PR. 
Isn't it technically possible (although practically very unlikely) that the `select` statement in
```
select {
case checkPassed <- condition():
default:
}
```
jumps to the `default:` branch because the `result` from previous invocation of `condition()` has not been read from `checkPassed` yet. In this way, the code could miss a `true` return value and fail.

(I came across this PR because I ran into the issue https://github.com/stretchr/testify/issues/805)

jszwec:
It's true, but without it you risk go routine leaks, these are tests, but still...

It seems that we did not decide what the final approach should be yet. Like I mentioned in previous comments, in my opinion we should not be calling `condition` multiple times in parallel. In this case I think the problem that you mentioned wouldn't exist. I wrote an example [above](https://play.golang.org/p/GhfafyX9epo)

zut:
Since you're no longer closing the `checkPassed` channel, I think this would also resolve https://github.com/stretchr/testify/issues/835.

jszwec:
@georgelesica-wf I updated the code with the approach that I proposed above. Tests are failing on Go tip, it's a go mod related failure unrelated to this PR.

PR #654, opened 2018-08-28
Fix race in Diff printer

gz-c:
Fixes #625 

Race conditions were triggered when Sprintf-ing certain objects when preparing a diff for mock failures.

This restores the order of sprintf(obj) calls to the order it was prior to the race condition appearing.

Incorporates a new test case from @bhcleek

PR #812, opened 2019-09-03
mock: refactor Diff to reduce risk of triggering race conditions

sjames-iberis:
This is an alternative to https://github.com/stretchr/testify/pull/654
and seeks to address https://github.com/stretchr/testify/issues/625, avoiding triggering race conditions when stringifying arguments.

The Diff method is refactored to decouple the generation of argument pass/fail reports from the argument checks, so that argument stringification only takes place after all arguments have been checked and only if arguments expectations are not met. 

It remains the case that if an argument expectation is not met, the subsequent argument stringification could trigger a race condition which might mask the original argument mismatch.

The changes incorporates the unit test added by  https://github.com/stretchr/testify/pull/654.
Without the changes to the Diff method, I found that race conditions were consistently reported by that test when running:
```
go test -race -count=10 github.com/stretchr/testify/mock
```

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -667,7 +667,6 @@ func (args Arguments) Is(objects ...interface{}) bool {
 func (args Arguments) Diff(objects []interface{}) (string, int) {
 	//TODO: could return string as error and nil for No difference
 
-	var output = "\n"
 	var differences int
 
 	var maxArgCount = len(args)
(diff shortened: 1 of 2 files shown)


perryao:
The changes in this PR seem to fix a race condition I've been trying to troubleshoot this weekend. For my particular issue, I naively moved this `mutex.Lock` https://github.com/stretchr/testify/blob/master/mock/mock.go#L346 after the call to `findExpectedCall`. That also helped, but from the looks of things, the changes here are likely closer to a real fix. 

Could this get reviewed soon?

perryao:
#654 fixed my particular issue as well. I see that one is in the next milestone so I'll use that for now.

PR #825, opened 2019-10-03
#562: assert.NoError() and assert.Error() now behave correctly when a nil error struct is supplied

### LINKED ISSUE #562: assert NoError behaves unexpectedly with nil struct
We have some internal functions that return a struct dedicated to wrapping customer-facing errors, and thus have more to it than the standard `error` interface. To enforce that we use these structs everywhere, as we want to, we're returning our custom `*errors.Error` type instead of `error`.

However, when passing the returned value to `assert.NoError(t, err)` it will **always** indicate an error, even when the value is `nil`. Our type implements `Error()`, so it can be passed places that `error` is expected, so testify's behavior here feels like a bug.

```go
func TestSomething(t *testing.T) {
  require.NoError(t, succeed())  // Received unexpected error: <nil>
}

type CustomError struct {}
func (e *CustomError) Error() string { return "boom" }

func succeed() *CustomError {
  return nil
}
```

This looks like an edge-case in the way that `nil` values are being detected, and the workaround for now is to use `require.Nil(t, err)` but I have been using `NoError()` for a long time elsewhere and will probably forget at some point. :stuck_out_tongue:

e-nikolov:
Fixes #562 

assert.Nil already does this correctly by checking via reflection if the supplied value isn't nil wrapped in an interface in order to not give false results. I think assert.NoError should do the same.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1197,7 +1197,7 @@ func NoError(t TestingT, err error, msgAndArgs ...interface{}) bool {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
-	if err != nil {
+	if !isNil(err) {
 		return Fail(t, fmt.Sprintf("Received unexpected error:\n%+v", err), msgAndArgs...)
 	}
 
@@ -1215,7 +1215,7 @@ func Error(t TestingT, err error, msgAndArgs ...interface{}) bool {
 		h.Helper()
 	}
 
-	if err == nil {
+	if isNil(err)  {
 		return Fail(t, "An error is expected but got nil.", msgAndArgs...)
 	}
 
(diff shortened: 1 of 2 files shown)


e-nikolov:
I can imagine that this could be considered as breaking backwards compatibility, so perhaps there could be new functions to do this instead of modifying the behaviour of the existing ones. But I think it's quite useful to be able to use something like assert.NoError(err) on custom error types.

boyan-soubachov:
Could you please add a unit test covering the use-case you're solving? If I understand correctly, this change only fixes for the following scenario:
```
type MyCustomError {
    err error
}

a := MyCustomError{err=nil}
```
?

ronaknnathani:
Any plans to merge this change?

Antonboom:
Looks like bad change leading to implicit bugs.
@dolmen I propose to close this PR

dolmen:
#562 is invalid. This patch will break correct behaviour instead of fixing.

Thanks @Antonboom for review.

PR #830, opened 2019-10-10
Add FileNotExists and DirNotExists assertions

JustSkiv:
Implement FileNotExists and DirNotExists assertions to check whether file/dir not exist in a given path.

This functions is the opposite of FileExists and DirExists. 

It is very useful in some cases. For example, if your tests creates some files while working, it would be nice to check if this files exist, before test. Or test can remove some files and we have to make sure that files are really removed.


JustSkiv:
@ernesto-jimenez can you take a look, please?

PR #838, opened 2019-11-09
assert.Eventually: fail the testcase if condition is not met

### LINKED ISSUE #837: assert.Eventually() does not fail test when the timer expires
assert.Eventually() returns false (and not Fail(...)) when the timer expires.

As a result, the test does not fail.

So for example, including  `assert.Eventually(t, func()bool{return false}, time.Second, time.Millisecond))` in your test does not cause it to fail.  It will poll the function for a second, and then pass.

Is this intentional?  It seems surprising, since the other assert functions cause the test to fail when the condition is not met.

dmacvicar:
Fixes #837

See https://github.com/stretchr/testify/issues/837#issuecomment-552088888 for an explanation.

Given that the original tests did not catch the problem, I am not sure how to enhance the tests to do it. But feedback is appreciated.

I am also not completely sure if I am missing something regarding the motivation for https://github.com/stretchr/testify/commit/f1bd0923b8320ddfb0a8a9d919b5aeb8b7a4cc40 change from `Fail`to `false`, or if it was just an oversight.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1514,7 +1514,7 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 	for tick := ticker.C; ; {
 		select {
 		case <-timer.C:
-			return false
+			return Fail(t, "Condition never satisfied", msgAndArgs...)
 		case <-tick:
 			tick = nil
 			go func() { ch <- condition() }()


boyan-soubachov:
@dmacvicar , please rebase onto master to make the tests pass on `tip`. This change makes the #805 unit test pass so I see no reason not to merge it. Would be happy to approve and merge once you rebase.

torkelrogstad:
This was superseded by #869 and should be closed

PR #839, opened 2019-11-10
Add PanicsWithError

### LINKED ISSUE #749: PanicsWithValue() should use equality, not ==
Panic with Error instead of string is a valid practice. Asserting the value of a panic using `!=` is not ok, because it results in a funny, but annoying test fails.
 
Example of a fail
```
=== RUN   Test
--- FAIL: Test (0.00s)
    test.go:27: 
        	Error Trace:	test.go:27
        	Error:      	func (assert.PanicTestFunc)(0x128adb0) should panic with value:	&errors.errorString{s:"message"}
        	            		Panic value:	&errors.errorString{s:"message"}
        	Test:       	Test
FAIL
```

The causing code (assertions.go:952):
```
if panicValue != expected {
    return Fail(t, fmt.Sprintf("func %#v should panic with value:\t%#v\n\tPanic value:\t%#v", f, expected, panicValue), msgAndArgs...)
}
```

leighmcculloch:
# What
Add `PanicsWithError` that behaves like `PanicsWithValue` but requires
the value to be an `error` and compares an error string against the
error in the same fashion as `EqualError`.

# Why
`PanicsWithValue` is very useful with the value that will panic is a
basic type that can be easily defined in a test, but is less useful for
validating an error because error values may not be easily recreated in
a test to exactly match the error that will be generated in the moment.
This can be because an error will contain fields that aren't exported
and the test may not be able to define the exact error. Most of the time
testing error objects just testing the string is sufficient. This
concept is already supported by this module in the existing `EqualError`
function.

Fixes #749


assert/assertions.go
```diff
@@ -985,6 +985,28 @@ func PanicsWithValue(t TestingT, expected interface{}, f PanicTestFunc, msgAndAr
...
+		h.Helper()
+	}
+
+	funcDidPanic, panicValue, panickedStack := didPanic(f)
+	if !funcDidPanic {
+		return Fail(t, fmt.Sprintf("func %#v should panic\n\tPanic value:\t%#v", f, panicValue), msgAndArgs...)
+	}
+	panicErr, ok := panicValue.(error)
+	if !ok || !Error(t, panicErr) || !EqualError(t, panicErr, errString) {
+		return Fail(t, fmt.Sprintf("func %#v should panic with error message:\t%#v\n\tPanic value:\t%#v\n\tPanic stack:\t%s", f, errString, panicValue, panickedStack), msgAndArgs...)
```
leighmcculloch: I'm failing a second time here with a more specific error message than that failed with in `Error` or `EqualError`. I'm not sure if there's a better way of doing this. I don't think I can fail twice, and I should probably just duplicate the logic from the other two functions here and return this more specific error. What do you think @georgelesica-wf?

leighmcculloch: Right now it results in two like this:
```
    **********_test.go:20: 
                Error Trace:
                Error:          Error message not equal:
                                expected: "myerror"
                                actual  : "myer ror"
                Test:           TestTest
    **********_test.go:20: 
                Error Trace:
                Error:          func (assert.PanicTestFunc)(0x70b760) should panic with error message:  "myerror"
                                        Panic value:    &errors.errorString{s:"myer ror"}
                                        Panic stack:    goroutine 20 [running]:
...
```

leighmcculloch:
The CI tests are failing only on `tip` and for reasons unrelated to this change.

assert/assertions.go
leighmcculloch: I removed the double error in e442a16.

leighmcculloch:
👋 @glesica This PR is green 💚 now and has all the code generated, which is verified by CI now that generation is fixed. Would love to use this if you're up for accepting it.

PR #844, opened 2019-11-13
Print more details in ElementsMatch

martin-sucha:
It is not very helpful to print that the lengths differ when an
assertion fails, since that does not reveal what the cause of the issue
might be.

Let's print which elements are extra in each list, that should convey
the relevant information to the user. Also use spew to format the
objects, similar to what Equal does, to make the output more readable.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -881,26 +881,38 @@ func ElementsMatch(t TestingT, listA, listB interface{}, msgAndArgs ...interface
 		return true
 	}
 
-	aKind := reflect.TypeOf(listA).Kind()
-	bKind := reflect.TypeOf(listB).Kind()
-
-	if aKind != reflect.Array && aKind != reflect.Slice {
-		return Fail(t, fmt.Sprintf("%q has an unsupported type %s", listA, aKind), msgAndArgs...)
+	if !isList(t, listA, msgAndArgs...) || !isList(t, listB, msgAndArgs...) {
+		return false
 	}
 
-	if bKind != reflect.Array && bKind != reflect.Slice {
-		return Fail(t, fmt.Sprintf("%q has an unsupported type %s", listB, bKind), msgAndArgs...)
+	extraA, extraB := diffLists(listA, listB)
+
+	if len(extraA) == 0 && len(extraB) == 0 {
+		return true
 	}
 
+	return Fail(t, formatListDiff(listA, listB, extraA, extraB), msgAndArgs...)
+}
+
+// isList checks that the provided value is array or slice.
+func isList(t TestingT, list interface{}, msgAndArgs ...interface{}) (ok bool) {
+	kind := reflect.TypeOf(list).Kind()
+	if kind != reflect.Array && kind != reflect.Slice {
+		return Fail(t, fmt.Sprintf("%q has an unsupported type %s, expecting array or slice", list, kind),
+			msgAndArgs...)
+	}
+	return true
+}
+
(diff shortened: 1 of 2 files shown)


assert/assertions.go
```diff
@@ -898,9 +898,7 @@ func ElementsMatch(t TestingT, listA, listB interface{}, msgAndArgs ...interface
 	aLen := aValue.Len()
 	bLen := bValue.Len()
 
-	if aLen != bLen {
-		return Fail(t, fmt.Sprintf("lengths don't match: %d != %d", aLen, bLen), msgAndArgs...)
-	}
+	var extraA, extraB []interface{}
```
boyan-soubachov: The function `ElementsMatch` looks a bit long and unwieldy. Would it make sense to refactor it into a few smaller, unit-tested functions?

martin-sucha: I have split `ElementsMatch` to smaller functions and added a test for the main one.

assert/assertions.go
```diff
@@ -918,11 +932,37 @@ func ElementsMatch(t TestingT, listA, listB interface{}, msgAndArgs ...interface
...
 		}
 		if !found {
-			return Fail(t, fmt.Sprintf("element %s appears more times in %s than in %s", element, aValue, bValue), msgAndArgs...)
+			extraA = append(extraA, element)
 		}
 	}
 
-	return true
+	for j := 0; j < bLen; j++ {
+		if !visited[j] {
```
boyan-soubachov: Would it be neater if we changed this to a guard clause? e.g.
```
for .... {
    if visited[j] {
        continue
    }
    extraB = append(...)
}
```

assert/assertions.go
```diff
@@ -881,27 +881,41 @@ func ElementsMatch(t TestingT, listA, listB interface{}, msgAndArgs ...interface
...
 	}
 
-	aKind := reflect.TypeOf(listA).Kind()
-	bKind := reflect.TypeOf(listB).Kind()
+	aOk := checkList(t, listA, msgAndArgs...)
+	bOk := checkList(t, listB, msgAndArgs...)
 
-	if aKind != reflect.Array && aKind != reflect.Slice {
-		return Fail(t, fmt.Sprintf("%q has an unsupported type %s", listA, aKind), msgAndArgs...)
+	if !aOk || !bOk {
```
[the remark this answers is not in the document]
martin-sucha: This check does not test whether an element is contained, but that listA and listB are of type list. I have renamed `checkList` to `isList` (and inlined the function call into the if condition).

assert/assertions.go
martin-sucha: Updated.

assert/assertions.go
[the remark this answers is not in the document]
boyan-soubachov: Good point, I must've confused them.

assert/assertions_test.go
```diff
@@ -788,6 +788,27 @@ func TestElementsMatch(t *testing.T) {
 	}
 }
 
+func TestDiffLists(t *testing.T) {
+	check := func(listA, listB interface{}, extraA, extraB []interface{}) {
```
boyan-soubachov: Thank you, one last thing.

Would it not be clearer if we convert these unit tests to table-driven tests? (https://github.com/golang/go/wiki/TableDrivenTests)

martin-sucha: Okay, updated.

martin-sucha:
@boyan-soubachov could this be merged please?

PR #845, opened 2019-11-15
Add ContainsValue for maps

jybp:
The current `Contains` func doesn't allow to check for specific values within a map.

Having an easy way to check for map values is really useful.
Especially when map keys are random or can't be known inside the test.
I added it outside of the current `Contains` func.


boyan-soubachov:
I am closing this PR due to inactivity. Please feel free to comment/reopen if you wish to address the issues.

PR #846, opened 2019-11-26
don't fail the whole build on go-tip failures

phemmer:
tip:
* Can change at any time
* Has no stability/compatability guarantee
* Is development code & can be broken
* Is out of control of testify contributors

I addition, go versions 1.12 and 1.13 are completely missing from the build matrix, as well as "release", yet "tip" is present...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ matrix:
     - go: "1.11.x"
       env: GO111MODULE=on
     - go: tip
+  allow_failures:
+    - go: tip
   script:
     - ./.travis.gogenerate.sh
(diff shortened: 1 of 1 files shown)


boyan-soubachov:
@phemmer , thank you for the contribution. We've recently had some changes in the `.travis.yml` file which add Go 1.11 - 1.13. 

I think this change is a good idea. Would it also not make sense to then add another travis build that is the latest official release without explicitly mentioning a version? I.e. a stable version of `tip`?

We can then have `tip` be allowed to fail but the latest release not so; this will ensure that we always support the latest official release when making any code changes.

Issue #841, opened 2019-11-10
go generate ./... broken on Go 1.12+


leighmcculloch:
The repository uses `go generate ./...` to generate the required and assertion formatter files, but when following the instructions in the `README.md` and following the example in `.travisci.gogenerate.sh` to generate the files with either Go 1.12.13 or 1.13.4, the following errors occur:

```
2019/11/09 22:18:23 open /home/leighmcculloch/devel/testify/src/github.com/stretchr/testify/assert/assertion_format.go: no such file or directory
exit status 1
assert/assertions.go:25: running "go": exit status 1
2019/11/09 22:18:24 open /home/leighmcculloch/devel/testify/src/github.com/stretchr/testify/assert/assertion_format.go: no such file or directory
exit status 1
require/forward_requirements.go:16: running "go": exit status 1
```

It looks like this code in the generator that looks up where the code lives is finding the code path incorrectly in these newer versions of Go that support Go modules. The code is located on my system outside a GOPATH.

https://github.com/stretchr/testify/blob/e442a1631a73174a4afde5b4c41ba21cde6f6889/_codegen/main.go#L174-L184

If I try moving the project to my GOPATH, I get a new set of errors:
```
2019/11/09 22:23:48 assertions.go:20:2: could not import github.com/davecgh/go-spew/spew (can't find import: "github.com/stretchr/testify/vendor/github.com/davecgh/go-spew/spew")
exit status 1
assert/assertions.go:25: running "go": exit status 1
2019/11/09 22:23:48 assertions.go:20:2: could not import github.com/davecgh/go-spew/spew (can't find import: "github.com/stretchr/testify/vendor/github.com/davecgh/go-spew/spew")
exit status 1
require/forward_requirements.go:16: running "go": exit status 1
```

boyan-soubachov:
As we spoke about in #848, happy to work on a solution for this to make it a bit less manual.

An initial thought I had is to add a git commit hook for any code modifying the 'base' files to automatically (and correctly) run `go generate`. That way, commits should automatically have up-to-date versions of the generated files.

leighmcculloch:
I saw your message on the PR but I haven't had a chance to retry the process as you described.

I think part of the issue is the generate process being dependent on a GOPATH and with Go Modules disabled. Testify is widely used, but still only runs its CI tests on Go 1.11 as the latest release which is pretty old and no longer supported. Rather than adding a githook I think it probably needs:

1. To be updated to work with Go Modules / Go 1.13+.
[...]

PR #848, opened 2019-11-29
Rerun go generate

boyan-soubachov:
* Added generated code for new assert & require tests

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -325,14 +325,6 @@ func JSONEqf(t TestingT, expected string, actual string, msg string, args ...int
 	return JSONEq(t, expected, actual, append([]interface{}{msg}, args...)...)
 }
 
-// YAMLEqf asserts that two YAML strings are equivalent.
-func YAMLEqf(t TestingT, expected string, actual string, msg string, args ...interface{}) bool {
-	if h, ok := t.(tHelper); ok {
-		h.Helper()
-	}
-	return YAMLEq(t, expected, actual, append([]interface{}{msg}, args...)...)
-}
-
 // Lenf asserts that the specified object has specific length.
 // Lenf also fails if the object has a type that len() not accept.
 //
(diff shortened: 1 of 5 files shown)


assert/assertion_format.go
```diff
@@ -557,6 +562,14 @@ func WithinDurationf(t TestingT, expected time.Time, actual time.Time, delta tim
 	return WithinDuration(t, expected, actual, delta, append([]interface{}{msg}, args...)...)
 }
 
+// YAMLEqf asserts that two YAML strings are equivalent.
+func YAMLEqf(t TestingT, expected string, actual string, msg string, args ...interface{}) bool {
```
boyan-soubachov: @georgelesica-wf , @ernesto-jimenez : May I ask what the logic behind having this function in the package is? 

It seems a bit too close to a specific use-case rather than general testing business logic, in my opinion.

leighmcculloch: Testing YAML and JSON equality is particularly difficult and that function and it's JSON counterpart makes it a lot easier and test output a lot easier to understand when two don't match. I don't think it is particularly business logic. Regardless, this function probably needs to stay if it's been added otherwise the change would be breaking?

leighmcculloch:
@boyan-soubachov What version of Go are you running these go generate commands with? I just opened #841 because I'm getting errors when attempting to do so. I'm wondering how you've managed to get it to go.

assert/assertion_format.go
boyan-soubachov: I understand the arguments/concerns about difficulty of testing JSON & YAML but my question arises from the fact that there are always new standards becoming the _de facto_ of the day :) 

I guess the argument I was trying to make is should the `testify` package also include standards-based formats or should be kept as a simple, low-level framework. I guess that's something for the steering group to decide :)

This won't be a breaking change by the way as one can only use the these functions in their code if they were `go generate`d. Since this code was never previously generated, it wouldn't have been used.

boyan-soubachov:
@leighmcculloch , I used go 1.13. The key-part is that go modules behave differently when in the `$GOPATH` and so one would get a lot of errors about missing imports. 

The way I solved it is by turning go-modules off for this repo (I kept it in my `$GOPATH`), `go get`ting  all the necessary imports and rerunning `go generate`

assert/assertion_format.go
glesica: Agreed that it's on the line, but in terms of philosophy I don't have strong feelings. Whatever the community finds helpful is probably fine as long as we don't go crazy.

boyan-soubachov:
Fair point. Thank you for the clarification and review 🙂

Issue #849, opened 2019-12-04
How to FailNow during SetupSuite?


AaronNBrock:
I'm writing a testing suite that spins up some resource on `SetupSuite` (in my case a virtual machine), runs some tests against it, then destroys the resource on `TearDownSuite`.

Now, sometimes the resource may fail to spin up or have some issue with it which makes running the tests useless, in which case I'd like to `suite.T().FailNow()`.  However, if you run `suite.T().FailNow()` during `SetupSuite` then the `TearDownSuite` never runs, which could lead to the VM not being destroyed & could lead to dangling/costly VMs.

Is this by design?  If so, how could I go about solving my issue.

glesica:
Would it make sense to let the setup finish and then have a test that runs immediately afterward and asserts that the environment is configured correctly? That being said, it seems reasonable that the teardown should at least run optionally if setup fails.

AaronNBrock:
@glesica the issue with doing it in the first test is I need the ability to pick and choose which tests I run (Some may take a long time).  I could maybe add this check during the `SetupTest`.  But, that's not an amazing solution because it leads to the tests being run over and over again, which is probably fine in _my_ case, but I could imagine that being a problem.

talha0324:
Any fix or a good workaround for this?

sneko:
For me this issue is still relevant. See https://github.com/stretchr/testify/issues/1123

PR #850, opened 2019-12-04
Move teardown defers to before setup

AaronNBrock:
A solution proposition for my issue #849


glesica:
I'm not opposed to this on principle, but I want to give it a few days to percolate in case this has consequences I'm not thinking of.

dmarkhas:
+1 for this, we're also hitting the same issue and would love to see this get fixed

AaronNBrock:
The only case I can see where this would cause a test to break is if they relied on the fact that the tear down doesn't run if the SetUp fails, e.g. SetUp creates a file & tear down tries to read it w/o verifying it exists.

But, that's probably bad practice anyhow.

boyan-soubachov:
> The only case I can see where this would cause a test to break is if they relied on the fact that the tear down doesn't run if the SetUp fails, e.g. SetUp creates a file & tear down tries to read it w/o verifying it exists.
> 
> But, that's probably bad practice anyhow.

👍 . Hopefully not too many people rely on this behaviour.

boyan-soubachov:
image
Oddly, the branch still has merge conflicts. Do you mind doing a rebase and force-push?

boyan-soubachov:
Have rebased in a separate PR (#931) due to demand and time.

mvdkleijn:
Closing this PR in favor of the new one at #931

PR #851, opened 2019-12-04
Remove vendor folder

boyan-soubachov:
* Removed vendor folder to get go generate working by default

This will allow contributors to be able to run `go generate ./...` on the default case (`GO111MODULE=auto` or unset, `stretchr/testify` in `$GOPATH/src`)


glesica:
OK, just to make sure I'm getting this... The only way to make `go generate` work properly, right now at least, is to clone the repo into your `GOPATH`, even though we are primarily relying on modules, or at least that was the plan. Is that accurate?

boyan-soubachov:
In summary:

If the `GO111MODULE` environment variable is unset or on `auto` (the default behaviour), `go generate ./...` should work perfectly fine inside our outside the `$GOPATH/src` directory as long as the `vendor/` folder isn't there as the `vendor/` folder forces go to use vendored imports.

If `GO111MODULE` is set to `off`, it will use at the `vendor/` folder if it exists or `$GOPATH` if it doesn't.

If `GO111MODULE` is set to `on`, the existence of the `vendor/` folder will override it and effectively force it to use vendored imports (which are incomplete), and if the the `vendor/` folder doesn't exist, it will automatically resolved depencencies and `go generate ./...` should work.

I'm not sure why the `vendor/` folder was included previously but it's not doing us any favours at the moment as it's out of date. It either needs to be updated (meaning we have to manually keep track of it) or removed (meaning we let go modules do the work).

My $0.02 :)

glesica:
Sounds good. Let's just drop it for now then. I'm going to let this sit for a day or so in case anyone else wants to weigh-in, but assuming no one has any strong feelings I'll merge it.

boyan-soubachov:
> Sounds good. Let's just drop it for now then. I'm going to let this sit for a day or so in case anyone else wants to weigh-in, but assuming no one has any strong feelings I'll merge it.

Makes sense to me :smile: :+1:

PR #852, opened 2019-12-07
Fix go generate for Go Modules

### LINKED ISSUE #841: go generate ./... broken on Go 1.12+

leighmcculloch:
Fix two issues with codegen that cause it not to work with Go Modules.

1. Changed `SrcRoot` and `ImportPath` to `Dir`. When parsing the code files the package SrcRoot and ImportPath were being joined which assumes the package is hosted inside a GOPATH src directory. Using those two fields is unnecessary because Dir tells us where the package lives.

2. When checking the types in the source files the default importer is being used. Unfortunately the default importer has not been updated to work with Go Modules and it only looks in the GOPATH for imports. Luckily the Go team were convinced to backport some logic from the newer go/packages package to go/build, and we can make use of the go/build importer which supports Go Modules.

In the future codegen should be rewritten to use go/packages once it is included in the stdlib. go/packages currently lives at golang.org/x/tools/go/packages and is under development.

I've confirmed this works on 1.11, 1.12, 1.13 with modules on, inside and outside of a GOPATH. Testify itself will still work in older versions of Go, but the codegen will need to be done in a version of Go supporting modules.

I tested the codegen with this Dockerfile:
<details><summary>Test Dockerfile</summary>

```Dockerfile
ARG REPO=https://github.com/leighmcculloch/stretchr--testify.git
ARG BRANCH=fix-go-generate-for-modules

# Go 1.13 inside GOPATH
FROM golang:1.13
ARG REPO
ARG BRANCH
ARG DIR=/go/src/github.com/stretchr/testify
RUN git clone --single-branch --branch $BRANCH $REPO $DIR
WORKDIR $DIR
RUN go generate ./...
RUN git diff --exit-code

# Go 1.13 outside GOPATH
FROM golang:1.13
ARG REPO
ARG BRANCH
ARG DIR=/testify
RUN git clone --single-branch --branch $BRANCH $REPO $DIR
WORKDIR $DIR
RUN go generate ./...
RUN git diff --exit-code

# Go 1.12 inside GOPATH with modules enabled
FROM golang:1.12
ARG REPO
ARG BRANCH
ARG DIR=/go/src/github.com/stretchr/testify
ENV GO111MODULE=on
RUN git clone --single-branch --branch $BRANCH $REPO $DIR
WORKDIR $DIR
RUN go generate ./...
RUN git diff --exit-code

# Go 1.12 outside GOPATH
FROM golang:1.12
ARG REPO
ARG BRANCH
ARG DIR=/testify
RUN git clone --single-branch --branch $BRANCH $REPO $DIR
WORKDIR $DIR
RUN go generate ./...
RUN git diff --exit-code

# Go 1.11 inside GOPATH with modules enabled
FROM golang:1.11
ARG REPO
ARG BRANCH
ARG DIR=/go/src/github.com/stretchr/testify
ENV GO111MODULE=on
RUN git clone --single-branch --branch $BRANCH $REPO $DIR
WORKDIR $DIR
RUN go generate ./...
RUN git diff --exit-code

# Go 1.11 outside GOPATH
FROM golang:1.11
ARG REPO
ARG BRANCH
ARG DIR=/testify
RUN git clone --single-branch --branch $BRANCH $REPO $DIR
WORKDIR $DIR
RUN go generate ./...
RUN git diff --exit-code
```
</details>

The codegen will require modules to be on, but you can still use testify itself without modules. Given that modules is always on in the latest version, I don't think this is a huge issue.

I've also fixed TravisCI the config so that it runs the custom scripts.

Fixes #841

### DIFF
--- a/.travis.gogenerate.sh
+++ b/.travis.gogenerate.sh
@@ -3,7 +3,7 @@
 # If GOMOD is defined we are running with Go Modules enabled, either
 # automatically or via the GO111MODULE=on environment variable. Codegen only
 # works with modules, so skip generation if modules is not in use.
-if [[ ! -z "$(go env GOMOD)" ]]; then
+if [[ -z "$(go env GOMOD)" ]]; then
   exit 0
 fi
 
--- a/.travis.gogenerate.sh
+++ b/.travis.gogenerate.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-go generate ./...
+GO111MODULE=on go generate ./...
 if [ -n "$(git diff)" ]; then
   echo "Go generate had not been run"
   git diff
--- a/.travis.gogenerate.sh
+++ b/.travis.gogenerate.sh
@@ -1,10 +1,13 @@
 #!/bin/bash
 
-if [[ "$TRAVIS_GO_VERSION" =~ ^1\.(12|13)(\..*)?$ ]]; then
+# If GOMOD is defined we are running with Go Modules enabled, either
+# automatically or via the GO111MODULE=on environment variable. Codegen only
+# works with modules, so skip generation if modules is not in use.
+if [[ ! -z "$(go env GOMOD)" ]]; then
   exit 0
 fi
 
-GO111MODULE=on go generate ./...
+go generate ./...
 if [ -n "$(git diff)" ]; then
   echo "Go generate had not been run"
   git diff
--- a/.travis.gogenerate.sh
+++ b/.travis.gogenerate.sh
@@ -1,10 +1,5 @@
 #!/bin/bash
 
-if [[ "$TRAVIS_GO_VERSION" =~ ^1\.[45](\..*)?$ ]]; then
-  exit 0
-fi
-
-go get github.com/ernesto-jimenez/gogen/imports
 go generate ./...
 if [ -n "$(git diff)" ]; then
   echo "Go generate had not been run"
--- a/.travis.gogenerate.sh
+++ b/.travis.gogenerate.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+if [[ "$TRAVIS_GO_VERSION" =~ ^1\.(12|13)(\..*)?$ ]]; then
+  exit 0
+fi
+
 GO111MODULE=on go generate ./...
 if [ -n "$(git diff)" ]; then
   echo "Go generate had not been run"
--- a/.travis.gogenerate.sh
+++ b/.travis.gogenerate.sh
@@ -4,6 +4,7 @@
 # automatically or via the GO111MODULE=on environment variable. Codegen only
 # works with modules, so skip generation if modules is not in use.
 if [[ -z "$(go env GOMOD)" ]]; then
+  echo "Skipping go generate because modules not enabled and required"
   exit 0
 fi
 
--- a/_codegen/go.mod
+++ b/_codegen/go.mod
@@ -1,5 +1,5 @@
 module github.com/stretchr/testify/_codegen
 
-go 1.13
+go 1.11
 
 require github.com/ernesto-jimenez/gogen v0.0.0-20180125220232-d7d4131e6607
--- a/_codegen/go.mod
+++ b/_codegen/go.mod
@@ -1,0 +1,5 @@
+module github.com/stretchr/testify/_codegen
+
+go 1.13
+
+require github.com/ernesto-jimenez/gogen v0.0.0-20180125220232-d7d4131e6607
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -22,7 +22,7 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
-//go:generate go run ../_codegen/main.go -output-package=assert -template=assertion_format.go.tmpl
+//go:generate sh -c "cd ../_codegen && go build && cd - && ../_codegen/_codegen -output-package=assert -template=assertion_format.go.tmpl"
 
 // TestingT is an interface wrapper around *testing.T
 type TestingT interface {
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -84,7 +84,7 @@ func Run(t *testing.T, suite TestingSuite) {
 	defer failOnPanic(t)
 
 	suiteSetupDone := false
-	
+
 	methodFinder := reflect.TypeOf(suite)
 	tests := []testing.InternalTest{}
 	for index := 0; index < methodFinder.NumMethod(); index++ {
--- a/.travis.govet.sh
+++ b/.travis.govet.sh
@@ -1,10 +1,5 @@
 #!/bin/bash
 
-cd "$(dirname $0)"
-DIRS=". assert require mock _codegen"
 set -e
-for subdir in $DIRS; do
-  pushd $subdir
-  go vet
-  popd
-done
+
+go vet ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,14 @@ matrix:
       env: GO111MODULE=off
     - go: "1.11.x"
       env: GO111MODULE=on
+    - go: "1.12.x"
+      env: GO111MODULE=off
+    - go: "1.12.x"
+      env: GO111MODULE=on
+    - go: "1.13.x"
+      env: GO111MODULE=off
+    - go: "1.13.x"
+      env: GO111MODULE=on
     - go: tip
   script:
     - ./.travis.gogenerate.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,8 @@ matrix:
     - go: "1.13.x"
       env: GO111MODULE=on
     - go: tip
-script:
-  - ./.travis.gogenerate.sh
-  - ./.travis.gofmt.sh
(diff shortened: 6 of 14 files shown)


assert/assertions.go
```diff
@@ -22,7 +22,7 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
-//go:generate go run ../_codegen/main.go -output-package=assert -template=assertion_format.go.tmpl
+//go:generate sh -c "cd ../_codegen && go build && cd - && ../_codegen/_codegen -output-package=assert -template=assertion_format.go.tmpl"
```
leighmcculloch: These changes are required because `go run` can no longer be used with a relative path now that the `_codegen` is its own module. This builds the _codegen binary, then executes it from the current directory.

_codegen/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/stretchr/testify/_codegen
+
+go 1.13
+
+require github.com/ernesto-jimenez/gogen v0.0.0-20180125220232-d7d4131e6607
```
leighmcculloch: I've added a `go.mod` to `_codegen` so that the `gogen` dependency doesn't show up in the `testify` `go.mod` file when we run go generate. When you run go generate on Go 1.12/1.13 it is discovering the dependency in the `_codegen` directory and adding it to the roots go.mod file. That dependency will become a constraint in any importers dependency graph so it's better to keep the code in here isolated.

leighmcculloch:
@glesica This fixes the codegen script so it works on Go Modules and is ready for 👀.

Note that I discovered in this process that testify's TravisCI config is broken. I've fixed it separately in #853. Once #853 is merged, and master is then merged into this PR, we should be able to see this PR run the `.travis.gogenerate.sh` script on CI.

.travis.gogenerate.sh
```diff
@@ -1,11 +1,10 @@
 #!/bin/bash
 
-if [[ "$TRAVIS_GO_VERSION" =~ ^1\.[45](\..*)?$ ]]; then
+if [[ "$TRAVIS_GO_VERSION" =~ ^1\.(12|13)(\..*)?$ ]]; then
   exit 0
 fi
 
-go get github.com/ernesto-jimenez/gogen/imports
-go generate ./...
+GO111MODULE=on go generate ./...
```
boyan-soubachov: Wouldn't this override the travis `env` setting? Should this line not just be `go generate ./...`?

_codegen/go.mod
boyan-soubachov: Should we have `_codegen` as a separate module? If I'm not mistaken, it is custom-written for `testify/assert` and is an integral requirement for generating the usable functions.

Would it make sense to try and keep it simple and only separate it if we have issues from the community about `_codegen` being included?

.travis.gogenerate.sh
leighmcculloch: Yup I'll change how this works. Thanks.

leighmcculloch: Updated in 1f912b6.

_codegen/go.mod
leighmcculloch: This is the perfect situation for a separate module. As you said it is custom-written for testify, which means we don't want its dependencies to pollute the dependency graph of people importing testify. Codegen is dependent on the `gogen` tool, and without making this its own module anyone importing testify will also have gogen included in their dependency graph.

There's also the problem that the `_codegen` folder starting with an underscore causes it to be ignored by the go mod tooling and so it doesn't show up in `go.mod`, but the moment we do `go run` on the file the `go.mod` gets updated to include it, which causes diffs everytime we run codegen. This is an important part of the change.

suite/suite.go
```diff
@@ -84,7 +84,7 @@ func Run(t *testing.T, suite TestingSuite) {
 	defer failOnPanic(t)
 
 	suiteSetupDone := false
-	
+
```
leighmcculloch: These changes are required because TravisCI is running the gofmt check again. A couple files weren't formatted properly.

_codegen/go.mod
leighmcculloch: @glesica Let me know if you'd prefer codegen not be isolated in its own module, and I'll work around that. If we do that I'll need to rename the directory to `codegen` because the underscore causes odd behavior with the go mod commands.

leighmcculloch:
🎉 The build is finally green 💚. @glesica I was a little premature earlier when I mentioned you. This is actually ready for 👀now. 😄

_codegen/go.mod
glesica: Personally, I like the idea of isolating it to keep development dependencies from leaking into applications.

boyan-soubachov: Fair enough, it's a good point @leighmcculloch

assert/assertions.go
boyan-soubachov: If we're rerunning this command several times in a few files, would it make more sense to separate the commands into a bash file and have the `go:generate` tool call the same file but with different parameters?

I think that would make the code cleaner & easier to maintain as well as reused with very little added extra complexity.

leighmcculloch: I think this line is low frequency enough that the value in doing that isn't high. It would mean another script floating around somewhere in the repo. I don't feel strongly though. Happy to add it if the maintainers would like it. @glesica?

boyan-soubachov: Fair point. I'd be happy with it either way as well then. I guess if it proves problematic we can look into separating it out. 👍

leighmcculloch:
That's a good idea.

I already tested the code generation on changes introduced in #839 and it generates the code as expected. I also tested deleting generated code and checking codegen returned it when run exactly the same byte-for-byte.

I just tested importing the latest commit of this PR and ran all the tests in a repo using testify that has 789 tests, and it worked perfectly 🎉.

boyan-soubachov:
Thank you! :)

.travis.gogenerate.sh
```diff
@@ -1,10 +1,13 @@
 #!/bin/bash
 
-if [[ "$TRAVIS_GO_VERSION" =~ ^1\.[45](\..*)?$ ]]; then
+# If GOMOD is defined we are running with Go Modules enabled, either
+# automatically or via the GO111MODULE=on environment variable. Codegen only
+# works with modules, so skip generation if modules is not in use.
+if [[ -z "$(go env GOMOD)" ]]; then
```
boyan-soubachov: Won't this make all of our travis tests which specify `GOMODULE=off` pointless?

assert/assertions.go
glesica: Sounds good to me.

glesica:
> In the future codegen should be rewritten to use go/packages once it is included in the stdlib. go/packages currently lives at golang.org/x/tools/go/packages and is under development.

@leighmcculloch would you mind opening an issue for this and include a little bit of background info?

.travis.gogenerate.sh
glesica: The generated code already exists, so I would imagine tests would pass still. We wouldn't catch situations where generate needed to be run and wasn't, but other builds should catch that. Maybe?

leighmcculloch: You're right, generated code would already exist just when modules if off we aren't checking that it regenerates the same. This is fine I think. The codegen code isn't worth trying to make work on both modules and not modules.

leighmcculloch:
@glesica I'd rather open an issue when go/build becomes a problem or when go/packages becomes part of the stdlib and go/build gets formally deprecated. go/package isn't in the stdlib yet, and until it is anything could happen.

glesica:
Sounds reasonable to me, thanks!

PR #853, opened 2019-12-07
Fix TravisCI config

leighmcculloch:
Fix the indenting in the TravisCI config that is causing the custom scripts to be ignored by TravisCI. None of the builds on testify are actually using the custom scripts, it's just using the default Go build behavior. This is because the `script` key is supposed to be at the top-level but it is nested under the matrix configs.

Some files are not go fmt'd correctly, so I've fixed them too to make the `.travis.gofmt.sh` script pass.

I also need to change the guard in the `.travis.gogenerate.sh` script so that it only runs when modules is turned off. This is because at the moment the codegen doesn't support modules. That is being fixed separately in #852.

### DIFF
--- a/.travis.gogenerate.sh
+++ b/.travis.gogenerate.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 
-if [[ "$TRAVIS_GO_VERSION" =~ ^1\.[45](\..*)?$ ]]; then
+# If GOMOD is defined we are running with Go Modules enabled, either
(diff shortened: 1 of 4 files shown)


leighmcculloch:
Going to close this PR and just fix the issue in #852. I was trying to do the "right thing" and isolate these two fixes, but it turns out this is fiddly and should be much easier to fix in #852 where the codegen is configured to run with modules.

PR #855, opened 2019-12-11
Added negative dir and file tests

### LINKED ISSUE #740: Introduce negated functions to FileExists and DirExists assertions
The `FileExists` and `DirExists` assertions help a lot. Sometimes you want to assert the opposite:

- This file should not exist.
- This directory should not exist.

I'd love to see functions like `FileNotExists` and `DirNotExists` as you otherwise have to build them yourself. Would you accept a pull request for such functionality? Happy to create one.

boyan-soubachov:
* Added NotFileExists test
* Added NotDirExists test
* Cleaned-up some comment formatting

Closes #740

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1391,9 +1391,6 @@ func NoFileExists(t TestingT, path string, msgAndArgs ...interface{}) bool {
 	}
 	info, err := os.Lstat(path)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return true
-		}
 		return true
 	}
 	if info.IsDir() {
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1364,7 +1364,8 @@ func NotZero(t TestingT, i interface{}, msgAndArgs ...interface{}) bool {
 	return true
 }
 
-// FileExists checks whether a file exists in the given path. It also fails if the path points to a directory or there is an error when trying to check the file.
+// FileExists checks whether a file exists in the given path. It also fails if
+// the path points to a directory or there is an error when trying to check the file.
 func FileExists(t TestingT, path string, msgAndArgs ...interface{}) bool {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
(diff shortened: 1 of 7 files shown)


assert/assertions.go
```diff
@@ -1382,7 +1383,27 @@ func FileExists(t TestingT, path string, msgAndArgs ...interface{}) bool {
...
+func NoFileExists(t TestingT, path string, msgAndArgs ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	info, err := os.Lstat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return true
+		}
+		return true
```
glesica: We just return `true` either way, is there a reason that the error check needs to exist?

assert/assertions.go
```diff
@@ -1382,7 +1383,27 @@ func FileExists(t TestingT, path string, msgAndArgs ...interface{}) bool {
...
+	}
+	if info.IsDir() {
+		return true
+	}
```
glesica: `FileMode` gives us a little more granularity if we decide that we want this to be more subtle. Right now a symlink would also pass the check. Maybe that's fine, not sure.

assert/assertions.go
```diff
@@ -1400,6 +1421,25 @@ func DirExists(t TestingT, path string, msgAndArgs ...interface{}) bool {
...
+	info, err := os.Lstat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return true
+		}
+		return true
+	}
+	if !info.IsDir() {
+		return true
+	}
```
glesica: Would a symlink to a directory fail this check? If so, we should probably try to make the file and dir versions consistent in how picky they are.

go.sum
```diff
@@ -1,5 +1,6 @@
 github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
```
glesica: Why'd this show up here?

glesica:
This looks great, thanks for doing it! I just had a couple questions about what we want to consider a file or a directory. They may not have good answers. In that case I'm fine with a best effort, so we can probably merge as-is, perhaps with some caveats in the documentation.

assert/assertions.go
boyan-soubachov: Good point, missed that :)

go.sum
boyan-soubachov: The go toolchain auto-added it. It's a lock-entry signifying `go-spew v1.1.1`'s hash.

assert/assertions.go
boyan-soubachov: Good point. I copied this from current behaviour so that the same rules are followed (i.e. symlinks are fine). If we want to have consistency and pick one for both File and Dir, we might have compatibility-breaking changes.

assert/assertions.go
boyan-soubachov: Agreed. Will check but it might be a breaking change if File and Dir are not consistent with regards to symlinks.

boyan-soubachov:
@glesica Added tests for the symlink scenarios we discussed. In summary:
* None of the 4 tests (`(No)FileExists`, `(No)DirExists`) follow symlinks
* `*DirExists` tests do not treat symlinks as directories as that's convention in Go's `os` package.

In its current state, the negated tests (`No*Exists`) behave exactly the opposite of the normal case tests. I would recommend we keep it that way as it's the simplest and most predictable (i.e. we follow standard Golang behaviour). If we find users/issues require more complex behaviour (i.e. following symlinks), then we can look at adding that as new functionality.

PR #859, opened 2019-12-20
Function comparisons for Equal

### LINKED ISSUE #182: assert.Equal() not work on functions
I want compare two functions are the same function, but failed:

fn := func() {}
assert.Equal(t, fn, fn) // expected passed, but failed

In go, functions are not comparable, but can testify do some special treatment on function type, such as
compare there pointer address, like reflect.Value(fn).Pointer(). Or add a new function like EqualFunc() not slow down Equal() function.

boyan-soubachov:
* Added `nil` function comparisons for `Equal`

Fixes #182

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -352,6 +352,19 @@ func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) 
 
 }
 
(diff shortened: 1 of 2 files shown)


boyan-soubachov:
@glesica

PR #866, opened 2020-01-10
Errors is

Fryuni:
Implements a `ErrorIs` as a test for error wrappers from [the new error methods in go 1.13](https://blog.golang.org/go1.13-errors)

```go
errA := errors.New("some error")
errB := fmt.Errorf("wrapping: %w", errA)
assert.ErrorIs(t, errA, errB)
```

### DIFF
--- a/assert/1.13_assertion_format.go
+++ b/assert/1.13_assertion_format.go
@@ -1,0 +1,17 @@
+// +build go1.13
+
+package assert
+
+// ErrorIsf asserts that a function returned an error chain
+// and that it is contains the provided error.
+// Mimicking errors.Is, providing `nil` as expected error
+// is the same as calling NoError
+//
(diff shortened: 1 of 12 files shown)


oralordos:
Since errors.Is comes from go1.13, you may wish to move the check over to a new conditionally compiled file.

Fryuni:
Yeah, I noticed that too after travis failed, I'll change it as soon as I have some time :)

Fryuni:
Hum, the code generation will add the new functions on the wrong files, probably will have to change the codegen... I'll think about it

nikandfor:
Did it in #884

boyan-soubachov:
This is a big change that has had a detailed discussion in #884 . Given the way we generate code, one of the options is to make this a testify/v2 major change.

PR #871, opened 2020-01-23
Add Never assertion

### LINKED ISSUE #870: Feature request: assert.Never
Proposal for new assertion to the API: `assert.Never(*testing.T, func() bool, time.Duration, time.Duration)`. It would be a direct mirror of `assert.Eventually`, checking that the given condition wasn't met within a certain timeframe.

Copying a bunch of code from `assert.Eventually`, the implementation could look like this: 

```go
func Never(t *testing.T, condition func() bool, waitFor time.Duration, tick time.Duration, msgAndArgs ...interface{}) bool {
	timer := time.NewTimer(waitFor)
	checkPassed := make(chan bool)
	defer timer.Stop()
	defer ticker.Stop()
	defer close(checkPassed)
	for {
		select {
		case <-timer.C:
			return true
		case result := <-checkPassed:
			if result {
				return assert.Fail(t, "Condition satisfied", msgAndArgs...)
			}
		case <-ticker.C:
			go func() {
				checkPassed <- condition()
			}()
		}
	}
}
```

If this is something the maintainers would merge, I'd be happy to make a PR with tests and documentation for it.

torkelrogstad:
Closes #870 

Never asserts that a conditon isn't met within a given timeframe.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1591,3 +1591,36 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 		}
 	}
 }
+
+// Never asserts that the given condition doesn't satisfy in waitFor time,
+// periodically checking the target function each tick.
+//
+//    assert.Never(t, func() bool { return false; }, time.Second, 10*time.Millisecond)
+func Never(t TestingT, condition func() bool, waitFor time.Duration, tick time.Duration, msgAndArgs ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+
+	ch := make(chan bool, 1)
+
+	timer := time.NewTimer(waitFor)
+	defer timer.Stop()
+
+	ticker := time.NewTicker(tick)
+	defer ticker.Stop()
+
+	for tick := ticker.C; ; {
+		select {
+		case <-timer.C:
+			return true
+		case <-tick:
+			tick = nil
+			go func() { ch <- condition() }()
+		case v := <-ch:
+			if v {
+				return Fail(t, "Condition satisfied", msgAndArgs...)
+			}
+			tick = ticker.C
+		}
+	}
+}
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -363,6 +363,17 @@ func LessOrEqualf(t TestingT, e1 interface{}, e2 interface{}, msg string, args .
 	return LessOrEqual(t, e1, e2, append([]interface{}{msg}, args...)...)
 }
 
+// Neverf asserts that the given condition doesn't satisfy in waitFor time,
+// periodically checking the target function each tick.
+//
+//    assert.Neverf(t, func() bool { return false; }, time.Second, 10*time.Millisecond, "error message %s", "formatted")
+func Neverf(t TestingT, condition func() bool, waitFor time.Duration, tick time.Duration, msg string, args ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	return Never(t, condition, waitFor, tick, append([]interface{}{msg}, args...)...)
+}
+
 // Nilf asserts that the specified object is nil.
 //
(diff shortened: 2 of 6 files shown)


assert/assertions.go
```diff
@@ -1591,3 +1591,36 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 		}
 	}
 }
+
+// Never asserts that the given condition doesn't satisy in waitFor time,
```
boyan-soubachov: Typo, "satisy" => "satisfy"

assert/assertions.go
```diff
@@ -1591,3 +1591,36 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
...
+// periodically checking target function each tick.
```
boyan-soubachov: "checking _the_ target function each tick"

assert/assertions.go
```diff
@@ -1591,3 +1591,36 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
...
+	ticker := time.NewTicker(tick)
+	defer ticker.Stop()
+
+	for tick := ticker.C; ; {
+		select {
+		case <-timer.C:
+			return true
+		case <-tick:
+			tick = nil
+			go func() { ch <- condition() }()
```
boyan-soubachov: If I am not mistaken, will this not end up spawning _timer_/_tick_ goroutines of the provided function? Should we not rather run the `condition()` function once and only do checks here?

torkelrogstad: What do you mean by running the condition function only once? I.e. where would you place that in the code, it is not clear to me. 

To be honest I just copied code from the existing `Eventually` function.  Here's the place where that part of the code was touched last: https://github.com/stretchr/testify/commit/f1bd0923b8320ddfb0a8a9d919b5aeb8b7a4cc40. The previous place was when the code was added. It looks like it has been done like this since the beginning.

boyan-soubachov: To clarify what I meant: It looks to me that on every timer tick, a goroutine is launched triggers calls the condition function. I'm not 100% sure if that's the case, but it does look like it to me. This means that we're re-running the function on every tick which would prove to be very inefficient and with possibly adverse side-effects if it's not a pure function/stateless.

> To be honest I just copied code from the existing Eventually function. Here's the place where that part of the code was touched last: f1bd092. The previous place was when the code was added. It looks like it has been done like this since the beginning.

Makes sense, maybe it's worth adding a unit test to check exactly that the existing `Eventually` logic doesn't call multiple copies of the same condition function. Thoughts?

I'd be happy to help where necessary. If you wish I could add the test for the existing, `Evenntually` function and we could work from there given the outcome?

torkelrogstad: From my understanding, rerunning the function every tick is the desired thing to do here. The function that's passed in is expected to not be pure, otherwise neither `Never` or `Eventually` would make any sense. If the functions are pure, `assert.Never` would be equivalent to `assert.False` and `assert.Eventually` would be equivalent to `assert.True`.

boyan-soubachov: :thinking: 

I think my understanding may be a bit off. Is the point of `Eventually` and `Never` to assert that a function completes/doesn't complete within a certain amount of time?

torkelrogstad: No, that seems to be a misunderstanding on your part.

`Eventually` asserts that the given function returns true at some point before the timeout
`Never` asserts that the given function _does not_ return true at some point before the timeout

boyan-soubachov: Got you. I'm still unsure of this but I'm happy to approve and commit. I'll spend some time figuring this out and maybe put out a refactoring PR; which I'd appreciate your feedback on :)

Happy for you to merge this and thank you for your contribution :)

boyan-soubachov:
@glesica , thoughts on this?

assert/assertions.go
torkelrogstad: Sure, just ping me if there's something you want me to look at! Thanks for taking a look at this.

PR #872, opened 2020-01-24
Add unsafe.Pointer to Nil/NotNil checks

### LINKED ISSUE #863: Expected nil, but got: (unsafe.Pointer)(nil)
Nil/NotNil assertions should work with unsafe.Pointer, since == nil and != nil do.

boyan-soubachov:
* Added checks for Nil/NotNil to handle the cases of `unsafe.Pointer` types
Fixes #863

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -515,7 +515,8 @@ func isNil(object interface{}) bool {
 		[]reflect.Kind{
 			reflect.Chan, reflect.Func,
 			reflect.Interface, reflect.Map,
-			reflect.Ptr, reflect.Slice},
+			reflect.Ptr, reflect.Slice,
+			reflect.UnsafePointer},
 		kind)
 
 	if isNilableKind && value.IsNil() {
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"testing"
 	"time"
+	"unsafe"
 )
 
 var (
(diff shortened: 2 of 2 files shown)


boyan-soubachov:
@georgelesica-wf , having had a look at the Travis build, it seems that there was a bug in Go 1.11 and earlier where `IsNil()` could not be called on types of `unsafe.Pointer`.  See https://go-review.googlesource.com/c/go/+/155797/

This leaves us with a bit of a quandary.

Do we want to:
a) Remove `testify` support for anything earlier than Go 1.12?
b) Hold off on this feature until Go 1.11 is sufficiently old enough? What is "sufficiently old enough"?

Thoughts? I'd like to say that I'm leaning towards `a` but Go 1.12 is still relatively new (just over a year old) and it seems quite unfair to drop support for something that new. Therefore I'm of the opinion that we let this soak for a bit and release it a few minor revisions down the line.

glesica:
What about pulling it out into its own file and using a compile flag to only provide it for versions newer than 1.12?

boyan-soubachov:
> What about pulling it out into its own file and using a compile flag to only provide it for versions newer than 1.12?

It's a good idea but it will require us to effectively maintain two separate copies of the `Nil`/`NotNil` functions if I am not mistaken. Given that this is an edge-case, I'm not sure it would be worth the added complexity.

Thoughts?

glesica:
Yeah, I definitely don't like working with conditionally compiled code, particularly since most IDEs don't make it a great experience in my experience. On the other hand, I definitely feel like 1.12 being a year old is a little too soon to drop support for it, particularly given that Go is starting to show up in "enterprise" settings.

glesica:
@boyan-soubachov do we want to get this into the next release? I'm thinking we should make the next release 2.0.0 so that would be a nice place to drop support for older versions of Go, I suppose.

boyan-soubachov:
I think that would make sense. I don’t think it warrants a whole separate ‘branch’ of code as we’ve discussed just to backport it. 

We should move this to 2.0.0 but I do think we have enough merged work for another release before that. I was working through the list of issues in the next milestone and was hoping we could see that out before 2.0.0

dolmen:
Support for `uintptr` has been fixed by #1319.

The issues raised about Go pre-1.11 are not relevant anymore as we do not support those old versions.

Closing.

PR #873, opened 2020-01-25
fix: error diff was not being show for large values

chirino:
If you were comparing values which when formatted were longer than about 64k then the actual, expected and diff messages were not printed to the console.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -429,14 +429,27 @@ func samePointers(first, second interface{}) bool {
 // to a type conversion in the Go grammar.
 func formatUnequalValues(expected, actual interface{}) (e string, a string) {
 	if reflect.TypeOf(expected) != reflect.TypeOf(actual) {
-		return fmt.Sprintf("%T(%#v)", expected, expected),
-			fmt.Sprintf("%T(%#v)", actual, actual)
+		return fmt.Sprintf("%T(%s)", expected, truncatingFormat(expected)),
+			fmt.Sprintf("%T(%s)", actual, truncatingFormat(actual))
 	}
 	switch expected.(type) {
 	case time.Duration:
 		return fmt.Sprintf("%v", expected), fmt.Sprintf("%v", actual)
 	}
-	return fmt.Sprintf("%#v", expected), fmt.Sprintf("%#v", actual)
+	return truncatingFormat(expected), truncatingFormat(actual)
+}
+
+// truncatingFormat formats the data and truncates it if it's too long.
+//
+// This helps keep formatted error messages lines from exceeding the
+// bufio.MaxScanTokenSize max line length that the go testing framework imposes.
+func truncatingFormat(data interface{}) string {
+	value := fmt.Sprintf("%#v", data)
+	max := bufio.MaxScanTokenSize - 100 // Give us some space the type info too if needed.
+	if len(value) > max {
+		value = value[0:max] + "<... truncated>"
+	}
+	return value
 }
 
 // EqualValues asserts that two objects are equal or convertable to the same types
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -1,6 +1,7 @@
 package assert
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/json"
 	"errors"
@@ -2187,5 +2188,20 @@ func Test_validateEqualArgs(t *testing.T) {
 
 	if validateEqualArgs(nil, nil) != nil {
 		t.Error("nil functions are equal")
+	}
+}
+
+func Test_truncatingFormat(t *testing.T) {
+
+	original := strings.Repeat("a", bufio.MaxScanTokenSize-102)
+	result := truncatingFormat(original)
+	Equal(t, fmt.Sprintf("%#v", original), result, "string should not be truncated")
+
+	original = original + "x"
+	result = truncatingFormat(original)
+	NotEqual(t, fmt.Sprintf("%#v", original), result, "string should have been truncated.")
+
+	if !strings.HasSuffix(result, "<... truncated>") {
+		t.Error("truncated string should have <... truncated> suffix")
 	}
 }


assert/assertions.go
```diff
@@ -429,14 +429,27 @@ func samePointers(first, second interface{}) bool {
...
 	}
-	return fmt.Sprintf("%#v", expected), fmt.Sprintf("%#v", actual)
+	return truncatingFormat(expected), truncatingFormat(actual)
+}
+
+// The output of the error message is eventually read a line at time by a bufio.Scanner
+// which has a max line length of bufio.MaxScanTokenSize, if the value we are displaying is too long
+// then the error message is not shown.  So lets trucate realy long values so we can
+// at least see some of the error message.
+func truncatingFormat(expected interface{}) string {
```
boyan-soubachov: Could you please add some unit tests for this function?

boyan-soubachov: Would it be possible/make sense to change this functions input argument to a strictly-defined type (e.g. `string` or `[]string`)?

assert/assertions.go
```diff
@@ -429,14 +429,27 @@ func samePointers(first, second interface{}) bool {
...
 	}
 	switch expected.(type) {
 	case time.Duration:
 		return fmt.Sprintf("%v", expected), fmt.Sprintf("%v", actual)
 	}
-	return fmt.Sprintf("%#v", expected), fmt.Sprintf("%#v", actual)
+	return truncatingFormat(expected), truncatingFormat(actual)
+}
+
+// The output of the error message is eventually read a line at time by a bufio.Scanner
```
boyan-soubachov: I think the function is clean, simple and easy to understand in terms of what it does. Do you think this comment is therefore redundant?

If you feel the need to have a comment here, would it not make more sense to follow the Go docstring format (i.e. starting with the function name) and describe 'why' instead of 'how'?

assert/assertions.go
chirino: It would greatly increase the amount of code in the only calling function.  But yes it can be done.

chirino: yep

assert/assertions.go
chirino: I clean it up a little.  I do think having a little comment on why is important since I don't think it's obvious to everyone that the go testing framework has a limit on line length for it's error messages.

boyan-soubachov: It's a good point. I guess to clarify a bit: IMO it's perfectly fine to have a comment stating 'why' about the code but it's a bit redundant if the code is clean & simple and the comment states 'how'. e.g.
```
// ParseUserInput gets the user input and parses it to ourSpec struct
func ParseUserInput(userInput string) ourSpec {
...
}
```
is redundant IMO, whereas:
```
// ParseUserInput guards against bad input and converts to our standard data struct
func ParseUserInput(userInput string) ourSpec {
...
}
```

assert/assertions.go
boyan-soubachov: Ah right, if it's not feasible then we could evaluate it as a standalone task. Thank you! 👍

assert/assertions_test.go
```diff
@@ -1,7 +1,8 @@
 package assert
 
 import (
-	"bytes"
+    "bufio"
```
boyan-soubachov: The indentation on these 2 new additions looks incorrect, could you please line them up with the rest of the imports? We use 4-character wide _tabs_

assert/assertions.go
```diff
@@ -429,14 +429,27 @@ func samePointers(first, second interface{}) bool {
...
 	case time.Duration:
 		return fmt.Sprintf("%v", expected), fmt.Sprintf("%v", actual)
 	}
-	return fmt.Sprintf("%#v", expected), fmt.Sprintf("%#v", actual)
+	return truncatingFormat(expected), truncatingFormat(actual)
+}
+
+// truncatingFormat formats the data and truncates it if it's too long.
+//
+// This helps keep formatted error messages lines from exceeding the
```
boyan-soubachov: Thank you, this is much better 😄 🙏

assert/assertions_test.go
```diff
@@ -2189,3 +2190,22 @@ func Test_validateEqualArgs(t *testing.T) {
...
+
+    original := strings.Repeat("a", bufio.MaxScanTokenSize-102)
+    result := truncatingFormat(original)
+    if fmt.Sprintf("%#v", original) != result {
+        t.Error("string should not be truncated")
+    }
+
+    original = original +"x"
+    result = truncatingFormat(original)
+    if fmt.Sprintf("%#v", original) == result {
```
boyan-soubachov: Could you please use the `Equal()` function used throughout the rest of the test file? We prefer to dogfood our own testing framework :)

assert/assertions_test.go
```diff
@@ -2189,3 +2190,22 @@ func Test_validateEqualArgs(t *testing.T) {
...
+        t.Error("string should have been truncated.")
+    }
+
+    if !strings.HasSuffix(result, "<... truncated>") {
```
boyan-soubachov: With the principle of dogfooding, you could use the `Contains()` function but I understand that it doesn't exactly cover the kind of comparison we want. Happy for you to make a call on this

boyan-soubachov:
Note, the builds are failing since you may have forgotten to run `gofmt` (or some other compatible formatting tool) on your source:

https://travis-ci.org/stretchr/testify/jobs/643376551?utm_medium=notification&utm_source=github_status

chirino:
Now using Equal and used gofmt

assert/assertions_test.go
```diff
@@ -2189,3 +2190,22 @@ func Test_validateEqualArgs(t *testing.T) {
 		t.Error("nil functions are equal")
 	}
 }
+
+func Test_truncatingFormat(t *testing.T) {
+
+	original := strings.Repeat("a", bufio.MaxScanTokenSize-102)
+	result := truncatingFormat(original)
+	if fmt.Sprintf("%#v", original) != result {
```
boyan-soubachov: Excuse me, I forgot to point out that just like you can use `Equal` below, you can use `NotEqual` here.

chirino:
fixed text

PR #875, opened 2020-01-28
Remove redundant files

boyan-soubachov:
* Removed redundant `dep` files as we now use Go moduels
* Correct import order for package_test.go


boyan-soubachov:
@glesica , minor clean-up

boyan-soubachov:
@glesica , minor clean-up

PR #877, opened 2020-01-28
Use latest Go master instead of dev

boyan-soubachov:
* Changed the latest Go version of the Travis build to be master instead of dev


boyan-soubachov:
Had a look earlier at the Travis build. It looks like they are. I still think `master` makes more sense than `tip` but happy to drop this PR either way.

glesica:
Go for it, master makes more sense to me, and I suspect to most people.

PR #880, opened 2020-01-29
Update License file copyright year

AngelKitty:


glesica:
There are quite a few contributors and AFAIK the project has never had a copyright assignment requirement. Should we add "and contributors" or something to the end of that?

AngelKitty:
@glesica It looks like no bad.

glesica:
Not sure what you mean but that.

AngelKitty:
@glesica I mean that it is more appropriate to remove "and contributors".

AngelKitty:
@glesica But this is the default format, Shall we change it?

glesica:
I'm not saying change the format, just that we should, perhaps, include a reference to the fact that many, many people hold a copyright over some portions of the code, at least as far as I know.

See https://opensource.stackexchange.com/questions/5508/what-does-and-contributors-in-the-copyright-byline-imply for some discussion.

mvdkleijn:
In my humble opinion, the line should read: `Copyright (c) 2012-2020 Mat Ryer, Tyler Bunnell and contributors`

This change would correctly point out the fact that Mat and Tyler are the original authors, but that there are many other contributors that have valid copyright claims on pieces of code.

AngelKitty:
@glesica OK, Could you please review it again?

PR #881, opened 2020-02-04
suite: make suite.TestingT satisfy mock.TestingT

hugelgupf:
...by adding Logf method.

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -22,6 +22,7 @@ type TestingT interface {
 	Errorf(format string, args ...interface{})
 	FailNow()
 	Log(args ...interface{})
+	Logf(format string, args ...interface{})
 	Skip(args ...interface{})
 }
 


hugelgupf:
> Thanks, clearly something we forgot :)
> 
> Is it possible to add a unit test for this? i.e. some kind of test that asserts that the `suite.TestingT` interface can satisfy the `mock.TestingT` interface?

I would prefer you do that -- I'm not super familiar with testify; I'm just trying to fix a dependency of a dependency for me

boyan-soubachov:
> > Thanks, clearly something we forgot :)
> > Is it possible to add a unit test for this? i.e. some kind of test that asserts that the `suite.TestingT` interface can satisfy the `mock.TestingT` interface?
> 
> I would prefer you do that -- I'm not super familiar with testify; I'm just trying to fix a dependency of a dependency for me

Fair enough, I'll add the test in parallel to this :)

hugelgupf:
Feel free to push it to the branch in this PR and merge it all at once

hugelgupf:
Thanks!

PR #883, opened 2020-02-05
Test go modules on and off for master builds

boyan-soubachov:


boyan-soubachov:
@glesica , review appreciated here :)

PR #886, opened 2020-02-14
Check method callable

Ogoyukin:
Added function 'IsMethodCallable', which checking that the method can be called.

### Example
```
if !userService.IsMethodCallable(t, "FindByID", userID) {
    return nil
}
args := s.MethodCalled(t, "FindByUserID", userID)
// TODO
```

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -556,14 +556,12 @@ func isArgsEqual(expected Arguments, args []interface{}) bool {
 	if len(args) != len(args) {
 		return false
 	}
-	var argsEqual = true
 	for i, v := range args {
 		if !reflect.DeepEqual(expected[i], v) {
-			argsEqual = false
-			continue
+			return false
 		}
 	}
-	return argsEqual
+	return true
 }
 
 func (m *Mock) methodWasCalled(methodName string, expected []interface{}) bool {
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -545,18 +545,25 @@ func (m *Mock) IsMethodCallable(t TestingT, methodName string, arguments ...inte
 		if v.Repeatability < v.totalCalls {
 			continue
 		}
-		var argsEqual = true
-		for i, arg := range v.Arguments {
-			if !reflect.DeepEqual(arguments[i], arg) {
-				argsEqual = false
-				continue
-			}
-		}
-		if argsEqual {
+		if isArgsEqual(v.Arguments, arguments) {
 			return true
 		}
 	}
 	return false
+}
+
+func isArgsEqual(expected Arguments, args []interface{}) bool {
+	if len(args) != len(args) {
+		return false
+	}
+	var argsEqual = true
+	for i, v := range args {
+		if !reflect.DeepEqual(expected[i], v) {
+			argsEqual = false
+			continue
+		}
+	}
+	return argsEqual
 }
 
 func (m *Mock) methodWasCalled(methodName string, expected []interface{}) bool {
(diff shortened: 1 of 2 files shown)


mock/mock.go
```diff
@@ -527,6 +527,47 @@ func (m *Mock) AssertNotCalled(t TestingT, methodName string, arguments ...inter
...
+		}
+		if len(arguments) != len(v.Arguments) {
+			continue
+		}
+		if v.Repeatability < v.totalCalls {
+			continue
+		}
+		if isArgsEqual(v.Arguments, arguments) {
+			isCallable = true
+			break
```
nelsam: Creating a variable (`isCallable` in this case) and then overwriting it in a narrower scope is _usually_ a sign that you should be returning instead.   Sometimes it's a sign that you need to extract the logic out to a separate function and set the variable to the return of that function.

In this case, it's perfectly reasonable to just `return true` here and delete the `isCallable` variable.

mock/mock.go
```diff
@@ -527,6 +527,47 @@ func (m *Mock) AssertNotCalled(t TestingT, methodName string, arguments ...inter
...
+		}
+	}
+	return isCallable
+}
+
+// Compares arguments
```
nelsam: Even though this is on an unexported function (and therefor it won't show up in `godoc`), it's probably better to include the function name in the documentation comment.

mock/mock.go
```diff
@@ -527,6 +527,47 @@ func (m *Mock) AssertNotCalled(t TestingT, methodName string, arguments ...inter
...
+func isArgsEqual(expected Arguments, args []interface{}) bool {
+	if len(args) != len(args) {
+		return false
```
nelsam: This code is unreachable.  `len(args)` will always be equal to `len(args)`.

glesica:
Agree with the comments by @nelsam

mock/mock.go
```diff
@@ -527,6 +527,47 @@ func (m *Mock) AssertNotCalled(t TestingT, methodName string, arguments ...inter
...
+// IsMethodCallable checking that the method can be called
+// If the method was called more than `Repeatability` return false
+func (m *Mock) IsMethodCallable(t TestingT, methodName string, arguments ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	var isCallable = false
```
boyan-soubachov: No need to explicitly set `isCallable` to false, it's a boolean's zero value.

mock/mock.go
boyan-soubachov: Agreed, we can just `return true` at line 551 and have no need for the `isCallable` variable.

mock/mock.go
boyan-soubachov: Go standard is to have the docstring start with the function's name

mock/mock.go
boyan-soubachov: Did you mean to test `len(expected)` vs `len(args)`?

mock/mock.go
```diff
@@ -527,6 +527,47 @@ func (m *Mock) AssertNotCalled(t TestingT, methodName string, arguments ...inter
...
+	return isCallable
+}
+
+// Compares arguments
+func isArgsEqual(expected Arguments, args []interface{}) bool {
+	if len(args) != len(args) {
+		return false
+	}
+	for i, v := range args {
+		if !reflect.DeepEqual(expected[i], v) {
```
boyan-soubachov: We can dogfood here by using our own `Equal()` :)

PR #887, opened 2020-02-14
invalid assertion result in NoError and Error is fixed using reflect value

princeparmar:
```
func Check() {
	checkNil((*Error)(nil))
}

type Error struct{}

func (*Error) Error() string {
	return "check ptr error"
}

func checkNil(err error) {
	if err == nil {
		fmt.Println("error is nil")
		return
	}

	fmt.Println("this is not nil")

}
```
**OUTPUT:** this is not nil

when we pass any object as any interface it creates pointer which points the data of that object.
when i pass my customised error to the error interface this error interface becomes the pointer to that nil object (which pointer is not considered as nil).

so to handle this i have added a nil check with reflect also which gives the desiered output.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1259,7 +1259,7 @@ func NoError(t TestingT, err error, msgAndArgs ...interface{}) bool {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
-	if err != nil {
+	if err != nil || !reflect.ValueOf(err).IsNil() {
 		return Fail(t, fmt.Sprintf("Received unexpected error:\n%+v", err), msgAndArgs...)
 	}
 
@@ -1277,7 +1277,7 @@ func Error(t TestingT, err error, msgAndArgs ...interface{}) bool {
 		h.Helper()
 	}
 
-	if err == nil {
+	if err == nil || reflect.ValueOf(err).IsNil() {
 		return Fail(t, "An error is expected but got nil.", msgAndArgs...)
 	}
 


princeparmar:
is this scenario intentionally added here.
`err = func() error {
		var err *customError
		if err != nil {
			t.Fatal("err should be nil here")
		}
		return err
	}()

	if err == nil { // err is not nil here!
		t.Errorf("Error should be nil due to empty interface: %s", err)
	}

	False(t, NoError(mockT, err), "NoError should fail with empty error interface")`

bacuse you actually expect this to be happen here in this testcase.

should i consider this as a feature ?

boyan-soubachov:
> is this scenario intentionally added here.
> 
> ```
> err = func() error {
> 		var err *customError
> 		if err != nil {
> 			t.Fatal("err should be nil here")
> 		}
> 		return err
> 	}()
> 
> 	if err == nil { // err is not nil here!
> 		t.Errorf("Error should be nil due to empty interface: %s", err)
> 	}
> 
> 	False(t, NoError(mockT, err), "NoError should fail with empty error interface")
> ```
> 
> bacuse you actually expect this to be happen here in this testcase.
> 
> should i consider this as a feature ?

@princeparmar , correct. That check is explicitly there because we also want to be able to test structs that conforms to the `error` interface in `NoError()`.

In your example, you're casting a `(*Error)` as `error`, which causes a "check ptr error", which itself becomes an error. This means that you are _in fact_ passing an error to `NoError()`, it's just that it's caused by your typecast.

boyan-soubachov:
I am closing this as it's not an issue with the framework but an incorrect use of it.

PR #891, opened 2020-02-19
test(suite): Assert suite.T compatiblity with gomock

shousper:
As requested 👌


boyan-soubachov:
I see that the tests are now failing. It's most likely due to the fact that we recently reverted the changes to the `suite.T` interface type.

@shousper, would you like to modify this PR to be compatible with those changes or abandon it altogether?

shousper:
Looks like you decided to revert the original breaking change in `master` that triggered my first PR + this PR, so both are redundant now. Closing.

PR #906, opened 2020-03-03
[WIP] Create testify version 2

boyan-soubachov:
This is a proposal to create a version 2 of testify.

This follows the recommended Go approach to major versioning for packages (https://blog.golang.org/v2-go-modules).

The aim of this PR is to start a discussion about the best way forward given how many breaking changes we have accumulated that need to be put into a v2.

@glesica , I would greatly appreciate your opinion and feedback.


boyan-soubachov:
P.S. This should be fully backwards compatible. All current and future package references to `github.com/stretchr/testify` should be working as they are now.

.travis.gotest.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+run_tests() {
+    go test -v -race ./...
+}
+
+run_tests
```
mvdkleijn, line 8: Wouldn't this mean that v1 tests are run when making changes / PRs for v2? That seems like work we shouldn't want to do?

boyan-soubachov: It does, that's a very good point. My thought was that as a first iteration we just run tests across the board and if we find that they're slowing down builds too much, we can then optimise it to run only tests on what has changed.

A big positive I can see with the current scenario is that it would prevent accidental regressions being backported into v1

boyan-soubachov:
> In general it LGTM. I did notice that some code from v1 is not in v2 (yet) so we'll probably need to decide how to do that.
> 
> Also a good point to think about: how to minimize maintenance overhead. This way, though recommended, is basically duplicating code which causes maintenance headaches if we want to keep v1 supported.

Valid point; my thought there was that once we're ready to release v2, we copy over all the logic from v1 into the `/v2` folder and merge. From that point on, v2 is officially out.

Further work would then only be done in the `/v2` folder and as such we only need to work/maintain in that folder. The only time we'll have to maintain 2 copies of something is if we want to explicitly backport it to v1 (usually due to some critical security flaws).

That being said, I think it would make great sense to also have an end-of-life deadline for v1 after which no further support will be provided. This means that all PRs should be in the `/v2` folder/module from that point onwards.

Thoughts?

.travis.gotest.sh
mvdkleijn: That's a fair point. Optimizing this later when needed is not a bad thought. :+1:

mvdkleijn:
> > In general it LGTM. I did notice that some code from v1 is not in v2 (yet) so we'll probably need to decide how to do that.
> > Also a good point to think about: how to minimize maintenance overhead. This way, though recommended, is basically duplicating code which causes maintenance headaches if we want to keep v1 supported.
> 
> Valid point; my thought there was that once we're ready to release v2, we copy over all the logic from v1 into the `/v2` folder and merge. From that point on, v2 is officially out.

OK. I'm assuming here that you mean to do an alpha release first of v2 so we can get some of the big, breaking changes in first before doing a full 2.0.0 release?

> Further work would then only be done in the `/v2` folder and as such we only need to work/maintain in that folder. The only time we'll have to maintain 2 copies of something is if we want to explicitly backport it to v1 (usually due to some critical security flaws).

Agreed, but then we should make that policy clear for users somehow.

> That being said, I think it would make great sense to also have an end-of-life deadline for v1 after which no further support will be provided. This means that all PRs should be in the `/v2` folder/module from that point onwards.
> 
> Thoughts?

Agreed. Having a clear end-of-life moment for older versions is always a good idea. That also means we need to clarify in CONTRIBUTING.md where PRs should be aimed at for certain goals.

One point I'm wondering though... the article you linked to makes some valid points but mentions nothing on cleaning up older EOL module versions? Does that mean we'll end up duplicating code into v3, v4, v5, v6 etc? That seems very painful to me.

boyan-soubachov:
> OK. I'm assuming here that you mean to do an alpha release first of v2 so we can get some of the big, breaking changes in first before doing a full 2.0.0 release?

Not an alpha release, just a cut-over. Let's say we decide that v2 gets released on date X, on that day, we copy over the latest state of v1 into v2 and merge into master.

> Agreed, but then we should make that policy clear for users somehow.

Yup, we could have something in the readme stating that v2 is now the latest version and to recommend its use.
 
> Agreed. Having a clear end-of-life moment for older versions is always a good idea. That also means we need to clarify in CONTRIBUTING.md where PRs should be aimed at for certain goals.

👍 

> One point I'm wondering though... the article you linked to makes some valid points but mentions nothing on cleaning up older EOL module versions? Does that mean we'll end up duplicating code into v3, v4, v5, v6 etc? That seems very painful to me.

We would duplicate the code we want to keep (i.e. if we want to remove deprecated features, we don't copy them over into the new version) and then resume development in the new version. 

I think 99% of the time we'll only be working on one version at a time with no duplications. Only in the special cases where we decide that something is absolutely critical and needs to be backported to a previous version will we need to make a PR that affects both versions (current & previous).

mvdkleijn:
@boyan-soubachov maybe I'm missing something but I'm unclear on how you envision getting the breaking changes into v2.0.0 if you release it immediately... unless we're having a miscommunication :smile:

I guess we could copy over everything to the v2 directory, then add our breaking changes before tagging it as 2.0.0

Otherwise if we'd be adding breaking changes in minor or patch releases, which would kinda defeat the purpose of going to a v2 I guess?

boyan-soubachov:
> @boyan-soubachov maybe I'm missing something but I'm unclear on how you envision getting the breaking changes into v2.0.0 if you release it immediately... unless we're having a miscommunication 😄

I see what you mean.

> I guess we could copy over everything to the v2 directory, then add our breaking changes before tagging it as 2.0.0

This was my thought as well.

> Otherwise if we'd be adding breaking changes in minor or patch releases, which would kinda defeat the purpose of going to a v2 I guess?

It wouldn't be semantic versioning if we were adding breaking changes in minor or patch releases. The code base makes no explicit mention of semantic versioning and whether we follow it, so if we think it best not to, we can follow this route.

One interesting issue to this approach I just thought of is that our git tags will lose the last v1 release version. Since we can only have one 'latest' tag on a branch (`master`) it would be the `v2` tag, so if someone wanted to find out what the latest v1 version is, that would be lost.

I'm not sure if that's a big problem but I think it's actually fine in our case since if we decide we're only maintaining one active version (the latest) of testify. The point then is that v1 solely exists for backwards compatibility and is effectively archived. To me that makes sense over maintaining multiple versions.

boyan-soubachov:
P.S. Another interesting approach I just noticed can be seen here: https://github.com/tealeg/xlsx/blob/master/go.mod

In summary:
No new folder is created for `/v2` but just the package is renamed to `.../v2`. That way, you keep one copy of the code base and the previous versions are still kept by default (as you have to explicitly add `/v2`), not breaking existing use cases.

The downside is that code cannot be backported if urgently needed to.

PR #908, opened 2020-03-03
Add stats to suites and tests

esdrasbeleza:
Hello! After the feedback I got in [my previous PR](https://github.com/stretchr/testify/pull/900), here comes a new one with the suggestions you asked.

In this new PR, if a suite has a callback method called `HandleStats(suiteName string, stats *suite.SuiteInformation)`, `suite.Run` will collect stats for the suite under execution that can be used later to measure test speed, flaky tests etc.

I wrote [an example](https://github.com/esdrasbeleza/testify-stats/blob/master/example/sample_test.go#L20-L33) which creates a log file with information for every test execution. In real life, we could use not only log files, but store `SuiteInformation` in a database for posterior analysis.

I'm not really sure about the names I used for my fields and functions, feedback is always welcome. 😄


boyan-soubachov:
I think you may need to rebase & fast forward given the large number of commits.

esdrasbeleza:
I was trying to fix a mistake in some commit messages, but my git skills are poor after a long day of work. Please delete this PR, I'm creating another one!

boyan-soubachov:
Not a problem, take your time :) Let me know if you need a hand with anything. Worst case, I'd be happy to open the PR if you send me a diff/patch

PR #909, opened 2020-03-03
Add stats to suites and tests

esdrasbeleza:
Hello! After the feedback I got in [my previous PR](https://github.com/stretchr/testify/pull/900), here comes a new one with the suggestions you asked.

In this new PR, if a suite has a callback method called `HandleStats(suiteName string, stats *suite.SuiteInformation)`, `suite.Run` will collect stats for the suite under execution that can be used later to measure test speed, flaky tests etc.

I wrote [an example](https://github.com/esdrasbeleza/testify-stats/blob/master/example/sample_test.go#L20-L33) which creates a log file with information for every test execution. In real life, we could use not only log files, but store `SuiteInformation` in a database for posterior analysis.

I'm not really sure about the names I used for my fields and functions, feedback is always welcome. 😄


suite/interfaces.go
```diff
@@ -1,6 +1,8 @@
 package suite
 
-import "testing"
+import (
```
boyan-soubachov: To maintain consistency, wouldn't it be better to remove the parenthesis when we're importing just one package?

esdrasbeleza: I think `goimports` added it automatically here, sorry! Fixing.

esdrasbeleza:
Any opinions about the structs, method names etc.? Even though they serve my purpose, I'm not 100% sure about the naming! :)

suite/suite.go
```diff
@@ -122,16 +146,34 @@ func Run(t *testing.T, suite TestingSuite) {
...
+					stats.start(method.Name)
+				}
+
 				defer func() {
+					if stats != nil {
+						passed := !t.Failed()
+
+						stats.end(method.Name, passed)
+
+						if !passed {
```
boyan-soubachov: If you change the `Passed` struct entry to a derived function instead of bool, this code would become redundant.

suite/suite.go
```diff
@@ -87,6 +88,14 @@ func Run(t *testing.T, suite TestingSuite) {
 
 	suiteSetupDone := false
 
+	var (
+		stats *SuiteInformation
+	)
+
+	if _, measureStats := suite.(WithStats); measureStats {
```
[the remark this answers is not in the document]
esdrasbeleza: Done

suite/suite.go
```diff
@@ -87,6 +88,14 @@ func Run(t *testing.T, suite TestingSuite) {
 
 	suiteSetupDone := false
 
+	var (
```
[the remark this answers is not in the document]
esdrasbeleza: Done

boyan-soubachov:
The new changes look good. Thank you :)

@glesica, any opinions & thoughts?

PR #914, opened 2020-03-09
Test for #913

### LINKED ISSUE #828: Using error in Contains fails with confusing message
https://play.golang.org/p/u7Dfo4UnxZ0
```go
assert.Contains(t, "error string", errors.New("error"))
```
```
Error Trace:	prog.go:25
Error:      	"error string" does not contain "error"
Test:       	TestCaseA
```

Whereas this is what's needed to pass:
```diff
-assert.Contains(t, "error string", errors.New("error"))
+assert.Contains(t, "error string", errors.New("error").Error())
```

I would prefer if the failure message explained that it was an error type, otherwise it looks like a string does not contain another string: I've fallen foul to this in the past, and just spent 5-10mins again not realising that I needed to pass a `string` type as the 3rd argument.

It would be really nice to have the type printed here when the value isn't a `Stringer` =)

henrahmagix:
Fixes #828

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -760,10 +760,10 @@ func Contains(t TestingT, s, contains interface{}, msgAndArgs ...interface{}) bo
 
 	ok, found := includeElement(s, contains)
 	if !ok {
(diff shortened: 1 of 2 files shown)


boyan-soubachov:
Thank you for your contribution!

PR #920, opened 2020-04-04
Add HTTPStatusCode assertion

### LINKED ISSUE #824: Adding HTTPStatusCodeEqual to assert if status code equals to expected value
Currently we have [`HTTPError`](https://godoc.org/github.com/stretchr/testify/assert#HTTPError) which check if the http response contains status code of 400 and above. I am wordering if we can add another function to check for specific status code like something below:

```go
func HTTPStatusCodeEqual(t TestingT, handler http.HandlerFunc, method, url string, values url.Values, statusCode int, msgAndArgs ...interface{}) bool
```

I can send a PR if you think this is something that can be included in the package. Thank you.

mvdkleijn:
@glesica @boyan-soubachov  Please consider this PR for merging.

**Motivation**
The current set of HTTP asserts do not cover the full range of status codes. This addition would allow a user to test for a specific status code.

As such it would cover the full range of status codes, including for example: `http.StatusSwitchingProtocols`.

**Example usage**
`assert.HTTPStatusCode(t, myHandler, "GET", "/", nil, http.StatusSwitchingProtocols)`

**Whether you intend to add / change behaviour or fix a bug**
Adds behaviour.

**Additionally**
- Would also close #824 
- Includes test
- Tests ran successfully

### DIFF
--- a/assert/http_assertions.go
+++ b/assert/http_assertions.go
@@ -90,6 +90,29 @@ func HTTPError(t TestingT, handler http.HandlerFunc, method, url string, values 
 	return isErrorCode
 }
 
+// HTTPStatusCode asserts that a specified handler returns a specified status code.
+//
+//  assert.HTTPStatusCode(t, myHandler, "GET", "/notImplemented", nil, 501)
+//
+// Returns whether the assertion was successful (true) or not (false).
+func HTTPStatusCode(t TestingT, handler http.HandlerFunc, method, url string, values url.Values, statuscode int, msgAndArgs ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	code, err := httpCode(handler, method, url, values)
+	if err != nil {
+		Fail(t, fmt.Sprintf("Failed to build test request, got error: %s", err))
+		return false
+	}
+
+	successful := code == statuscode
+	if !successful {
+		Fail(t, fmt.Sprintf("Expected HTTP status code %d for %q but received %d", statuscode, url+"?"+values.Encode(), code))
+	}
+
+	return successful
+}
+
 // HTTPBody is a helper that returns HTTP body of the response. It returns
 // empty string if building a new request fails.
 func HTTPBody(handler http.HandlerFunc, method, url string, values url.Values) string {
--- a/assert/http_assertions.go
+++ b/assert/http_assertions.go
@@ -33,7 +33,6 @@ func HTTPSuccess(t TestingT, handler http.HandlerFunc, method, url string, value
 	code, err := httpCode(handler, method, url, values)
 	if err != nil {
 		Fail(t, fmt.Sprintf("Failed to build test request, got error: %s", err))
-		return false
 	}
 
 	isSuccessCode := code >= http.StatusOK && code <= http.StatusPartialContent
@@ -56,7 +55,6 @@ func HTTPRedirect(t TestingT, handler http.HandlerFunc, method, url string, valu
 	code, err := httpCode(handler, method, url, values)
 	if err != nil {
 		Fail(t, fmt.Sprintf("Failed to build test request, got error: %s", err))
-		return false
 	}
 
 	isRedirectCode := code >= http.StatusMultipleChoices && code <= http.StatusTemporaryRedirect
@@ -79,7 +77,6 @@ func HTTPError(t TestingT, handler http.HandlerFunc, method, url string, values 
 	code, err := httpCode(handler, method, url, values)
 	if err != nil {
 		Fail(t, fmt.Sprintf("Failed to build test request, got error: %s", err))
-		return false
 	}
 
 	isErrorCode := code >= http.StatusBadRequest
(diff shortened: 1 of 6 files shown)


mvdkleijn:
Sorry for the extraneous commits, I'm fairly new at this. I believe this last commit will fix everything that was missing.

assert/http_assertions.go
```diff
@@ -90,6 +90,29 @@ func HTTPError(t TestingT, handler http.HandlerFunc, method, url string, values
...
+// Returns whether the assertion was successful (true) or not (false).
+func HTTPStatusCode(t TestingT, handler http.HandlerFunc, method, url string, values url.Values, statuscode int, msgAndArgs ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	code, err := httpCode(handler, method, url, values)
+	if err != nil {
+		Fail(t, fmt.Sprintf("Failed to build test request, got error: %s", err))
+		return false
+	}
```
boyan-soubachov, line 106: This logic is common to this function and at least the `HTTPRedirect`, `HTTPSuccess` and the `HTTPError` functions. Would it not make sense to extract them into one function and the reference them in there?

assert/http_assertions.go
```diff
@@ -90,6 +90,29 @@ func HTTPError(t TestingT, handler http.HandlerFunc, method, url string, values
 	return isErrorCode
 }
 
+// HTTPStatusCode asserts that a specified handler returns a specified status code.
+//
+//  assert.HTTPStatusCode(t, myHandler, "GET", "/notImplemented", nil, 501)
+//
+// Returns whether the assertion was successful (true) or not (false).
+func HTTPStatusCode(t TestingT, handler http.HandlerFunc, method, url string, values url.Values, statuscode int, msgAndArgs ...interface{}) bool {
```
boyan-soubachov, line 98: Would it not make sense to change the `method` argument (in both this and the other functions) into an enum? This would break backwards compatibility so it would have to wait till v2 but in my opinion makes a lot more sense as there are only a finite amount of verbs supported in HTTP standards.

assert/http_assertions.go
```diff
@@ -90,6 +90,29 @@ func HTTPError(t TestingT, handler http.HandlerFunc, method, url string, values
...
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	code, err := httpCode(handler, method, url, values)
+	if err != nil {
+		Fail(t, fmt.Sprintf("Failed to build test request, got error: %s", err))
+		return false
```
boyan-soubachov, line 105: We don't need to `return false` on a fail if I'm not mistaken. The `Fail` automatically breaks out of the calling function and stops execution

assert/http_assertions.go
mvdkleijn: The comment you made makes sense to me for a v2... however, that feels like a form of scope creep to me?

May I suggest a middle ground for this?

My suggestion would be to merge this PR so people can start using it.

I can make a new PR with your change that can be labelled v2 or something or merged into a v2 branch. (didn't check if there already is one)

assert/http_assertions.go
mvdkleijn: I'm away from my laptop right now, but I'll check thia later and rip out the return false if needed.

assert/http_assertions.go
mvdkleijn: Will take a look.

assert/http_assertions.go
mvdkleijn: I checked and it appears the return false is indeed not needed. I've removed it.

assert/http_assertions.go
mvdkleijn: @boyan-soubachov I looked at this and I agree this doesn't feel very DRY. However, wrapping the call to httpCode and the error handling into a separate function that has TestingT as an argument feel kinda silly to me... maybe that's because I'm new to Go, dunno.

I could handle the error checking in httpCode itself, but that would also require adding TestingT as an argument since we'd need it to Fail().

As for the h.Helper() code... I'm not entirely clear on what it does but that code's all over the place in Testify, so that makes no sense for me to extract from these particular functions.

I'd say leave it as it is now or move the error checking into httpCode()... please let me know what you want to do.

assert/http_assertions.go
boyan-soubachov: Good point. It would make this PR too big and lose focus. 👍

assert/http_assertions.go
boyan-soubachov: Fair enough, having had a second look, I think you're correct. I think it's about as simple as it should be at the moment :)

mvdkleijn:
@boyan-soubachov you're welcome and thanks for your approval.

Are you guys planning on doing a v2 branch by the way? Didn't see anything that looked like it.

mvdkleijn:
By the way, totally off topic here so feel free to ignore this... the number of maintainers is looking a little thin here. If you're interested, I can help out a little by cleaning up the issue list for example?

boyan-soubachov:
> @boyan-soubachov you're welcome and thanks for your approval.
> 
> Are you guys planning on doing a v2 branch by the way? Didn't see anything that looked like it.

We are indeed. One of the submitted proposals is #906 .

> By the way, totally off topic here so feel free to ignore this... the number of maintainers is looking a little thin here. If you're interested, I can help out a little by cleaning up the issue list for example?

We'd love to have more hands on deck to review and help out. @glesica has admin rights to the repo if I'm not mistaken so he should be able to add contributors. Check #testify-dev on Gophers slack (https://invite.slack.golangbridge.org/)

PR #921, opened 2020-04-04
mock: fix the assert message on AssertCalled

alexrocco:
**Motivation:**

Wrong message on AssertCalled when one of the mocked methods are not called

**Example:**

```
package mock

import (
	"testing"
)

type (
	ObjA interface {
		DoA()
	}
	ObjBC interface {
		DoB(name string)
		DoC(name string)
	}
)

type objA struct {
	objBC ObjBC
}

func (o *objA) DoA() {
        // Only calling DoB, never DoC
	o.objBC.DoB("B")
}

type mockObjBC struct {
	Mock
}

func (m *mockObjBC) DoB(name string) {
	m.Called(name)
}

func (m *mockObjBC) DoC(name string) {
	m.Called(name)
}

func Test_AssertMethodMessage(t *testing.T) {
	mockObjBC := new(mockObjBC)

	mockObjBC.On("DoB", "B").Return()
	mockObjBC.On("DoC", "C").Return()

	objA := objA{objBC: mockObjBC}

	objA.DoA()

	mockObjBC.AssertCalled(t,"DoB", "B")
	mockObjBC.AssertCalled(t,"DoC", "C")
}
```
Output:
```
=== RUN   Test_AssertMethodMessage
    Test_AssertMethodMessage: example_test.go:48: 
        	Error Trace:	
        	Error:      	Should have called with given arguments
        	Test:       	Test_AssertMethodMessage
        	Messages:   	Expected "DoC" to have been called with:
        	            	[C]
        	            	but actual calls were:
        	            	        [B]
--- FAIL: Test_AssertMethodMessage (0.00s)
FAIL
```
Method `DoC` was never called.

The validation is right but the message is not filtering for the right method, its getting all the calls.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -500,7 +500,9 @@ func (m *Mock) AssertCalled(t TestingT, methodName string, arguments ...interfac
 	if !m.methodWasCalled(methodName, arguments) {
 		var calledWithArgs []string
 		for _, call := range m.calls() {
-			calledWithArgs = append(calledWithArgs, fmt.Sprintf("%v", call.Arguments))
+			if call.Method == methodName {
+				calledWithArgs = append(calledWithArgs, fmt.Sprintf("%v", call.Arguments))
+			}
 		}
 		if len(calledWithArgs) == 0 {
 			return assert.Fail(t, "Should have called with given arguments",


mock/mock.go
```diff
@@ -500,7 +500,9 @@ func (m *Mock) AssertCalled(t TestingT, methodName string, arguments ...interfac
 	if !m.methodWasCalled(methodName, arguments) {
 		var calledWithArgs []string
 		for _, call := range m.calls() {
-			calledWithArgs = append(calledWithArgs, fmt.Sprintf("%v", call.Arguments))
+			if call.Method == methodName {
```
boyan-soubachov, line 503: Hmm. Would we have false information here if two methods in the call stack shared the same name but from different packages? i.e.

pkg1.FuncA() => pkg2.FuncB() => pkg2.FuncA()

alexrocco: I don't think so, because It would be another mock instance.

boyan-soubachov: Fair enough.

Could we add unit tests for this new logic? I think it would make sense if we at least cover the use-case mentioned in the PR description as well as the same-function name one I mentioned above. Other cases would also be welcome :)

boyan-soubachov (changes requested):

PR #924, opened 2020-04-07
Add GitHub pull request template

boyan-soubachov:
Add a simple pull request template for future contributions to the repository


boyan-soubachov:
> I'd suggest to also add the entries from the CONTRIBUTING.md file that we already have:
> ## Motivation
> ### Example usage (if applicable)
> 
> Does this **add** behaviour, **change** behaviour or **fix** a bug:

That's a good point. Missed that.

PR #926, opened 2020-04-17
mock: improve failure messages for AssertCalled and AssertNotCalled

### LINKED ISSUE #896: AssertCalled makes deep comparison but shallow print
I have a `func Dispatch(something interface{})` which is used by the code I want to test. That code looks something like this:

```
dispatcher.Dispatch(&Something{Name: "foo"})
```

Calling `mockDispatcher.AssertCalled(t, "Dispatch", &Something{Name: "foo"})` passes, as expected.

But when I write a test that calls `mockDispatcher.AssertCalled(t, "Dispatch", &Something{Name: "bar"})` it fails with an error like:

```
        	Messages:   	Expected "Dispatch" to have been called with:
        	            	[0xc00009c800]
        	            	but actual calls were:
        	            	        [0xc000083030]
```

This is misleading, as I thought the problem was that testify does not do a deep comparison (which is wrong, judging by the passing test).

I'd expect the printed message to be a bit more verbose so the user can tell it's a problem with the values referenced by the pointer, not the pointer itself.

AdRiley:
## Summary
Fixes issue 896 by doing deep printing arguments for failures of AssertCalled and AssertNotCalled.

## Changes
Extract formatArguments from callString.
Use new method in AssertCalled and AssertNotCalled

I haven't added tests, but let me know if you think I need any.

## Motivation
Fixes issue 896

## Related issues
Closes #896

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -300,15 +300,19 @@ func (m *Mock) findClosestCall(method string, arguments ...interface{}) (*Call, 
 	return closestCall, err
 }
 
+func formatArguments(arguments Arguments) string {
+	var argVals []string
+	for argIndex, arg := range arguments {
+		argVals = append(argVals, fmt.Sprintf("%d: %#v", argIndex, arg))
+	}
+	return fmt.Sprintf("\n\t\t%s", strings.Join(argVals, "\n\t\t"))
+}
+
 func callString(method string, arguments Arguments, includeArgumentValues bool) string {
 
 	var argValsString string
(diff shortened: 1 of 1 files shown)


mvdkleijn (changes requested):
Hi, thanks for your efforts!

It looks good to me but I'd love to have some basic tests around the new function. Though not a huge deal if we don't have them, I'd really like to cover as much as possible... if only for the reason of what Testify itself is. :smile:

Other than that though, looks good.

PR #928, opened 2020-04-19
Add warning in log when no tests are run within a suite

### LINKED ISSUE #862: -testify.m reports nothing if no tests found
Will be great to add some alerting if no tests were found using testify filter.
For example, standard go test displays `testing: warning: no tests to run` at `go test -run unesistingTest`

For now testify works as follow:
```
go test -failfast -v -run Test_SuiteSomeOperation_Execute -testify.m AAAAA
=== RUN   Test_SuiteSomeOperation_Execute
--- PASS: Test_SuiteSomeOperation_Execute (0.00s)
PASS
ok  	github.com/some/path/to/package	0.005s
```

AdRiley:
## Summary
Add a warning message when no tests are run within a suite. Before this change it would just report passed with no indication that nothing was actually run.

## Changes
Added a check that there are tests to actually run and a warning to testing.Log if there are not

## Motivation
See issue 862. It is easy to assume that all your tests have passed when in fact you didn't run any because of perhaps a typo in your regex.

## Related issues
Closes #862

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -179,16 +179,20 @@ func Run(t *testing.T, suite TestingSuite) {
 }
 
 func runTests(t testing.TB, tests []testing.InternalTest) {
-	r, ok := t.(runner)
-	if !ok { // backwards compatibility with Go 1.6 and below
-		if !testing.RunTests(allTestsFilter, tests) {
-			t.Fail()
+	if len(tests) == 0 {
+		t.Log("warning: no tests to run")
+	} else {
+		r, ok := t.(runner)
+		if !ok { // backwards compatibility with Go 1.6 and below
+			if !testing.RunTests(allTestsFilter, tests) {
+				t.Fail()
+			}
+			return
 		}
-		return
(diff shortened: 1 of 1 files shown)


suite/suite.go
```diff
@@ -179,16 +179,20 @@ func Run(t *testing.T, suite TestingSuite) {
 }
 
 func runTests(t testing.TB, tests []testing.InternalTest) {
-	r, ok := t.(runner)
-	if !ok { // backwards compatibility with Go 1.6 and below
-		if !testing.RunTests(allTestsFilter, tests) {
-			t.Fail()
+	if len(tests) == 0 {
```
boyan-soubachov, line 182: Would it make sense to implement this as a guard clause instead? 

Something like:
```
if len(tests) == 0 {
    t.Log(...)
    return
}
```

That way we don't have to indent the rest of the function in an `else` block

AdRiley: Works for me. Done.

PR #929, opened 2020-04-20
Change how tearDownAllSuite is run to aviod deadlock when running with failfast

### LINKED ISSUE #917: TearDownSuite() not working with -failfast flag
To reproduce the bug, run the test below with `-failfast` flag, it will get stuck.

```go
package testsuite

import (
	"testing"

)

type TestSuite struct {
	suite.Suite
}

func (ts *TestSuite) TestA() {
	ts.Require().True(false)
}

func (ts *TestSuite) TestB() {
	ts.Require().True(true)
}

func (ts *TestSuite) TearDownSuite() {
}

func TestTestSuite(t *testing.T) {
	suite.Run(t, new(TestSuite))
}
```
After removing `-failfast` flag or `TearDownSuite` method, the suite fails correctly.

Versions: testify 1.15.1, go 1.13

AdRiley:
## Summary
Fixes a deadlock when there is a failing test in a suite when run with failfast

## Changes
This commit https://github.com/stretchr/testify/commit/0d3c8ce9f8c3be984ccd5d2e3438b173ea722267 attempted to fix an ordering issue with suite teardown, but introduced a deadlock when a test fails when running with failfast.

The problem is that the sync.WaitGroup was being incremented for each test which was expected to be run, but with failfast not all of those tests were actually run. Then the first test is trying to run tearDownAllSuite after itself but waiting on the sync.WaitGroup which never wakes up.

I think we can just move the tearDownAllSuite  into a defer which will run after we have run whatever tests are going to be run, without the need for a sync.WaitGroup at all.

## Motivation
Fixes a defect

## Related issues
Closes #917

### DIFF
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -516,7 +516,7 @@ func TestSuiteWithStats(t *testing.T) {
 	assert.True(t, testStats.Passed)
 }
 
-// Suite to test behavior when running with the failfast flag
+// FailfastSuite will test the behavior when running with the failfast flag
 // It logs calls in the callOrder slice which we then use to assert the correct calls were made
 type FailfastSuite struct {
 	Suite
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -580,7 +580,7 @@ func (s *FailfastSuite) TearDownTest() {
 
 func (s *FailfastSuite) Test_A_Fails() {
 	s.call("Test A Fails")
-	s.Require().True(false)
+	s.T().Error("Test A meant to fail")
 }
 
 func (s *FailfastSuite) Test_B_Passes() {
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -2,6 +2,7 @@ package suite
 
 import (
 	"errors"
+	"flag"
 	"io/ioutil"
 	"math/rand"
 	"os"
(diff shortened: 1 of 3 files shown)


boyan-soubachov:
Hi, do you mind rebasing this and resolving the conflicts so we could have a clear picture of the exact implementation of the change?

AdRiley:
Yes no problem. Done.

AdRiley:
If you have any ideas on how I could replicate failfast from the test suite, it would be great to add a unit test for this, but I couldn't find a way.

mvdkleijn:
@AdRiley You mentioned a possible race on #929 in a comment, did you get that figured out?

AdRiley:
No... It is this build https://travis-ci.org/github/stretchr/testify/builds/677634469 that is bothering me. All pass except Go: Master with GO111MODULE=on. All of the other tests pass and now I have rebased, all of the tests pass. 

With the failing tests it is difficult to see exactly which test fails. Any ideas?

I need to work out how to pull Go master into my dev environment to investigate any further.

AdRiley:
OK I worked out why that build failed and I don't think it is related to my change.

`main.go:27:2: cannot find package "github.com/ernesto-jimenez/gogen/imports" in any of:
	/home/travis/.gimme/versions/go/src/github.com/ernesto-jimenez/gogen/imports (from $GOROOT)
	/home/travis/gopath/src/github.com/ernesto-jimenez/gogen/imports (from $GOPATH)
assert/assertions.go:25: running "sh": exit status 1
main.go:27:2: cannot find package "github.com/ernesto-jimenez/gogen/imports" in any of:
	/home/travis/.gimme/versions/go/src/github.com/ernesto-jimenez/gogen/imports (from $GOROOT)
	/home/travis/gopath/src/github.com/ernesto-jimenez/gogen/imports (from $GOPATH)
require/forward_requirements.go:16: running "sh": exit status 1`

Not sure why it got that error though.

So no race. I just confused myself with tests that pass but output "failed" in the log because they are testing failing tests. I might look if I can make those tests quieter when they pass.

AdRiley:
Agreed. So I've added a unit test, but of course failfast was introduced in Go 1.10 so the test doesn't play nice on Go 1.8 and 1.9 which we still run against on Travis. How do you want handle this?

1. Stop testing against 1.8 and 1.9?
2. Add extra complication to the Travis yml or the unit test?

I'm happy to do either, but just looking for a direction before I do.

boyan-soubachov:
> Agreed. So I've added a unit test, but of course failfast was introduced in Go 1.10 so the test doesn't play nice on Go 1.8 and 1.9 which we still run against on Travis. How do you want handle this?
> 
>     1. Stop testing against 1.8 and 1.9?
> 
>     2. Add extra complication to the Travis yml or the unit test?
> 
> 
> I'm happy to do either, but just looking for a direction before I do.

Thank you.

I'm happy to drop support for 1.8 and 1.9. It's outdated and unsupported by the Go team.

As for the test, is it possible that we call it from within one of the test files rather than Travis? That way we have all tests in one place and it won't be puzzling to future engineers why there's a random test call in the Travis file :)

AdRiley:
Thanks I'll drop 1.8 and 1.9 then.

I can't work out a way to turn on failfast from inside. Any ideas? Looking at https://golang.org/src/testing/testing.go it pulls it out of flags and then uses it. Unless I am able to set flags from inside? I'll take a look and see if that is possible.

boyan-soubachov:
> Thanks I'll drop 1.8 and 1.9 then.
> 
> I can't work out a way to turn on failfast from inside. Any ideas? Looking at https://golang.org/src/testing/testing.go it pulls it out of flags and then uses it. Unless I am able to set flags from inside? I'll take a look and see if that is possible.

Fair enough, I had a rough thought that we could maybe do it by forking into a separate process with `exec.Command()`.

If that proves infeasible however, then we could leave it in the `travis.yml` file as long as we have a comment there. Maybe it would make even more sense to generalise the failfast to run the whole suite instead of that specific test so that we replicate the use-case scenario for our whole package?

AdRiley:
Wondering if I hit this issue in 1.10. https://github.com/golang/go/issues/24412 Will try changing the way that test fails to see if I can work around.

suite/suite_test.go
```diff
@@ -512,3 +515,75 @@ func TestSuiteWithStats(t *testing.T) {
 	assert.NotZero(t, testStats.End)
 	assert.True(t, testStats.Passed)
 }
+
+// Suite to test behavior when running with the failfast flag
```
boyan-soubachov, line 519: One small nit, could we have this docstring follow standard Go convention for exported funcs and types (i.e it starts with the object name, `FailfastSuite`) ?

AdRiley: No problem. Done :)

PR #932, opened 2020-04-26
Fix suite argument matching

boyan-soubachov:
## Summary
Custom test function names for suite, specified by `testify.m` flag, currently do not work correctly. This PR fixes that

## Changes
* Fixed argument matching for custom test function name patterns (i.e. using the 'testify.m' flag).
* Refactored custom test function name matching logic

## Motivation
Non-functioning custom test function names

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -2,8 +2,7 @@ package suite
 
 import (
 	"flag"
-	"fmt"
-	"os"
+	"log"
 	"reflect"
 	"regexp"
 	"runtime/debug"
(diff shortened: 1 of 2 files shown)


suite/suite.go
```diff
@@ -95,22 +87,21 @@ func Run(t *testing.T, suite TestingSuite) {
...
-			os.Exit(1)
+			log.Fatalf("testify: invalid regexp for -m: %s\n", err)
 		}
 
 		if !ok {
 			continue
 		}
 
-		suiteName := methodFinder.Elem().Name()
+		suiteName := suiteInstance.Elem().Name()
```
dolmen, line 104: This must be moved outside of the loop.

Issue #925, opened 2020-04-16
Http Assertions does not allow the creation of a body


ihulsbus:
Hi, 

It seems like the HTTP Assert package does not allow me to specify a body to pass in a POST request for example. Looking at the `HTTPBody()` function, the body argument is set to `nil`: 
`req, err := http.NewRequest(method, url+"?"+values.Encode(), nil)` 

Is this by design, or am I looking at a missing feature? 

Cheers,

mvdkleijn:
We feel that it's reasonable to want to be able to set the body in certain cases. If you'd like to help out, feel free to open a PR.

gohargasparyan:
Hi @mvdkleijn , I just made changes for the issue, tried to push to a new branch and got 403, do I need some permissions to make PR?

mvdkleijn:
> Hi @mvdkleijn , I just made changes for the issue, tried to push to a new branch and got 403, do I need some permissions to make PR?

This sounds like you're new or perhaps I'm misunderstanding? :smile:

The way you create a PR is by forking the repository to your own copy, create a branch with your changes pushed to it and then use the Github UI to create the PR. Basically that will be a request for us to merge changes from your fork into our original copy.

It sounds like you were trying to push to a branch in the main Testify repository which is indeed off limits to all but the maintainers.

Hope that helps!

gohargasparyan:
Yeah first time trying to contribute to open source, thanks, it helps :)

mvdkleijn:
> Yeah first time trying to contribute to open source, thanks, it helps :)

No worries, feel free to ask more if desired. The Github docs also provide a lot of information on how to do things.

Once you have the PR in place, I'll take a look and see if anything needs adjustment and once satisfied all looks good, we can merge it. :+1:

gohargasparyan:
No worries, it was straight forward. https://github.com/stretchr/testify/pull/938

fazypng:
@mvdkleijn hi i need help. very much help. i just joined the Github community yesterday and i don't understand a single thing I've been reading here 😩

mvdkleijn:
> @mvdkleijn hi i need help. very much help. i just joined the Github community yesterday and i don't understand a single thing I've been reading here 😩 

I'm not sure what exactly your question is that I could help with...?

lwlach:
should this issue be closed?

mvdkleijn:
Not really, since the merged PR was reverted by @boyan-soubachov and aa such the issue is still valid.

[...]

PR #938, opened 2020-04-29
allow body for HTTPBodyContains and HTTPBodyNotContains for POST

gohargasparyan:
## Summary
Issue  #925
Http Assertions does not allow the creation of a body

## Changes
* Added body argument in HTTPBodyContains and HTTPBodyNotContains 
* in tests added httpPostHandler which reads request body and writes to response
* In HTTPBody checking if there are url params , otherwise not adding to url

## Related issues
#925

### DIFF
--- a/assert/assertion_forward.go
+++ b/assert/assertion_forward.go
@@ -6,6 +6,7 @@
 package assert
 
 import (
+	"io"
 	http "net/http"
 	url "net/url"
 	time "time"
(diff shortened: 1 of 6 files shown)


assert/http_assertions.go
```diff
@@ -111,9 +112,13 @@ func HTTPStatusCode(t TestingT, handler http.HandlerFunc, method, url string, va
...
 // empty string if building a new request fails.
-func HTTPBody(handler http.HandlerFunc, method, url string, values url.Values) string {
+func HTTPBody(handler http.HandlerFunc, method, url string, values url.Values, body io.Reader) string {
 	w := httptest.NewRecorder()
-	req, err := http.NewRequest(method, url+"?"+values.Encode(), nil)
+
+	if values != nil {
+		url = url + "?" + values.Encode()
+	}
+	req, err := http.NewRequest(method, url, body)
```
boyan-soubachov, line 121: Given that this logic has been added, would it not make sense to add new test cases to check for it?

gohargasparyan: Sure, I will add some tests

dpajkovic:
Would it be possible to give example for the simple GET request in GoDoc/pkg.go.dev documentation? The example given

`assert.HTTPBodyContains(t, myHandler, "GET", "www.google.com", nil, "I'm Feeling Lucky")`

no longer works, and more frustratingly none of the test scripts I've written work without rolling back the package version to 1.5.1.

boyan-soubachov:
@dpajkovic , apologies. I just realised that I should have tagged this for v2 rather than 1.6.0 😞 .

I will revert this change and release a 1.6.1 quickly.

gohargasparyan:
oh, sorry @dpajkovic missed that part, so what happens to this changes now @boyan-soubachov ?

boyan-soubachov:
@gohargasparyan, I guess the following options are available to us:
a) Do this work again but only merge it in v2 as it's a breaking change
b) Implement this as a new function/feature
c) Try and make the change backwards-compatible

My preference (in order) would be: c -> a -> b :)

assert/assertion_forward.go
```diff
@@ -6,6 +6,7 @@
 package assert
 
 import (
+	io "io"
```
dolmen, line 9: Remove the alias

dolmen (commented):
This can't be merged in v1.

And v2 is very far away.

PR #942, opened 2020-05-07
assertion_compare: fix format string for numbers

zjx20:
fix format string for numbers. without this patch, we got something like

```
Error:      	"%!s(int=382)" is not less than "%!s(int=100)"
```

### DIFF
--- a/assert/assertion_compare_test.go
+++ b/assert/assertion_compare_test.go
@@ -1,9 +1,25 @@
 package assert
 
 import (
+	"bytes"
+	"fmt"
 	"reflect"
 	"testing"
 )
+
+type outputT struct {
+	buf *bytes.Buffer
+}
+
+// Implements TestingT
+func (t *outputT) Errorf(format string, args ...interface{}) {
+	s := fmt.Sprintf(format, args...)
+	t.buf.WriteString(s)
+}
+
+func (t *outputT) get() string {
+	return string(t.buf.Bytes())
+}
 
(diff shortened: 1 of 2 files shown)


assert/assertion_compare_test.go
```diff
@@ -1,10 +1,26 @@
...
+// Implements TestingT
+func (t *outputT) Errorf(format string, args ...interface{}) {
+	s := fmt.Sprintf(format, args...)
+	t.buf.WriteString(s)
+}
+
+func (t *outputT) get() string {
+	return string(t.buf.Bytes())
+}
+
```
boyan-soubachov, line 23: Given that we only use this logic for the tests you just added, do you mind moving it closer to them?

assert/assertion_compare_test.go
```diff
@@ -99,21 +143,49 @@ func TestLess(t *testing.T) {
...
+	}
+
+	// Check error report
+	for _, currCase := range []struct {
+		less    interface{}
+		greater interface{}
+		msg     string
+	}{
+		{less: "a", greater: "b", msg: `"b" is not less than or equal to "a"`},
+		{less: 1, greater: 2, msg: `"2" is not less than or equal to "1"`},
```
boyan-soubachov, line 184: Could you also please add some more complex type examples (i.e. floats an bools) to make sure we don't adversely affect them? :)

zjx20: done. FYI bools are not comparable

PR #949, opened 2020-05-22
Allow comparing custom numeric types

torkelrogstad:
## Summary
Allow comparing custom numeric types without panics. 

## Changes
Prior to this commit, this would not work:

```go
type myInt int
assert.Less(t, myInt(1), myInt(2)) // panic!
```

We add some type conversions, similarly to how `ObjectsAreEqualValues`
work. This allows us to compare `myInt` to `myInt`, as well as `myInt`
to `int`.

## Motivation
It's a pain doing `assert.Less(t, int(myCustomIntValue), 8)` every time I want to compare numbers.

### DIFF
--- a/assert/assertion_compare.go
+++ b/assert/assertion_compare.go
@@ -13,12 +13,42 @@ const (
 	compareGreater
 )
 
+var (
+	intType   = reflect.TypeOf(int(1))
+	int8Type  = reflect.TypeOf(int8(1))
+	int16Type = reflect.TypeOf(int16(1))
+	int32Type = reflect.TypeOf(int32(1))
+	int64Type = reflect.TypeOf(int64(1))
+
+	uintType   = reflect.TypeOf(uint(1))
+	uint8Type  = reflect.TypeOf(uint8(1))
+	uint16Type = reflect.TypeOf(uint16(1))
+	uint32Type = reflect.TypeOf(uint32(1))
+	uint64Type = reflect.TypeOf(uint64(1))
+
+	float32Type = reflect.TypeOf(float32(1))
+	float64Type = reflect.TypeOf(float64(1))
+
+	stringType = reflect.TypeOf("")
+)
+
 func compare(obj1, obj2 interface{}, kind reflect.Kind) (CompareType, bool) {
+	obj1Value := reflect.ValueOf(obj1)
+	obj2Value := reflect.ValueOf(obj2)
+
+	// throughout this switch we try and avoid calling .Convert() if possible,
+	// as this has a pretty big performance impact
 	switch kind {
 	case reflect.Int:
 		{
-			intobj1 := obj1.(int)
-			intobj2 := obj2.(int)
+			intobj1, ok := obj1.(int)
+			if !ok {
+				intobj1 = obj1Value.Convert(intType).Interface().(int)
+			}
+			intobj2, ok := obj2.(int)
+			if !ok {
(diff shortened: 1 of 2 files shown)


assert/assertion_compare.go
```diff
@@ -31,8 +51,8 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (CompareType, bool) {
 		}
 	case reflect.Int8:
 		{
-			int8obj1 := obj1.(int8)
-			int8obj2 := obj2.(int8)
+			int8obj1 := obj1Value.Convert(int8Type).Interface().(int8)
```
boyan-soubachov, line 54: Would these calls to `Convert()` not actually modify the type being tested and as such, have side-effects?

torkelrogstad: Consider this snippet: 

```go
type myInt int
assert.Less(t, int(1), myInt(2))
```

In my mind, this should work, and return `true`. Right now it panics. Is your point above that this snippet should not work, because they are different types?

boyan-soubachov: I see what you mean. Makes sense to me then. Just slightly worried that the reflection would have a non-negligible effect on performance. 

Would you mind running a benchmark (using `testing.B`) to compare the before and after on the working (i.e. non-custom type) cases?

torkelrogstad: I'm not much experienced with benchmarks, so LMK if I'm doing something wrong here. But here's an attempt: 

My PR, with TestCompare renamed to BenchmarkCompare, left as-is prior to this PR, except wrapping the body in `for i := 0; i < t.N; i++ { <test here> }`:

```
Running tool: /Users/torkel/go/go1.14.1/bin/go test -benchmem -run=^$ github.com/stretchr/testify/assert -bench ^(BenchmarkCompare)$

goos: darwin
goarch: amd64
pkg: github.com/stretchr/testify/assert
BenchmarkCompare-8   	  437868	      2691 ns/op	     312 B/op	      66 allocs/op
PASS
ok  	github.com/stretchr/testify/assert	1.317
```

Master: 
```
Running tool: /Users/torkel/go/go1.14.1/bin/go test -benchmem -run=^$ github.com/stretchr/testify/assert -bench ^(BenchmarkCompare)$

goos: darwin
goarch: amd64
pkg: github.com/stretchr/testify/assert
BenchmarkCompare-8   	 4445959	       282 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/stretchr/testify/assert	1.880s
```

That's a quite unfortunate difference in performance... Might be possible to optimize this a bit? I'll take a look

torkelrogstad:
Changed this up a bit: we only do the `Convert` call if we're given types that need conversion. We also make it possible to convert string types. 

Benchmarks with most recent changes: 

```
BenchmarkCompare
BenchmarkCompare-8   	 2755450	       426 ns/op
PASS
Process finished with exit code 0
```

Still a hit compared to master, but I think it's worth it. I don't believe this code path is the most critical one?

torkelrogstad:
Any chance of getting this in?

boyan-soubachov:
> Any chance of getting this in?

Having a look and will merge if all is good :)

boyan-soubachov:
Looks great. Thanks for optimizing with the branching logic. This way, current uses (i.e. non-custom types) should experience a negligible performance impact.

Issue #640, opened 2018-07-23
Add NotEqualValues assertion and FailNowDiff


0xdevalias:
Currently there is an `EqualValues` assertion, which seems good for checking structs are the same. But if I want to test the negative case (`NotEqualValues`), there doesn't seem to be a good method. It would be basically the same as the `EqualValues`, though just changing the inner check:

**From:**
```
func EqualValues(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool {
..snip..
if !ObjectsAreEqualValues(expected, actual) {
..snip..
```

**To:**
```
func NotEqualValues(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool {
..snip..
if ObjectsAreEqualValues(expected, actual) {
..snip..
```

Perhaps a refactor to share the common code in an unexported method with a flag on the test case would be better too.

I can make my own using `ObjectsAreEqual` and `Condition`, but then I don't get the niceness of the diff format in the output, which I can't access to make my own similar (without copying code) as the diff functions aren't exported.

It would also be nice if there was a wrapper function to access the diff features, perhaps something like (based on the error diff code in `EqualValues`:

```
func FailNowDiff(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool {
	diff := diff(expected, actual)
	expected, actual = formatUnequalValues(expected, actual)
	return Fail(t, fmt.Sprintf("Not equal: \n"+
		"expected: %s\n"+
		"actual  : %s%s", expected, actual, diff), msgAndArgs...)
}
```

mvdkleijn:
Having something like `NotEqualValues` makes sense to me. A PR would be welcome I think.

PR #953, opened 2020-05-22
Issue640 add NotEqualValues()

iivvoo:
## Summary
This PR implements NotEqualValues(), as requested (partially) in #640

## Changes
Changes consist of adding the implementation of NotEqualValues(), similar to how currently EqualValues() and NotEquals() are implemented.

## Motivation
When asserting that values do not equal using `NotEquals`, the assertion will already be true if types differ. This can happen easily when using numeric literals, e.g.

```
assert.NotEquals(t, 10, SomeFuncReturningUInt())
```
will hold even if the func returns 10 since the type assigned by the compiler / returned by the func differ.

## Related issues
Implements #640  in part.

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -222,6 +222,10 @@ func TestEqual(t *testing.T) {
 	if Equal(mockT, myType("1"), myType("2")) {
 		t.Error("Equal should return false")
 	}
+	// A case that might be confusing, especially with numeric literals
+	if Equal(mockT, 10, uint(10)) {
+		t.Error("Equal should return false")
+	}
 }
 
 func ptr(i int) *int {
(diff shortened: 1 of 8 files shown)


assert/assertions_test.go
```diff
@@ -544,6 +548,70 @@ func TestNotEqual(t *testing.T) {
...
+	}
+	if !NotEqualValues(mockT, int(10), int(11)) {
+		t.Error("NotEqualValues should return true")
+	}
+	if NotEqualValues(mockT, int(10), uint(10)) {
+		t.Error("NotEqualValues should return false")
+	}
+	if NotEqualValues(mockT, struct{}{}, struct{}{}) {
+		t.Error("NotEqualValues should return false")
+	}
```
boyan-soubachov, line 614: This is not a blocker, but would these tests be neater if we converted them to table-driven tests? (e.g. https://dave.cheney.net/2019/05/07/prefer-table-driven-tests)

Happy to leave it as your call if you want to change them.

iivvoo: I agree - I just copied the testing patterns used elsewhere (and partially just copied the `NotEqual` tests), and I tend to overengineer things :)

And if I would improve it here, why not improve other tests as well, etc.

But I'll give it a shot - expect an update later today.

boyan-soubachov: If you would like to have a shot at converting other cases to table-driven tests, we'd greatly appreciate the time and effort. Just a key point, as you mentioned as well, is to keep them in separate PRs. It makes the review process much easier and quick.

If you would like, I'd be happy to merge this PR and you can refactor the tests in separate PRs?

iivvoo: Let's do that then. And I will try to refactor at least some tests in a separate PR

PR #955, opened 2020-05-27
Refactor unit tests to table-driven tests

iivvoo:
## Summary

This refactors some of the tests but does not add any new ones.

## Changes

- define test cases in tables where sensible
- combine testcases for `Foo()` and `NotFoo()`, specifically there were testcases were already the same
- make some types local to funcs if not used elsewhere

## Motivation

- table based tests are the way to go.

## Related issues

As discussed in #953

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -594,6 +594,7 @@ func TestContainsNotContains(t *testing.T) {
 			}
 		}
 	})
+
 	t.Run("TestNotContains", func(t *testing.T) {
 		mockT := new(testing.T)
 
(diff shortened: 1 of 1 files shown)


assert/assertions_test.go
```diff
@@ -182,49 +177,37 @@ func TestIsType(t *testing.T) {
 
 }
 
-type myType string
```
iivvoo, line 185: I don't like to pollute the top namespace if not necessary

assert/assertions_test.go
```diff
@@ -182,49 +177,37 @@ func TestIsType(t *testing.T) {
...
+	cases := []TestCase{
+		{"Hello World", "Hello World", true},
+		{123, 123, true},
+		{123.5, 123.5, true},
+		{[]byte("Hello World"), []byte("Hello World"), true},
+		{nil, nil, true},
+		{int32(123), int32(123), true},
+		{uint64(123), uint64(123), true},
+		{myType("1"), myType("1"), true},
+		{&struct{}{}, &struct{}{}, true}, // pointer equality is based on equality of underlying value
```
iivvoo, line 195: This is one case where there was a specific message about the test. Not sure what to do with that (though I do add testcase specific messages further down below)

assert/assertions_test.go
```diff
@@ -672,91 +625,56 @@ func TestContainsFailMessage(t *testing.T) {
...
-	}
-	if NotContains(mockT, list, "Foo") {
-		t.Error("NotContains should return false: \"[\"Foo\", \"Bar\"]\" contains \"Foo\"")
-	}
-	if NotContains(mockT, simpleMap, "Foo") {
-		t.Error("Contains should return true: \"{\"Foo\": \"Bar\"}\" contains \"Foo\"")
-	}
-	if !NotContains(mockT, simpleMap, "Bar") {
-		t.Error("Contains should return false: \"{\"Foo\": \"Bar\"}\" does not contains \"Bar\"")
+func TestSubsetNotSubset(t *testing.T) {
```
iivvoo, line 628: In this case, in my opinion, it makes sense to combine TestSubset and TestNotSubset with the same set of cases to avoid duplocation. Also, if a case ever gets added we get it for free for the opposite test.

assert/assertions_test.go
```diff
@@ -482,146 +465,101 @@ func TestExactly(t *testing.T) {
...
+		res := NotEqualValues(mockT, c.expected, c.actual)
+
+		if res != c.result {
+			t.Errorf("NotEqualValues(%#v, %#v) should return %#v", c.expected, c.actual, c.result)
+		}
+	}
 }
 
-func TestContains(t *testing.T) {
+func TestContainsNotContains(t *testing.T) {
```
iivvoo, line 556: In this case, in my opinion, it makes sense to combine TestContains and TestNoContains with the same set of cases to avoid duplocation. Also, if a case ever gets added we get it for free for the opposite test.

assert/assertions_test.go
```diff
@@ -630,34 +568,49 @@ func TestContains(t *testing.T) {
...
-	}
-	if Contains(mockT, "Hello World", "Salut") {
-		t.Error("Contains should return false: \"Hello World\" does not contain \"Salut\"")
-	}
+	cases := []TestCase{
+		{"Hello World", "Hello", true},
+		{"Hello World", "Salut", false},
+		{list, "Bar", true},
+		{list, "Salut", false},
+		{complexList, &A{"g", "h"}, true},
```
iivvoo, line 576: When failing, the error is slighly more cryptic than the explicit strings originally printed, since %#v doesn't fully print the complexList and its contents

assert/assertions_test.go
```diff
@@ -100,47 +100,42 @@ func (a *AssertionTesterConformingObject) TestMethod() {
 type AssertionTesterNonConformingObject struct {
 }
 
-func TestObjectsAreEqual(t *testing.T) {
+// TestCase holds the expected/actual values to be passed to most checks and their expected result (true/false)
+type TestCase struct {
+	expected interface{}
+	actual   interface{}
+	result   bool
+}
```
boyan-soubachov, line 108: I'm not sure this provides much value as it's incredibly abstract. It basically says 'a test case can expect anything, compared with anything, and the result can be success or fail' :)

In my opinion it's better to just do them on an individual basis (i.e. per-test) and actually have them be specific to the function we're testing. Thoughts?

assert/assertions_test.go
boyan-soubachov: Agreed, good one.

assert/assertions_test.go
boyan-soubachov: I think we can handle that if we define the `TestCase` struct on a per-test level rather than package level, as mentioned in the comment above.

assert/assertions_test.go
boyan-soubachov: [This](https://github.com/golang/go/wiki/TableDrivenTests#example-of-a-table-driven-test) tends to be the _de facto_ template for table-driven tests in Go. In my opinion we should try and stick to it :)

assert/assertions_test.go
boyan-soubachov: Agreed.

assert/assertions_test.go
```diff
@@ -630,34 +568,49 @@ func TestContains(t *testing.T) {
...
+			// Not that we negate the result, since we expect the opposite
+			if res != !c.result {
+				if res {
+					t.Errorf("NotContains(%#v, %#v) should return true:\n\t%#v does not contains %#v", c.expected, c.actual, c.expected, c.actual)
+				} else {
+					t.Errorf("NotContains(%#v, %#v) should return false:\n\t%#v contains %#v", c.expected, c.actual, c.expected, c.actual)
+				}
+			}
+		}
+	})
```
boyan-soubachov, line 613: I think you could actually go one step further and solidify the expected behaviour that `Contains() == !Contains()` in the test above by removing this.

E.g.

Remove these lines and on line 587 assert that `res == !NotContains(...)`.

That way you'll test that `NotContains()` is always the boolean opposite of `Contains()` which is, by definition, its behaviour.

assert/assertions_test.go
boyan-soubachov: Agreed, I think we can do the same as above with `Contains()` and `NotContains()`

assert/assertions_test.go
iivvoo: It's modeled after the signature of most asserts - they take an expected, an actual and return a (bool) result.

I'm fine with defining them on  per assert-test basis but it will still look mostly the same (except that you don't have to provide a name for the struct anymore - it can be entirely inline)

Also, following the "de facto template" looks good, but that change is mosty about wrapping the test in a t.Run() (which would probably need a descriptive string per test to properly name the func)

assert/assertions_test.go
iivvoo: Good point, will look into it.

iivvoo:
- Global TestCase struct removed (now local, anonymous in test functions
- Rewrote testcase loop to t.Run() individual tests
- Asserted that Foo != NotFoo

PR #963, opened 2020-06-12
assert: replace type assertions with reflect.ValueOf()

### LINKED ISSUE #780: Cannot compare order of time.Duration
`Less` and `Greater` asserts (and their variants) use reflection and type assertion to later get the value. `time.Duration` is an alias for `int64`, but type assertion for `int64` fails. 

`reflect.ValueOf(e1).Int()` should be used instead. It would be even better to add a special case for `time.Duration` before falling back to reflection to have nicer error formatting since `fmt.Sprintf` handles `time.Duration` very well.

Pr0Ger:
## Summary
Replace type assertions with `reflect.ValueOf(...).Int()` to fix comparison for aliased types

## Changes
- `obj1.(int)` => `reflect.ValueOf(obj1).Int()`
- Reduce number of cases since we can handle different types in one block
- Add tests for comparing order of `time.Duration` (alias of `int64`)
## Motivation
From `reflect` perspective both `int64` and `time.Duration` (or any other aliased type) have type `reflect.Int64` but type assert will not work because it's different types. This PR supposed to fix this.

```go
assert.Greater(t, time.Second, time.Millisecond)
...
panic: interface conversion: interface {} is time.Duration, not int64
```

## Related issues
Fixes #780

### DIFF
--- a/assert/assertion_compare.go
+++ b/assert/assertion_compare.go
@@ -15,10 +15,10 @@ const (
 
 func compare(obj1, obj2 interface{}, kind reflect.Kind) (CompareType, bool) {
 	switch kind {
-	case reflect.Int:
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
 		{
-			intobj1 := obj1.(int)
-			intobj2 := obj2.(int)
+			intobj1 := reflect.ValueOf(obj1).Int()
+			intobj2 := reflect.ValueOf(obj2).Int()
 			if intobj1 > intobj2 {
 				return compareGreater, true
 			}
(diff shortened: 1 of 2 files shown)


mvdkleijn:
@boyan-soubachov this should go into 2.0 right?

boyan-soubachov:
I think I may have wrongly labelled this change. Having a second look, it doesn't actually look like a breaking change. My only concern is given that it's a hot path in the code, the reflection would most likely result in a significant performance hit.

Would it not be a good idea to add some `testing.TB` style benchmarks to this PR so that we can compare?

mvdkleijn:
> Would it not be a good idea to add some `testing.TB` style benchmarks to this PR so that we can compare?

Agreed.

Pr0Ger:
I think it's unlikely two more reflection calls will cause a significant performance hit since this code already uses reflection to determine type of objects.

Anyway, I'm not sure if this PR should be merged now since 51595dcf940cfb09eb91b6b78babba78bb2bfb25 fixed comparison of aliased types so feel free to close it. Otherwise I can resolve conflicts and add some benchmarks for this code

boyan-soubachov:
> I think it's unlikely two more reflection calls will cause a significant performance hit since this code already uses reflection to determine type of objects.
> 
> Anyway, I'm not sure if this PR should be merged now since [51595dc](https://github.com/stretchr/testify/commit/51595dcf940cfb09eb91b6b78babba78bb2bfb25) fixed comparison of aliased types so feel free to close it. Otherwise I can resolve conflicts and add some benchmarks for this code

Probably not but we'd still like to have data to be sure of the impact :) 

Up to you, but the way the switch has been cleaned up seems like a very useful thing to merge, even if we leave out any fixes and new features.

PR #965, opened 2020-06-16
mock: get call count in thread safe way

### LINKED ISSUE #964: mock - get call count/list in thread safe way
**What**
Currently retrieving the calls via `mock.Calls` results in a data race condition when the test is using multiple go routines.

**Expected**
A thread-safe way to obtain the call count/list.

**Use Case**
A poller object uses a mocked client that is called in an infinite polling loop until a private flag is set on the poller object to exit it. Testing that upon setting the private flag, the infinite loop is no longer invoking the mocked client's method.
This would be trivial to do if we'd have access to the call-count after the flag is set, and if the call count stays the same after 1 second.

nbaztec:
## Summary
Allows returning number of calls to a specific method in a thread safe way.

## Changes
Adds a new method on the `Mock` object that returns the call count.

## Motivation
There was a need to get the call count information on a given mock. Accessing the `Calls` member leads to a race condition as the lock is neglected. 

```go
type client interface {
	Do(x int)
}

type owner struct {
	c client
}

func (o owner) Start() chan struct{} {
	done := make(chan struct{})
	go func() {
		for {
			select {
			case <-done:
				return
			case <-time.After(2 * time.Millisecond):
				o.c.Do(1)
			}
		}
	}()

	return done
}

type clientMock struct {
	mock.Mock
}

func (m *clientMock) Do(x int) {
	m.Called(x)
}

func TestOwnerCloses(t *testing.T) {
	m := &clientMock{}
	m.On("Do", mock.AnythingOfType("int"))
	defer m.AssertExpectations(t)

	o := owner{c: m}
	done := o.Start()
	time.Sleep(5*time.Millisecond)	// wait for some time, the client gets called finite amount of times
	close(done)
	time.Sleep(2*time.Millisecond)	// allow closing

	expected := m.NumberOfCalls("Do")
	time.Sleep(10*time.Millisecond)	// test if the client was called during the time
	actual := m.NumberOfCalls("Do")

	assert.Equal(t, expected, actual)
}
```

## Related issues
Closes #964

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -578,6 +578,20 @@ func (m *Mock) IsMethodCallable(t TestingT, methodName string, arguments ...inte
 	return false
 }
 
+
+// NumberOfCalls returns the number of times a method was called
+func (m *Mock) NumberOfCalls(methodName string) int {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	var actualCalls int
+	for _, call := range m.calls() {
+		if call.Method == methodName {
+			actualCalls++
+		}
+	}
+	return actualCalls
+}
+
 // isArgsEqual compares arguments
 func isArgsEqual(expected Arguments, args []interface{}) bool {
 	if len(expected) != len(args) {
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -578,7 +578,6 @@ func (m *Mock) IsMethodCallable(t TestingT, methodName string, arguments ...inte
 	return false
 }
 
-
 // NumberOfCalls returns the number of times a method was called
 func (m *Mock) NumberOfCalls(methodName string) int {
 	m.mutex.Lock()
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -1476,6 +1476,17 @@ func Test_MockReturnAndCalledConcurrent(t *testing.T) {
 	wg.Wait()
 }
 
+func Test_Mock_NumberOfCalls(t *testing.T) {
+	var mockedService = new(TestExampleImplementation)
+	mockedService.On("Test_Mock_NumberOfCalls", 1, 2, 3).Return(5, 6, 7)
+
+	mockedService.Called(1, 2, 3)
+	assert.Equal(t, mockedService.NumberOfCalls("Test_Mock_NumberOfCalls"), 1)
+	mockedService.Called(1, 2, 3)
+
+	assert.Equal(t, mockedService.NumberOfCalls("Test_Mock_NumberOfCalls"), 2)
+}
+
 type timer struct{ Mock }
 
 func (s *timer) GetTime(i int) string {
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -1124,37 +1124,6 @@ func Test_Mock_AssertCalled_WithAnythingOfTypeArgument(t *testing.T) {
 
 }
 
-func Test_Mock_AssertNumberOfCallsIsThreadSafe(t *testing.T) {
-
-	var mockedService = new(TestExampleImplementation)
-
-	mockedService.On("TheExampleMethod", 1, 2, 3).Return(5, 6, 7)
-
-	done := make(chan struct{})
-	shutdown := func() {
-		time.Sleep(10 * time.Millisecond)
-		close(done)
-		time.Sleep(10 * time.Millisecond)
-	}
-
-	go func() {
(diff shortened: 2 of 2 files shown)


mock/mock.go
```diff
@@ -578,6 +578,19 @@ func (m *Mock) IsMethodCallable(t TestingT, methodName string, arguments ...inte
 	return false
 }
 
+// NumberOfCalls returns the number of times a method was called
+func (m *Mock) NumberOfCalls(methodName string) int {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+	var actualCalls int
+	for _, call := range m.calls() {
```
boyan-soubachov, line 586: Please move the mutex locking right before the start of the critical section (i.e. before this line)

mock/mock.go
```diff
@@ -578,6 +578,19 @@ func (m *Mock) IsMethodCallable(t TestingT, methodName string, arguments ...inte
 	return false
 }
 
+// NumberOfCalls returns the number of times a method was called
+func (m *Mock) NumberOfCalls(methodName string) int {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
```
boyan-soubachov, line 584: Would it not make sense to move this unlock to right after the critical section (i.e. between lines 590,591)? My thought is that it being explicitly there will help the critical section unnecessarily ballooning by accident.

boyan-soubachov (changes requested):

mock/mock.go
nbaztec: I'd ideally go with `defer Unlock` since it's more "failsafe" in case the function body panics, however in this case, I don't think the critical section will panic at all. As such the result is identical the `Unlock` can be moved to end. Though, as long as we're also aware of the fact that any newly introduced code prone to `panic` will introduce a potential bug in future.

thoughts?

boyan-soubachov:
That’s a very good point. Fine to leave it as is then 🙂

mvdkleijn (dismissed):
LGTM

mvdkleijn:
@boyan-soubachov If it looks good to you too, feel free to merge

mock/mock_test.go
```diff
@@ -1476,6 +1476,17 @@ func Test_MockReturnAndCalledConcurrent(t *testing.T) {
 	wg.Wait()
 }
 
+func Test_Mock_NumberOfCalls(t *testing.T) {
```
boyan-soubachov, line 1479: Could you add a test which covers the issue you fixed. I would imagine it would fail when run with `go test -race` and _without_ the fix you've added in this PR?

boyan-soubachov:
> @boyan-soubachov If it looks good to you too, feel free to merge

The code looks good to me but I think we should have a test that demonstrates the issue (without the fix) and how it's fixed.

mock/mock_test.go
nbaztec: Sure, added the test to demonstrate data race. 

Commenting out the mutex yields a data race with `go test -v -race -run Test_Mock_NumberOfCallsIsThreadSafe ./...`

boyan-soubachov (approved):
Thank you for your time and effort. @mvdkleijn , thoughts?

mvdkleijn (approved):
lgtm

nbaztec:
Is this safe to get merged?

ubaltaci:
Can we merge this? It seems impl agreed and tests there, also `NumberOfCalls` method itself should be nice addition to API?

mock/mock.go
```diff
@@ -578,6 +578,19 @@ func (m *Mock) IsMethodCallable(t TestingT, methodName string, arguments ...inte
 	return false
 }
 
+// NumberOfCalls returns the number of times a method was called
+func (m *Mock) NumberOfCalls(methodName string) int {
```
brackendawson, line 582: This is not actually `NumberOfCalls`, it's more like `NumberOfCallsWithMethodName`. This PR is useful, but do we have to add another method here for people to count calls with certain arguments, or number of calls regardless of method name?

Perhaps something like I mentioned in #1128 is more useful?

brackendawson (changes requested):

dolmen:
We now have [`Mock.AssertNumberOfCalls`](https://pkg.go.dev/github.com/stretchr/testify/mock#Mock.AssertNumberOfCalls). I think that this fits the need expressed in the description, so closing.

PR #967, opened 2020-06-19
Give mock.Arguments a method to assign arguments to named return values.

brackendawson:
## Summary
Allow users to return values **and** nil pointers in a mock using only one method.

## Changes
* Give `mock.Arguments` a `Returns(v ...interface{})` method to assign arguments into named return values.

## Motivation
To return a nill pointer via mock.Called().Get().(*type) requires a two-value type assertion because nil fails type assertion to pointer types:
```
func (m *MockedType) Method() (*ReturnType, error) {
  args := m.Called()
  var r *ReturnType
  r, _ := args.Get(0).(*ReturnType)
  return r, args.Error(1)
}
```
This also stands a chance of being confusing if the type assertion fails because of a different reason, such as a missing `&` in the call to `Return()`.

Instead, passing named return values to a method by reference means we can avoid using type assertion. This also shrinks the mock method a bit as the user doesn't have to write out a line of type assertions:
```
func (m *MockedType) Method() (r *ReturnType, err error) {
  m.Called().Returns(&r, &err)
  return
}
```

Further; we can give more useful error messages to the user, such as when more arguments are passed to Return() than are unpacked with Get(), which is a silent fault today.

Yet more; users don't have to count the argument indices!

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -731,6 +731,46 @@ func (args Arguments) Get(index int) interface{} {
 	return args[index]
 }
 
+// Returns assigns the arguments to the values pointed to by v. Used to return
+// values in a mocked method.
+//
+// For example:
+//     func (o *MyTestObject) SavePersonDetails(firstname, lastname string, age int) (n int, err error) {
+//       o.Called(firstname, lastname, age).Returns(&n, &err)
+//       return
+//     }
+func (args Arguments) Returns(v ...interface{}) {
+	if len(v) != len(args) {
+		panic(fmt.Sprintf("assert: arguments: Cannot call Returns() with %d value(s) because there are %d argument(s).", len(v), len(args)))
+	}
+
+	for i := range v {
+		val := reflect.ValueOf(v[i])
+		if val.Kind() != reflect.Ptr || !val.Elem().CanSet() {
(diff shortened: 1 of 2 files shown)


brackendawson:
Performance wise, Returns() is almost 1 order of magnitude slower than a Get() and an Error():
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ github.com/stretchr/testify/mock -bench ^(Benchmark_Arguments_Returns_Vs_GetAndError)$

goos: darwin
goarch: amd64
pkg: github.com/stretchr/testify/mock
Benchmark_Arguments_Returns_Vs_GetAndError/Returns-12         	30244458	        33.4 ns/op	       0 B/op	       0 allocs/op
Benchmark_Arguments_Returns_Vs_GetAndError/Get-12             	249771368	         4.81 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/stretchr/testify/mock	2.994s
```

Benchmark: https://github.com/brackendawson/testify/commit/f35dfff8aed5e4480be0141f5f7685e47acfcf4e

brackendawson:
I closed this by initially opening the PR from the master branch on my fork, and then updating that branch to a newer HEAD of the upstream repo. This was an error.

PR #968, opened 2020-06-22
add IsErrorType assert

nbaztec:
## Summary
Adds `assert.IsErrorType(t, expected, actual)` to the assertions library

## Changes
Adds a new function `IsErrorType` to the `assert` and `require` packages.

## Motivation
Since go1.13 `errors.Is()` and `fmt.Errorf()` have been the recommended way to test for errors. Checking for error strings is largely considered an anti-pattern within the community.

```go

var (
	errInputMissing   = errors.New("missing input")
	errInputTooShort  = errors.New("too short input")
	errInputMalformed = errors.New("missing prefix")
)

func validate(input interface{}) error {
	if input == nil {
		return errInputMissing
	}

	s := input.(string)
	if len(s) < 5 {
		return fmt.Errorf("%w: got %d characters", errInputTooShort, len(s))
	}
	if !strings.HasPrefix(s, "foo") {
		return fmt.Errorf("%w: wanted prefix '%s' got %s", errInputMalformed, "foo", s)
	}

	return nil
}

func main() {
	cases := []error{
		validate(nil),
		validate("123"),
		validate("barbaz"),
		validate("foobar"),
	}

	for _, err := range cases {
		switch {
		case errors.Is(err, errInputMissing):
			fmt.Printf("MISSING: %s\n", err)
		case errors.Is(err, errInputTooShort):
			fmt.Printf("SHORT: %s\n", err)
		case errors.Is(err, errInputMalformed):
			fmt.Printf("MALFORMED: %s\n", err)
		default:
			fmt.Printf("GOOD\n")
		}
	}
}
```

with the new functions the tests can be written as:

```go
func TestInputTooShortError(t *testing.T) {
	err := validate("123")
	assert.IsErrorType(t, errInputTooShort, err)
}
```

## Related issues
#497

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -321,6 +321,18 @@ func InEpsilonSlicef(t TestingT, expected interface{}, actual interface{}, epsil
 	return InEpsilonSlice(t, expected, actual, epsilon, append([]interface{}{msg}, args...)...)
 }
 
+// IsErrorTypef asserts that a function returned an error (i.e. not `nil`)
+// and that it is equal to the provided error as evaluated by `errors.Is()`.
+//
+//   actualObj, err := SomeFunction()
+//   assert.IsErrorTypef(t, expectedErr, err, "error message %s", "formatted")
+func IsErrorTypef(t TestingT, expected error, actual error, msg string, args ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	return IsErrorType(t, expected, actual, append([]interface{}{msg}, args...)...)
+}
+
 // IsTypef asserts that the specified objects are of the same type.
 func IsTypef(t TestingT, expectedType interface{}, object interface{}, msg string, args ...interface{}) bool {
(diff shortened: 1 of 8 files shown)


nbaztec:
this currently fails on go1.10 

Is there a planned release where the support for go1.10 would be dropped? Or is it imperative to support go1.10?

nbaztec:
Closing in favor of #884

boyan-soubachov:
We try and follow the Go team's support schedule of the two-latest releases. I.e. we currently support 1.14 and 1.13.

That being said, we still have 1.10 to test pre-modules code. I think it probably makes sense to drop it now and assume modules are universally available given how (relatively) long ago 1.11 was.

nbaztec:
Thanks. If that's the prevailing policy to support last two releases and the change is acceptable or at the least ready to be discussed I'd be willing to re-open the PR.

boyan-soubachov:
Closing this PR in favour of a more comprehensive one done in #972 . Thank you for your time and effort :)

Issue #387, opened 2017-01-11
mock.AnythingOfType("[]byte") doesn't quite work


kjlubick:
```
func (m *mockObj) SetFileContents(path string, contents []byte) error {
	args := m.Called(path, contents)
	return args.Error(0)
}

// in testing
m := testutils.NewMockGCSClient()
m.On("SetFileContents", mock.AnythingOfType("string"), 
     mock.AnythingOfType("[]byte")).Return(nil)
m.SetFileContents("foo", []byte("bar"))  
```
This fails with: 
```
SetFileContents(string,[]uint8)
		0: "foo"
		1: []byte{...}

The closest call I have is: 

SetFileContents(mock.AnythingOfTypeArgument,mock.AnythingOfTypeArgument)
		0: "string"
		1: "[]byte"
```

This was pretty confusing, until I saw that the Mock Function seems to want []uint8 (And fmt.Sprintf("%T",[]byte{}) yields []uint8

This is worked around with 
```
m.On("SetFileContents", mock.AnythingOfType("string"), 
     mock.AnythingOfType("[]uint8")).Return(nil)
```
but should really work as written, using "[]byte"

xild:
Same here. trying to mock a []byte object only works with this work around.

charlessachet:
is there any update for that issue?

vespian:
Ping?

lootek:
Same here

bjoernHeneka:
Is this issue somehow solved? It still occurs on my side too

boyan-soubachov:
We haven't had a chance (the time) to get around to it :) If anyone wants to take a stab at it, I'd be happy to review and provide feedback :)

georgysavva:
Hi. Is there any update on this one?

luca-aki:
It would be awesome if we could pass a regex in mock.AnythingOfType. For cases like following: 
`V2Estimate(*context.emptyCtx,*api.APIRequest)
		0: (*context.emptyCtx)(0xc000126008)
		1: &api.APIRequest{NewRequests:[]api.NewRequest{}, Language:"en"}

The closest call I have is: 

V2Estimate(mock.AnythingOfTypeArgument,mock.AnythingOfTypeArgument)
		0: "context"
		1: "api.APIRequest"`

luca-aki:
It would be awesome if we could pass a regex in mock.AnythingOfType. For cases like following:
```
SomeMethod(*context.emptyCtx,*api.APIRequest)
[...]

PR #969, opened 2020-06-25
mock: fix matching []byte vs []uint8 #387

thanakritju:
## Summary
This is a bug reported in #387 . The reason is that according to [this](https://golang.org/pkg/builtin/#byte), that says byte is an alias of uint8, that's why the issue can work around by calling `AnythingOfType("[]uint8")` works

because of the implementation of `Diff` uses `reflect`
This is not a really good fix but it is a quick fix and less likely to break the code.

I do not implement the case where you use call with `[]byte("byte")` as an argument and try to match it with `AnythingOfType("[]uint8")` . Because this create more complex in the code and might require rewriting the whole matching logics.

## Changes
Add a bool of checking if `AnythingOfType("[]byte")` is used and if the actual reflection is `[]uint8`, it matches.
Add a test

## Motivation
because it is on current milestone and no one picks it.

## Related issues
#387

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -785,8 +785,11 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			}
 		} else if reflect.TypeOf(expected) == reflect.TypeOf((*AnythingOfTypeArgument)(nil)).Elem() {
 
+			// check for []byte and []uint8
+			var matchedBtye = string(expected.(AnythingOfTypeArgument)) == "[]byte" && reflect.TypeOf(actual).String() == "[]uint8"
+
 			// type checking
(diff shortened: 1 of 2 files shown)


mvdkleijn:
@boyan-soubachov could you take a look?

boyan-soubachov:
Whilst this is a bug, it's fundamental and highly-used in the API. I think this would be a breaking change, even if for the better. One of our big-ticket items seems to be moving to `go-cmp` for v2. That would most likely fix/obviate this problem.

mvdkleijn:
> One of our big-ticket items seems to be moving to `go-cmp` for v2.

Yes, I believe we have several PRs / issues that are related to `go-cmp` and/or are conflicting with the `go-cmp` move. We'll have to do some organizing.

dolmen:
We have [`mock.IsType`](https://pkg.go.dev/github.com/stretchr/testify/mock#IsType):

```go
mock.IsType([]byte(nil))
```

=> closing.

PR #974, opened 2020-06-30
add assertion order

### LINKED ISSUE #710: Assert value order comparison.
I have created a test collection comparisons there https://github.com/alexpantyukhin/assertord . I believe many of them are redundant but it's just for demo. Any thoughts could it be good idea to integrate that functionality into the `testify`? Also I guess it would be good to have `assert.IsGreater`, `assert.IsGreaterOrEqual`, `assert.IsLess`, `assert.IsLessOrEqual` in the `testify`.

alexpantyukhin:
## Summary
Add assertion of collections ordering.

## Motivation
It could help to check collection on specific order.

```go
 assert.IsIncreasing(t, []int{1, 2, 3})
```

## Related issues
Closes #710

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -321,6 +321,54 @@ func InEpsilonSlicef(t TestingT, expected interface{}, actual interface{}, epsil
 	return InEpsilonSlice(t, expected, actual, epsilon, append([]interface{}{msg}, args...)...)
 }
 
+// IsDecreasingf asserts that the collection is decreasing
+//
+//    assert.IsDecreasingf(t, []int{2, 1, 0}, "error message %s", "formatted")
+//    assert.IsDecreasingf(t, []float{2, 1}, "error message %s", "formatted")
+//    assert.IsDecreasingf(t, []string{"b", "a"}, "error message %s", "formatted")
+func IsDecreasingf(t TestingT, object interface{}, msg string, args ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	return IsDecreasing(t, object, append([]interface{}{msg}, args...)...)
+}
+
+// IsIncreasingf asserts that the collection is increasing
+//
+//    assert.IsIncreasingf(t, []int{1, 2, 3}, "error message %s", "formatted")
+//    assert.IsIncreasingf(t, []float{1, 2}, "error message %s", "formatted")
+//    assert.IsIncreasingf(t, []string{"a", "b"}, "error message %s", "formatted")
+func IsIncreasingf(t TestingT, object interface{}, msg string, args ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	return IsIncreasing(t, object, append([]interface{}{msg}, args...)...)
+}
+
+// IsNonDecreasingf asserts that the collection is not decreasing
+//
+//    assert.IsNonDecreasingf(t, []int{1, 1, 2}, "error message %s", "formatted")
+//    assert.IsNonDecreasingf(t, []float{1, 2}, "error message %s", "formatted")
+//    assert.IsNonDecreasingf(t, []string{"a", "b"}, "error message %s", "formatted")
+func IsNonDecreasingf(t TestingT, object interface{}, msg string, args ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	return IsNonDecreasing(t, object, append([]interface{}{msg}, args...)...)
+}
+
+// IsNonIncreasingf asserts that the collection is not increasing
+//
+//    assert.IsNonIncreasingf(t, []int{2, 1, 1}, "error message %s", "formatted")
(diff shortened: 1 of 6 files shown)


assert/assertion_order.go
```diff
@@ -0,0 +1,81 @@
...
+	objValue := reflect.ValueOf(object)
+	objLen := objValue.Len()
+
+	if objLen <= 1 {
+		return true
+	}
+
+	value := objValue.Index(0)
+	valueInterface := value.Interface()
+	firstValueKind := value.Kind()
```
boyan-soubachov, line 24: Given that in lines 18-20 we've guaranteed that the array will be of at least length 2 by the time we get to this point, would it not be clearer if we just moved this code inside the for loop and referenced the previous index?

e.g.
```
for i:= 1; i < objLen; i++ {
    prevValue := objValue.Index(i-1)
    value := objValue.Index(i)
    ...
}
...
```

boyan-soubachov: You could also inline the assignments that way in cases where it would improve clarity.

assert/assertion_order.go
```diff
@@ -0,0 +1,81 @@
...
+		prevValue := value
+		prevValueInterface := valueInterface
+
+		value = objValue.Index(i)
+		valueInterface = value.Interface()
+
+		compareResult, isComparable := compare(prevValueInterface, valueInterface, firstValueKind)
+
+		if !isComparable {
+			return Fail(t, fmt.Sprintf("Can not compare type \"%s\"", reflect.TypeOf(value)), msgAndArgs...)
```
boyan-soubachov, line 36: Would it not be better if we printed both the types of `value` and `previousValue`? If I'm not mistaken, `[]interface{}` can have each entry be of a different type?

assert/assertion_order_test.go
```diff
@@ -0,0 +1,178 @@
...
+		{collection: []string{"b", "a"}, msg: `"b" is not less than "a"`},
+		{collection: []int{2, 1}, msg: `"2" is not less than "1"`},
+		{collection: []int8{2, 1}, msg: `"2" is not less than "1"`},
+		{collection: []int16{2, 1}, msg: `"2" is not less than "1"`},
+		{collection: []int32{2, 1}, msg: `"2" is not less than "1"`},
+		{collection: []int64{2, 1}, msg: `"2" is not less than "1"`},
+		{collection: []uint8{2, 1}, msg: `"2" is not less than "1"`},
+		{collection: []uint16{2, 1}, msg: `"2" is not less than "1"`},
+		{collection: []uint32{2, 1}, msg: `"2" is not less than "1"`},
+		{collection: []uint64{2, 1}, msg: `"2" is not less than "1"`},
```
boyan-soubachov, line 41: Do you mind adding some tests to show that this works for more than 2 entries? Doesn't have to be for every type but just to prove by induction that the function works for _n_ entries.

alexpantyukhin:
@boyan-soubachov I have fixed 2/3 points. Have question-discussion about the first one

assert/assertion_format.go
```diff
@@ -321,6 +321,54 @@ func InEpsilonSlicef(t TestingT, expected interface{}, actual interface{}, epsil
 	return InEpsilonSlice(t, expected, actual, epsilon, append([]interface{}{msg}, args...)...)
 }
 
+// IsDecreasingf asserts that the collection is decreasing
+//
+//    assert.IsDecreasingf(t, []int{2, 1, 0}, "error message %s", "formatted")
+//    assert.IsDecreasingf(t, []float{2, 1}, "error message %s", "formatted")
+//    assert.IsDecreasingf(t, []string{"b", "a"}, "error message %s", "formatted")
```
devdinu, line 328: Do we need `IsIncreasingf` with format given `IsIncreasing` can be used directly for the same? and similarly for other functions

alexpantyukhin:
@devdinu It is required I guess. It doesn't pass CI without `f` methods.

boyan-soubachov:
> @boyan-soubachov I have fixed 2/3 points. Have question-discussion about the first one

Sure, what specifically?

alexpantyukhin:
@boyan-soubachov In the current version we have only one call of `.Index(i)` and getting `.Interface()` per loop iteration.
You suggest more clear solution:
```go
for i:= 1; i < objLen; i++ {
    prevValue := objValue.Index(i-1)
    value := objValue.Index(i)
    ...
}
```
But this case we will need to call `Index` and `Interface`  2 * (length of array) times. Am I correctly understand your idea?

boyan-soubachov:
> @boyan-soubachov In the current version we have only one call of `.Index(i)` and getting `.Interface()` per loop iteration.
> You suggest more clear solution:
> 
> ```go
> for i:= 1; i < objLen; i++ {
>     prevValue := objValue.Index(i-1)
>     value := objValue.Index(i)
>     ...
> }
> ```
> 
> But this case we will need to call `Index` and `Interface` 2 * (length of array) times. Am I correctly understand your idea?

I see what you mean, good point.

I guess the code is clear enough that it's not worth the extra reflection calls 👍

alexpantyukhin:
@boyan-soubachov What do you think, should we merge it?

PR #979, opened 2020-07-20
Compare time.Time objects for equality correctly

senyast4745:
Adding test

## Summary
Fix assert.ObjectsAreEqual bug with not equals time.Time before and after serializing\deserializing

## Changes
Added a check in ObectsAreEqual in case the arguments are a time object

## Motivation
Fix bug to use test in docker 

## Related issues
Closes #950 
Fix #951

PR #985, opened 2020-07-27
Fix compairsion of struct containing time.Time. 

### LINKED ISSUE #984: Struct containing a time.Time can't be compared correctly.
#979 fixed comparison of `time.Time`, but struct containing `time.Time` still can't be compared correctly because still, we will use `reflect.DeepEqual`.

I'm working on a PR to fix this.

leoleoasd:
## Summary
#979 fixed comparison of `time.Time`, but struct containing `time.Time` still can't be compared correctly because still, we will use `reflect.DeepEqual`.
This PR is trying to fix it.

## Related issues
Closes #984

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -16,6 +16,7 @@ import (
 	"time"
 	"unicode"
 	"unicode/utf8"
+	"unsafe"
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/pmezard/go-difflib/difflib"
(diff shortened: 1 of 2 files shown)


leoleoasd:
I have a workaround here, but it only works with "struct containing time.Time", not with "struct containing struct containing time.Time".
```go
func ObjectsAreEqual(expected, actual interface{}) bool {
	if expected == nil || actual == nil {
		return expected == actual
	}

	switch expected.(type) {
	case time.Time:
		if tAct, ok := actual.(time.Time); ok {
			return expected.(time.Time).Equal(tAct)
		}
		return false
	case []byte:
		if act, ok := actual.([]byte); ok {
			exp := expected.([]byte)
			if exp == nil || act == nil {
				return exp == nil && act == nil
			}
			return bytes.Equal(exp, act)
		}
		return false
	}
	v1 := reflect.ValueOf(expected)
	v2 := reflect.ValueOf(actual)
	if v1.Kind() == reflect.Struct {
		if !v1.IsValid() || !v2.IsValid() {
			return v1.IsValid() == v2.IsValid()
		}
		if v1.Type() != v2.Type() {
			return false
		}
		// Cause we won't call ObjectsAreEqual again
		// So the "visit" param of reflect.deepValueEqual is not required.
		for i, n := 0, v1.NumField(); i < n; i++ {
			timeV1, ok1 := v1.Field(i).Interface().(time.Time)
			timeV2, ok2 := v2.Field(i).Interface().(time.Time)
			if ok1 && ok2 {
				if !timeV1.Equal(timeV2) {
					return false
				}
			} else {
				if !reflect.DeepEqual(v1.Field(i).Interface(), v2.Field(i).Interface()) {
					return false
				}
			}
		}
		return true
	}
	return reflect.DeepEqual(expected, actual)
}
```

If we call `ObjectsAreEqual` again on each field, we should deal with recursive types that the visited parameter of `reflect.deepValueEqual` are used for.

boyan-soubachov:
I see. I'll revert the merged commit then to give us more room to work out a better approach.

leoleoasd:
It occurs to me that the building the `visited` map requires unexported field of `reflect.Value`.

boyan-soubachov:
I've reverted the previous merge. No rush, feel free to scope out a better solution should you wish :)

leoleoasd:
@boyan-soubachov Added a workaround to deal with non-recursive types. I think we can do nothing about recursive types, cause we can't look into the unexported field of reflect.Value.

Something weird is that when I'm running `TestObjectsAreEqual_StructContainingTime` only, it will succeed, but if I run all tests together, it will fail.

Also, is it normal for the tests of the "suite" package kept failing?

leoleoasd:
And......
The tests running from travis-ci succeed.....
I'm using a mac,  `go test ./assert/...` kept failing, running only the `TestObjectsAreEqual_StructContainingTime` test from Goland IDE kept succeed. I've tried multiple times with the same result.

leoleoasd:
@boyan-soubachov ready for review.

boyan-soubachov:
Hmm. I'm not sure the complexity and safety (the use of `unsafe` pointers) is worth the benefits. We'll be converting to use `go-cmp` in testify v2 so for a quick-fix this seems a bit too much.

leoleoasd:
> Hmm. I'm not sure the complexity and safety (the use of `unsafe` pointers) is worth the benefits. 
The unsafe pointers are only used to prevent dead-recurse. I'm using it exactly the same way which `reflect.DeepEqual` uses it, so it should be safe. 

> Thank you for your time and effort but I think it's best we hold off on this. Thoughts?
For most cases(and in my case), support for "struct containing time.Time" is fairly enough. So, maybe [this](https://github.com/stretchr/testify/pull/985#issuecomment-664355701) fix is acceptable?
Of course, it needs some modification to support un-addressable variables.

PR #986, opened 2020-07-28
mock: speed up mock.findExpectedCall by skipping diff if not necessary

### LINKED ISSUE #983: Speed up mock.findExpectedCall by skipping diff if not necessary
If the call should not be used anymore due to `call.Repeatability < 0`, there is no need to calculate arguments diff. This should save some unnecessary computations.

mrahs:
## Summary
Speed up `mock.findExpectedCall` by skipping argument diffing if not necessary

## Changes
Modified the function to skip over used up expected calls.
Everything else remains the same from the user's perspective, including the associated special printed message for such cases.

## Motivation
I have test suites with many one-time expected calls. I have noticed that the suite keeps accumulating such expired calls and evaluates every one of them before reaching the few usable ones.
Ideally, such calls should be discarded to save space and cycles, but that's an optimization for another day.

## Related issues
Closes #983

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -280,21 +280,17 @@ func (m *Mock) On(methodName string, arguments ...interface{}) *Call {
 // */
 
 func (m *Mock) findExpectedCall(method string, arguments ...interface{}) (int, *Call) {
-	var expectedCall *Call
-
 	for i, call := range m.ExpectedCalls {
-		if call.Method == method {
-			_, diffCount := call.Arguments.Diff(arguments)
-			if diffCount == 0 {
-				expectedCall = call
-				if call.Repeatability > -1 {
-					return i, call
-				}
-			}
+		if call.Repeatability < 0 || call.Method != method {
+			continue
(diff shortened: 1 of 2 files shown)


mrahs:
@boyan-soubachov I wonder if you can take a look?

boyan-soubachov (changes requested):
Thank you for this contribution.

I agree with simplifying (and optimising) the logic in `findExpectedCall` but it also looks to me like you're changing the behaviour of the exported function (`args.Diff`).

This means that this change would not be backwards compatible and would have to wait for testify v2.0.0.

Would it make more sense to split this into 2 PRs, one for optimising the iteration logic in `findExpectedCall` and the other for reducing the return vars and simplifying the upstream call-chain?

mrahs:
@boyan-soubachov thank you for the review. I have updated the code to preserve visible behavior and a added a test case for expected error when a method has been used up to ensure consistent behavior before and after this change.

PR #642, opened 2018-07-28
assert: s/rune/r/ to avoid "rune" predeclared ident shadowing

quasilyte:
Found using https://go-critic.github.io/overview#builtinShadow-ref

PR #987, opened 2020-07-31
assert: s/rune/r/ to avoid "rune" predeclared ident shadowing #642

mvdkleijn:
## Summary

Since its a simple fix, let's just do the rebase ourselves as it were. Thanks for the original PR @quasllyte (#642)


mvdkleijn:
@boyan-soubachov Please be so kind to hit approve and merge. :smile:

PR #990, opened 2020-08-01
Add ErrorsMatch assertion

swithek:
## Summary
This PR adds `assert.ErrorsMatch` for comparing errors of different kinds.

## Changes
- Add ErrorsMatch assertion

## Motivation
There are times when I have to write table-driven tests and check errors of different kinds. `assert.Equal` works well, though not all errors are comparable like that, mostly because they are not sentinels. Of course, I can use `assert.Error` for this and check whether the actual error is non-nil, but then I need to mark my test cases somehow, so that they wouldn't use `assert.Equal` like others do. `assert.AnError` seems like a good 'marker' for such cases, but there is no function in `assert` package that would use it as a 'marker' to check whether the actual error is non-nil (and use `assert.Error` internally). The new `assert.ErrorsMatch` would be able to accept `assert.AnError` to check whether the actual error is non-nil, and compare errors directly when the expected error is not `assert.AnError`. Example:

```go
func TestCalc(t *testing.T) {
    cases := []struct{
        Name string
        Input int
        Err error
    }{
        {
            Name: "Unknown error",
            Input: -10,
            Err: assert.AnError,
        },
        {
            Name: "io.ErrUnexpectedEOF error",
            Input: 200,
            Err: io.ErrUnexpectedEOF,
        },
        {
            Name: "No error",
            Input: 3000,
            Err: nil,
        },
    }

    for _, case := range cases {
        t.Run(case.Name, func(t *testing.T) {
              err := calc(case.Input)
              assert.ErrorsMatch(t, c.Err, err)
        })
    }
}

func calc(input int) error {
    if input == -10 {
        return otherCalc()
    }

    if input == 200 {
        return io.ErrUnexpectedEOF
    }

    return nil
}
```

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -134,7 +134,7 @@ func ErrorIsf(t TestingT, err error, target error, msg string, args ...interface
 
 // ErrorsMatchf asserts that two errors are equal, unless the expected error
 // is AnError, in which case the actual error can be any non-nil error.
-// Errors are compared directly, without errors.Is
+// Errors are compared directly, without errors.Is().
 //
 //    assert.ErrorsMatchf(t, io.ErrUnexpectedEOF, err, "error message %s", "formatted")
 //    assert.ErrorsMatchf(t, nil, error(nil), "error message %s", "formatted")
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -132,6 +132,20 @@ func ErrorIsf(t TestingT, err error, target error, msg string, args ...interface
 	return ErrorIs(t, err, target, append([]interface{}{msg}, args...)...)
 }
 
+// ErrorsMatchf asserts that two errors are equal, unless the expected error
+// is AnError, in which case the actual error can be any non-nil error.
+// Errors are compared directly, without errors.Is
+//
+//    assert.ErrorsMatchf(t, io.ErrUnexpectedEOF, err, "error message %s", "formatted")
+//    assert.ErrorsMatchf(t, nil, error(nil), "error message %s", "formatted")
+//    assert.ErrorsMatchf(t, assert.AnError, errors.New("any error"), "error message %s", "formatted")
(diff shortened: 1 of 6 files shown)


boyan-soubachov:
We now have Go 1.13's `ErrorIs` in `master`. Would that not solve your issue?

swithek:
Not exactly. The main goals of `assert.ErrorsMatch` are:
- To check whether both errors are the same (no unwrapping is being done). This is useful when you want to check if the function you test returns a correct sentinel error (`io.ErrUnexpectedEOF` in the example).
- To check whether the actual error is non-nil by making use of the already existing `assert.AnError`.

`assert.ErrorsMatch` is essentially `assert.Error`, `assert.NoError` and `assert.Equal` combined into one. It is especially handy when you have to check many errors of different kinds: some concrete sentinel errors, and some unknown, returned from other internally called functions.

swithek:
@boyan-soubachov is there anything that needs to be changed in this PR?

boyan-soubachov:
Hmm. So if I understand correctly, this is effectively acting as syntactic sugar for
```
if expectedErr == nil {
    assert.NoError(got)
} else {
   assert.Error(got)
   assert.Equal(got, expectedErr)
}
```
?

If so, would it not be better to just adapt `Equal()` to handle empty values (or even leave it as is)?

swithek:
Yes, you're right. But it would also handle one additional case:
```go
if expectedErr == assert.AnError {
    assert.Error(t, got)
}
```
This assertion function is marely a syntactic sugar for these 3 different kinds of error comparison, which is especially useful when you have to test multiple cases of the same logic with multiple possible error values.

I have considered extending `asser.Equal`, but I think there are users who might be returning `assert.AnError` from their mocked functions and expecting `assert.Equal` to perform `gotErr(assert.AnError) == expectedErr(assert.AnError)` comparison instead of `gotErr(assert.AnError) != nil`.

PR #993, opened 2020-08-10
Revise test case name on README.md of mock section

RyuseiNomi:
## Summary
On README.md, there are many examples as useful.
But I was a little worried about a test case name on Mock section.
I think test case name should contain a target method name accurately and indicate what would you want to test.
So I change two test case name on readme.

## Changes
I change two test case name as below.

1. Match method name to sample method.
`func TestSomething(t *testing.T)` → `func TestDoSomething(t *testing.T)`

2. Match method name to sample method and indicate what is the different with above.
`func TestSomethingElse(t *testing.T)` → `func TestDoSomethingWithPlaceholder(t *testing.T)`

## Motivation
Test case name should indicate which method is targeted and what you want to test.
It's important not only production code but also sample code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ func (m *MyMockedObject) DoSomething(number int) (bool, error) {
 
 // TestSomething is an example of how to use our test object to
 // make assertions about some target code we are testing.
-func TestDoSomething(t *testing.T) {
+func TestSomething(t *testing.T) {
 
   // create an instance of our test object
   testObj := new(MyMockedObject)
@@ -169,12 +169,12 @@ func TestDoSomething(t *testing.T) {
 
 }
 
-// TestSomethingElse is a second example of how to use our test object to
+// TestSomethingWithPlaceholder is a second example of how to use our test object to
 // make assertions about some target code we are testing.
 // This time using a placeholder. Placeholders might be used when the
 // data being passed in is normally dynamically generated and cannot be
 // predicted beforehand (eg. containing hashes that are time sensitive)
-func TestDoSomethingWithPlaceholder(t *testing.T) {
+func TestSomethingWithPlaceholder(t *testing.T) {
 
   // create an instance of our test object
   testObj := new(MyMockedObject)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -152,7 +152,7 @@ func (m *MyMockedObject) DoSomething(number int) (bool, error) {
 
 // TestSomething is an example of how to use our test object to
 // make assertions about some target code we are testing.
-func TestSomething(t *testing.T) {
+func TestDoSomething(t *testing.T) {
```
boyan-soubachov, line 155: I am not sure what problem this solves :/

README.md
```diff
@@ -174,7 +174,7 @@ func TestSomething(t *testing.T) {
 // This time using a placeholder. Placeholders might be used when the
 // data being passed in is normally dynamically generated and cannot be
 // predicted beforehand (eg. containing hashes that are time sensitive)
-func TestSomethingElse(t *testing.T) {
+func TestDoSomethingWithPlaceholder(t *testing.T) {
```
boyan-soubachov, line 177: This makes sense. Not sure why we have the `Do` part but I can see how having `TestSomethingWithPlaceholder` makes things a little clearer.

Importantly, the docstrings must start with the (new) function names.

README.md
RyuseiNomi: I read CONTRIBUTING.md again and I understood this difference will not solve any problem.
[I undo the difference](9e16738c413647272784610361e4694bbaa27609) so please don't worry about it sorry.

README.md
RyuseiNomi: Thanks for giving a positive opinion :)
Surely, I forget to revise the docstrings... [I revised it](9e16738c413647272784610361e4694bbaa27609) :)
Whether to have the `Do` part, I'll follow the previous example.

README.md
boyan-soubachov: No worries. We welcome all contributions, just wanted to get more context on what problem you're trying to solve with this :)

RyuseiNomi: > just wanted to get more context

I got it! 
In my opinion, method name is better to be able to indicate which method is targeted so I suggested the revision.
But for this now, the example code is simple and short so it wouldn't be confuse :)

boyan-soubachov:
LGTM as well. Thank you!

PR #999, opened 2020-09-10
add API Off for Mock, which is used for clearing repeatability of the specific call during table test case ;

sinngae:


brackendawson:
It's always possible to know before a test if an expectation is expected, so this is a no on the grounds that it's unnecessary and promotes bad testing practice. Sorry if that's disappointing.

An Unset method was added to Arguments, rather than the whole mock, but I also discourage its use.

PR #1011, opened 2020-10-17
Use (time.Time).Equal for time objects

### LINKED ISSUE #1010: Testify `assert.Equal` between two times doesn't work same as `(time.Time).Equal`.
### Motivation

Before using `testify` I used to test my `time.Time` objects like this:
```
var (
    expected time.Time
    got time.Time
)

// ... 

if !expected.Equal(got) {
    t.Error(...)
}
```

Once I started using testify, my intuition was replace the above test with 
```
var (
    expected time.Time
    got time.Time
)

// ... 

assert.Equal(t, expected, got)
```

Which mostly worked fine.... until not. I notice that two objects with different locations but representing same time instant were considered as not equal... I stopped using `assert.Equal` to compare objects `time.Time` but instead, using 
```
assert.True(t, expected.Equal(got))
```

This way to use the library feels tricky, uncomfortable and lose the idiomatic meaningful. I don't want to assert a boolean, but two `time.Time` objects instead 😔

I propose to use built-in Equal for `time.Time` objects.

Considering this case could be translatable to any object with a method `<T>.Equal(<T>) bool`. Maybe checking existence of such method and using it, is a good approach.

Thanks for your consideration

gfelixc:
## Summary
Use built-in Equal method when compare `time.Time` objects

## Changes
When use Equal, try to cast given objects into `time.Time` and uses (time.Time).Equal in case of true

## Motivation
See #1010 

## Related issues
Closes #1010 

Breaking change

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -60,6 +60,12 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
 		return expected == actual
 	}
 
+	if e, ok := objectToTime(expected); ok {
+		if a, ok := objectToTime(actual); ok {
+			return e.Equal(a)
+		}
+	}
+
 	exp, ok := expected.([]byte)
 	if !ok {
 		return reflect.DeepEqual(expected, actual)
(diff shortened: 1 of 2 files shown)


assert/assertions.go
```diff
@@ -75,6 +81,20 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
...
 
+func objectToTime(o interface{}) (time.Time, bool) {
+	s, ok := o.(time.Time)
+	if ok {
+		return s, ok
+	}
+
+	p, ok := o.(*time.Time)
+	if ok {
+		return *p, true
```
colega, line 92: `p` could be a nil pointer, please check that `p != nil` before dereferencing it.

leoleoasd:
Does the exact same thing with #979.
#979 is reverted shortly after merge, because, as #984 mentions, this only fixes comparing between time. a struct containing time still cannot be compared. #985 tried to fix this by creating something like reflect.DeepEqual and call time.Time.Equal when needed, but used unsafe pointers.

brackendawson:
Earlier duplicate of #1464

PR #1014, opened 2020-10-22
[Fix] Limit the spew recursion

grahamjenson:
## Summary

The latest update to golang/protobuf (I am on v1.4.2) break the spew circular data structure detection. This means that when calling `assert.Equal(t, proto1, proto2)`, if it fails it will enter an infinite recursion.

By limiting the recursion with `MaxDepth` set to a large number (10 seems reasonable as a struct that is 10 deep will be pretty large) this means that if there is an issue with recursion (as there is with protobuf) the test will still return a failure in a reasonable time instead of just a timeout.

## Changes
Adding MaxDepth to spew output.

## Motivation
Stop infinite loop if spew fails to detect circular recusion.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1622,7 +1622,7 @@ var spewConfig = spew.ConfigState{
 	DisableCapacities:       true,
 	SortKeys:                true,
 	DisableMethods:          true,
-	MaxDepth:                100,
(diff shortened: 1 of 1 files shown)


boyan-soubachov (approved):
Looks good, thank you!

PR #1018, opened 2020-10-31
Add positive/negative assertion

rinchsan:
## Summary

Adding positive/negative assertion functions

## Changes

What do you think about these changes?

- Adding function `Positive` to assert that the specified element is positive
- Adding function `Negative` to assert that the specified element is negative

## Motivation

When we want to assert that some value is positive or negative, we test like below.

```go
func TestHoge(t *testing.T) {
    assert.Greater(t, 1, 0) // want to assert positive
    assert.Greater(t, 1.23, 0.0) // want to assert positive
    assert.Less(t, -1, 0) // want to assert negative
    assert.Less(t, -1.23, 0) // want to assert negative
}
```

By adding `Positive`/`Negative` assert functions, we can simplify it like an example usage below.

```go
func TestHoge(t *testing.T) {
    assert.Positive(t, 1)
    assert.Positive(t, 1.23)
    assert.Negative(t, -1)
    assert.Negative(t, -1.23)

    assert := assert.New(t)
    assert.Positive(1)
    assert.Positive(1.23)
    assert.Negative(-1)
    assert.Negative(-1.23)
}
```

## Related issues

nothing

### DIFF
--- a/assert/assertion_compare.go
+++ b/assert/assertion_compare.go
@@ -342,6 +342,24 @@ func LessOrEqual(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...inter
 	return compareTwoValues(t, e1, e2, []CompareType{compareLess, compareEqual}, "\"%v\" is not less than or equal to \"%v\"", msgAndArgs)
 }
 
+// Positive asserts that the specified element is positive
+//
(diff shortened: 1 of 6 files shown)


boyan-soubachov (approved):
This looks great, thank you!

PR #1021, opened 2020-11-05
Fix custom messages in comparison-assertions

benjaminjkraft:
## Summary
This PR fixes custom messages in comparison-assertions.

## Changes
The assesrtions now pass varargs rather than a single slice arg.

## Motivation
The comparison-assertions were passing msgAndArgs as a slice rather than
varargs, which meant that the entire message-and-arguments got printed
as a single slice.  Now it gets printf-ed correctly.  See test cases for examples.

Edit: Fixes the rest of #1034 (the other part was fixed by #1026).

### DIFF
--- a/assert/assertion_compare.go
+++ b/assert/assertion_compare.go
@@ -310,7 +310,7 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (CompareType, bool) {
 //    assert.Greater(t, float64(2), float64(1))
 //    assert.Greater(t, "b", "a")
 func Greater(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...interface{}) bool {
-	return compareTwoValues(t, e1, e2, []CompareType{compareGreater}, "\"%v\" is not greater than \"%v\"", msgAndArgs)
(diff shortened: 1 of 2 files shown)


boyan-soubachov (changes requested):
Looks great, thanks for catching this! Is it possible to add a test to cover the regression caused by not expanding the varargs?

benjaminjkraft:
Hi @boyan-soubachov -- thanks for the look!  The test I added should cover the regression, I think -- if you run it without the other changes it fails with e.g.
```
    assertion_compare_test.go:266: 
                Error Trace:
                Error:          "\n\tError Trace:\t\n\tError:      \t\"1\" is not greater than \"2\"\n\tMessages:   \t[%v wasn't greater 1]\n" does not contain "1 wasn't greater"
                Test:           TestComparisonCustomMessages
```
Let me know if there was something different you wanted, I'm happy to add it!

codyohlsen:
can we merge this? :) thanks

AlekSi:
@boyan-soubachov PTAL

lambdanis:
Looks like the issue was fixed in https://github.com/stretchr/testify/pull/1150/, so this PR can be closed.

dolmen:
@benjaminjkraft Thanks for the patch. Unfortunately it hasn't been noticed earlier.

@lambdanis Thanks for the duplicate report.

PR #1022, opened 2020-11-09
Add ErrorContains

alxn:
## Summary
It's common practice to wrap errors, so that you do not have a transitive dependency on the error format of a third party library, but also so that you are able to exactly match you are testing a specific error case correctly.

This change adds an `ErrorContains()` method to check that both an error is reurned, and it contains a provided string.

## Changes
Added `ErrorContains()` and `ErrorContainsf()`.

## Motivation

When I wrap errors returned from third party modules:

```go
func method(f foo, b bar) error {
        if err := f(); err != nil {
                return fmt.Errorf("foo failed: %w", err)
        }
        if err := b(); err != nil {
                return fmt.Errorf("bar failed: %w", err)
        }
        return nil
}
```

This then forces me to write:

```go
f := func() error { return assert.AnError }
b := func() error { return nil }
err := method(f, b)
require.Error(t, err)
assert.Contains(t, err.Error(), "foo failed")

f = func() error { return nil }
b = func() error { return assert.AnError }
err = method(f, b)
require.Error(t, err)
assert.Contains(t, err.Error(), "bar failed")
```
This change removes the `require` and `Error()` unpacking:

```go
f := func() error { return assert.AnError }
b := func() error { return nil }
err := method(f, b)
assert.ErrorContains(t, err, "foo failed")

f = func() error { return nil }
b = func() error { return assert.AnError }
err = method(f, b)
assert.ErrorContains(t, err, "bar failed")
```

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -123,6 +123,18 @@ func ErrorAsf(t TestingT, err error, target interface{}, msg string, args ...int
 	return ErrorAs(t, err, target, append([]interface{}{msg}, args...)...)
 }
 
+// ErrorContainsf asserts that a function returned an error (i.e. not `nil`)
+// and that the error contains the specified substring.
+//
+//   actualObj, err := SomeFunction()
+//   assert.ErrorContainsf(t, err,  expectedErrorSubString, "error message %s", "formatted")
+func ErrorContainsf(t TestingT, theError error, contains string, msg string, args ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	return ErrorContains(t, theError, contains, append([]interface{}{msg}, args...)...)
+}
+
 // ErrorIsf asserts that at least one of the errors in err's chain matches target.
 // This is a wrapper for errors.Is.
 func ErrorIsf(t TestingT, err error, target error, msg string, args ...interface{}) bool {
--- a/assert/assertion_forward.go
+++ b/assert/assertion_forward.go
@@ -222,6 +222,30 @@ func (a *Assertions) ErrorAsf(err error, target interface{}, msg string, args ..
 	return ErrorAsf(a.t, err, target, msg, args...)
 }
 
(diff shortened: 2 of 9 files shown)


boyan-soubachov:
I'd love to get community feedback on this. I'm not sure the value this extra function adds is greater than the confusion/complexity it creates by having yet another function in the `assert`/`require` packages.

In fact, how is this different from the `EqualError` assertions?

alxn:
`EqualError` is different because it requires an exact error match. Here, I have the problem where I have a decorated error. I don't know the underlying error, only the wrapping.

boyan-soubachov:
I see. Would it not make sense to rename this function to `ContainsError` to be consistent with `EqualError` then?

alxn:
```
assert.ErrorContains(t, err, "bar failed")
```
Semantically I'm asserting that the _Error_ `err` _Contains_ `"bar failed"`.

`ContainsError` scans more like:

```
assert.ContainsError(t, something, err)
```

I'd written `ErrorContains` to be a variant of `ErrorIs`:

```
func EqualError(t TestingT, theError error, errString string, msgAndArgs ...interface{}) bool
func Error(t TestingT, err error, msgAndArgs ...interface{}) bool
func ErrorAs(t TestingT, err error, target interface{}, msgAndArgs ...interface{}) bool
func ErrorIs(t TestingT, err, target error, msgAndArgs ...interface{}) bool
func NoError(t TestingT, err error, msgAndArgs ...interface{}) bool
func NotErrorIs(t TestingT, err, target error, msgAndArgs ...interface{}) bool
func PanicsWithError(t TestingT, errString string, f PanicTestFunc, msgAndArgs ...interface{}) bool
```

boyan-soubachov:
Fair point, that makes sense :)

boyan-soubachov (approved):

yarikk:
This was merged 6 months ago but still unavailable in @latest which currently resolves to v1.7.0. When to expect it version-tagged?

webmaster128:
1.7.1 is realeased and seems to include this feature. Thank you everyone.

Issue #1006, opened 2020-10-07
Unable to mock variadic function with function argument


BeyondEvil:
I have a function with this signature:
```go
func (api *Client) PostMessage(channelID string, options ...MsgOption) (string, string, error)`
```

Where `MsgOption` is a function-type:
```go
type MsgOption func(*sendConfig) error
```

The mock:
```go
func (mock *slackAPIMock) PostMessage(channelID string, options ...slack.MsgOption) (string, string, error) {
	args := mock.Called(channelID, options)
	return args.String(0), args.String(1), args.Error(2)
}
```

Turning it `On`:
```go
	options := []slack.MsgOption{
		slack.MsgOptionAttachments(expectedAttachment),
		slack.MsgOptionAsUser(true),
	}

	mock.On(
		"PostMessage",
		mock.AnythingOfTypeArgument("string"),
		options,
	).Return(mockedChannel, timestamp, nil)
```

And the assert:
```go
	mock.AssertCalled(
		t,
		"PostMessage",
		mockedChannel,
		slack.MsgOptionAttachments(expectedAttachment),
		slack.MsgOptionAsUser(true),
	)
```

Result:

```
Diff: 1: FAIL:  ([]slack.MsgOption=[0x1376370 0x1376180]) != ([]slack.MsgOption=[0x1376370 0x1376180])
```

They look the same to me. What am I missing here? 🤔 

<details>

<summary>Full output</summary>

```
panic: 

mock: Unexpected Method Call
-----------------------------

PostMessage(string,[]slack.MsgOption)
		0: "some_channel"
		1: []slack.MsgOption{(slack.MsgOption)(0x1376370), (slack.MsgOption)(0x1376180)}

The closest call I have is: 

PostMessage(mock.AnythingOfTypeArgument,[]slack.MsgOption)
		0: "string"
		1: []slack.MsgOption{(slack.MsgOption)(0x1376370), (slack.MsgOption)(0x1376180)}


Diff: 1: FAIL:  ([]slack.MsgOption=[0x1376370 0x1376180]) != ([]slack.MsgOption=[0x1376370 0x1376180]) [recovered]
	panic: 

mock: Unexpected Method Call
-----------------------------

PostMessage(string,[]slack.MsgOption)
		0: "some_channel"
		1: []slack.MsgOption{(slack.MsgOption)(0x1376370), (slack.MsgOption)(0x1376180)}

The closest call I have is: 

[...]

PR #1023, opened 2020-11-09
allow testing for functional options

### LINKED ISSUE #997: Mock Matcher for functional options
Hi,

In most go libraries nowadays (etcd, kubernets, aws, etc.) the [functional options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) is being used:

```go
c := client.New(client.WithTimeout(), client.WithLabel("bar"))
```

The mocks for these functional options are unfortunately unable to match since the option usually is implemented so:

```go
type opt struct {
  timeout bool
  label string
}

type ClientOption func(*opt)

func WithTimeout() ClientOption {
  return func(o *opt) {
    o.timeout = true
  }
}

func WithLabel(v string) ClientOption {
  return func(o *opt) {
    o.label = v
  }
}

func (c *client) New(opts ...ClientOption) *someClient {
  options := &opt{}
  for _, f := range opts {
    f(options)
  }

  return &someClient{options.timeout, options.label}
}
```

`testify.Mock` matcher will not match `x := WithTimeout()` with `y := WithTimeout()` since these are actually distinct functions. Using `mock.AnythingOfType` defeats the purpose since these options are usually significant and even critical to behavior.

As a workaround currently, I'm using reflection and the `Call.Run()` (alt. `mock.MatchedBy`) function to match these `opts` and log a proper diff (different options, different type, wrong values, etc.). I thought about it and it would've been ideal if we can introduce a matcher type `mock.OptionTypeArgument` or something since the MatchedBy func is inadequate for printing meaningful diffs as it will only true/false info of the failure scenario.

Is this something that the library could/should support? If yes, I can open a PR to elicit the approach.

nbaztec:
## Summary

This PR fixes #997 

In most go libraries nowadays (etcd, kubernets, aws, etc.) the [functional options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) is being used:

```go
c := client.New(client.WithTimeout(), client.WithLabel("bar"))
```

```go
type opt struct {
  timeout bool
  label string
}

type ClientOption func(*opt)

func WithTimeout() ClientOption {
  return func(o *opt) {
    o.timeout = true
  }
}

func WithLabel(v string) ClientOption {
  return func(o *opt) {
    o.label = v
  }
}

func (c *client) New(opts ...ClientOption) *someClient {
  options := &opt{}
  for _, f := range opts {
    f(options)
  }

  return &someClient{options.timeout, options.label}
}
```

## Changes
Adds a new argument type `FunctionalOptions`

## Motivation
This wasn't possible before.

## Related issues
Closes #997 #1006

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -665,15 +665,23 @@ func IsType(t interface{}) *IsTypeArgument {
 	return &IsTypeArgument{t: t}
 }
 
+// FunctionalOptionsArgument is a struct that contains the type and value of an functional option argument
+// for use when type checking.
 type FunctionalOptionsArgument struct {
 	name  string
 	value interface{}
 }
 
+// String returns the string representation of FunctionalOptionsArgument
 func (f *FunctionalOptionsArgument) String() string {
 	return strings.Replace(fmt.Sprintf("%#v", f.value), "[]interface {}", f.name, 1)
 }
 
+// FunctionalOptions returns an FunctionalOptionsArgument object containing the functional option type
+// and the values to check of
+//
+// For example:
+// Assert(t, FunctionalOptions("[]foo.FunctionalOption", foo.Opt1(), foo.Opt2()))
 func FunctionalOptions(name string, value ...interface{}) *FunctionalOptionsArgument {
 	return &FunctionalOptionsArgument{
 		name:  name,
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -3,6 +3,7 @@ package mock
 import (
 	"errors"
 	"fmt"
+	"path"
 	"reflect"
 	"regexp"
 	"runtime"
@@ -13,6 +14,7 @@ import (
 	"github.com/davecgh/go-spew/spew"
 	"github.com/pmezard/go-difflib/difflib"
 	"github.com/stretchr/objx"
+
 	"github.com/stretchr/testify/assert"
 )
 
@@ -324,6 +326,10 @@ func callString(method string, arguments Arguments, includeArgumentValues bool) 
 	if includeArgumentValues {
 		var argVals []string
 		for argIndex, arg := range arguments {
+			if _, ok := arg.(*FunctionalOptionsArgument); ok {
+				argVals = append(argVals, fmt.Sprintf("%d: %s", argIndex, arg))
+				continue
+			}
 			argVals = append(argVals, fmt.Sprintf("%d: %#v", argIndex, arg))
 		}
 		argValsString = fmt.Sprintf("\n\t\t%s", strings.Join(argVals, "\n\t\t"))
@@ -659,6 +665,22 @@ func IsType(t interface{}) *IsTypeArgument {
 	return &IsTypeArgument{t: t}
 }
 
+type FunctionalOptionsArgument struct {
+	name  string
+	value interface{}
+}
+
+func (f *FunctionalOptionsArgument) String() string {
+	return strings.Replace(fmt.Sprintf("%#v", f.value), "[]interface {}", f.name, 1)
+}
+
+func FunctionalOptions(name string, value ...interface{}) *FunctionalOptionsArgument {
+	return &FunctionalOptionsArgument{
+		name:  name,
+		value: value,
+	}
+}
+
 // argumentMatcher performs custom argument matching, returning whether or
 // not the argument is matched by the expectation fixture function.
 type argumentMatcher struct {
@@ -797,6 +819,23 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			if reflect.TypeOf(t) != reflect.TypeOf(actual) {
 				differences++
 				output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type %s - %s\n", output, i, reflect.TypeOf(t).Name(), reflect.TypeOf(actual).Name(), actualFmt)
+			}
+		} else if reflect.TypeOf(expected) == reflect.TypeOf((*FunctionalOptionsArgument)(nil)) {
+			name := expected.(*FunctionalOptionsArgument).name
+			t := expected.(*FunctionalOptionsArgument).value
+			tName := reflect.TypeOf(t).Name()
+			if name != reflect.TypeOf(actual).String() {
+				differences++
+				output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type %s - %s\n", output, i, tName, reflect.TypeOf(actual).Name(), actualFmt)
+			} else {
+				if ef, af := assertOpts(t, actual); ef == "" && af == "" {
+					// match
+					output = fmt.Sprintf("%s\t%d: PASS:  %s == %s\n", output, i, tName, tName)
+				} else {
+					// not match
+					differences++
+					output = fmt.Sprintf("%s\t%d: FAIL:  %s != %s\n", output, i, af, ef)
+				}
 			}
 		} else {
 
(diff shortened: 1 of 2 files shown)


alameddinc (dismissed):
I think It should be inserted into the mock library 👍

dillonstreator:
Any update on this? Would love to see functional option testing support.

mock/mock.go
```diff
@@ -798,6 +828,23 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 				differences++
 				output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type %s - %s\n", output, i, reflect.TypeOf(t).Name(), reflect.TypeOf(actual).Name(), actualFmt)
 			}
+		} else if reflect.TypeOf(expected) == reflect.TypeOf((*FunctionalOptionsArgument)(nil)) {
+			name := expected.(*FunctionalOptionsArgument).name
+			t := expected.(*FunctionalOptionsArgument).value
+			tName := reflect.TypeOf(t).Name()
+			if name != reflect.TypeOf(actual).String() {
```
dillonstreator, line 835: We should be derive the `name` field on `FunctionalOptionsArgument` with a bit more reflection and simplify the `FunctionalOptions()` signature

```suggestion
			t := expected.(*FunctionalOptionsArgument).value

			var name string
			tValue := reflect.ValueOf(t)
			if tValue.Len() > 0 {
				name = "[]" + reflect.TypeOf(tValue.Index(0).Interface()).String()
			}

			tName := reflect.TypeOf(t).Name()
			if name != reflect.TypeOf(actual).String() && tValue.Len() != 0 {
```

mock/mock.go
```diff
@@ -659,6 +665,30 @@ func IsType(t interface{}) *IsTypeArgument {
 	return &IsTypeArgument{t: t}
 }
 
+// FunctionalOptionsArgument is a struct that contains the type and value of an functional option argument
+// for use when type checking.
+type FunctionalOptionsArgument struct {
+	name  string
```
dillonstreator, line 671:
```suggestion
```

mock/mock.go
```diff
@@ -659,6 +665,30 @@ func IsType(t interface{}) *IsTypeArgument {
...
+	value interface{}
+}
+
+// String returns the string representation of FunctionalOptionsArgument
+func (f *FunctionalOptionsArgument) String() string {
+	return strings.Replace(fmt.Sprintf("%#v", f.value), "[]interface {}", f.name, 1)
```
dillonstreator, line 677:
```suggestion
	var name string
	tValue := reflect.ValueOf(f.value)
	if tValue.Len() > 0 {
		name = "[]" + reflect.TypeOf(tValue.Index(0).Interface()).String()
	}

	return strings.Replace(fmt.Sprintf("%#v", f.value), "[]interface {}", name, 1)
```

mock/mock.go
```diff
@@ -659,6 +665,30 @@ func IsType(t interface{}) *IsTypeArgument {
...
+// FunctionalOptions returns an FunctionalOptionsArgument object containing the functional option type
+// and the values to check of
+//
+// For example:
+// Assert(t, FunctionalOptions("[]foo.FunctionalOption", foo.Opt1(), foo.Opt2()))
+func FunctionalOptions(name string, value ...interface{}) *FunctionalOptionsArgument {
+	return &FunctionalOptionsArgument{
+		name:  name,
+		value: value,
+	}
```
dillonstreator, line 689:
```suggestion
func FunctionalOptions(value ...interface{}) *FunctionalOptionsArgument {
	return &FunctionalOptionsArgument{
		value: value,
	}
```

mock/mock_test.go
```diff
@@ -1044,6 +1067,23 @@ func Test_Mock_AssertExpectationsCustomType(t *testing.T) {
 
 }
 
+func Test_Mock_AssertExpectationsFunctionalOptionsType(t *testing.T) {
+
+	var mockedService = new(TestExampleImplementation)
+
+	mockedService.On("TheExampleMethodFunctionalOptions", "test", FunctionalOptions("[]mock.OptionFn", OpNum(1), OpStr("foo"))).Return(nil).Once()
```
dillonstreator, line 1074:
```suggestion
	mockedService.On("TheExampleMethodFunctionalOptions", "test", FunctionalOptions(OpNum(1), OpStr("foo"))).Return(nil).Once()
```

mock/mock_test.go
```diff
@@ -1044,6 +1067,23 @@ func Test_Mock_AssertExpectationsCustomType(t *testing.T) {
...
+	assert.False(t, mockedService.AssertExpectations(tt))
+
+	// make the call now
+	mockedService.TheExampleMethodFunctionalOptions("test", OpNum(1), OpStr("foo"))
+
+	// now assert expectations
+	assert.True(t, mockedService.AssertExpectations(tt))
+
+}
+
```
dillonstreator, line 1086: Add test for empty functional options

```suggestion

func Test_Mock_AssertExpectationsFunctionalOptionsType_Empty(t *testing.T) {

	var mockedService = new(TestExampleImplementation)

	mockedService.On("TheExampleMethodFunctionalOptions", "test", FunctionalOptions()).Return(nil).Once()

	tt := new(testing.T)
	assert.False(t, mockedService.AssertExpectations(tt))

	// make the call now
	mockedService.TheExampleMethodFunctionalOptions("test")

	// now assert expectations
	assert.True(t, mockedService.AssertExpectations(tt))

}

```

dillonstreator (changes requested):
@nbaztec thoughts on these changes to simplify the signature?

dillonstreator:
@boyan-soubachov can you please provide feedback on this PR?

nbaztec:
@dillonstreator The suggestions look good, I've incorporated them. 
I was a bit hesitant at first if there could arise a situation during a refactor/upgrade, where the first parameter could be an incorrect type in the assertion, thereby using an incorrect type - but that seems like a non-issue, since the same can happen when specifying a manual string.

boyan-soubachov (approved):
Thank you :)

mock/mock.go
```diff
@@ -798,6 +832,29 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
...
+			var name string
+			tValue := reflect.ValueOf(t)
+			if tValue.Len() > 0 {
+				name = "[]" + reflect.TypeOf(tValue.Index(0).Interface()).String()
+			}
+
+			tName := reflect.TypeOf(t).Name()
+			if name != reflect.TypeOf(actual).String() && tValue.Len() != 0 {
+				differences++
+				output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type %s - %s\n", output, i, tName, reflect.TypeOf(actual).Name(), actualFmt)
```
bhingston-va, line 847: I believe I'm following the test setup nearly the same (options are from another package) and I noticed 2 things.
1. `tName` is empty for me but `reflect.TypeOf(tValue.Index(0).Interface()).String()` has the same type as `reflect.TypeOf(actual).String()`
2. `name` has the prefix `[]` so I'm hitting this branch

Curious the choice for `tName` and `reflect.TypeOf(actual).Name()` in the log here instead of `name` and `reflect.TypeOf(actual).String()` that is used in the conditional.

It was hard for me to debug as the logs showed empty string is not the same type as my option type. When debugging I found that `name` was really close except for the prefixed `[]`

I might follow up with a PR to fix the logged out line here if it was just an accident.

PR #1026, opened 2020-11-20
fix the output source file in test output

fefe982:
## Summary
The file name in the failed summary shows files inside the `testify` package, instead the real testing file. E.ge
```
--- FAIL: TestCurrentCache (0.12s)
    assertion_compare.go:221: 
                Error Trace:    xxxx_test.go:352
                Error:          Elements should be the same type
                Test:           TestXXXX
                Messages:       []
```
The `assertion_compare.go` is a file inside `testify` package. It should also show `xxxx_test.go:352` as the `Error Trace`.

## Changes
Make functions inside `testify` package as test helpers with:
```
	if h, ok := t.(tHelper); ok {
		h.Helper()
	}
```

## Motivation

With this PR, the filename and number would correctly show `xxxx_test.go:352`

### DIFF
--- a/assert/assertion_compare.go
+++ b/assert/assertion_compare.go
@@ -310,6 +310,9 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (CompareType, bool) {
 //    assert.Greater(t, float64(2), float64(1))
 //    assert.Greater(t, "b", "a")
 func Greater(t TestingT, e1 interface{}, e2 interface{}, msgAndArgs ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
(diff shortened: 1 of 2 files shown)


boyan-soubachov (commented):
Thanks for this. Unfortunately this will slow down test execution speed but it has caught a reversion. 

Could you possibly add a unit test to cover this problem and prevent the reversion from happening again in the future?

fefe982:
Add some test. The implementation of the `Helper` function in the test is directly copied for the `test` package. The test ensures that the `Helper` function is called.

boyan-soubachov (approved):
Looks great. Thank you!

PR #1035, opened 2020-12-24
Modify testify to allow for testing proto equality

edpark11:
The default testify library breaks on assert.Equal for new protobuf messages.

This modifies testify to bring in go-cmp; it then ignores the protobuf-generated fields.


edpark11:
I am very sorry-- this was an internal hack I was playing around with that was meant to be merged into a private draft fork until we figured out a better way to manage this. I accidentally forgot to change the base creating the PR. Anyways maybe it'll be helpful to someone else.

PR #1041, opened 2021-01-15
optimize isBetterMatchThan

bestgopher:

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -304,24 +304,12 @@ type matchCandidate struct {
 }
 
 func (c matchCandidate) isBetterMatchThan(other matchCandidate) bool {
-	if c.call == nil {
-		return false
-	}
-	if other.call == nil {
-		return true
-	}
-
-	if c.diffCount > other.diffCount {
-		return false
-	}
-	if c.diffCount < other.diffCount {
-		return true
-	}
-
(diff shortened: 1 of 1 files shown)


arsham:
But, why?

bestgopher:
> But, why?
I don't like to judge the Boolean of the return value of an expression, and then return a Boolean. We can directly return the value of this expression. This is just my personal habit. 

For example, in this function, These two sentences
```
if c.call.Repeatability > 0 && other.call.Repeatability <= 0 {	
		return true	
	}
return false
```
can be written like this
```
return c.call.Repeatability > 0 && other.call.Repeatability <= 0
```

Of course, if you don’t approve, I can also close this pr.

arsham:
I'm not a maintainer in this repo therefore it's not my call.

Just to quote [Rob Pike](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=14m35s)
```
Clear is better than clever.
```

bestgopher:
> I'm not a maintainer in this repo therefore it's not my call.
> 
> Just to quote [Rob Pike](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=14m35s)
> 
> ```
> Clear is better than clever.
> ```

Thank you for your suggestions and comments. @arsham

PR #1046, opened 2021-02-02
#750: added NoErrors to assert and require

### LINKED ISSUE #750: Assert multiple errors at once
A pattern that I've found pretty common when writing tests is to call many functions that return an error:
```go
err := f1()
require.NoError(t, err)

err = f2()
require.NoError(t, err)

// and so on
```

How feasible would it be to have `func NoErrors(t TestingT, errs ...error) bool` so that we can
```go
require.NoErrors(t, f1(), f2() ... fN())
```

In my head, the naive implementation would look like:
```go
// NoErrors asserts that none of the functions returned an error (i.e. `nil).
// NoErrors is a convenienve function that wraps NoError.
//   assert.NoErrors(t, fun1(), fun2())
func NoErrors(t TestingT, errs ...error) bool {
	for _, err := range errs {
		if !NoError(t, err) {
			return false
		}
	}
	return true
}
```

We would lose the possibility to annotate the errors, but we would gain readability when writing the tests.

SaeeSaadat:
## Summary
added NoErrors to assert and require

## Changes
NoErrors function in assertions.go and require.go

## Example usage 
assert.NoErrors(t, "message", possibleError1, possibleError2, possibleError3)

## Related issues
Closes #750

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1335,6 +1335,17 @@ func NoError(t TestingT, err error, msgAndArgs ...interface{}) bool {
 	return true
 }
 
+// NoErrors asserts that multiple functions return no errors
+func NoErrors(t TestingT, msg string, errors ...error) bool {
+	for index, err := range errors {
+		message := fmt.Sprintf("error found in index %d -> %s", index, msg)
+		if !NoError(t, err, message) {
+			return false
+		}
+	}
+	return true
+}
+
 // Error asserts that a function returned an error (i.e. not `nil`).
 //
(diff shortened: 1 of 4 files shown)


dolmen:
1. This is a shortcut that anyone can implement in its own test suite
2. The function signature doesn't allow to pass a customized error message like the rest of the `assert` API, so it doesn't fit well.
3. This PR is 4 years old and nobody even commented or emoji-voted on it

For those reasons, I don't see much value in adding this function, so I'm closing.

PR #1048, opened 2021-02-05
time equal added

matinFT:
assert.Equal is now able to compare time.Time objects (issue #1010)


dolmen:
That would be a breaking change as this wasn't the existing behaviour since the start of the project.

So closing.

PR #1053, opened 2021-02-21
mock: fix Maybe().Times(n) and little refactoring

grongor:
## Summary
Allow optional calls with fixed repetitions: `m.Maybe().Once() == at most once`

## Changes
Besides little refactoring for readability, the change is simple: only return errors for `m.Repeatibility > 0` if `m.optional` is false

## Motivation
I just tried to create such a mock and was surprised that there is an error. I want to have a mock of which method maybe called, and if it is, it should be called only once.

See the example in the new test.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -650,11 +650,13 @@ func (m *Mock) AssertExpectations(t TestingT) bool {
 }
 
 func (m *Mock) checkExpectation(call *Call) (bool, string) {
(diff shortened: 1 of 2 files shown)


grongor:
Hello @boyan-soubachov , could you please find some time and look at this? :)

The code is pretty old, but I just tested the functionality on the latest version, and this still doesn't work (you can't say `m.Maybe().Once() == at most once`).
If you think this should be supported, please let me know, and I'll rebase this PR onto master. 

Cheers! :tada:

grongor:
:wave: @dolmen I've rebased it

PR #1055, opened 2021-02-25
Update versions supported to include `go 1.16`

jeffwidman:
Rather than directly adding `1.16`, I thought better to reword to state the minimum version supported... this way the line doesn't have to get updated with every new release, but instead only when dropping support for old releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,7 +323,7 @@ To update Testify to the latest version, use `go get -u github.com/stretchr/test
 Supported go versions
 ==================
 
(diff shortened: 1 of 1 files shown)


jeffwidman:
nudge @boyan-soubachov

jeffwidman:
Nudge @boyan-soubachov

boyan-soubachov:
LGTM :). Thank you

boyan-soubachov (approved):

jeffwidman:
Thx!

PR #1058, opened 2021-03-01
assert.ObjectsAreEqual: improve list/map support

### LINKED ISSUE #1057: Testify `assert.EqualValues` should equate nested []interface{} by values
For slices of type []interface{}, if elements have the same value but different type, `assert.EqualValues` would fail.

e.g.
```go
// assertion would fail
assert.EqualValues(t, []interface{}{1}, []interface{}{int64(1)})
```
Now as a workaround I have to iterate over objects and compare element by element. It will be handy if `assert.EqualValues` compare two nested objects by element value.

harryzcy:
## Summary

Function `ObjectsAreEqual` now compares two objects of type array/slice/map by iterating over each element. When two objects contain the same value but have different types, the function will now return `true`.

## Changes

- The function `ObjectsAreEqualValues` in `assertion.go` is changed.
- Two helper functions `listsAreEqualValues` and `mapsAreEqualValues` are added.
- Tests are added accordingly.

## Motivation

Currently, `assert.EqualValues` would fail when two objects are of type array/slice/map and contain values of different types.

```go
// assert would fail
assert.EqualValues(t, []interface{}{1}, []interface{}{int64(1)})
```

But, `assert.EqualValues` should always compare objects by value. So the above code should idealy pass.

## Related issues

Closes #1057


joselitofilho:
Hi @harryzcy,

Thanks for your fix. Do you have any idea when this fix will be released?

harryzcy:
> Hi @harryzcy,
> 
> Thanks for your fix. Do you have any idea when this fix will be released?

No, I don't know when maintainers will review them.

assert/assertions.go
```diff
@@ -202,6 +213,68 @@ func isNumericType(t reflect.Type) bool {
 	return t.Kind() >= reflect.Int && t.Kind() <= reflect.Complex128
 }
 
+// listsAreEqualValues gets whether two lists(arrays, slices) are equal, or if their
+// values are equal.
+//
+// This function should only be used by ObjectsAreEqualValues.
+func listsAreEqualValues(expected, actual interface{}) bool {
+	expectedValue := reflect.ValueOf(expected)
+	actualValue := reflect.ValueOf(actual)
```
brackendawson, line 222: These reflect calls aren't needed until after the length checking, can we move them down.

assert/assertions.go
```diff
@@ -174,6 +174,17 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
 
 	expectedType := expectedValue.Type()
 	actualType := actualValue.Type()
+
+	if expectedType.Kind() == actualType.Kind() {
```
brackendawson, line 178: I think in the spirit of EqualValues, array and slice should also pass:
```go
assert.EqualValues(t, []interface{}{1}, [1]interface{}{int64(1)})
```

But this comparison prevents it.

assert/assertions.go
```diff
@@ -202,6 +213,68 @@ func isNumericType(t reflect.Type) bool {
 	return t.Kind() >= reflect.Int && t.Kind() <= reflect.Complex128
 }
 
+// listsAreEqualValues gets whether two lists(arrays, slices) are equal, or if their
```
brackendawson, line 216: It doesn't get if lists are equal because lists aren't comparable.

```suggestion
// listsAreEqualValues returns true if the expected and actual lists (arrays or slices) are the same length, and each index in both lists is convertible to the larger type and equal.
```

assert/assertions.go
```diff
@@ -202,6 +213,68 @@ func isNumericType(t reflect.Type) bool {
...
+	// Iterate over elements and compare
+	for i := 0; i < expectedLen; i++ {
+		if !ObjectsAreEqualValues(expectedValue.Index(i).Interface(), actualValue.Index(i).Interface()) {
+			return false
+		}
+	}
+	return true
+}
+
+// mapsAreEqualValues gets whether two maps are equal, or if their
```
brackendawson, line 243: As above:

```suggestion
// mapsAreEqualValues returns true if all the keys in the expected and actual maps are convertible to the larger type and equal.
```

brackendawson (changes requested):
This is a similar change to #1586 and I think it can be a welcome one.

Currently the doc for EqualValues says:
> EqualValues asserts that two objects are equal or convertible to the larger type and equal. 

That would need to change because this behavior is quite different. Probably a second sentence describing the behavior for lists and maps. Though I've often been hesitant to have a function with different behavior based on the value of its arguments. Would it be better to have a new assertion for this? Genuine question as I'm undecided.

Apologies for the delay in this review. I advise you read my last comment first.

assert/assertions.go
harryzcy: I updated it to support equality between slice and array.

assert/assertions.go
```diff
@@ -202,6 +213,68 @@ func isNumericType(t reflect.Type) bool {
...
+	expectedKeys := expectedValue.MapKeys()
+	actualKeys := actualValue.MapKeys()
+	if len(expectedKeys) != len(actualKeys) {
+		return false
+	}
+
+	// Key types should be convertible
+	expectedKeyType := expectedValue.Type().Key()
+	actualKeyType := actualValue.Type().Key()
+	if !expectedKeyType.ConvertibleTo(actualKeyType) {
```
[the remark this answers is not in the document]
harryzcy: Good question. I updated this part and do direct comparison on keys only

harryzcy:
I am very late to addressing those comments. Apologies.

I have addresses those comments, and the PR is ready for review again.

> Currently the doc for EqualValues says:
> 
> > EqualValues asserts that two objects are equal or convertible to the larger type and equal.
> 
> That would need to change because this behavior is quite different. Probably a second sentence describing the behavior for lists and maps. Though I've often been hesitant to have a function with different behavior based on the value of its arguments. Would it be better to have a new assertion for this? Genuine question as I'm undecided.

I'm also not very sure on this. It does seems like slightly different behavior for different types are confusing. Would like to hear what other people think? If more people are leaning toward having a new assertion, I'm happy to do that instead.

PR #1060, opened 2021-03-05
assert.InDeltaSlice: fix panic

### LINKED ISSUE #1059: Panic in assert.InDeltaSlice
`assert.InDeltaSlice` panics when the number of elements in expected slice is fewer than that in actual.

go: 1.16
testify : 1.7.0

~~~go
package main

import (
	"testing"

)

func TestInDeltaSlice(t *testing.T) {
	expect := []float32{1.0, 2.0}
	actual := []float32{1.03, 1.98, 3.01}

	assert.InDeltaSlice(t, expect, actual, 0.03)
}
~~~

expect has 2 elements, and actual has 3 elements.

~~~
--- FAIL: TestInDeltaSlice (0.00s)
panic: reflect: slice index out of range [recovered]
        panic: reflect: slice index out of range

goroutine 35 [running]:
testing.tRunner.func1.2(0x11aa5e0, 0x1224ae0)
        /Users/egawata/git/github.com/golang/go/src/testing/testing.go:1144 +0x332
testing.tRunner.func1(0xc000184600)
        /Users/egawata/git/github.com/golang/go/src/testing/testing.go:1147 +0x4b6
panic(0x11aa5e0, 0x1224ae0)
        /Users/egawata/git/github.com/golang/go/src/runtime/panic.go:965 +0x1b9
reflect.Value.Index(0x11a6c20, 0xc00011a000, 0x97, 0x2, 0x11a9320, 0xc00011602c, 0x18d)
        /Users/egawata/git/github.com/golang/go/src/reflect/value.go:962 +0x1f2
github.com/stretchr/testify/assert.InDeltaSlice(0x1227c40, 0xc000184600, 0x11a6c20, 0xc00011a000, 0x11
a6c20, 0xc00011a018, 0x3f9eb851eb851eb8, 0x0, 0x0, 0x0, ...)
        /Users/egawata/go/pkg/mod/github.com/stretchr/testify@v1.7.0/assert/assertions.go:1198 +0x405
command-line-arguments.TestInDeltaSlice(0xc000184600)
        /tmp/in_delta_slice/in_delta_slice_test.go:13 +0x145
testing.tRunner(0xc000184600, 0x11fb298)
        /Users/egawata/git/github.com/golang/go/src/testing/testing.go:1194 +0xef
created by testing.(*T).Run
        /Users/egawata/git/github.com/golang/go/src/testing/testing.go:1239 +0x2b3
FAIL    command-line-arguments  0.290s
FAIL
~~~

egawata:
## Summary

Fix assert.InDeltaSlice to fail instead of to panic when num of elements in expected slice is fewer than actual.

## Changes

When num of elements in expected is fewer than actual, `assert.InDeltaSlice` will fail rather than panic.

The result of test in https://github.com/stretchr/testify/issues/1059#issue-823147438 would be like this.

~~~
--- FAIL: TestInDeltaSlice (0.00s)
    in_delta_slice_test.go:13:
                Error Trace:    in_delta_slice_test.go:13
                Error:          Too few elements in expected slice
                Test:           TestInDeltaSlice
FAIL
FAIL    command-line-arguments  0.224s
FAIL
~~~

## Motivation

Test would stop immediately when one of test panics without recover. At that time the rest of test cases would be skipped.  
Instead, we could continue to run tests by failing.

## Related issues

Closes #1059 
Related to #1263

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1497,6 +1497,10 @@ func InDeltaSlice(t TestingT, expected, actual interface{}, delta float64, msgAn
 	actualSlice := reflect.ValueOf(actual)
 	expectedSlice := reflect.ValueOf(expected)
 
+	if expectedSlice.Len() < actualSlice.Len() {
+		return Fail(t, fmt.Sprintf("Too few elements in expected slice"), msgAndArgs...)
+	}
+
 	for i := 0; i < actualSlice.Len(); i++ {
 		result := InDelta(t, actualSlice.Index(i).Interface(), expectedSlice.Index(i).Interface(), delta, msgAndArgs...)
 		if !result {
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1497,14 +1497,14 @@ func InDeltaSlice(t TestingT, expected, actual interface{}, delta float64, msgAn
 	actualSlice := reflect.ValueOf(actual)
 	expectedSlice := reflect.ValueOf(expected)
 
-	if expectedSlice.Len() < actualSlice.Len() {
-		return Fail(t, fmt.Sprintf("Too few elements in expected slice"), msgAndArgs...)
+	expectedLen := expectedSlice.Len()
+	if !Len(t, actual, expectedLen, msgAndArgs...) {
+		return false
 	}
 
-	for i := 0; i < actualSlice.Len(); i++ {
-		result := InDelta(t, actualSlice.Index(i).Interface(), expectedSlice.Index(i).Interface(), delta, msgAndArgs...)
-		if !result {
-			return result
+	for i := 0; i < expectedLen; i++ {
+		if !InDelta(t, actualSlice.Index(i).Interface(), expectedSlice.Index(i).Interface(), delta, msgAndArgs...) {
+			return false
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


assert/assertions.go
```diff
@@ -1194,6 +1194,10 @@ func InDeltaSlice(t TestingT, expected, actual interface{}, delta float64, msgAn
 	actualSlice := reflect.ValueOf(actual)
 	expectedSlice := reflect.ValueOf(expected)
 
+	if expectedSlice.Len() < actualSlice.Len() {
```
brackendawson, line 1197: Taking line 1225 below as an example, `InDeltaMapValues` tests that the expected and actual maps are the same length, and then because of line 1237 that the maps have identical keys.

I think it would make assert more consistent if this tested that both slices have equal length. The downside is that this could cause tests that currently pass to fail.

dolmen: We should definitely check for exact same length: that's what we do in `InEpsilonSlice`.
That issue has been reported as #1263.

dolmen (changes requested):
Must check for exact length. Use the same code as in `InEpsilonSlice`.

egawata:
@dolmen Thanks for the review!
I've updated the code to check for exact length using Len(), following the same approach as InEpsilonSlice.
https://github.com/stretchr/testify/pull/1060/commits/e495998c98af0056cd70fb2b6acc340107be6654
Could you take another look?

PR #1074, opened 2021-05-04
Improve mock.MatchedBy failed comparison Diff message

perrydunn:
## Summary
Improve the Diff message for `mock.MatchedBy` functions taking a reference type.

## Changes
The `argumentMatcher.String` method internally now uses the `reflect.Type.String` method instead of the `reflect.Type.Name` method.

## Motivation
Prior to this change a `mock.MatchedBy` function taking a reference type, such as a slice (`func([]int) bool`), would have the Diff presented as `... func() bool`; now the Diff will accurately be presented as `... func([]int) bool`.

## Related issues
N/A

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -728,7 +728,7 @@ func (f argumentMatcher) Matches(argument interface{}) bool {
 }
 
 func (f argumentMatcher) String() string {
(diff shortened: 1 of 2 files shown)


perrydunn:
@boyan-soubachov Please could you take a look?

boyan-soubachov (approved):
Thank you for your contribution :)

Issue #1078, opened 2021-05-22
Diffs on time.Time objects nested inside structs produce hard to read diffs


shawc71:
We recently observed tough to read diffs produced by testify on assertion failures on structs when the structs contain nested `time.Time` objects.

I have created this repo with this testfile that does a minimal repro of the issue https://github.com/shawc71/testify-issue-demo/blob/master/example_test.go

In summary, given a struct like:

```
type structWithTime struct {
        	someTime   *time.Time
        	someString string
}
```

the following produces a [diff that is 1220 lines+](https://github.com/stretchr/testify/files/6527166/test_output.txt) and unparsable for humans:

```
		expectedTime, err := time.Parse("2006-01-02", "2020-05-22")
		assert.NoError(t, err)
		expected := &structWithTime{
			someTime:   &expectedTime,
			someString: "some val",
		}

		loc, err := time.LoadLocation("America/New_York")
		assert.NoError(t, err)
		actualTime, err := time.ParseInLocation("2006-01-02", "2020-05-21", loc)
		actual := &structWithTime{
			someTime:   &actualTime,
			someString: "some val",
		}
		assert.Equal(t, expected, actual)
```
      
I believe this was an unintended side-effect of https://github.com/stretchr/testify/pull/895. 

This makes doing assertions on structs with time.Time pretty rough. I imagine having a time.Time in a struct and being able to do assertions on those structs is a pretty common use case so I think we should make this nicer. :)

I am working on a PR to address this :)

shawc71:
Here's my attempt at fixing it: https://github.com/stretchr/testify/pull/1079

ccoVeille:
I have opened another PR to address this issue, I would like to get feedback from people who were involved in this issue :point_up: 

- #1829

PR #1079, opened 2021-05-22
Make spew config selection configurable to allow for helpful diffs on time.Time objects nested in structs.

shawc71:
## Summary
Make time.Time objects nested inside structs produce reasonable diffs by making spewConfig configurable

## Changes
https://github.com/stretchr/testify/pull/895 had some unintended side-effects on what diffs of `time.Time` see https://github.com/stretchr/testify/issues/1078.

This attempts to solve the problem by allowing the user to select if they want to use the SpewConfig that has Stringer methods enabled. The default behavior remains the same. But if the user runs their tests after setting the environment variable `TESTIFY_SPEW_STRINGER_ENABLE` to `TRUE`, testify will pick the stringer config that does not disable the Stringer methods and thus produces nice diffs for time.Time nested inside struct. 

[This](https://github.com/shawc71/testify-issue-demo/blob/master/example_test.go) contains a minimal repro of this issue, right now those tests produce an almost 1500 line diff on test failure. After this change and  when `TESTIFY_SPEW_STRINGER_ENABLE` env var is set to `TRUE` it will produce a user friendly diff like:

```
--- FAIL: TestDemo (0.00s)
    --- FAIL: TestDemo/same_tzs (0.00s)
        example_test.go:29: 
                Error Trace:    example_test.go:29
                Error:          Not equal: 
                                expected: &testify_issue_demo.structWithTime{someTime:(*time.Time)(0xc00000e460), someString:"some val"}
                                actual  : &testify_issue_demo.structWithTime{someTime:(*time.Time)(0xc00000e4a0), someString:"some val"}
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,3 @@
                                 (*testify_issue_demo.structWithTime)({
                                - someTime: (*time.Time)(2020-05-22 00:00:00 +0000 UTC),
                                + someTime: (*time.Time)(2020-05-21 00:00:00 +0000 UTC),
                                  someString: (string) (len=8) "some val"
                Test:           TestDemo/same_tzs
    --- FAIL: TestDemo/different_tzs (0.00s)
        example_test.go:47: 
                Error Trace:    example_test.go:47
                Error:          Not equal: 
                                expected: &testify_issue_demo.structWithTime{someTime:(*time.Time)(0xc00000e740), someString:"some val"}
                                actual  : &testify_issue_demo.structWithTime{someTime:(*time.Time)(0xc00000e780), someString:"some val"}
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,3 @@
                                 (*testify_issue_demo.structWithTime)({
                                - someTime: (*time.Time)(2020-05-22 00:00:00 +0000 UTC),
                                + someTime: (*time.Time)(2020-05-21 00:00:00 -0400 EDT),
                                  someString: (string) (len=8) "some val"
                Test:           TestDemo/different_tzs
FAIL
FAIL    testify_issue_demo      0.006s

```

## Motivation
Please see https://github.com/stretchr/testify/issues/1078 which describes the issue in detail

`export TESTIFY_SPEW_STRINGER_ENABLE=TRUE && go test ./...`

## Related issues
https://github.com/stretchr/testify/issues/1078

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1642,7 +1642,9 @@ func isFunction(arg interface{}) bool {
 	return reflect.TypeOf(arg).Kind() == reflect.Func
 }
 
-var spewConfig = spew.ConfigState{
+const stringerEnabledEnvVarName = "TESTIFY_SPEW_STRINGER_ENABLE"
+
+var spewConfigStringerDisabled = spew.ConfigState{
 	Indent:                  " ",
 	DisablePointerAddresses: true,
 	DisableCapacities:       true,
@@ -1657,6 +1659,20 @@ var spewConfigStringerEnabled = spew.ConfigState{
 	DisableCapacities:       true,
 	SortKeys:                true,
 	MaxDepth:                10,
+}
+
+var spewConfig spew.ConfigState
+
+func setupSpewConfig() {
+	if os.Getenv(stringerEnabledEnvVarName) == "TRUE" {
+		spewConfig = spewConfigStringerEnabled
+		return
+	}
+	spewConfig = spewConfigStringerDisabled
+}
+
+func init() {
+	setupSpewConfig()
 }
 
 type tHelper interface {
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -2062,6 +2062,54 @@ func TestDiffRace(t *testing.T) {
 	}
 }
 
+func TestDiffEnableStringer(t *testing.T) {
+	if err := os.Setenv(stringerEnabledEnvVarName, "TRUE"); err != nil {
+		t.Fatal(fmt.Sprintf("unable to set <%s> env var", stringerEnabledEnvVarName))
+	}
+	defer func() {
+		err := os.Unsetenv(stringerEnabledEnvVarName)
+		if err != nil {
+			t.Fatal(fmt.Sprintf("unable to unset <%s> env var", stringerEnabledEnvVarName))
+		}
+	}()
+	setupSpewConfig()
+
+	loc, err := time.LoadLocation("America/New_York")
+	if err != nil {
+		t.Fatal("unable to load location")
+	}
+
+	type structWithTime struct {
+		someTime   time.Time
+		someString string
+	}
+
+	structE := structWithTime{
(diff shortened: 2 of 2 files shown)


shawc71:
@boyan-soubachov would it be possible to take a look at this. I think this is a nice usability improvement.

boyan-soubachov:
Hmm. It seems like most of the feedback after #895 seems to be that `time.Time` is not very human-readable. I think it would make sense to revert back to the old behaviour and have some kind of option to enable what is the current behaviour. 

In other words:
1) The logic in this PR should become default (reverting to the previous behaviour)
2) A control should be added to enable the (what is now) current printing mode.

Would it make more sense if the control is something that's injected in the code rather than an env var. I think it would make more sense if it's something explicit and isn't coupled to environment configuration since this is a unit testing framework, after all :)

tjanez:
> In other words:
> 
>     1. The logic in this PR should become default (reverting to the previous behaviour)
>
>     2. A control should be added to enable the (what is now) current printing mode.

Agreed.
 
> Would it make more sense if the control is something that's injected in the code rather than an env var.

Agreed. I don't think this is something one would want to change depending on a particular environment, but rather an option a project could use to control how diffs are printed in a static manner.

SchumacherFM:
Can this be please merged as it is? Wouldn't change the logic as it breaks backwards compatibility.

assert/assertions.go
```diff
@@ -1642,7 +1642,9 @@ func isFunction(arg interface{}) bool {
 	return reflect.TypeOf(arg).Kind() == reflect.Func
 }
 
-var spewConfig = spew.ConfigState{
+const stringerEnabledEnvVarName = "TESTIFY_SPEW_STRINGER_ENABLE"
```
brackendawson, line 1645: Don't expose the library name in the environment variable. Something like "TESTIFY_DIFF_STRINGER_ENABLE" would be better, if we use an environment variable at all.

ccoVeille:
I have opened another PR to address this issue, I would like to get feedback from people who were involved in this current PR :point_up: 

- #1829

PR #1084, opened 2021-05-30
Fix #1083 (Nil-pointer dereference)

### LINKED ISSUE #1083: Prevent nil pointer dereference
Contains*-Assertions panic when a `nil` instead of a string, slice, map etc. is injected.
Minimal example here:

```go
var myMap map[string]interface{} // empty
// try an assertion on nested keys
// which effectively leads to a passing of nil
require.Contains(t, myMap["some-key"], "expected-nested-key") // <-- panic
```

Assertions should be robust against nil-injection and gracefully fail instead of panicking.

mvrahden:
The fix gracefully fails the assertion when given `nil` as list value instead of panicking unresolved.

Fixes #1083


mvrahden:
@ernesto-jimenez can anybody give a pass on this please? this is a trivial fix IMHO

mvrahden:
As recommended in #1095 I'll kindly ask @boyan-soubachov to review this fix and provide feedback or merge.

assert/assertions.go
```diff
@@ -721,7 +721,11 @@ func NotEqualValues(t TestingT, expected, actual interface{}, msgAndArgs ...inte
 func includeElement(list interface{}, element interface{}) (ok, found bool) {
```
boyan-soubachov, line 721: Something seems wrong with the docstring of this function... Do you mind updating the function name in line 717?

boyan-soubachov (changes requested):
Hmm. I think we need to migrate to GitHub actions since Travis killed support for open source a while ago if I recall correctly :/

boyan-soubachov:
I've been a bit swamped with life for the past year or so (who knew 2020 and 2021 would be such shockers :D)

Thanks for your contribution. I'll first migrate the builds to GitHub actions and we can then rebase and merge your PR once the builds are done :)

assert/assertions.go
mvrahden: yes, they are not synced. I fixed that and fixed further linting issues in the `assert` package.

mvrahden:
All good :) the last year was a bummer for many I guess 😅

I addressed your request with two additional commits.

There are further linting issues in other packages.
I might get back to them when I have some time and will make another PR for them.

boyan-soubachov:
Hello, do you mind merging the latest master or rebasing this PR to trigger the CI builds

mvrahden:
should be rebased now 👍

boyan-soubachov (approved):

boyan-soubachov:
LGTM, thank you for your contribution!

mvrahden:
happy to do so!

PR #1085, opened 2021-05-30
Fix NaN similarity checks

### LINKED ISSUE #1080: Make Delta* and Epsilon* funcs robust to NaN-Similarity
Currently the following funcs fail:

```golang
require.Delta(t, math.NaN(), math.NaN(), 1e-3)
require.Epsilon(t, math.NaN(), math.NaN(), 1e-3)
```

This makes comparing timeseries on similarity very tough.

Where I agree, that Equality checks on NaN-values should follow the Go-Spec and thus fail.
I disagree in the cases where we check for Similarity.

- A NaN-value is always semantically different to numerical values and thus a check like `NaN == NaN` should **fail** (see: [Issue #12025](https://github.com/golang/go/issues/12025#issuecomment-162643745)).
- In contrast: A NaN-value is not semantically much different to another NaN and thus a check like `NaN ≈ NaN` should **pass**.

Same logic applies to their `slice` and `map` derivatives.

Please have a look at #1081

mvrahden:
With this fix, NaN-similarity checks are tolerated.

Fixes #1080


mvrahden:
As recommended in #1095 I'll kindly ask @boyan-soubachov to review this fix and provide feedback or merge.

boyan-soubachov:
Hello, do you mind merging the latest master or rebasing this PR to trigger the CI builds

mvrahden:
should be rebased now 👍

assert/assertions.go
```diff
@@ -1250,19 +1254,19 @@ func InDeltaMapValues(t TestingT, expected, actual interface{}, delta float64, m
 
 func calcRelativeError(expected, actual interface{}) (float64, error) {
 	af, aok := toFloat(expected)
-	if !aok {
-		return 0, fmt.Errorf("expected value %q cannot be converted to float", expected)
+	bf, bok := toFloat(actual)
```
boyan-soubachov, line 1257: I'm concerned that this might break some expected behaviour. I'm fine to let it slide for now as master but will tag it for version 1.8.0

boyan-soubachov (approved):
LGTM, thank you for your contribution

PR #1086, opened 2021-06-02
Use %v when formatting objects in assert.Len

benjaminjkraft:
## Summary

Fix `assert.Len`'s error message for non-`%s`able values.

## Details

It was using %s, which doesn't work if the thing being len'ed is not a
`[]string` or similar.  For example, `assert.Len(t, []int{1}, 0)` says
that `[%!s(int=1)] should have 1 item(s), but has 0`.  Now it uses `%v`
which will work for almost anything.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -637,11 +637,11 @@ func Len(t TestingT, object interface{}, length int, msgAndArgs ...interface{}) 
 	}
 	ok, l := getLen(object)
 	if !ok {
-		return Fail(t, fmt.Sprintf("\"%s\" could not be applied builtin len()", object), msgAndArgs...)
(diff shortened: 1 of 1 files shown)


k1ng440 (approved):

wwade (commented):
Looks good and matches the change I was planning to make :smile:

wwade (approved):

wwade:
@glesica @boyan-soubachov @mvdkleijn could someone please have a look at this one?

dolmen:
@benjaminjkraft @k1ng440 @wwade Could you instead review #1485 which I plan to merge soon?

k1ng440:
Thank you for merging this PR.

PR #1092, opened 2021-06-22
Add argument matcher for JSON content in mocks

esdrasbeleza:
## Summary
A matcher to use when we need to check if a parameter passed to a mocked function is a JSON we expect the function to receive.

## Changes
- A function `JSONEq(expectedJSON string) argumentMatcher` in the mock package
- Add a `description` field to `argumentMatcher`, so we can use a friendly description for this new JSON matcher

## Motivation
I needed to implement some mocks for functions where a JSON string was received. The JSON fields could come in different orders, so matching wasn't very trivial and was becoming repetitive. I needed to implement custom matchers for this purpose, so I decided to open a PR with this feature. :)

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -1,6 +1,7 @@
 package mock
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
@@ -698,7 +699,8 @@ func IsType(t interface{}) *IsTypeArgument {
 // not the argument is matched by the expectation fixture function.
 type argumentMatcher struct {
 	// fn is a function which accepts one argument, and returns a bool.
-	fn reflect.Value
+	fn          reflect.Value
+	description string
 }
 
 func (f argumentMatcher) Matches(argument interface{}) bool {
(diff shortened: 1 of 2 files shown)


ratkins:
I need/want exactly this feature. Seems this PR is quite old, has something superseding it been merged already? Or is there a different/better approach?

ratkins:
Actually, I've just realised that I can use `mock.MatchedBy()` to build an ersatz argument captor and that works fine for my purposes.

esdrasbeleza:
> I need/want exactly this feature. Seems this PR is quite old, has something superseding it been merged already? Or is there a different/better approach?

Hi! I opened this PR two years ago. 

What I've been doing when I need this feature is using `Run` to run a function that gets that parameter and uses `assert.JSONEq` to verify if it's the JSON I expect:

Imagine that my interface has a function `HasName(body []byte) (bool, error)` that receives a body that might (or might not be) a valid JSON:

```go
usernameExtractor.On("HasName", mock.Anything).
		Run(func(args mock.Arguments) {
			arg := args.Get(0).([]byte)
			expectedInput := `{"name": "Esdras"}`
			assert.JSONEq(t, expectedInput, string(arg)) // Here's the matcher
		}).Return(true, nil)
```

Another approach is to use a custom matcher:

```go
	jsonMatcher := func(body []byte) bool {
		return assert.JSONEq(t, `{"name": "Esdras"}`, string(body))
	}

	// set up expectations
	usernameExtractor.On("HasName", mock.MatchedBy(jsonMatcher)).
		Return(true, nil)
```

dolmen:
At this point (maintenance resources are low), I prefer to avoid extending the API surface.

So I'm closing this issue as "won't fix".

PR #1093, opened 2021-06-28
impl EqualValues 2 objects have `(T) Equal(T) bool`

rubensayshi:
## Summary
when using `EqualValues` for 2 objects which are of the same type and implement an `(T) Equal(T) bool` method, use the method to compare.

Tbh I'd almost put this in `Equal`, but `EqualValues` is more correct...

I chose not to implement supporting `EqualValues(Foo{1}, &Foo{1})`, though I do have a branch with it implemented ...

## Motivation
This seems very much inline with many many other languages and makes it much easier to compare custom types, ranging from application level types to `decimal.Decimal`.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -179,6 +179,11 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
 	}
 
 	if !isNumericType(expectedType) || !isNumericType(actualType) {
+		if isEqualler(actualType) {
+			res := reflect.ValueOf(actual).MethodByName(equalMethod).Call([]reflect.Value{expectedValue})
+			return res[0].Bool()
+		}
+
 		// Attempt comparison after type conversion
 		return reflect.DeepEqual(
 			expectedValue.Convert(actualType).Interface(), actual,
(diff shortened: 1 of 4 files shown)


arvenil:
We had to switch to https://github.com/google/go-cmp for the same reason.

rowland66:
> We had to switch to https://github.com/google/go-cmp for the same reason.

Do you mean that you integrated google's go-cmp with stretchr, or that you used go-cmp instead. I don't think that go-cmp is a replacement for stretchr.

rubensayshi:
rebased this old MR, I still think this is a good change! we've been using a fork for years now to have this change...

we use `decimal.Decimal` extensively and `DeepEqual` doesn't work well for those because the same decimal can be represented in different ways...

brackendawson:
I am firmly against the use of this Equal method in testify's comparisons because there are examples of existing Equal methods which do not test that the instance's are "equal", including in the standard library (see time.Time).

The Equal interface has no relation to the purpose of EqualValues which was to allow people to continue to try to compare types which are not comparable, accepting the false positive and negative matches which occur when you try to do that.

rubensayshi:
@brackendawson so you might be open for a variant of his with `EqualValues` interface instead?

brackendawson:
No, we will not be changing the logic behind Equal functions any more.

PR #1094, opened 2021-06-29
Make the error message for missing argument show the method call with…

haroldcampbell:
Error message shown when a parameter is missing is not clear

## Summary
Make it easier to find argument call with a missing parameter.

## Changes
These changes were all made to the mock.go file.
Added a new `Mock#identifySource() string` function (line 763) to identify the source of the error
Added a "source" of the error to the panic in `Mock#Get(int) interface{}` method (line 792).
Added a new test called `TestArgumentCallMismatchedShowsAssertionInformation` in the mock_test.go file (line 1572).

## Motivation

If you have code like the following:
`service.On("SelectedDatastores", ActiveDatastoreGUIDs).Return(ActivatedDatastores, nil)`

If either `ActivatedDatastores` or `nil` are accidentally omitted the error message doesn't communicate that the error is due to one of those parameters being omitted .

The change is necessary as I feel it helps with my productivity when I can quickly read the log and find the problem.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -758,10 +758,40 @@ func MatchedBy(fn interface{}) argumentMatcher {
 	return argumentMatcher{fn: reflect.ValueOf(fn)}
 }
 
+// Identify the file in the callstack
+// https://gist.github.com/swdunlop/9629168
+func identifySource() string {
+	var name, file string
+	var line int
+	var pc [16]uintptr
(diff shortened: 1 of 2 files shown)


brackendawson:
This is just duplicating information that is already available in the panic stacktrace. It also is not the line where the bug is.

PR #1105, opened 2021-07-21
Feat: new assertion SameAddress close #1076

### LINKED ISSUE #1076: Assert equal pointers
I try to test some properties of zap log and I got this:

```
assert.Same(t, zapcore.ISO8601TimeEncoder, got.EncoderConfig.EncodeTime)

       Error:      Not same:
        	       expected: 0x6bb840 (func(time.Time, zapcore.PrimitiveArrayEncoder))(0x6bb840)
        	       actual  : 0x6bb840 (zapcore.TimeEncoder)(0x6bb840)
```

In this test the object types are different, but the pointer (address) are same.
Internally the function `Same` call function `samePointers`, but this function compare two object pointers and not if the address of each object are equal.
https://github.com/stretchr/testify/blob/6241f9ab994219cafb009b160a20acf4a62063aa/assert/assertions.go#L409

For me its ok this behavior, but I propose a new assert, something like `assert.SameAddress(t TestingT, referenceObject, actualObject interface{})`

This new assert just check if de `uintptr` is the same value of **referenceObject** and **actualObject** using the function `reflect.ValueOf(i interface{}).Pointer()`

black-desk:
## Summary

add a new assertion "SameAddress" which only compare address
## Changes

add a new parameter considerType, let samePointer can temporarily ignore type.

## Related issues

Closes #1076

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -705,6 +705,19 @@ func Samef(t TestingT, expected interface{}, actual interface{}, msg string, arg
 	return Same(t, expected, actual, append([]interface{}{msg}, args...)...)
 }
 
+// SameAddressf asserts that two pointers reference to the same address.
+//
+//    assert.SameAddressf(t, ptr1, ptr2, "error message %s", "formatted")
+//
+// Both arguments must be pointer variables. Pointer variable sameness is
+// determined based on the equality of value only.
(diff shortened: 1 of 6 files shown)


black-desk:
ping @mvdkleijn

PR #1110, opened 2021-08-11
assert: add partial equal assertion for JSON

### LINKED ISSUE #833: Feature Request: Support regex in JSONEq
In many situations when comparing JSON we're unable to know the exact contents of the JSON, however we can test the structure and the format. As an example we may have an API that returns a particular JSON that contains a number that is random or changing and not something that is easily mocked.

There's probably a few different approaches that could be taken that I can think of:

- Could we make it easier to partially test JSON by writing additional JSON functions that partially test a JSON, similar to how [Rspec's `include_json`](https://relishapp.com/waterlink/rspec-json-expectations/docs/json-expectations/include-json-matcher) works.

- Could we make it possible to define regular expressions for string fields in a JSON when comparing? The comparison JSON would probably need to be a type or map of some kind, and then it would be comparing two maps where strings were compared to regular expressions, and non-string values are converted to strings for comparison to a regex if one is present.

There are probably many other ways we could make it easier to test JSON.

alan2wong:
## Summary
Allows for the assertion of partial expected JSON against the actual.

## Changes
- Added function `JSONPartialEqf` to assert that the expected JSON construct exists within the actual JSON construct. This is done through the usage of the `prune` function that first removes all branches of the actual JSON that do not exist in the expected. The remaining branches are then compared with `DeepEqual`.

## Motivation
When using Terratest to test against either the returned plan JSON or the returned state JSON, it is impossible in all cases to provide a 100% complete expected JSON to compare against the actual due to the amount of data returned. This new assert function will allow for partial equivalence to allow for validating only a key portion of the plan/state against what is expected without needing to further refine the returned actual.

This simple example shows that a partial tree can be checked for existence with the actual return:

```
a.JSONPartialEq(`{"foo": {"foo": "bar"}}`, `{"hello": "world", "foo": {"foo": "bar", "hello": "world"}}`)
```

## Related issues
This may partially resolve #833.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1611,8 +1611,8 @@ func prune(a, b *interface{}) {
 	case []interface{}:
 		temp := []interface{}{}
 
-		for i, aValue := range aTyped {
-			if bArray, ok := (*b).([]interface{}); ok {
+		if bArray, ok := (*b).([]interface{}); ok {
+			for i, aValue := range aTyped {
 				if len(bArray) > i {
 					prune(&aValue, &bArray[i])
 					temp = append(temp, aValue)
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1565,6 +1565,65 @@ func JSONEq(t TestingT, expected string, actual string, msgAndArgs ...interface{
 	return Equal(t, expectedJSONAsInterface, actualJSONAsInterface, msgAndArgs...)
 }
 
+// JSONPartialEq asserts that the expected JSON construct exists within
+// the actual. Note that while the order of the properties does not
+// matter, the order of elements within an array does matter. This can
+// be circumvented by using an empty map to properly offset the expected.
+//
+//    assert.JSONPartialEq(t, `{"foo": {"foo": "bar"}}`, `{"hello": "world", "foo": {"foo": "bar", "hello": "world"}}`)
+//    assert.JSONPartialEq(t, `{"foo": [{}, {"hello": "world"}]}`, `{"hello": "world", "foo": [{"foo": "bar"}, {"hello": "world"}]}`)
+func JSONPartialEq(t TestingT, expected string, actual string, msgAndArgs ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+
+	var expectedJSONAsInterface, actualJSONAsInterface interface{}
+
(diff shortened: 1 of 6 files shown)


wuhan005 (dismissed):

wuhan005:
@boyan-soubachov Cloud you please review this PR and provide feedback or merge? Thanks a lot.

noetix:
Could someone please review this? LGTM

breeswish:
Maybe this can be called JSONContains?

wuhan005 (approved):

empire:
@boyan-soubachov Could you please review this PR and provide a feedback? Thanks a lot.
CC @glesica, @mvdkleijn

yordis:
Any opportunity to merge this one?

assert/assertions.go
```diff
@@ -1565,6 +1565,65 @@ func JSONEq(t TestingT, expected string, actual string, msgAndArgs ...interface{
...
+	switch aTyped := (*a).(type) {
+	case map[string]interface{}:
+		if bMap, ok := (*b).(map[string]interface{}); ok {
+			for k, aValue := range aTyped {
+				if bValue, ok := bMap[k]; ok {
+					prune(&aValue, &bValue)
+					aTyped[k] = aValue
+				} else {
+					delete(aTyped, k)
+				}
```
yordis, line 1608: We gave it a try in our local setup. But we removed this because it was hard to see exactly what was passed since anything that wasn't specified is showing up in the assertion message, so we can't tell what we got and compared it to what was expected.

yordis (changes requested):

assert/assertions.go
empire: It seems we can check them partially without pruning them. Laravel has implemented a similar assert function.
See the following link:
https://github.com/laravel/framework/blob/b9203fca96960ef9cd8860cb4ec99d1279353a8d/src/Illuminate/Testing/TestResponse.php#L772

alan2wong: The reason that I chose to prune was so that I could reuse an existing method to do the actual JSON compare. If you remove the pruning (specifically the key/value deletion) it may not correctly compare the two JSON objects based on the unit tests that are included.

The alternative is to write a partial JSON comparison from the ground up without relying on `Equal()`.

yordis: @alan2wong I see, maybe that is exactly what we needed or trying to end up doing since the intention was to ignore Timestamps and IDs from the comparison since they aren't important.

PR #1112, opened 2021-09-02
Fix interpolation using %s instead of %#v for interfaces

gustavoefeiche:
## Summary
Fix string interpolation issues using `%s` instead of `%#v`

## Changes
Switch %s with %#v inside error messages in func `Len()`
use `err.Error()` instead of only `err` to log the error message in func `NotEqual()`
Switch %s with %#v inside error messages in func `NotContains()`
Switch %s with %#v inside error messages in func `Subset()`
Switch %s with %#v inside error messages in func `NotSubset()`

## Motivation
The error messages were not clear on what was happening because the element is not being formatted correctly.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -637,11 +637,11 @@ func Len(t TestingT, object interface{}, length int, msgAndArgs ...interface{}) 
 	}
 	ok, l := getLen(object)
 	if !ok {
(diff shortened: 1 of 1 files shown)


dittrichlucas (approved):

dolmen:
@effeix  Please rebase and fix conflicts.

PR #1114, opened 2021-09-07
Clarify what's being linked

makusu2:
## Summary
Clarify that the link in question points to the default testing framework, not `testify`.

## Changes
Add clause `using its default testing framework`

## Motivation
It wasn't immediately clear to me that the link in question was bringing me to a *different* testing library's documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Features include:
 Get started:
 
   * Install testify with [one line of code](#installation), or [update it with another](#staying-up-to-date)
-  * For an introduction to writing test code in Go, see http://golang.org/doc/code.html#Testing
+  * For an introduction to writing test code in Go using its default testing framework, see http://golang.org/doc/code.html#Testing
   * Check out the API Documentation http://godoc.org/github.com/stretchr/testify
   * To make your testing life easier, check out our other project, [gorc](http://github.com/stretchr/gorc)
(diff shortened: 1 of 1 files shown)


dittrichlucas (approved):

dittrichlucas (approved):

README.md
```diff
@@ -16,7 +16,7 @@ Features include:
 Get started:
 
   * Install testify with [one line of code](#installation), or [update it with another](#staying-up-to-date)
-  * For an introduction to writing test code in Go, see https://go.dev/doc/code#Testing
+  * For an introduction to writing test code in Go using its default testing framework, see https://go.dev/doc/code#Testing
```
dolmen, line 19: I don't think that the wording is that much better. Because in fact knowing the `testing` framework is requirement to be able to use `Testify`. So this is not about `testing` being the *default* but being the *base* framework.

dolmen (changes requested):
Needs more work.

brackendawson:
> It wasn't immediately clear to me that the link in question was bringing me to a different testing library's documentation.

What information were you looking for when you had this experience?

I see that bullet's purpose as to show new gophers how to make a test function, because nothing in testify is about doing that, unlike for example pytest. As has been mentioned, the stdlib testing package and testify do different things and testify (usually) depends on the former.

PR #1119, opened 2021-10-11
mock: log callerInfo along with closest call debug output

jdavisclark:
## Summary

Adds callerInfo metadata to closest call information in unfound method calls. This is useful in instances where tests may be a bit large or unwieldy, as it makes it much more clear which mock registrations are being referenced by the debug output.

## Changes
Just output the existing callerInfo attached to mock registrations.

## Motivation
Its a nice quality of life improvement in certain types of projects.


dolmen (changes requested):
Missing tests.

PR #1120, opened 2021-10-12
Replace Apache 2.0 licensed gopkg.in/yaml.v3 with MIT licensed github.com/goccy/go-yaml

Al2Klimov:
... to be compatible with GPLv2 applications.


Al2Klimov:
We at @Icinga are developing applications under GPLv2. Unfortunately due to its viral copyleft nature it's not compatible with the Apache 2.0 license. And even more unfortunately at the moment your MIT licensed lib depends on the Apache 2.0 licensed gopkg.in/yaml.v3 which makes it unusable for us (Icinga/icingadb#381). This PR shall fix that... mostly. Now there are only permissively licensed non-golang.org/x deps ex. gopkg.in/yaml.v2 (`go mod graph |cut -d ' ' -f 2- |cut -d @ -f 1 |cut -d / -f 1-3 |sort -u`). That is only required by github.com/stretchr/testify@v1.4.0 (`go mod graph |grep -Fwe gopkg.in/yaml.v2`) which is only required by github.com/leodido/go-urn@v1.2.0 (`go mod graph |grep -Fwe github.com/stretchr/testify@v1.4.0`). Only after yet another release of this lib I can continue with fixing github.com/leodido/go-urn, etc. and make another PR for you which removes the gopkg.in/yaml.v2 dep.

Al2Klimov:
Now with goccy/go-yaml#260 it should be feature-equivalent.

dolmen:
Testify got contributions from hundreds of contributors. I don't think it is possible changing the license at this point.

=> `wontfix`

dolmen (commented):
[goccy/go-yaml](https://pkg.go.dev/github.com/goccy/go-yaml) brings [too many dependencies](https://github.com/goccy/go-yaml/blob/v1.11.3/go.mod) which should not be needed for YAML serialization:  `go-colorable`, `go-isatty`...

=> `wontfix`

PR #1124, opened 2021-10-29
assert.IsIncreasing et al: support sort.Interface

### LINKED ISSUE #1122: Support order assertions on `sort.Interface` types
Given the following type implementing [the sort interface](https://pkg.go.dev/sort#Interface)

```
type segment struct {
    left int
    right int
}

type segments []segment

func (l *segments) Len() int {
	return len([]segment(*l))
}

func (l *segments) Less(i, j int) bool {
	data := []segment(*l)
	return data[i].left < data[j].left  || (data[i].left == data[j].left && data[i].right < data[j].right)
}

func (l *segments) Swap(i, j int) {
	data := []segment(*l)
	data[i], data[j] = data[j], data[i]
}
```

I should be able to perform sort assertions

```
sample := segments([]segment{
{1, 2},
{3, 4},
})
assert.IsIncreasing(t, &sample)
```

brackendawson:
## Summary
Support collections which implement `sort.Interface` as objects to `IsIncreasing`, `IsNonDecreasing`, `IsDecreasing` & `IsNonIncreasing`.

## Changes
* Fixed unrelated bug in `IsIncreasing`, `IsNonDecreasing`, `IsDecreasing` & `IsNonIncreasing`, if a non-collection is passed as object then the assertion returns false without failing the test. 😱
* Make `IsIncreasing`, `IsNonDecreasing`, `IsDecreasing` & `IsNonIncreasing` into helper functions, they already should be and their generated siblings already are.
* If object passed to `IsIncreasing`, `IsNonDecreasing`, `IsDecreasing` or `IsNonIncreasing` implements `sort.Interface`, check its ordering using that interface. This is done first as a performance concern because the algorithm does not use reflection.

## Motivation
If you have a collection which implements `sort.Interface`, you can now assert the four ordering primitives against it. For `IsIncreasing` this could have been expressed as `assert.True(t, sort.IsSorted(object))`, but for `IsNonDecreasing`, `IsDecreasing` or `IsNonIncreasing` you would have to write the algorithm in the test.

## Related issues
closes #1122

### DIFF
--- a/assert/assertion_order.go
+++ b/assert/assertion_order.go
@@ -9,7 +9,7 @@ import (
 func isOrdered(t TestingT, object interface{}, allowedComparesResults []CompareType, failMessage string, msgAndArgs ...interface{}) bool {
 	objKind := reflect.TypeOf(object).Kind()
 	if objKind != reflect.Slice && objKind != reflect.Array {
-		return false
+		return Fail(t, fmt.Sprintf("object %T is not a collection", object), msgAndArgs...)
 	}
 
 	objValue := reflect.ValueOf(object)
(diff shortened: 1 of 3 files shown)


dittrichlucas (approved):

dolmen:
This will have to be rebased once #1430 is merged.

PR #1126, opened 2021-11-09
arrays value types in a zero-initialized state are considered empty

adamluzsi:
## Summary

With this pull request, arrays value types in a zero-initialized state are considered empty.
Before that, only the zero-length arrays were considered empty.
The new behavior is backward compatible with the existing currently test coverage.

## Changes

- added test cases for `assert.Empty` with array value types.
- added test cases for `assert.NotEmpty` with array value types.
- Adjust implementation according to the new test cases
- minor refactoring

## Motivation

When I was reading the code source for learning purposes.
I discovered by chance that the `assert.isEmpty / case reflect.Array` code part doesn't have coverage.

```go
assert.Empty(t, [1]int{})
assert.NotEmpty(t, [1]int{42})
```

## Related issues

I'm not aware of any, but I didn't deeply check through the issues.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -565,10 +565,6 @@ func isEmpty(object interface{}) bool {
 	// collection types are empty when they have no element
 	case reflect.Chan, reflect.Map, reflect.Slice:
 		return objValue.Len() == 0
-	// array types are empty when they match their zero-initialized state
-	case reflect.Array:
-		zero := reflect.Zero(objValue.Type()).Interface()
-		return ObjectsAreEqual(zero, object)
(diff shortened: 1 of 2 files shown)


adamluzsi:
Any news on reviewing this PR?

adamluzsi:
@evanphx @ernesto-jimenez @andreas @vkryukov

boyan-soubachov (approved):
Makes sense. Thank you for your contribution :)

funvit:
Note: `uuid.Nil` from `github.com/google/uuid` now works as expected with `assert.Empty`.

PR #994, opened 2020-08-20
Colorize 'expected' and 'actual'

hcmlinj:
## Summary

Colorize "expected" and "actual"

## Changes

Resolves #946 

## Motivation

1. @boyan-soubachov says "It's an interesting idea."

1. @mvdkleijn says "I like this idea."

```go
package testify

import (
	"errors"
	"testing"

	"github.com/stretchr/testify/assert"
)

func TestExamples(t *testing.T) {
	wrapT := assert.WrapColorize(t)
	newAssert := assert.New(wrapT)

	assert.Equal(wrapT, 1, 2)
	assert.Equal(wrapT, "A", "b")
	newAssert.Equal(3, 4)

	assert.Same(wrapT, new(assert.Colorize), new(assert.Colorize))
	newAssert.Same(new(assert.Colorize), new(assert.Colorize))

	assert.EqualValues(wrapT, uint32(123), int32(124))
	newAssert.EqualValues(uint32(123), int32(124))

	assert.EqualError(wrapT, errors.New("foo"), "bar")
	newAssert.EqualError(errors.New("foo"), "bar")
}
```

## Related issues

N/A

### TestColorize
TestColorize

### Equal
Equal

### Same
Same

### EqualValues
EqualValues

### EqualError
EqualError

PR #1127, opened 2021-11-23
Color output

### LINKED ISSUE #946: Colorize "expected" and "actual"
Can we have an option to colorize "expected" and "actual" in test output? Like this:
Screenshot from 2020-05-13 18-27-59

0_ZTerrO_akuDWSV8h

For me the colors helps because it allows me to locate the diff faster. This kind of coloring is very common in Node.js test frameworks.

walles:
## Summary

If `TESTIFY_COLOR=true` is in the environment, then `expected` and `actual` values will be coloured in green and red respectively.

## Motivation

Stolen from #994:
> 1. @boyan-soubachov says "It's an interesting idea."
> 2. @mvdkleijn says "I like this idea."

Unlike #994, this PR is configured once for the whole run by setting an environment variable.

testicolor

## Related issues
* Closes #946 
* Related to #994

### DIFF
--- a/assert/colorize.go
+++ b/assert/colorize.go
@@ -1,12 +1,11 @@
 package assert
 
 import (
+	"fmt"
 	"os"
 	"regexp"
 	"strconv"
 	"sync"
-
-	"golang.org/x/term"
 )
 
 const (
@@ -20,8 +19,10 @@ var (
 )
 
 func figureOutShouldColorize() bool {
-	// Auto: Color terminal output but not piped output
-	autoResponse := term.IsTerminal(int(os.Stdout.Fd()))
+	// FIXME: The auto response should be true on terminals and false on
+	// not-terminals, but my attempts at checking os.Stdout() terminalness have
+	// all failed. Patches welcome!
+	autoResponse := false
 
 	hint := os.Getenv("TESTIFY_COLOR")
 	if len(hint) == 0 {
@@ -38,7 +39,7 @@ func figureOutShouldColorize() bool {
 		return value
 	}
 
-	// FIXME: Print a warning here
+	fmt.Printf("WARNING: TESTIFY_COLOR should be either true or false, was: \"%s\"\n", hint)
 
 	return autoResponse
 }
--- a/assert/colorize.go
+++ b/assert/colorize.go
@@ -1,0 +1,21 @@
+package assert
+
+import "regexp"
+
+const (
+	expected_colon = "expected:"
+	actual_colon   = "actual  :"
+)
+
+func colorize(plainText string) string {
+	re := regexp.MustCompile(`(.*)\n` + expected_colon + `(.*)\n` + actual_colon + `(.*)`)
+
+	return string(
+		re.ReplaceAll(
+			[]byte(plainText),
+			[]byte("\033[33m${1}\033[0m\n"+
+				expected_colon+"\033[32m${2}\033[0m\n"+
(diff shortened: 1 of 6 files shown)


walles:
@boyan-soubachov could you have a look at this?

walles:
@alxn you had opinions on the sibling PR which should not be an issue with this one.

Do you think this one would work?

alxn:
Using an env-var to turn it on seems fine. At a later date someone can try to do the `isatty` check.

vegerot:
Any updates?

vegerot:
While I understand the concerns about portability, file redirection, etc., it feels like all the other test runners I know (GoogleTest, Jest, Deno Test, etc.) all show colors when appropriate (jest does mess up sometimes when piping into other programs).  

While I know this is a tricky problem to solve, it is definitely not intractable given these other frameworks

assert/colorize.go
```diff
@@ -0,0 +1,72 @@
...
+	if shouldColorize {
+		return doColorize(plainText)
+	}
+	return plainText
+}
+
+// Colorize unconditionally. You probably should call colorize() instead of this
+// function.
+func doColorize(plainText string) string {
+	re := regexp.MustCompile(`(.*)\n` + expected_colon + `(.*)\n` + actual_colon + `(.*)`)
```
ysmood, line 62: Will it be false positive the content of output that contains those keywords?

walles: Getting the colors wrong would be possible if:
* `failureMessage` contains linefeeds
* one or more lines in `failureMessage` **starts with** either `expected:` or `actual  :` (note `:` and spacing)

Just having the words `expected` or `actual` in the `failureMessage` would not trigger any miscoloring.

So yes, getting the colors wrong here is possible.

assert/colorize.go
```diff
@@ -0,0 +1,72 @@
...
+	setShouldColorize sync.Once
+)
+
+func figureOutShouldColorize() bool {
+	// FIXME: The auto response should be true on terminals and false on
+	// not-terminals, but my attempts at checking os.Stdout() terminalness have
+	// all failed. Patches welcome!
+	autoResponse := false
+
+	hint := os.Getenv("TESTIFY_COLOR")
```
GerardRodes, line 27: Instead if defining a new env variably could we define this behavior based on the current environment?
Maybe with something like this: https://wiki.archlinux.org/title/Color_output_in_console#Enumerate_supported_colors

This is how people is doing it in js:
https://github.com/chalk/supports-color/blob/ee88ebcfca649233bb35c9b0db226059883a77b8/index.js#L46

walles: That JS function takes a `streamIsTTY` parameter:

```
function _supportsColor(haveStream, {streamIsTTY, sniffFlags = true} = {}) {
```

And I don't know how to figure that out :(

https://github.com/stretchr/testify/blob/c602661a25d7372ccd53be44c8dd6445068eb336/assert/colorize.go#L21-L25

... so I can't really do it like that unfortunately.

walles: And in the long run, somebody will want to override the default behavior, and then the environment variable is likely to be needed anyway.

walles: @GerardRodes is this OK or do you have some suggestion for how to improve it?

walles:
I think leaving this open doesn't provide any value, closing.

Feel free to reopen if a reviewer becomes available!

Issue #757, opened 2019-04-25
go mod inconsistencies


AdamSLevy:
When updating this package using `go get -u github.com/stretchr/testify` the following is added to the `go.mod` file:

```
require (
        github.com/stretchr/objx v0.2.0 // indirect
        github.com/stretchr/testify v1.3.0
)
```

The indirect dependency is surprising because the go.mod file declared by testify uses objx v0.1.0

More odd is that after running `go mod tidy` I get the following:

```
require (
        github.com/davecgh/go-spew v1.1.1 // indirect
        github.com/stretchr/testify v1.3.0
)
```

I would like to see this repo's go.mod file to be updated using `go mod tidy` so that this indirect dependency won't clutter up other go.mod files that import testify. This will also require a minor version bump.

AeroHil:
I am also inquiring about this. It's been a year, has anyone got a work around for this?

boyan-soubachov:
If I upgrade the dependencies, `go-spew` moves from 1.1.0 to 1.1.1, `objx` moves from `0.1.0` to `0.2.0` and `gopkg.in/yaml.v3` moves from `v3.0.0-20200313102051-9f266ea9e77c` to `v3.0.0-20200506231410-2ff61e1afc86`.

Are there any explicit reasons you would like us to upgrade the dependency versions? We prefer to stick to known versions until/unless an issue has been discovered :) 

I'd be happy to bump the versions up if there are any documented issues causing problems with the repo or other packages depending on us.

dolmen:
Old issue no longer relevant.

PR #1130, opened 2021-12-03
chore(gomod): use stable yaml version

georgettica:
this might solve all of the erros of go-get

partially fixes https://github.com/stretchr/testify/issues/757

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -19,7 +19,7 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/pmezard/go-difflib/difflib"
-	yaml "gopkg.in/yaml.v3"
+	yaml "gopkg.in/yaml.v2"
 )
 
 //go:generate sh -c "cd ../_codegen && go build && cd - && ../_codegen/_codegen -output-package=assert -template=assertion_format.go.tmpl"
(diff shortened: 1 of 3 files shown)


SVilgelm:
I'm just curious what issues do you have with go get so you want to use v2 instead of v3?

georgettica:
I am going through security suits, and one of the things is using tagged versions and not the `go get *@main`
Also, this makes updating more confusing

georgettica:
The tests I was running were on github.com/ossf/scorecard

SVilgelm:
hmm, what is wrong with `go get *@main`, why is the `go get *@vN` considered more secure? It only depends how a maintainer releases the projects. There are a lot of packages that have no tags or branches (except main) at all, but it does not make them less secure.

The problem here is that the maintainer of go-yaml does not follow best practices in how to release the libs. Actually he didn't release any versions for v3, also v2 branch has no updates since Nov 17, 2020.

Is there any ope CVE against yaml.v3, that are not fixed in v3.0.0-20200313102051-9f266ea9e77c?

SVilgelm:
And v2 branch has two open issues: https://osv.dev/list?q=yaml&affected_only=true&page=1&ecosystem=Go
and those issues are fixed in v3 and cannot be fixed in v2 with our breaking backwards compatibility

PR #1133, opened 2021-12-16
mock.AssertCalled: fix error output

monkslc:
## Summary
Fixes the error output for [`Mock.AssertCalled`](https://pkg.go.dev/github.com/stretchr/testify#Mock.AssertCalled).

## Changes
Filter by method name when iterating over method calls on the mock instance.

## Motivation
When AssertCalled can't find a call matching the method name and arguments the caller was looking for, it generates an error message for the caller showing all calls on the mock instance. When generating this error message, it doesn't filter by method name. This means that calls to other mocked methods during the test pollute the output. This commit adds filtering by method name when generating the error message so the caller only sees calls to the method being asserted on.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -560,7 +560,9 @@ func (m *Mock) AssertCalled(t TestingT, methodName string, arguments ...interfac
 	if !m.methodWasCalled(methodName, arguments) {
 		var calledWithArgs []string
 		for _, call := range m.calls() {
-			calledWithArgs = append(calledWithArgs, fmt.Sprintf("%v", call.Arguments))
+			if call.Method == methodName {
+				calledWithArgs = append(calledWithArgs, fmt.Sprintf("%v", call.Arguments))
+			}
 		}
 		if len(calledWithArgs) == 0 {
 			return assert.Fail(t, "Should have called with given arguments",
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -96,6 +96,7 @@ func (i *TestExampleImplementation) TheExampleMethodFuncType(fn ExampleFuncType)
 // MockTestingT mocks a test struct
 type MockTestingT struct {
 	logfCount, errorfCount, failNowCount int
+	errorfCalls                          []Arguments
 }
 
 const mockTestingTFailNowCalled = "FailNow was called"
@@ -104,8 +105,9 @@ func (m *MockTestingT) Logf(string, ...interface{}) {
 	m.logfCount++
 }
 
-func (m *MockTestingT) Errorf(string, ...interface{}) {
+func (m *MockTestingT) Errorf(format string, args ...interface{}) {
 	m.errorfCount++
+	m.errorfCalls = append(m.errorfCalls, append(Arguments{format}, args...))
 }
 
 // FailNow mocks the FailNow call.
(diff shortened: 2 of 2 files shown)


dolmen:
Please show a test case running on https://go.dev/play .

DMRobertsonOctopus:
> Please show a test case running on https://go.dev/play .

I'm not sure how to use go.dev/play for this, since I don't think it can pull in external packages. But I hope https://go.dev/play/p/CmCN89PvjKj illustrates what the original reporter had in mind.

<details><summary>Source code</summary>

```go
package internal_test

import (
	"testing"

)

type Person interface {
	SetName(string)
	GetName()
}

type MockPerson struct {
	mock.Mock
}

func (m *MockPerson) SetName(name string) {
	m.Called(name)
}

func (m *MockPerson) GetName() {
	m.Called()
}

var _ Person = &MockPerson{}

func Test_PR1133(t *testing.T) {
	// Create a no-op MockPerson.
	m := &MockPerson{}
	m.Test(t)
	m.On("SetName", mock.IsType("")).Maybe()
	m.On("GetName").Maybe()

	var p Person = m
	p.GetName()
	p.SetName("John")
	p.GetName()
	p.SetName("Alice")
	p.GetName()

	m.AssertCalled(t, "SetName", 1)
}
```
</details>

To summarise: the test body makes 5 calls to a `mock.Mock`, then uses `AssertCalled` to make an assertion about those calls.

```go
	// Create a no-op MockPerson.
	m := &MockPerson{}
	m.Test(t)
	m.On("SetName", mock.IsType("")).Maybe()
	m.On("GetName").Maybe()

	var p Person = m
	p.GetName()
	p.SetName("John")
	p.GetName()
	p.SetName("Alice")
	p.GetName()

	m.AssertCalled(t, "SetName", "Bob)
```

The test fails (as expected), but the message explaining why the assertion failed is quite verbose:

```shellsession
$ go test internal/example_test.go
--- FAIL: Test_PR1133 (0.00s)
    example_test.go:42:
        	Error Trace:	/Users/dmr/workspace/investments-api/internal/example_test.go:42
        	Error:      	Should have called with given arguments
        	Test:       	Test_PR1133
        	Messages:   	Expected "SetName" to have been called with:
        	            	[Bob]
        	            	but actual calls were:
        	            	        []
        	            	[John]
        	            	[]
        	            	[Alice]
        	            	[]
FAIL
FAIL	command-line-arguments	0.351s
FAIL
```

This message describes calls to `SetName`, but also to `GetName`. (It's not too bad here, but when there are calls with many arguments or arguments whose string representation is large, it gets hard to read.) Speaking as user, I'd find it helpful if the `GetName` calls were omitted from this message, i.e.

```
Expected "SetName" to have been called with:
        	            	[Bob]
        	            	but actual calls were:
        	            	[John]
        	            	[Alice]
```

An alternative might be to prefix each call with the function name:

```
Expected "SetName" to have been called with:
        	            	[Bob]
        	            	but actual calls were:
        	            	        GetName: []
        	            	SetName: [John]
        	            	GetName: []
        	            	SetName: [Alice]
        	            	GetName: []
```

dolmen:
Fixed reproducer: https://go.dev/play/p/wJXHNukHWYt

Unfortunately play.golang.org timeouts.

PR #1137, opened 2021-12-21
assert.NotErrorAs: new assertion

### LINKED ISSUE #1066: testify `{assert|require}` packages seem to be missing `NotErrorAs` methods.
When using this gem of a library I realized that the `assert` and `require` package seem to be missing a convenience function for `NotErrorAs(...)`.

`{assert|require}.NotErrorIs(...)` does exist so the absence of the `error.As` equivalent feels weird.

If there is a reason for this, which I do not see, I would love to be enlightened, else I would happily create a PR for this (:

Right now one has to do the following steps to check for this (which kinda breaks the feel IMHO):

```Golang
var customErr *package.CustomErr
assert.Error(t, err)
assert.False(t, errors.As(err, &customErr))
```

or similar.

titusjaka:
## Summary

Add `NotErrorAs` methods to check that none of the errors in err's chain matches target.

## Changes
- add `NotErrorAs` and `NotErrorAsf` assertions to `assert` and `require` packages

## Motivation

Sometimes, it's essential to check that error does not match an exact type.

```
	err := errors.New("test error")

	var notFoundErr ErrNotFound
	assert.NotErrorAs(t, err, &notFoundErr)

	err = WrapNotFoundError(err)
	assert.ErrorAs(t, err, &notFoundErr)
}
```

## Related issues

Closes #1066

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -565,6 +565,15 @@ func NotEqualValuesf(t TestingT, expected interface{}, actual interface{}, msg s
 	return NotEqualValues(t, expected, actual, append([]interface{}{msg}, args...)...)
 }
 
+// NotErrorAsf asserts that at none of the errors in err's chain matches target, otherwise, sets target to that error value.
+// This is a wrapper for errors.As.
+func NotErrorAsf(t TestingT, err error, target interface{}, msg string, args ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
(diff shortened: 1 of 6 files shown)


johanneswuerbach:
I was just searching for this function, to discover it is missing.

Maybe @boyan-soubachov could have a look? :-)

Antonboom:
+1 for this PR.
and take a look to https://github.com/stretchr/testify/pull/1129

PR #1142, opened 2022-01-07
Experimental - Fuzz test support

asaha123:
## Summary

Add support for fuzz tests as introduced in Go 1.18 (still beta) and documented [here](https://go.dev/doc/fuzz/)

## Changes

TBD

## Motivation

Support fuzz tests in testify

## Example usage

Example test:

```
package foo

import (
	"testing"

)

	suite.Suite
	VariableThatShouldStartAtFive int
}

func (suite *ExampleTestSuite) SetupTest() {
	suite.VariableThatShouldStartAtFive = 5
}

	assert.Equal(suite.T(), 5, suite.VariableThatShouldStartAtFive)
	suite.Equal(5, suite.VariableThatShouldStartAtFive)
}

type ExampleFuzzSuite struct {
	suite.Suite
}

func (suite *ExampleFuzzSuite) FuzzAFunc() {
	f := suite.F()
	f.Add("foo")
	f.Fuzz(func(t *testing.T, aStr string) {
		// this fails the test run when run in fuzz mode
		//t.Fail()

		// this doesn't fail the test
		suite.EqualError(nil, "an error")
	})
}

}

func FuzzSuite(f *testing.F) {
	suite.Run(f, new(ExampleFuzzSuite))
}
```

With this modification to `testify`, the above test run in fuzz mode looks as follows:

```
$ go1.18beta1 test -fuzz=FuzzSuite -fuzztime=1s
fuzz: elapsed: 0s, gathering baseline coverage: 0/300 completed
fuzz: elapsed: 0s, gathering baseline coverage: 300/300 completed, now fuzzing with 8 workers
fuzz: elapsed: 1s, execs: 18028 (16323/sec), new interesting: 0 (total: 298)
PASS
ok  	test	1.308s
```

I expected the test to fail since i have `suite.EqualError(nil, "an error')`.

In "unit test" mode, the test fails:

```
% go1.18beta1 test                             
--- FAIL: FuzzAFunc (0.00s)
    f_test.go:36: 
        	Error Trace:	f_test.go:36
        	            				value.go:556
        	            				value.go:339
        	            				fuzz.go:332
        	Error:      	An error is expected but got nil.
        	Test:       	FuzzAFunc
    f_test.go:36: 
        	Error Trace:	f_test.go:36
        	            				value.go:556
        	            				value.go:339
        	            				fuzz.go:332
        	Error:      	An error is expected but got nil.
        	Test:       	FuzzAFunc
--- FAIL: FuzzSuite (0.00s)
FAIL
exit status 1
FAIL	test	0.165s
```


suite/suite.go
```diff
@@ -23,24 +23,49 @@ type Suite struct {
...
 	suite.t = t
 	suite.Assertions = assert.New(t)
 	suite.require = require.New(t)
 }
 
+// SetF sets the current *testing.F context.
+func (suite *Suite) SetF(f *testing.F) {
+	suite.f = f
+	suite.Assertions = assert.New(f)
+	suite.require = require.New(f)
```
boyan-soubachov, line 50: I think this might be problematic. If we called `suite.SetT(t)` then called `suite.SetF()`, would that not lose the "T" we set in `SetT(t)`?

If so, it seems like that would be the wrong behaviour? Should we not find a way to add an "F" or "T" if the relevant Assertions and requires have already been instantiated?

suite/suite.go
```diff
@@ -23,24 +23,49 @@ type Suite struct {
...
 // Require returns a require context for suite.
 func (suite *Suite) Require() *require.Assertions {
 	if suite.require == nil {
-		suite.require = require.New(suite.T())
+
+		var tT require.TestingT
+		if suite.T() != nil {
+			tT = suite.T()
+		}
+		if suite.F() != nil {
```
boyan-soubachov, line 61: Why are we checking for the existence of `suite.F` and `suite.T` when previously we weren't?

Is it possible that the existence of both of those should be guaranteed before the code reaches this point (i.e. when suite is first instantiated)?

suite/suite.go
```diff
@@ -23,24 +23,49 @@ type Suite struct {
...
+			tT = suite.F()
+		}
+
+		if tT == nil {
+			panic("Neither T nor F was non-nil")
+		}
```
boyan-soubachov, line 67: I'm not sure why we want to test for this? Is it not possible for a suite to have both non-nil `T` and `F`?

suite/suite.go
```diff
@@ -52,12 +77,19 @@ func (suite *Suite) Require() *require.Assertions {
...
+		if suite.F() != nil {
+			tT = suite.F()
+		}
+		suite.Assertions = assert.New(tT)
 	}
 	return suite.Assertions
 }
 
-func failOnPanic(t *testing.T) {
+func failOnPanic(t testing.TB) {
```
boyan-soubachov, line 92: What's the logic for expanding the interface to include benchmarks and removing the pointer here?

suite/suite.go
```diff
@@ -80,10 +112,17 @@ func (suite *Suite) Run(name string, subtest func()) bool {
 
 // Run takes a testing suite and runs all of the tests attached
 // to it.
-func Run(t *testing.T, suite TestingSuite) {
+func Run(t testing.TB, suite TestingSuite) {
```
boyan-soubachov, line 115: Same question as line 92

suite/suite.go
```diff
@@ -193,12 +232,13 @@ func runTests(t testing.TB, tests []testing.InternalTest) {
 
 	r, ok := t.(runner)
 	if !ok { // backwards compatibility with Go 1.6 and below
```
boyan-soubachov, line 234: We don't officially support Go 1.6 anymore. Can we remove this check altogether or at least remove the old backwards compatibility change and update this to reflect 1.18 and fuzzing?

boyan-soubachov (changes requested):
Thanks for the contribution! 🙂

I've put down some questions to better understand the decisions you've made.

suite/suite.go
asaha123: Good point. The current change assumes that you will not have a situation like that. From my current very limited understanding of both testify and [go fuzz tests](https://go.dev/doc/fuzz/), it seems like for having a suite of fuzz tests, you will only have a "top-level" `testing.F` context. 

The [fuzz target](https://go.dev/doc/fuzz/#glos-fuzz-target) has a `testing.T` context, but that's not relevant i think to `testify` as that gets invoked by the go test runner.

suite/suite.go
asaha123: At this stage, the current suite under test could either be a fuzz test suite (either in normal/fuzz mode) or a non-fuzz test suite. Hence, depending on which mode we are in, initialize `suite.require` appropriately. Hence, the current `panic()` where i assert that at least one of them is set. 

Again, this assumes that during a single run, for *a suite*, it can only have `testing.T` or `testing.F` set.

suite/suite.go
asaha123: Again, same assumption as above.

suite/suite.go
asaha123: The reason for replacing `*testing.T` by `testing.TB` is so that `failOnPanic()` can be called with a `testing.F` value as well (`testing.TB` interface is satisfied by `T`, `F` and `B` [source](https://sourcegraph.com/github.com/golang/go/-/blob/src/testing/testing.go?L743))

suite/suite.go
asaha123: Sounds like a good idea. Should I create a separate change to remove this and rebase on top it?

suite/suite.go
asaha123: Answered above.

suite/suite.go
```diff
@@ -193,12 +232,13 @@ func runTests(t testing.TB, tests []testing.InternalTest) {
...
+		// t.F doesn't have Run() so Fuzz tests also
+		// take this branch
 		if !testing.RunTests(allTestsFilter, tests) {
 			t.Fail()
```
asaha123, line 238: What i am observing is this which explains the behavior I am seeing (which may have been clear to you folks, but i am seeing why now):
- When the fuzz test is run in "unit test" mode, the function call `RunTests()` returns with a `false` value and hence the test fails (considering the sample test in the PR)
- When the fuzz test is run in fuzz mode, this function returns `true` 
-

asaha123: I am going to next experiment with updating the test discovery and running logic using https://pkg.go.dev/testing@go1.18beta2#InternalFuzzTarget  for the fuzz tests. i think that may be required and certainly feels correct.

asaha123: > Okay, I think i have found the issue. The problem seems to be that when we call the `suite.EqualError()` method, it reports any failures using the testing context's (`testing.T` value) methods. However, the problem is since we are running a Fuzz test, we only have a `testing.F` injected via `go test`. The `testing.T` value is created and injected into the fuzz target [as part of the test run](https://sourcegraph.com/github.com/golang/go@38174b3a3514629b84dcd76878b2f536b189dd7b/-/blob/src/testing/fuzz.go?L302). So, we need to make this `testing.T` created as part of the test run available to testify.

Hence, when i update my test as follows to make a call to the `SetT()` method, it fails as expected in fuzz mode as well:

```
type ExampleFuzzSuite struct {
	suite.Suite
}

func (suite *ExampleFuzzSuite) FuzzAFunc() {
	f := suite.F()
	f.Add("foo")
	f.Fuzz(func(t *testing.T, aStr string) {
		fmt.Println("I am in Fuzz target")
		// this fails the test run when run in fuzz mode
		//t.Fail()

		// this doesn't fail the test when run in fuzz mode if i comment
		// out the next line: https://github.com/stretchr/testify/pull/1142/files#r821269235
		suite.SetT(t)
		suite.EqualError(nil, "an error")
	})
}

}

func FuzzSuite(f *testing.F) {
	suite.Run(f, new(ExampleFuzzSuite))
}
```

asaha123:
> Thanks for the contribution! slightly_smiling_face
> 
> I've put down some questions to better understand the decisions you've made.

Thanks for the review so far. I have replied to the comments. 

I suspect we may have run into an issue with how we handle assertions and report failures using `testing.T` value currently and it's workings in the context of fuzz tests: https://github.com/stretchr/testify/pull/1142/files#r821324429

mcauto:
How's it going?

Go supports fuzzing in its standard toolchain beginning in Go 1.18. Native Go fuzz tests are [supported by OSS-Fuzz](https://google.github.io/oss-fuzz/getting-started/new-project-guide/go-lang/#native-go-fuzzing-support).

https://go.dev/security/fuzz/

PR #1149, opened 2022-01-17
assert: allow comparing time.Time

torkelrogstad:
## Summary
Add support for comparing `time>Time` and custom types derived from it.  

## Changes
When comparing types, see if we're comparing `time.Time`, or a custom type based on `time.Time`. If that's the case, compare their Unix timestamps, with the smallest possible resolution. 

## Motivation
This now allows sticking a `time.Time` into `assert.Greater` and `assert.Less`, which is nice. Time stamps have a natural ordering, so I'd expect this to work.

### DIFF
--- a/assert/assertion_compare.go
+++ b/assert/assertion_compare.go
@@ -306,7 +306,7 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (CompareType, bool) {
 	case reflect.Struct:
 		{
 			// All structs enter here. We're not interested in most types.
-			if !obj1Value.CanConvert(timeType) {
+			if !canConvert(obj1Value, timeType) {
 				break
 			}
 
--- a/assert/assertion_compare.go
+++ b/assert/assertion_compare.go
@@ -3,6 +3,7 @@ package assert
 import (
 	"fmt"
 	"reflect"
+	"time"
 )
 
 type CompareType int
@@ -30,6 +31,8 @@ var (
 	float64Type = reflect.TypeOf(float64(1))
 
 	stringType = reflect.TypeOf("")
+
+	timeType = reflect.TypeOf(time.Time{})
 )
 
 func compare(obj1, obj2 interface{}, kind reflect.Kind) (CompareType, bool) {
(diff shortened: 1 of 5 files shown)


torkelrogstad:
Had to fiddle a bit with build tags to get this working across versions. Nw tests pass locally for me with both a new (1.17.6) and old (1.14.15). Looks like CI is fine as well.

assert/assertion_compare.go
```diff
@@ -299,6 +302,27 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (CompareType, bool) {
...
+			// All structs enter here. We're not interested in most types.
+			if !canConvert(obj1Value, timeType) {
+				break
+			}
+
+			// time.Time can compared!
+			timeObj1, ok := obj1.(time.Time)
+			if !ok {
+				timeObj1 = obj1Value.Convert(timeType).Interface().(time.Time)
+			}
```
boyan-soubachov, line 317: Why are we converting to `timeType` (i.e. `time`), then interface, then back to time?

assert/assertion_compare_legacy.go
```diff
@@ -0,0 +1,11 @@
+// +build !go1.17
```
boyan-soubachov, line 1: Could you please add TODO comments in all these version-dependent stating that we can merge them back in once we're done supporting Go 1.17?

boyan-soubachov (changes requested):

assert/assertion_compare.go
torkelrogstad: `Convert()` returns a `reflect.Value`, `Interface()` takes us _out_ of the reflection wrappers, and then finally we do a type assertion. It looks a bit convoluted, but it's all necessary!

torkelrogstad:
Rebased, and added doc comments. Assumed you meant 1.16, as that's what's holding up the build tag stuff.

torkelrogstad:
Okay, this is a bit awkward... `go fmt` wants me to use a syntax that's not supported by earlier versions of Go, so I don't quite know how to make all Go versions happy here

boyan-soubachov (approved):
Thank you :)

PR #1161, opened 2022-02-28
Include locations of subtests in test failure stack traces

krobelus:
We @symflower have been using a slightly modified version of testify for years. We'd like to upstream several changes that we have found very useful for everyday testing. This is the first change (working on publishing the rest).

This change improves stack traces for subtests (as created by t.Run()) by adding the call trace of t.Run(). Hence, instead of seeing only locations inside of t.Run() the user also sees locations of their own code.

(see [our blog post](https://symflower.com/en/company/blog/2022/better-table-driven-testing/) for our testing style)

cc @zimmski


dolmen:
Your @symflower testing style doesn't use the `testing.T.Run` in the way it was designed, and so you have to bend Testify to fit get proper reporting. I'm definitely not convinced.

I wanted to show you how to refactor your `TestIsSorted`, but your blog post shows code in image, so I can't get the text :(. It would have been helpful if you had provided code that runs on https://go.dev/play .

Anyway, here is how to refactor `TestIsSorted` to NOT need the feature of this pull request:

```go
check := func(value []int, expected bool) func(t *testing.T) {
    return func(t *testing.T) {
        actual := IsSorted(value)
        assert.Equal(t, expected, actual, "%#v", value)
    }
} 

t.Run("Empty", check([]int{}, false))
t.Run("SingleElement", check([]int{0}, true))
```

assert/assertions.go
```diff
@@ -102,60 +104,72 @@ the problem actually occurred in calling code.*/
 // CallerInfo returns an array of strings containing the file and line number
 // of each stack frame leading from the current test to the assert call that
 // failed.
-func CallerInfo() []string {
+func CallerInfo(t TestingT) []string {
```
dolmen, line 107: This is a breaking change as the signature of the function changes.

assert/assertions.go
```diff
@@ -13,9 +13,11 @@ import (
 	"runtime"
 	"runtime/debug"
 	"strings"
+	"testing"
 	"time"
 	"unicode"
 	"unicode/utf8"
+	"unsafe"
```
dolmen, line 20: We don't want `unsafe` code. `Testify` must not dig more into Go implementation details because we have to maintain it on the long term.

dolmen (commented):
I think that the purpose isn't good.

The implementation isn't good either as it breaks the existing API surface and add `unsafe` code.

PR #1174, opened 2022-04-05
feat: IsError ErrorAssertionFunc

robdefeo:
## Summary
[gotest -template=testify](https://github.com/cweill/gotests/tree/develop/templates/testify) has an `assert.ErrorAssertionFunc` as one of the parameters. Testify has `assert.NoError` and `assert.Error` which asserts the functions returns an error or not. Usually this is enough but sometimes checking for a specific error is more appropriate. 

## Changes
Add `IsError` that takes an `expectedErr` as a parameter and returns an `assert.ErrorAssertionFunc` 

## Motivation
While `assert.EqualError` checks for specific errors, it is not an `assert.ErrorAssertionFunc` so can not be used in the template table tests. A new function that takes an `expectedErr` and returns an `assert.ErrorAssertionFunc` makes testing for specific errors easier and clearer.

```
func Example(input int) (bool, error) {
	switch input {
	case 1:
		return false, errors.New("1 not allowed")
	case 2:
		return false, errors.New("2 not allowed")
	default:
		return true, nil
	}
}
```
Example test
```
	type args struct {
		input int
	}
	tests := []struct {
		name      string
		args      args
		want      bool
		assertion assert.ErrorAssertionFunc
	}{
		{
			"condition-1",
			args{1},
			true,
			assert.IsError(errors.New("2 not allowed")),
		},
		{
			"condition-2",
			args{2},
			true,
			assert.IsError(errors.New("2 not allowed")),
		},
		{
			"condition-other",
			args{99},
			true,
			assert.NoError,
		},
	}
	for _, tt := range tests {
			got, err := Example(tt.args.input)
			tt.assertion(t, err)
			assert.Equal(t, tt.want, got)
		})
	}
}
```

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1337,6 +1337,21 @@ func NoError(t TestingT, err error, msgAndArgs ...interface{}) bool {
 	return true
 }
 
+// IsError asserts that a function returned a specific error.
+//
+//   actualObj, err := SomeFunction()
+//   if assert.IsError(expectedErr)(t, err) {
+//	   assert.Equal(t, expectedObj, actualObj)
+//   }
+func IsError(theError error) func(t TestingT, err error, msgAndArgs ...interface{}) bool {
+	return func(t TestingT, err error, msgAndArgs ...interface{}) bool {
+		if theError == nil {
+			return Fail(t, "An expected error is required but got nil.")
+		}
+		return EqualError(t, err, theError.Error(), msgAndArgs...)
+	}
+}
+
 // Error asserts that a function returned an error (i.e. not `nil`).
 //
 //   actualObj, err := SomeFunction()
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -1061,6 +1061,44 @@ func TestNoError(t *testing.T) {
 	False(t, NoError(mockT, err), "NoError should fail with empty error interface")
 }
 
+func TestIsError(t *testing.T) {
+
+	mockT := new(testing.T)
+
+	// start with a nil error
+	var expetedErr error
+	var err error
+
+	False(t, IsError(expetedErr)(mockT, err), "IsError should return True for nil expected error")
+
+	// now set an expected error
+	expetedErr = errors.New("expected error")
+
+	False(t, IsError(expetedErr)(mockT, err), "IsError should return True for nil arg")
+
+	// now set error to same as expected error
+	err = errors.New("expected error")
+
+	True(t, IsError(expetedErr)(mockT, err), "IsError with error should return True for same error")
+
+	// now set error to another error
+	err = errors.New("another error")
+
+	False(t, IsError(expetedErr)(mockT, err), "IsError with error should return False for different error")
+
+	// returning an empty error interface
+	err = func() error {
+		var err *customError
+		return err
+	}()
+
+	if err == nil { // err is not nil here!
+		t.Errorf("Error should be nil due to empty interface: %s", err)
+	}
+
+	False(t, IsError(expetedErr)(mockT, err), "IsError should fail with empty error interface")
+}
+
 type customError struct{}
 
 func (*customError) Error() string { return "fail" }


robdefeo:
Let me know @glesica, @boyan-soubachov, or @mvdkleijn if you have any feedback on this PR. At least one thing I wasn't sure about is the most appropriate name for this function. 😄

Antonboom:
@robdefeo why is [assert.ErrorIs](https://pkg.go.dev/github.com/stretchr/testify@v1.7.1/assert#ErrorIs) not enough?

robdefeo:
`ErrorIs` is useful, I didn't see it as I routinely used v1.6.1 😊 but it solves the problem in a different place. The naming is confusing immediately making the name I chose a poor choice. 

Let me explain the intended case again and try to clean up the confusion between the two. Most my tests are written with a `wantErr` and they checking for that error in the tests case.

If you want to use the `assertion assert.ErrorAssertionFunc` because it expects a function, you only have the option of `assert.NoError` and `assert.Error`. 
You can check for specific errors with `ErrorIs`, however you need to call the function which means it can only be done in test body not in the test case definition. Using `ErrorIs` would mean adding another param to the tests case like `wantError` which is a reasonable solution but it doesn't fit the pattern of `assert.ErrorAssertionFunc`.
I've started writing tests with `assert.ErrorAssertionFunc` and I find the tests are clearer to read as the assertion is the the test case definition, there is not way to specify the error through without a function that returns a function, hence this PR. 

If accepted, it will need a new name so open to suggests. Thanks

Antonboom:
I see. I think it would be better just to add

```go
package assertions

type ErrorIsAssertionFunc func(t TestingT, err, target error, msgAndArgs ...interface{}) bool
```
(and for require too)

And use `assert.ErrorIs`.

Your tests written in bad pattern: it's better to just check for an error (how was before) than to use its text for comparison. `EqualError`, `ErrorContains` are examples of an API that should not be used universally and is only useful in very rare cases.

Antonboom:
I see. 

Your tests written in bad pattern: it's better to just check for an error (how was before) than to use its text for comparison. `EqualError`, `ErrorContains` are examples of an API that should not be used universally and is only useful in very rare cases.

What your tests should look like in modern Go without changes in `testify`:
```go
var (
	ErrNotAllowed1 = errors.New("1 not allowed") // can be private `errNotAllowed`
	ErrNotAllowed2 = errors.New("2 not allowed")
)

func Example(input int) (bool, error) {
	switch input {
	case 1:
		return false, ErrNotAllowed1
	case 2:
		return false, ErrNotAllowed2
	default:
		return true, nil
	}
}
```

```go
	type args struct {
		input int
	}
	tests := []struct {
		name     string
		args     args
		want     bool
		expError error
	}{
		{
			"condition-1",
			args{1},
			false,
			ErrNotAllowed2,
		},
		{
			"condition-2",
			args{2},
			false,
			ErrNotAllowed2,
		},
		{
			"condition-other",
			args{99},
			true,
			nil,
		},
	}
	for _, tt := range tests {
			got, err := Example(tt.args.input)
			require.ErrorIs(t, err, tt.expError)
			assert.Equal(t, tt.want, got)
		})
	}
}
```

boyan-soubachov:
I'm with @Antonboom on this one. It's a simpler, more commonly used pattern that achieves the same legibility (maybe even better), IMO.

dolmen:
I agree with @Antonboom (https://github.com/stretchr/testify/pull/1174#issuecomment-1104725016).

I'm closing this PR as "wont fix".

Wani4ka:
Hi! Let me reincarnate this discussion :D
I got some argument for an assertion like this. Suppose we have a function, that creates a new error every time it is supposed to return an error (for example, [github.com/google/uuid.Parse](https://github.com/google/uuid/blob/0f11ee6918f41a04c201eceeadf612a377bc7fbc/uuid.go#L77)). Then, I have a function that uses it:
```go
type Request struct {
    UUID string
}
var ErrRequestIsMissing = errors.New("request is missing")

func Do(r *Request) error {
    if r == nil {
        return ErrRequestIsMissing
    }
    if _, err := uuid.Parse(r.UUID); err != nil {
        return fmt.Errorf("failed to parse uuid: %w", err")
    }
    return nil
}
```

In order to test the first scenario (request is nil), I should use `assert.ErrorIs(t, err, ErrRequestIsMissing)`. But for the second one, I can only check if error is present (`assert.Error(t, err)`). The second one satisfies `assert.ErrorAssertionFunc` signature, but the first one doesn't. So, when making table tests for this function, I have to combine both methods to keep it simple, something like this:
```go
func TestDo(t *testing.T) {
    tests := []struct{
        name string
        r *Request

        checkErr assert.ErrorAssertionFunc // for the first scenario
        wantErr error                      // for the second scenario
    }{{
        name: "nil",
        r: nil,
        wantErr: ErrRequestIsMissing,
    }, {
        name: "invalid uuid",
        r: &Request{UUID: "invalid"},
        checkErr: assert.Error,
    }, {
        name: "success",
        r: &Request{UUID: uuid.NewString()},
    }}
    for _, tt := range tests {
            err := Do(tt.r)
            if tt.checkErr != nil {
                tt.checkErr(t, err)
            } else {
                assert.ErrorIs(t, err, tt.wantErr)
            }
        })
    }
}
```

Notice the boilerplate with two branches, which could be removed if we had an ErrorAssertionFunc that uses errors.Is inside. Same applies to ErrorAs, ErrorContains, EqualError, NotErrorAs, NotErrorIs. This happens pretty often when I have to test functions, that call both controllable code (I can return assert.AnError from them using a mock) and uncontrollable (json.Unmarshal, uuid.Parse, url.Parse, etc.)

The function in question would look like this:
```go
func ErrorIsFunc(target error) assert.ErrorAssertionFunc {
    return func(t TestingT, err error, args ...interface{}) bool {
        assert.ErrorIs(t, err, target, args...)
    }
}
```
The test would look like this:
```go
func TestDo(t *testing.T) {
    tests := []struct{
        name string
        r *Request
        checkErr assert.ErrorAssertionFunc // for both scenarios
    }{{
        name: "nil",
        r: nil,
        checkErr: assert.ErrorIsFunc(ErrRequestIsMissing),
    }, {
        name: "invalid uuid",
        r: &Request{UUID: "invalid"},
        checkErr: assert.Error,
    }, {
        name: "success",
        r: &Request{UUID: uuid.NewString()},
        checkErr: assert.NoError,
    }}
    for _, tt := range tests {
            t.checkErr(t, Do(tt.r))
        })
    }
}
```

If think this can be a good idea, I can make a PR with such additions for both assert and require packages

Antonboom:
Hi, @Wani4ka 👋 

you just need to introduce your own error and be aligned with layered architecture approach

```go
var ErrInvalidUUID = errors.New("failed to parse uuid")

func Do(r *Request) error {
    if _, err := uuid.Parse(r.UUID); err != nil {
        return fmt.Errorf("%w: %v", ErrInvalidUUID, err")
    }
    return nil
}
```

Another yet option, to combine "error fact" and "specific error" (how you described above):
```go
    tests := []struct{
        name string
        wantErr bool
        errExpected error
    }

    if tt.wantErr {
        require.Error(t, err)
        if tt.errExpected != nil {
            require.ErrorIs(t, err, tt.errExpected)
        }
    } else {
        require.NoError(t, err)
    }
```

P.S. But if you want to support all possible scenarios (`errors.Is`, `errors.As`, plain error), then a separate function might actually come in handy, you are right. But I have never met functions that return such heterogeneous errors at once 🤔

ccoVeille:
I had this use case a couple of time

Here are the 2 solutions we used:

- create a local assertion helper, and use it in your loop

```go
func checkError(t *testing.T, err, want error) {
   t.Helper()

   if want == nil {
      require.NoError(t, err)
      return
   }

   require.ErrorIs(t, err, want)
}
```

- separate tests in 2 loops

```go
   success := []struct{
        name string
        wantErr bool
        errExpected error
    }

    for _, tt := range success {
        require.NoError(t, err)
    }

    failures := []struct{
        name string
        errExpected error
    }

    for _, tt := range failures {
        require.Error(t, err)
        if tt.errExpected != nil {
            require.ErrorIs(t, err, tt.errExpected)
        }
    }
```

Using the second option often brings clarity.

brackendawson:
Here's a functional pattern for error assertions I've settled on for many years now. It doesn't have boilerplate in the test function, it strictly checks the error (types of errors returned by your API are part of your API), and it always checks both the return type and the error.

```go
package kata_test

import (
	"testing"

	"github.com/brackendawson/kata"
)

func TestIfy(t *testing.T) {
	for name, test := range map[string]struct {
		dividend, divisor, expected int
		requireError                require.ErrorAssertionFunc
	}{
		"10/2=5": {
			dividend:     10,
			divisor:      2,
			expected:     5,
			requireError: require.NoError,
		},
		"1/0=:-S": {
			dividend:     1,
			divisor:      0,
			expected:     0,
			requireError: errorEqual(kata.ErrDivideByZero),
		},
	} {
			t.Parallel()
			actual, err := kata.Divide(test.dividend, test.divisor)
			test.requireError(t, err)
			require.Equal(t, test.expected, actual)
		})
	}
}

func errorEqual(expected error) require.ErrorAssertionFunc {
	return func(t require.TestingT, err error, msgAndArgs ...interface{}) {
		require.Equal(t, expected, err, msgAndArgs...)
	}
}
```

There is valid argument that testify could include such high order functions for error assertions whose signature doesn't match ErrorAssertionFunc.

Wani4ka:
> you just need to introduce your own error and be aligned with layered architecture approach
> 
> ```go
> var ErrInvalidUUID = errors.New("failed to parse uuid")
> 
> func Do(r *Request) error {
>     if _, err := uuid.Parse(r.UUID); err != nil {
>         return fmt.Errorf("%w: %v", ErrInvalidUUID, err")
>     }
>     return nil
> }
> ```

@Antonboom This is actually a good idea, but I still think such error assertion functions may come in handy, because such an approach may not be that obvious and conventional to everyone when writing tests :thinking:

Issue #1173, opened 2022-04-01
Would love a MapSubset


SleepyBrett:
or perhaps expand subset so it can also handle maps.

given
```
a := map[string]string{
		"one": "foo",
		"two": "bar",
		"three": "baz",
	 }

b := map[string]string{
		"one": "foo",
		"three": "baz",
	 }

assert.MapSubset(t, a, b, "a does not contain b") // will pass
assert.MapSubset(t, b, a, "b does not contain a") // will fail
```

brackendawson:
Enhancing Subset would definitely be better IMO.

One question, should this assert true?
```go
a := map[string]string{
		"one": "foo",
		"two": "bar",
	 }

b := map[string]string{
		"one": "baz",
	 }

assert.Subset(t, a, b)
```

Does it test that the key and value are in the superset, or does it follow two value dereference (`_, ok := a["one"]`) which only checks the key? It would need to be documented.

SleepyBrett:
it should test both the key and the value.

OladapoAjala:
@brackendawson What is the process to having this proposal accepted? If accepted @SleepyBrett can I work on it?

brackendawson:
That would be open a pull request and then hope a maintainer returns.

OladapoAjala:
@brackendawson sure thanks, would create a PR.

OladapoAjala:
Hello @brackendawson @SleepyBrett I created a PR for this issue here: https://github.com/stretchr/testify/pull/1178

tisonkun:
FYI testify `Contains` on Map type validates by map keys, not entries.

eikenb:
While looking for this exact functionality I came across this ticket and #704 which looks like a duplicate.

PR #1178, opened 2022-04-20
fix: assert.MapSubset (or just support maps in assert.Subset)

OladapoAjala:
## Summary
Extended Subset & NotSubset to handle maps

## Changes
Added some logic to check if each subsets key-value pair are present in the list.

## Motivation
To handle maps in Subset & NotSubset

## Related issues
#1173


boyan-soubachov (approved):
Looks good, thank you for your contribution :)

PR #1188, opened 2022-05-23
Add WithinTimeRange method

moolmanruan:
## Summary
Add `WithinTimeRange` and `WithinTimeRangef` assertion methods.

## Changes
Adds the `WithinTimeRange` assertion function that checks whether the _expected_ time is on or after the _start_ time and on or before the _end_ time. If _end_ is before _start_ the function also fails.

Adds `WithinTimeRange` and `WithTimeRangef` methods on `*Assertions` that calls the `WithinTimeRange` function internally. 

## Motivation
There are cases where one would like to test that a time value is within a certain range, for example when the code being tested updates a timestamp:
```
before := time.Now()
actual := Foo()
after := time.Now()

assert.WithinTimeRange(t, actual.timeStamp, before, after)
```

Currently, the only time-related assertion method is `WithinDuration`, but in the above mentioned situation it leads to one of the following solutions:
- Using two `assert.True` checks against the `before` and `after` times.
 ```
before := time.Now()
actual := Foo()
after := time.Now()

assert.True(t, actual.timeStamp.After(before))
assert.True(t, actual.timeStamp.Before(after))
```
- Deriving a _duration_ from the `before` and `after` times to use in the `WithDuration` assertion.
```
before := time.Now()
actual := Foo()
after := time.Now()
halfDur := after.Sub(before) / 2

assert.WithinDuration(t, actual.timeStamp, before.Add(halfDur), halfDur)
```
- Using `WithDuration` to assert against `time.Now()` with an arbitrary duration, leading to flaky tests.
```
before := time.Now()
actual := Foo()
after := time.Now()

assert.WithinDuration(t, actual.timeStamp, time.Now(), time.Second)
```

## Related issues
https://github.com/stretchr/testify/pull/689

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1118,7 +1118,7 @@ func WithinRange(t TestingT, actual, start, end time.Time, msgAndArgs ...interfa
 	}
 
 	if end.Before(start) {
-		return Fail(t, "start should be before end", msgAndArgs...)
+		return Fail(t, "Start should be before end", msgAndArgs...)
 	}
 
 	if actual.Before(start) {
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1112,7 +1112,7 @@ func WithinDuration(t TestingT, expected, actual time.Time, delta time.Duration,
 // WithinRange asserts that a time is within a time range (inclusive).
 //
 //   assert.WithinRange(t, time.Now(), time.Now(), time.Now())
-func WithinRange(t TestingT, expected, start, end time.Time, msgAndArgs ...interface{}) bool {
+func WithinRange(t TestingT, actual, start, end time.Time, msgAndArgs ...interface{}) bool {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: assert/assertions.go [hunk 1]
```diff
@@ -1109,6 +1109,27 @@ func WithinDuration(t TestingT, expected, actual time.Time, delta time.Duration,
 	return true
 }
 
+// WithinTimeRange asserts that a time is within a time range (inclusive).
+//
+//   assert.WithinTimeRange(t, time.Now(), time.Now(), time.Now())
```
boyan-soubachov, line 1114: I'm of the opinion that this example isn't the most intuitive. Would it make more sense if we changed it to something like:
```
assert.WithinTimeRange(t, time.Now(), time.Now().Sub(time.Second()), time.Now().Add(time.Second()))
```

If you can think of a simpler example that would explicitly demonstrate the differences between start and end, that would be awesome :)

assert/assertions.go
```diff
@@ -1109,6 +1109,27 @@ func WithinDuration(t TestingT, expected, actual time.Time, delta time.Duration,
...
+func WithinTimeRange(t TestingT, expected, start, end time.Time, msgAndArgs ...interface{}) bool {
```
boyan-soubachov, line 1115: 1. For consistency with `WithinDuration`, should this function not be named `WithinRange`, the `time.Time` arguments explicitly state that the range being tested is a time range?
2. Semantically, the 2nd argument should be called `actual`, not `expected` since it's the value we're testing

assert/assertions.go
```diff
@@ -1109,6 +1109,27 @@ func WithinDuration(t TestingT, expected, actual time.Time, delta time.Duration,
...
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+
+	if end.Before(start) {
+		return Fail(t, "start should be before end", msgAndArgs...)
```
boyan-soubachov, line 1121: Please capitalise for consistency

boyan-soubachov (changes requested):

assert/assertions.go
moolmanruan: Yeah, I'll think about it a bit and have a look at other examples to see if there's a better way to present it.

moolmanruan: Changed to `assert.WithinTimeRange(t, time.Now(), time.Now().Add(-time.Second), time.Now().Add(time.Second))`.

boyan-soubachov (approved):
Looks great! Thank you for your contribution.

I hope you're enjoying your time at Luno, I had a blast there :)

PR #1198, opened 2022-06-05
integrate github.com/pmezard/go-difflib

### LINKED ISSUE #736: Remove unmaintained dependency go-difflib
The package `github.com/pmezard/go-difflib` is as of this commit no longer maintained: https://github.com/pmezard/go-difflib/commit/1ef1646f464bff6853dfa4c32750886762dc023a

Maybe we could move the used functionality into this repo and stop depending on the `difflib` package all together?

The functions used are [`difflib.GetUnifiedDiffString()`](https://github.com/pmezard/go-difflib/blob/5d4384ee4fb2527b0a1256a821ebfc92f91efefc/difflib/difflib.go#L635-L639) and [`difflib.SplitLines()`](https://github.com/pmezard/go-difflib/blob/5d4384ee4fb2527b0a1256a821ebfc92f91efefc/difflib/difflib.go#L768-L772).

### LINKED ISSUE #1159: Find an alternative for go-difflib
As github.com/pmezard/go-difflib is no longer maintained (EOLed), kindly find an alternative.

### LINKED ISSUE #1187: go-difflib is no longer maintained
It appears that https://github.com/pmezard/go-difflib is no longer maintained, but is still a dependency of this project. Are there any plans to remove it/fork and maintain a copy of it?

obsti8383:
## Summary
Since go-difflib is unmaintained since quite some time, the required functions have been taken over into a separate testify package.

## Motivation
Unmaintained packages might vanish or be taken over by attackers. 

## Related issues
Closes #1187 
Closes #1159 
Closes #736

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,3 +344,5 @@ License
 =======
 
 This project is licensed under the terms of the MIT license.
+
+difflib/ is Copyright (c) 2013, Patrick Mezard, see [difflib/LICENSE](difflib/LICENSE).
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -18,7 +18,8 @@ import (
 	"unicode/utf8"
 
 	"github.com/davecgh/go-spew/spew"
-	"github.com/pmezard/go-difflib/difflib"
+	"github.com/stretchr/testify/difflib"
+
 	yaml "gopkg.in/yaml.v3"
 )
 
--- a/difflib/LICENSE
+++ b/difflib/LICENSE
@@ -1,0 +1,27 @@
+Copyright (c) 2013, Patrick Mezard
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+    Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+    Redistributions in binary form must reproduce the above copyright
+notice, this list of conditions and the following disclaimer in the
+documentation and/or other materials provided with the distribution.
+    The names of its contributors may not be used to endorse or promote
+products derived from this software without specific prior written
+permission.
+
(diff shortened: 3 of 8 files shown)


obsti8383:
> It's a good idea but would it not be easier to just pin to the latest version in our `go.mod` file and break on any hash changes (`go.sum`, we already do that in the CI jobs).

go.mod already PINs the version, so that doesn't change anything. Pulling everything into this repo gives the advantage that you don't upgrade to a malicious version by accident.

Main advantage of moving all used code into this repo is that you get rid of concerns by using an unmaintained source code, by maintaining it yourself (see all three issues).

I've also included only the used functions, which makes the code base smaller.
 
> I get the problem of what happens if the package vanishes, would vendoring not be an easier solution then?

As far as I understand vendoring it doesn't bring any advantages here, since command like "go get" will still use the original repository and you don't get rid of the dependency.

miguelalcantar (commented):
Everything seems good to me. Also I need this upgrade

miguelalcantar:
@boyan-soubachov could we have a look at this?

ilkka:
Would love to see this integrated as well.

TReadOnly:
Looking forward to having github.com/pmezard/go-difflib integrated.

bmv126:
Any update on this ?
Can this be merged.

dolmen:
We have enough code to maintain and not enough maintainer.

We should instead investigate using another library that provides the same features.

bmv126:
How about https://github.com/martinohmann/go-difflib ?

dolmen (commented):
Move package difflib to internal/difflib.

dolmen:
@miguelalcantar Why do you "need" that change?

obsti8383:
> Move package difflib to internal/difflib.

@dolmen You closed this PR 3 weeks ago. What do you mean?

Issue #1210, opened 2022-06-22
Mock.On no longer acquires a lock


creachadair:
As part of the change in #1182 to change the Mock mutex to a pointer, the `Mock.On` method was changed to re-construct the mutex each time it is called. 

This appears to have also introduced another regression (related to but in addition to #1208): Previously it was safe to call `On` concurrently with other operations on a Mock, but now because `On` resets that pointer, concurrent calls can race (observed in v1.7.4).

A small repro is attached to tickle the basic issue (`go test -race` should suffice).

Arguably it's silly for us to be calling `On` concurrently with other operations, but regrettably we have some code that relies on the ability to do so, and it worked before v1.7.4 but no longer does.

[repro.zip](https://github.com/stretchr/testify/files/8958555/repro.zip)

creachadair:
This appears to have been addressed by #1212, thanks!

Issue #1208, opened 2022-06-21
AssertNotCalled() panics because mutex is nil


gkdr:
hello,

thank you for your continued work on this library.

i have an issue that for some reason only came up today, but i believe this is a similar issue to #1206. i have one test which only checks that a function was not called using `AssertNotCalled()`, which now fails because the mutex is nil and testify panics.

i instantiate the mock using `new`, is that wrong?

i am willing to prepare a PR for you, but i am not sure what the fix is. in #1207, a check is added:
```go
	if m.mutex == nil {
		m.mutex = &sync.Mutex{}
	}
```

does this need to be added to all of `AssertNotCalled()`, `AssertCalled()`, `AssertNumberOfCalls()`? is that all or are there any other places?

PrafulFP:
I think a fix is also needed for `IsMethodCallable` and `MethodCalled`.

Kryvchun:
The same about `AssertNumberOfCalls`

psmarcin:
Using `yourmock.Test` method before `yourmock.AssertNumberOfCalls` populates `mutex`. It's just temporary workaround because this change looks like breaking change for me.

yurishkuro:
Having a similar issue - upgrading to 1.7.4 causes some tests to panic.

Just curious, what was the motivation for changing mutex to a pointer in #1182 ?  The PR did not provide any reasoning, and considering that the mocks are usually initialized as plain structs, having a pointer is asking for trouble.

peczenyj:
> I think a fix is also needed for `IsMethodCallable` and `MethodCalled`.

Indeed, I have a m.Called() without On on my programs -- the called was not expected, should fail, not panic on nil reference

boyan-soubachov:
Fixed in latest version (1.7.5), please confirm and let me know if all is good :)

PrafulFP:
Working fine now. Thanks 👍🏻

gkdr:
reverting the change to a pointer fixed the issue. thank you!

PR #1212, opened 2022-06-23
fixing panic in calls to assertion with nil m.mutex

wenlock:
## Summary

@Cleancut and I found these changes that should be reverted as they are causing panics after library updates.

This reverts a change that was made in https://github.com/stretchr/testify/pull/1182
That PR makes m.mutex a pointer which now needs to be checked but it's not checked for nil everywhere.
It also reset the mutex in the `.On` function when it could actually have locks from other callers.

## Changes
- revert `m.mutex` back to a non pointer value
- don't throw away the lock that might be used concurrently in the `.On` function

## Motivation
Fixing panics and crashes in the library after a minor update.

## Related issues
This should also help with these issues:
- https://github.com/stretchr/testify/issues/1208
- https://github.com/stretchr/testify/issues/1210
- https://github.com/stretchr/testify/issues/1206 (introduces a workaround while keeping the new pointer)

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -521,7 +521,7 @@ func AssertExpectationsForObjects(t TestingT, testObjects ...interface{}) bool {
 	for _, obj := range testObjects {
 		if m, ok := obj.(*Mock); ok {
 			t.Logf("Deprecated mock.AssertExpectationsForObjects(myMock.Mock) use mock.AssertExpectationsForObjects(myMock)")
-			obj = &m
+			obj = m
 		}
 		m := obj.(assertExpectationser)
 		if !m.AssertExpectations(t) {
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -255,7 +255,7 @@ type Mock struct {
 	// this data completely allowing you to do whatever you like with it.
 	testData objx.Map
 
-	mutex *sync.Mutex
+	mutex sync.Mutex
 }
 
 // String provides a %v format string for Mock.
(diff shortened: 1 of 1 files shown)


wenlock:
Hey @boyan-soubachov sorry to ping you directly but since you were the original author for https://github.com/stretchr/testify/pull/1182 I thought I'd bring this one to your attention. Thanks for any review you might be able to offer.

mock/mock.go
```diff
@@ -526,9 +519,9 @@ func AssertExpectationsForObjects(t TestingT, testObjects ...interface{}) bool {
 		h.Helper()
 	}
 	for _, obj := range testObjects {
-		if m, ok := obj.(Mock); ok {
+		if m, ok := obj.(*Mock); ok {
```
boyan-soubachov, line 522: Are you sure this should be changed? IIRC I didn't change this logic in the PR I merged?

wenlock: Yes, because we want to refer to the same instance of sync.Mutex for the copy that is a type check here.

wenlock: Also, `go vet ./...` will show that warning for us ( I think it was caught in CI).

boyan-soubachov: Fair enough; I see it's a test helper anyways.

Thank you for catching this!

boyan-soubachov (approved):

mock/mock.go
boyan-soubachov: 🤦 . I must've missed this when doing the Go 1.18 PR (which broke this) and 'fixed it' the long way around. Thanks

brackendawson: This was in fact incorrect. It makes the entire if statement into a no-op and introduced a new panic in #1227

PR #1219, opened 2022-06-29
Nil actual type made AnythingOfType matcher panic

### LINKED ISSUE #1209: Mock panics when argument of nil type checked against AnythingOfType
Reported by @cfiderer: https://github.com/stretchr/testify/issues/1155#issuecomment-1126159634

Raising a new issue because I think the fix to 1155 will not fix this.

This deadlocks in testify v1.7.4:
```go
package kata

import (
	"testing"

)

type mockThing struct {
	mock.Mock
}

func (m *mockThing) Do(arg interface{}) string {
	return m.Called(arg).String(0)
}

func TestTestify(t *testing.T) {
	m := &mockThing{}
	defer m.AssertExpectations(t)
	m.On("Do", mock.AnythingOfType("string")).Return("Nae")
	m.Do(nil)
}
```

brackendawson:
## Summary
Passing an actual value of type nil via a Mock into the argument matcher for AnythingOfType in a Mock causes a panic.

## Changes
Handle nil types up front when checking actual types against mock.AnythingOfType before calling any methods on the reflect.Type, which [will be nil](https://pkg.go.dev/reflect#TypeOf) if the value was a nil interface.

## Motivation
Someone may expect their mock to be called with a nil interface, we should not panic in this case.

## Related issues
Closes #1209

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -909,10 +909,15 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			}
 		} else if reflect.TypeOf(expected) == reflect.TypeOf((*AnythingOfTypeArgument)(nil)).Elem() {
 			// type checking
-			if reflect.TypeOf(actual).Name() != string(expected.(AnythingOfTypeArgument)) && reflect.TypeOf(actual).String() != string(expected.(AnythingOfTypeArgument)) {
+			if reflect.TypeOf(actual) == nil {
+				if expected != AnythingOfTypeArgument("<nil>") {
+					differences++
+					output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type <nil> - %s\n", output, i, expected, actualFmt)
+				}
+			} else if reflect.TypeOf(actual).Name() != string(expected.(AnythingOfTypeArgument)) && reflect.TypeOf(actual).String() != string(expected.(AnythingOfTypeArgument)) {
 				// not match
 				differences++
-				output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type %s - %s\n", output, i, expected, reflect.TypeOf(actual).Name(), actualFmt)
+				output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type %s - %s\n", output, i, expected, reflect.TypeOf(actual).String(), actualFmt)
 			}
 		} else if reflect.TypeOf(expected) == reflect.TypeOf((*IsTypeArgument)(nil)) {
 			t := expected.(*IsTypeArgument).t
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -1613,6 +1613,50 @@ func Test_Arguments_Diff_WithAnythingOfTypeArgument_Failing(t *testing.T) {
 
 }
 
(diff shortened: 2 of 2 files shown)


mock/mock_test.go
```diff
@@ -1613,6 +1613,50 @@ func Test_Arguments_Diff_WithAnythingOfTypeArgument_Failing(t *testing.T) {
 
 }
 
+func Test_Arguments_Diff__WithAnythingOfTypeArgument_NilType(t *testing.T) {
```
boyan-soubachov, line 1616: You have an extra underscore here, please look at test names above to maintain consistency :)

mock/mock_test.go
```diff
@@ -1613,6 +1613,50 @@ func Test_Arguments_Diff_WithAnythingOfTypeArgument_Failing(t *testing.T) {
...
+
+	var args = Arguments([]interface{}{AnythingOfType("<nil>")})
+	var count int
+	_, count = args.Diff([]interface{}{nil})
+
+	assert.Zero(t, count)
+
+}
+
+func Test_Arguments_Diff__WithAnythingOfTypeArgument_NilType_Failing(t *testing.T) {
```
boyan-soubachov, line 1626: See comment on line 1616

mock/mock.go
```diff
@@ -909,10 +909,15 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			}
 		} else if reflect.TypeOf(expected) == reflect.TypeOf((*AnythingOfTypeArgument)(nil)).Elem() {
 			// type checking
-			if reflect.TypeOf(actual).Name() != string(expected.(AnythingOfTypeArgument)) && reflect.TypeOf(actual).String() != string(expected.(AnythingOfTypeArgument)) {
+			if reflect.TypeOf(actual) == nil {
+				if expected != AnythingOfTypeArgument("<nil>") {
+					differences++
+					output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type <nil> - %s\n", output, i, expected, actualFmt)
+				}
+			} else if reflect.TypeOf(actual).Name() != string(expected.(AnythingOfTypeArgument)) && reflect.TypeOf(actual).String() != string(expected.(AnythingOfTypeArgument)) {
```
boyan-soubachov, line 917: According to the `String()` function [documentation](https://pkg.go.dev/reflect#Type):
```
	// String returns a string representation of the type.
	// The string representation may use shortened package names
	// (e.g., base64 instead of "encoding/base64") and is not
	// guaranteed to be unique among types. To test for type identity,
	// compare the Types directly.
```

we should not be using it for comparison.

Do you mind updating and fixing the logic to reflect this?

boyan-soubachov (changes requested):

greenstatic:
I ran into this issue as well. Is there any activity on this PR fixing this issue?

brackendawson:
> I ran into this issue as well. Is there any activity on this PR fixing this issue?

I believe the state is that fixing this *well* would be a breaking change, so not until Testify 2.0, and I think we should do that. But the comments above object to something which is not a regression, so I think this change can be merged into 1.x.

brackendawson:
#1799 is a better approach.

Issue #912, opened 2020-03-09
Add `EqualEnum` function


Nokel81:
### Problem:
Currently when asserting a specific enum value is set I have been using either `assert.Equal(t, val1, val1)` or `assert.Equal(t, val1.String(), val2.String())`.

The first is, imo, better from the point of view of correctness but the second one produces better error messages. I know that I could implement `GoString()` but that has several problems. 
1. It seems like changing code for an implementation detail in my testing package.
2. Some of the enums I am checking are defined in my dependencies which are out of my control, so I can't even define `GoString()` even if I wanted to.

### Proposal:
Add the following functions in both `testify/assert` and `testify/require`:
```golang
func EqualEnum(t *testing.T, expected interface{}, actual interface{}, msgAndArgs ...interface{}) {
    ...
}

func EqualEnumf(t *testing.T, expected interface{}, actual interface{}, msg string, args ...interface{}) {
    ...
}

func NotEqualEnum(t *testing.T, expected interface{}, actual interface{}, msgAndArgs ...interface{}) {
    ...
}

func NotEqualEnumf(t *testing.T, expected interface{}, actual interface{}, msg string, args ...interface{}) {
    ...
}
```

These functions would act very similarly to `Equal` currently, but in the error output formatting would use the `String()` function ("%v") not the `GoString()` function.

boyan-soubachov:
Can you tell us what `val1` and `val2` look like?

As it stands, Go doesn't have an enum type. It's usually implemented as a collection of constants with the same type so I think there could be multiple definitions of "enum equality". Can you explain in the detail how the comparison would work?

Nokel81:
`val1` and `val2` are of `type MyVal int` or `type MyVal uint` or something like that.

I know that go doesn't have an enum type, only the constants. I would say that the comparison should be `val1 == val2` but print using `%v` instead of `%#v`. So it would display something like `MyVar1 != MyVar2` if the command was `assert.EnumEqual(t, MyVar2, act)` and `act = MyVar1`

brackendawson:
It's not exactly a duplicate. But if #400 ends up setting the format for **All numeric types** to `"%[1]T(%[1]v)"` then this is solved also.

PR #1223, opened 2022-07-07
Clean uint formatting on mismatch

### LINKED ISSUE #400: Equal diff prints ints in hex format
`Equal` and `EqualValues` print ints in hex format when showing diff. It would be much more useful to see `1701`, instead of `0x6a5`.

```
Error Trace:    equality_test:34
        Error:          Not equal:
                        expected: float64(1700)
                        received: uint(0x6a5)
        Messages:       Expecting prices to be equal
```

mnotti:
## Summary

Cleans the formatting on uint mismatches.

## Changes

I added uint types to switch on type in `formatUnequalValues` to use `%v` instead of `%#v` which was giving weird hex display.

I also refactored the accompanying unit test to use a more conventional testCase struct loop.

## Motivation

All uints show weird hex value when displaying diffs which is confusing and annoying when debugging failed tests. 
https://go.dev/play/p/hEUdfo-pVlP 

Seems to be a known issue since 2017 and no one has done anything so I thought I'd take a quick stab at it. 
(https://github.com/stretchr/testify/issues/400) 

## Related issues

Closes #400


mnotti:
@boyan-soubachov could i get some quick eyes on this one if you get some cycles?

dolmen:
LGTM.

@mnotti Could you rebase your branch on current master?

assert/assertions.go
```diff
@@ -435,7 +435,7 @@ func formatUnequalValues(expected, actual interface{}) (e string, a string) {
 			fmt.Sprintf("%T(%s)", actual, truncatingFormat(actual))
 	}
 	switch expected.(type) {
-	case time.Duration:
+	case time.Duration, uint, uint8, uint16, uint32, uint64:
 		return fmt.Sprintf("%v", expected), fmt.Sprintf("%v", actual)
```
dolmen, line 439: Use `fmt.Sprint(x)` instead of `fmt.Sprintf("%v", x)`.

dolmen (changes requested):

dolmen:
Simplified test case to confirm the issue on latest release: https://go.dev/play/p/6xEleN5v45P

assert/assertions.go
mnotti: Done.

mnotti:
> @mnotti Could you rebase your branch on current master?

Done.

vitorfalcaor:
Hey @dolmen would you be able to take a look at this PR, please? Just spent some time debugging a test case thinking that `assert.Equal`, for some mysterious reason, was comparing memory addresses, then I realized it was actually a formatting issue. Not sure if @mnotti is still able to rebase and update the PR, so happy to take this to the finish line.

assert/assertions.go
```diff
@@ -504,8 +504,8 @@ func formatUnequalValues(expected, actual interface{}) (e string, a string) {
 			fmt.Sprintf("%T(%s)", actual, truncatingFormat(actual))
 	}
 	switch expected.(type) {
-	case time.Duration:
-		return fmt.Sprintf("%v", expected), fmt.Sprintf("%v", actual)
+	case time.Duration, uint, uint8, uint16, uint32, uint64:
+		return fmt.Sprint(expected), fmt.Sprint(actual)
```
brackendawson, line 508: Formatting these with "%[1]T(%[1]v)" would be better so that we don't say two apparently identical numbers are different. Otherwise we regress `assert.Equal(t, int(1), uint(1))`.

brackendawson (changes requested):

assert/assertions.go
brackendawson: Also, using this format for all numeric types would fix #912, but this is creeping somewhat, perhaps better in a follow up PR.

brackendawson:
I'm happy doing the work to review that the changes in the unit test are correct in this PR. I see an argument where these additions tip the threshold of making this into a tabled test. My previous review comments are still un-addressed though.

PR #1235, opened 2022-07-20
assert.Len: avoid printing the value of very large objects

Inuart:
## Summary
When testing the `Len` of very large objects, the error message can appear very long or just not appear at all. Having a "printing length limit" helps in keeping the console clean and ensures the message `should have %d item(s), but has %d` appears and can be read.

## Changes
A failed `Len` assertion will print the type of the object instead of its value if the actual length of the object is greater than `100` items. `100` is an arbitrary value and I'm open to changing it.

## Motivation
In my tests some slices have thousands of items and I find myself avoiding

`assert.Len(t, items, expectedItems)`

in favour of

`assert.True(t, len(items) == expectedItems, "expected %d items to be set", expectedItems)`

just because the console gets filled with thousands of values that take all available space and are thus not useful for debugging.

In very large slices no message appears, leaving an empty error behind.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -644,6 +644,9 @@ func Len(t TestingT, object interface{}, length int, msgAndArgs ...interface{}) 
 	}
 
 	if l != length {
+		if l > 99 {
+			return Fail(t, fmt.Sprintf("%T should have %d item(s), but has %d", object, length, l), msgAndArgs...)
+		}
 		return Fail(t, fmt.Sprintf("\"%s\" should have %d item(s), but has %d", object, length, l), msgAndArgs...)
 	}
 	return true


fgimian:
An interesting issue I'm experiencing due to this issue is that no error is printed at all when the string representation of the slice is too long:

```
=== RUN   TestGetProducts
    products_test.go:48:
                Error Trace:    C:/Users/Fots/source/downloads/api/products_test.go:48
                Error:
                Test:           TestGetProducts
--- FAIL: TestGetProducts (0.01s)
```

In my case, the string representation had 248,649 characters.

Cheers
Fotis

brackendawson:
#1646 is kind of related, but it addresses output so long that it isn't returned at all. It could be modified or built upon to reduce output to a sane limit rather than a possible limit.

ccoVeille:
I would expect the change made here for Len to be included in #1646, and we close this old PR

Inuart:
It's been.. a while since I created this PR and it got ignored. Hopefully #1646 can solve it.

PR #1237, opened 2022-07-22
This vet fix regressed Mock

### LINKED ISSUE #1227: mock.Mock is not mock.assertExpectationser: missing method AssertExpectations
Upgrading from v1.7.1 to v1.8.0 results in this for my tests which call `AssertExpectationsForObjects`:
```
mock.Mock is not mock.assertExpectationser: missing method AssertExpectations
```

Called like this:
```
mock.AssertExpectationsForObjects(t, someRepository.Mock, otherRepository.Mock)
```

Where `someRespository` and `otherRepository` are defined like such:
```
type someRepository struct {
   mock.Mock
}
```

Looking at v1.8.0, I can see the `assertExpectationser` interface is defined [here](https://github.com/stretchr/testify/blob/cf1284f8dd6f0bf34eed4ab8808ef88f40f7d00f/mock/mock.go#L552-L554), and implemented by `mock.Mock` [here](https://github.com/stretchr/testify/blob/cf1284f8dd6f0bf34eed4ab8808ef88f40f7d00f/mock/mock.go#L580).

I see that passing in `myMock.Mock` is deprecated, and we should pass in `myMock` instead, but this still results in the same error, just kicks the can up the stack.

brackendawson:
## Summary
The attempt to fix a legitimate go vet finding (https://github.com/stretchr/testify/commit/b5ce16571001d6e96e1693ac891fed5c2510c651) breaks the referenced deprecated use case in v1.8.0. This vet finding can't be fixed until v2. The mutex is only copied when people use the deprecated case.

## Changes
Revert invalid change from https://github.com/stretchr/testify/commit/b5ce16571001d6e96e1693ac891fed5c2510c651

## Motivation
This caused a regression in v1.8.0 for people using a deprecated call to AssertExpectationsForObjects

## Related issues
fixes #1227

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -562,9 +562,9 @@ func AssertExpectationsForObjects(t TestingT, testObjects ...interface{}) bool {
 		h.Helper()
 	}
 	for _, obj := range testObjects {
-		if m, ok := obj.(*Mock); ok {
+		if m, ok := obj.(Mock); ok {
 			t.Logf("Deprecated mock.AssertExpectationsForObjects(myMock.Mock) use mock.AssertExpectationsForObjects(myMock)")
-			obj = m
+			obj = &m
 		}
 		m := obj.(assertExpectationser)
 		if !m.AssertExpectations(t) {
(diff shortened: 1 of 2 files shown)


dolmen:
@brackendawson Could you improve the commit message and rebase?

brackendawson:
The simple revert by this PR is problematic because it fails vet, as the prior implementation did. There is no good way to fix this without failing vet. This was broken between v1.7.2 and v1.7.3. There has bee one user report of an issue since then in #1227. I think given the time since then and the lack of complaint it might be better to keep the unintended breaking change. But I can improve the panic by promoting the deprecation warning into a test failure, I'll do that in another PR.

Issue #1160, opened 2022-02-22
Feature: Fail fast for time comparisons


distinctdan:
I'm running into an issue with time comparisons failing when the tests are run in github actions, due to internal timezone differences in the time structs. I understand why they're failing per this issue: https://github.com/stretchr/testify/issues/666. However, I'd like a way to catch these kind of things locally so that we can prevent coding any time comparisons in the first place.

Would it be possible to globally fail any assertion that tries to compare 2 times directly? Ideally I'm thinking of a global config setting that can be enabled, something like `disableTimeCompare`, that makes any test fail that directly compares 2 time.Time's.

Example:
```
type MyStruct struct {
    Now time.Time
}

var a, b MyStruct

// Bad comparison - should fail because of time compare.
assert.Equal(t, a, b)
```

Alternatively, is there already a way to do this? Thanks!

brackendawson:
I'm struggling to come up with a test which is sensitive to the system TimeZone, can you show a minimum reproducible example? In the one you have given `time.Time{} == time.Time{}` is true in all zones.

A possible workaround might be you can select one of the more obscure time zones in your tests using an init function:
```go
func init() {
    os.Setenv("TZ", "Australia/Broken_Hill")
}
```

Also, nice issue given the significance of today's date.

distinctdan:
So the actual cases for this that I've come across are in integration tests, either when we convert a struct to json and back, or when we read times from the database. 

Here's a playground example where I convert a time to json and back. The time shows as having the exact same time, and using the time package's `Equal` function returns true. However, directly comparing the structs fails:
https://go.dev/play/p/rMxB3oq4QBy

Also, that's a creative idea of setting an odd timezone to make any comparisons fail, thanks, I'll have to try that.

jalevin:
I'm currently running into this in integration testing with xorm.

``` 
type myTime struct {
  time time.Time
}

t := time.Now().UTC()

expected := myTime {
  time: time.Now().UTC().round(0) // remove wall clock
}

_ := functionToInsertRecordIntoOrm(mt)

db actual := functionToRetrieveRecordFromORM()

assert(t, expected, actual)


// output

     Time: (time.Time) {
                                -  wall: (uint64) 0,
                                +  wall: (uint64) 901192000,
                                   ext: (int64) 63791013547,
//
```

Specific to my case, when the ORM is recreating the time, the wall clock is added back in. t.Equal automatically handles this comparison however I haven't seen a graceful way to handle this with testify

PR #1240, opened 2022-07-31
Add TimeEqual function to better compare times

annarosenthal:
## Summary
In order to fix JSON-unmarshalled comparison (https://github.com/stretchr/testify/issues/1160), we've added a new function `TimeEqual()` that compares timestamps instead of `time.Time` fields. This also allows you to compare times with different time zones.

## Changes
- added `TimeEqual` to compare by timestamps

## Motivation
To fix https://github.com/stretchr/testify/issues/1160 and to allow comparison with different time zones.

## Related issues
Closes https://github.com/stretchr/testify/issues/1160


dolmen:
`assert.Equal()` already supports time.Time values.

dolmen:
It is easy to do `assert.True(t, t1.Equal(t2), "got(%v), expected(%v)", t2, t1)`.

We also have [`assert.WithinDuration`](https://pkg.go.dev/github.com/stretchr/testify/assert#WithinDuration).

ccoVeille:
> It is easy to do `assert.True(t, t1.Equal(t2), "got(%v), expected(%v)", t2, t1)`.
> 
> We also have [`assert.WithinDuration`](https://pkg.go.dev/github.com/stretchr/testify/assert#WithinDuration).

WithinDuration is indeed a good idea.

Just to confirm it's about this, right?

```go
assert.WithinDuration(t, t1, t2, 0)
```

EDIT: confirmed https://github.com/stretchr/testify/issues/1010#issuecomment-975822073

ccoVeille:
I would close this PR as solution and workaround are provided

brackendawson:
assert.WithinDuration with a duration of zero is what you're looking for here.

PR #1244, opened 2022-08-08
Fixed FailfastSuite failing when test.count > 1

eisandbar:
FailfastSuite used a callOrder slice that gets populated by the suite methods and compared it to a string literal. When test.count is > 1 callOrder is overpopulated and does not match the expected string. Fixed this by having TearDownSuite dump the callOrder into a new slice callOrderPerRun and have tests assert each run separately.

## Summary
Fixes TestFailfastSuite failing when test.count is set > 1

## Changes
FailfastSuite now dumps callOrder into callOrderPerRun when TearDownSuite is called.
TestFailfastSuite now compares the string literal to each entry of callOrderPerRun

## Motivation
Previously running go test ./... test.count=2 resulted in failure

### DIFF
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -537,7 +537,8 @@ func TestSuiteWithStats(t *testing.T) {
 // It logs calls in the callOrder slice which we then use to assert the correct calls were made
 type FailfastSuite struct {
 	Suite
-	callOrder []string
+	callOrder       []string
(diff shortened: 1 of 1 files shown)


dolmen:
Is it fixed?
```console
$ go test ./suite -count=2
ok  	github.com/stretchr/testify/suite	9.404s
```

PR #1109, opened 2021-08-04
suite: add support for parallel sub-tests and remove `suite.Suite` pseudo-interitance

maroux:
## Summary
Add support for parallel sub-tests and remove `suite.Suite` pseudo-interitance

## Changes
- Run sub-tests in a group so `TearDownSuite` is called in the right order
- Remove `suite.Suite` interface / struct because it's a bad attempt at
  inheritance which is causing problems.

## Motivation
Currently, parallel sub-tests are broken with `suite`. Go runs parallel tests by deferring
them until the end of parent test. `testing.T.Run` returns early, and thus [the deferred calls](https://github.com/stretchr/testify/blob/cd58006fe60b8ead9b2172ac361b197f476de83c/suite/suite.go#L164)
in suite fire off, such as `TearDownSuite` (the test itself runs in a separate go-routine).

Additionally, since there's only one `T` object per suite instance, this means during parallel
tests, this object gets overridden by subsequent tests's [`F` method](https://github.com/stretchr/testify/blob/cd58006fe60b8ead9b2172ac361b197f476de83c/suite/suite.go#L128), which leads to test output being associated with the wrong test,
and potentially other problems because `testing.T` is NOT concurrent-safe.

## Example usage (if applicable)

```
type ExampleTestSuite struct {}

func (suite *ExampleTestSuite) TestExample(t *suite.T) {
    t.Equal(5, 5)
}
```

## Related issues

Fixes #934, #187.

PR #1255, opened 2022-08-18
Allow suite methods to take *testing.T parameter

grongor:
## Summary
Allow suite methods to take `*testing.T` parameter (as normal tests do). This is BC, and we are already messing with reflection there, so no penalties either.

## Changes
Extend allowed suite method signature to include optional `*testing.T`.
Fail with a friendly error when the signature doesn't match (before it panicked).

## Motivation
There are many places where you need `t`, like for example for mocks and other testing libraries. And typing `s.T()` get's tedious after a while, and is completely avoidable at no cost :)

```
type MyTestSuite struct {
	suite.Suite
}

func (s *MyTestSuite) TestLorem(t *testing.T) {
	s.Same(t, s.T())

	mockA := mocks.NewMockA(t)
	mockB := mocks.NewMockB(t)
	mockIHaveTooManyMocks := mocks.NewMockMyLife()

	testStuff(mockA, mockB, mockIHaveTooManyMocks)
}

func (s *MyTestSuite) TestSimple() {
	s.NotEqual("this is", "backwards compatible")
}
```

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -172,7 +172,18 @@ func Run(t *testing.T, suite TestingSuite) {
 					stats.start(method.Name)
 				}
 
-				method.Func.Call([]reflect.Value{reflect.ValueOf(suite)})
+				switch method.Type.NumIn() {
+				case 1:
+					method.Func.Call([]reflect.Value{reflect.ValueOf(suite)})
+				case 2:
+					if method.Type.In(1) == reflect.TypeOf(&testing.T{}) {
+						method.Func.Call([]reflect.Value{reflect.ValueOf(suite), reflect.ValueOf(t)})
+						break
+					}
+					fallthrough
+				default:
+					t.Error("invalid test method signature: expecting no parameters, or single *testing.T")
+				}
 			},
 		}
 		tests = append(tests, test)
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -216,10 +216,10 @@ func (suite *SuiteTester) TestOne() {
 }
 
 // TestTwo is another example of a test.
-func (suite *SuiteTester) TestTwo() {
+func (suite *SuiteTester) TestTwo(t *testing.T) {
 	beforeCount := suite.TestTwoRunCount
 	suite.TestTwoRunCount++
-	assert.NotEqual(suite.T(), suite.TestTwoRunCount, beforeCount)
+	assert.NotEqual(t, suite.TestTwoRunCount, beforeCount)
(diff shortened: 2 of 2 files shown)


suite/suite.go
```diff
@@ -172,7 +172,18 @@ func Run(t *testing.T, suite TestingSuite) {
 					stats.start(method.Name)
 				}
 
-				method.Func.Call([]reflect.Value{reflect.ValueOf(suite)})
+				switch method.Type.NumIn() {
+				case 1:
+					method.Func.Call([]reflect.Value{reflect.ValueOf(suite)})
+				case 2:
+					if method.Type.In(1) == reflect.TypeOf(&testing.T{}) {
```
dolmen:
```suggestion
					if method.Type.In(1) == reflect.TypeOf((*testing.T)(nil)) {
```

grongor:
Just a few days ago I saw this PR and though that it's a shame it didn't get merged, so...thank you for taking another look! :bow: :rocket:

dolmen:
I think we should go further and also allow `*assert.Assert` as an alternative version of the parameter, because there is a high chance that our assertions will be used.

grongor:
Sure, could be useful :) But then:
- not everyone is using the `assert` package; I personally prefer `require`...so maybe both?
- we still need access to the `*testing.T` for stuff like test helpers, mock constructors,...

So to make this truly universal and flexible we'd need to compare the list of parameters with the "list of allowed values" (containing `*testing.T`, `*assert.Assertions` and `*require.Assertions` for now) in arbitrary order. It can be done and it wouldn't even be that difficult...It is a bit unorthodox/"not something you see every day", but on the other hand this is a lib for testing - a bit of magic is kinda expected. 

If you want I can look into it ;) If you choose to keep it simple with "just the T", it's imo perfectly reasonable too :)

brackendawson:
The panic on mis-matched signature can be resolved by #1665, this PR only addresses changing the signature change.

Changing the signature would conflict with #1109 if in the future we fix the pseudo-inheritance in suite.

Is `s.T()` arduous? Why are we adding a second way to access an interface that we have a defined way to access, this is not typical in Go.

Adding *assert.Assert as an optional argument as well as *testing.T; what about both? Would the order matter? This seems a little bit like like pretending Go has keyword args.

The suite package already suffers from an incompatibility with Parallel testing because suite employed a pseudo-inheritance structure. The assert package suffers from unclear argument type validation because the `msgAndArgs ...interface{}` argument mis-uses variadics to give each assertion pseudo-optional arguments. I'm not currently convinced that optional parameters is a good idea.

PR #1264, opened 2022-09-13
add EventuallyWithT assertion

### LINKED ISSUE #902: Better alternative to Eventually()
I've found that `Eventually` is not really useful:
1. I can't use assertions inside `condition`.
1. Error message doesn't help to understand failure cause: `Condition never satisfied`.

So I use func `WaitFor` instead: [gist](https://gist.github.com/maratori/010bfbf05639aa3a5ba832cdd75320ec).

The only disadvantage compared to `Eventually`:
* `WaitFor` will not stop after timeout (immediately). It will call function one more time to pass error message from assertions inside that function to the test.

```go
// WaitFor calls fn (once in tick) until it passes or timeout expired.
//
// Attempt is failed if any of following methods called in fn:
//  - Fail/FailNow
//  - Error/Errorf
//  - Fatal/Fatalf
// Otherwise attempt is considered as successful and WaitFor stops.
//
// WaitFor receives test object as the first argument.
// It is passed to fn only on the last attempt. All other attempts use fakeT.
// So the test will fail only after timeout expired.
//
// Note: t.Log() and other methods will print to stdout only on the last attempt.
func WaitFor(t TestingT, timeout time.Duration, tick time.Duration, fn func(t TestingT)) {
	timer := time.NewTimer(timeout)
	defer timer.Stop()
	defer ticker.Stop()
	for {
		ft := &fakeT{name: t.Name()}
		didPanic := false
		func() {
			defer func() {
				if recover() != nil {
					didPanic = true
				}
			}()
			fn(ft)
		}()
		if !ft.Failed() && !didPanic {
			return
		}

		select {
		case <-timer.C:
			fn(t)
			return
		case <-ticker.C:
		}
	}
}
```
The rest code (with examples) can be found [in this gist](https://gist.github.com/maratori/010bfbf05639aa3a5ba832cdd75320ec).

It would be great to see the same feature in testify.

tobikris:
## Summary
This adds a version of Eventually that allows to do `assertions` in the condition.

## Changes
* Added `EventuallyWithT` assertion.
* Added `CollectT` helper struct.

## Motivation
The current `Eventually` only returns one summary error in case of a failure. This is not very helpful if several conditions need to be matched in the condition function. This version collects all failures of a tick and replays the last tick's failures after the timeout.
Thus the last iteration's errors can be shown to the user.

## Example Usage
```
externalValue := false
go func() {
	time.Sleep(8*time.Second)
	externalValue = true
}()
assert.EventuallyWithTf(t, func(c *assert.CollectT) {
	// add assertions as needed; any assertion failure will fail the current tick
	assert.True(c, externalValue, "expected 'externalValue' to be true")
}, 1*time.Second, 10*time.Second, "external state has not changed to 'true'; still: %v", externalState)
```

## Related issues
Closes #902

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -164,15 +164,15 @@ func Eventuallyf(t TestingT, condition func() bool, waitFor time.Duration, tick 
 // If the condition is not met before waitFor, the collected errors of
 // the last tick are copied to t.
 //
-//	externalValue := false
-//	go func() {
-//		time.Sleep(8*time.Second)
-//		externalValue = true
-//	}()
-//	assert.EventuallyWithTf(t, func(c *assert.CollectT) {
-//		// add assertions as needed; any assertion failure will fail the current tick
-//		assert.True(c, externalValue, "expected 'externalValue' to be true")
-//	}, 1*time.Second, 10*time.Second, "external state has not changed to 'true'; still: %v", externalState)
+// 	externalValue := false
+// 	go func() {
+// 		time.Sleep(8*time.Second)
+// 		externalValue = true
+// 	}()
+// 	assert.EventuallyWithTf(t, func(c *assert.CollectT, "error message %s", "formatted") {
+// 		// add assertions as needed; any assertion failure will fail the current tick
+// 		assert.True(c, externalValue, "expected 'externalValue' to be true")
+// 	}, 1*time.Second, 10*time.Second, "external state has not changed to 'true'; still false")
 func EventuallyWithTf(t TestingT, condition func(collect *CollectT), waitFor time.Duration, tick time.Duration, msg string, args ...interface{}) bool {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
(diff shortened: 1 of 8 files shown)


mrIncompetent:
@boyan-soubachov if you have a minute 🙏

l-abels:
I implemented something similar about a year ago, this looks like a good solution. What needs to happen to get this merged? I'm running up against this use case again and would love to have this available.

boyan-soubachov (approved):
Thank you

dolmen:
Methods `Reset` and `Copy` of `CollectT` should have been made private. But they are now part of the public API. :(

dolmen:
I can't understand how adding such a feature has gone into a patch release of testify. Does this gives us (maintainers) the permission to also fix it in a patch release?

Cc: @brackendawson

brackendawson:
Probably the minor version should have been bumped when this was released. Fixing it should be a patch unless we chnage the API. I think we should do the right things, perhaps this discussion deserves an issue to tackle release higyne? I'm also somewhat interested in some automated testing before releases using a lot of real modules that import testify, there have been some regressions in recent releases.

Issue #1180, opened 2022-04-21
Deprecate ObjectsAreEqual and ObjectsAreEqualValues from the assert package and move them in a separate package for reuse


nikolavp:
The two methods are currently located in the `assert` package making them easily mistaken as "assert" methods when in fact they only return a boolean which indicates if the objects are equal or equal values. 

There is [lots](https://github.com/search?l=&q=ObjectsAreEqual+language%3AGo&type=code) and [lots](https://github.com/search?p=3&q=ObjectsAreEqualValues+language%3AGo&type=Code) of code that incorrectly assumes that these methods are asserting equality.

Although it is clear from their signature (they don't accept a `testing.T` object), one assumes that everything in the assert package is asserting something about the arguments.

Can we move these in a separate package called "equal" and deprecate the instances in `assert` so at least people don't make the mistake again.

MichaelMallett:
Just had a long discussion about this at work, and realised that this has caught quite a few of us out multiple times, so I'm in complete agreement that this should be moved to a seperate package that makes their intended use clear.

MichaelMallett:
Created a PR for this
https://github.com/stretchr/testify/pull/1279

k3n:
I was bit by this today, I'm new to Go and getting my feet wet with testing and kept scratching my head why this method wasn't failing even when I passed in an obviously incorrect comparison object.

Looks like the PR is stuck in review limbo, hopefully this gets resolved soon.

PR #1279, opened 2022-10-10
Move object equality functions to another package for clarity

mminklet:
## Summary
Move none assertion functions into a seperate package to make it clear that these do no do any assertions. There is a a lot of existing code out there with useless tests because people are assuming that assert.Foobar functions will do assertions.

Happy to take guidance on comments/naming etc.

## Changes
Moved ObjectsAreEqual and ObjectsAreEqualValues and tests into a new sub package `equals`. Label existing functions as deprecated and call new functions instead. Call new methods anywhere in assert where the previous functions are called.

## Motivation
Currently very confusing that these are in the assert package when they don't do any assertions at all.

change `assert.ObjectsAreEqual` to `equal.ObjectsAreEqual` etc

## Related issues
Closes 1180


nikolavp (dismissed):

MichaelMallett:
> If the aim of this is to prevent people from thinking that `ObjectsAreEqual` is an assertion that they should use, then I think this new `equal` package should be internal:
> 
> ```
> testify/
> |-- assert
> |   `-- assertions.go
> |-- internal
> |   `-- equal
> |       `-- equal.go
> |-- mock
> |   `-- mock.go
> ```

I was actually thinking that people would still continue to use these methods, as they are useful, they're just not immediately obvious that the bool needs to be checked to have a working test. But perhaps another name/location outside of assert makes more sense?

brackendawson:
> > If the aim of this is to prevent people from thinking that `ObjectsAreEqual` is an assertion that they should use, then I think this new `equal` package should be internal:
> > ```
> > testify/
> > |-- assert
> > |   `-- assertions.go
> > |-- internal
> > |   `-- equal
> > |       `-- equal.go
> > |-- mock
> > |   `-- mock.go
> > ```
> 
> I was actually thinking that people would still continue to use these methods, as they are useful, they're just not immediately obvious that the bool needs to be checked to have a working test. But perhaps another name/location outside of assert makes more sense?

While it's true that they might be useful to someone, I think people will still misuse them in exactly the same way. Testify is not a utility library, I think if people are looking for a comparison library then they should be looking at go-cmp instead.

MichaelMallett:
Yeah that's a good point. Tbh then I'm a bit reticent about introducing this change then, as I don't want to remove useful functionality even if it is currently being misused

mminklet:
Is there any progress on this? If there are suggestions that are required, please let me know and I'll gladly make them.  [People are still being caught by this](https://github.com/stretchr/testify/issues/1180#issuecomment-1485540940)

dolmen:
Please rebase this branch on top of master, to fix gofmt issues.

dolmen:
Replaced by #1431.

mminklet:
Nice! Sorry I didn’t have time to rebase, it had been so long that there were more changes that needed more work. Glad to see the change got in anyway!

PR #1284, opened 2022-10-17
subsequent YAML documents were ignored

### LINKED ISSUE #1281: Asserting nested yaml documents are working as expected
Asserting nested yaml documents will always pass. 

I would expect following test to fail.
`
package main

import (
	"testing"
)

func TestMultipleYAMLDocuments(t *testing.T) {

	expectedYaml := `
Director:
  name: Spielberg
  Movies:
    - Movie:
        title: E.T.
        year: 1975
    - Movie:
        title: Jaws
        year: 1982
---
Director:
  name: Spielberg
  Movies:
    - Movie:
        title: E.T.
        year: 1975
    - Movie:
        title: Jaws
        year: 1982

`
	actualYaml := `
Director:
  name: Spielberg
  Movies:
    - Movie:
        title: E.T.
        year: 1975
    - Movie:
        title: Jaws
        year: 1982
---
Director:
  name: Spielberg
  Movies:
    - Movie:
        title: E.T.
        year: 1975

`
	assert.YAMLEq(t, expectedYaml, actualYaml, "yaml should be the same")
}

`

brackendawson:
## Summary
In YAMLEq(), yaml documents in the input after the first document were silently not considered in the comparison. Make it compare all the docs.

I checked JSONEq() and it always fails if you try to pass it JSON Lines documents, even if they are the same.

## Changes
- Use a `yaml.Deocoder.Docode()` loop into a `[]interface{}` rather than a `yaml.Unmarshal()` into an `interface{}`.

## Motivation
Unlike with passing JSON Lines inputs to `JSONEq()`, which always fails, passing multiple YAML documents with differences after the first document into `YAMLEq()` silently passes. 😱 

## Related issues
Closes #1281


brackendawson:
Sorry about all the comment changes [gofmt changed](https://tip.golang.org/doc/go1.19), let me know if you want me to do it again with an older toolchain.

dolmen (commented):
LGTM. Please rebase on master.

dolmen:
This patch has been broken by #1579 which made the YAML library pluggable.

Updating the patch requires changing the interface for pluggable YAML lib as `yaml.Unmarshal` doesn't support multidocs: this patch uses a [`yaml.Decoder`](https://pkg.go.dev/gopkg.in/yaml.v3#Decoder), so rebasing is not trvial.

brackendawson:
Because #1579 exposed the internal implementation of YAMLEq on testify's API, it is not possible to solve this issue without either partially implementing a YAML parser in testify itself or changing testify's API in one of two ways:
* Change `assert.yaml.Unmarshal` to a decoder. This would be a breaking change so no.
* Add another YAML assertion (like YAMLEqAll) that uses decode. I dislike this because we shouldn't have added YAMLEq in the first place.

I'm closing this PR because it's no longer tenable.

PR #1287, opened 2022-10-21
Print underlying object in ptr comparison (issue #1118)

### LINKED ISSUE #1118: Print underlying objects when comparing pointers
[Example](https://play.golang.org/p/4PB_uXG8-Zh) 

```go
package main

import (
	"testing"
)

func TestPointerToValue(t *testing.T) {
	var a int = 1
	var b int = 2
	require.EqualValues(t, &a, &b, "The two numbers should be the same.")

}
```

The output is 

```
=== RUN   TestPointerToValue
    prog.go:13: 
        	Error Trace:	prog.go:13
        	Error:      	Not equal: 
        	            	expected: (*int)(0xc0000189b8)
        	            	actual  : (*int)(0xc0000189c0)
        	Test:       	TestPointerToValue
        	Messages:   	The two numbers should be the same.
--- FAIL: TestPointerToValue (0.00s)
```

It would be incredible to have the actual values compared printed as well. For instance 
`expected: (*int)(0xc0000189b8)` could be `expected: (*int)(0xc0000189b8) 1`.

Currently, I am abusing this helper

```
func equalValuesPrettyPrint(t testing.TB, expected, actual interface{}) bool {
	return assert.EqualValues(t, expected, actual, "exp=%v, got=%v", reflect.Indirect(reflect.ValueOf(expected)), reflect.Indirect(reflect.ValueOf(actual)))
} 
```

### LINKED ISSUE #1273: assert.Equal with pointers prints address instead of value on failure
testify version: v1.8.0

Comparing pointers to values with `Equal` prints the address of the value if the comparison fails instead of the value. This can be confusing to users as it makes it seem like the addresses are being compared instead of the values.

I have verified this behavior for both `*string` and `*int` types. Comparisons between `nil` and some non-nil pointer also prints the address of the non-nil value

Example:
```go
func EqualOutputString(t *testing.T) {
  e := "hello"
  g := "hola"
  ep := &e
  gp := &g

  assert.Equal(t, ep, gp)
  assert.Equal(t, ep, nil)
}

func EqualOutputInt(t *testing.T) {
  e := 7
  g := 42
  ep := &e
  gp := &g

  assert.Equal(t, ep, gp)
}
```

test output:
```
=== RUN   TestEqualOutputString
    equal_output_test.go:15: 
                Error Trace:    /go/src/output_test/equal_output_test.go:15
                Error:          Not equal: 
                                expected: (*string)(0x4000573530)
                                actual  : (*string)(0x4000573540)
                            
                                Diff:
                Test:           TestEqualOutputString
    equal_output_test.go:16: 
                Error Trace:    /go/src/output_test/equal_output_test.go:16
                Error:          Not equal: 
                                expected: *string((*string)(0x400006aec0))
                                actual  : <nil>(<nil>)
                Test:           TestEqualOutputString
--- FAIL: TestEqualOutputString (0.00s)
=== RUN   TestEqualOutputInt
    equal_output_test.go:24: 
                Error Trace:    /go/src/output_test/equal_output_test.go:24
                Error:          Not equal: 
                                expected: (*int)(0x4000546b10)
                                actual  : (*int)(0x4000546b18)
                Test:           TestEqualOutputInt
--- FAIL: TestEqualOutputInt (0.00s)
```

mchlp:
## Summary
Prints out the underlying value of a pointer to a basic type in addition to the pointer address when the comparison between two basic type pointers fails. 

## Changes
- Updated `truncatingFormat` function in `assert/assertions.go` with a flag that control whether or not the underlying value of basic types are also printed in addition to their address
- Updated `assert/assertions_test.go` to fix unit tests that were broken and added additional unit test to cover added code

## Motivation
As described in Issue #1118, when the comparison between two pointers fails, the underlying value is not always printed. This gives the impression that it is the address of the pointers that are being compared, which is incorrect, since it is actually the underlying values of the pointers that are being compared. 

After further research, we discovered that non-basic types such as maps and structs already print the underlying value when the `%#v` specifier is used for formatting, so their underlying value is already being printed. However, for basic types such as int, bool, and string, the `%#v` specifier only prints out the pointer address. Our change will ensure that the underlying value of these basic type pointers are printed as well.

For example, this is the original output when comparing two int pointers:
```
Error:          Not equal: 
                    expected: (*int)(0x14000120df0)
                    actual  : (*int)(0x14000120df8)
```
And this is the new output when comparing two int pointers:
```
Error:          Not equal: 
                                expected: (*int)(0x14000120df0) 5
                                actual  : (*int)(0x14000120df8) 0
```

Similarly, this is the new output when comparing two string and bool pointers:
```
Error:          Not equal: 
                                expected: (*string)(0x140001132d0) hello
                                actual  : (*string)(0x140001132e0) 
```
```
 Error:          Not equal: 
                                expected: (*bool)(0x14000120fa8) true
                                actual  : (*bool)(0x14000120fa9) false
```

However, we do not print out the underlying value of basic type pointers when a comparison is being made between two different pointer types. This is the same behaviour as before. For example, for a comparison between an int pointer and a string pointer, this is the output:
```
Error:          Not equal: 
                                expected: *string((*string)(0x140000997c0))
                                actual  : *int((*int)(0x140000a7608))
```

## Related issues
Closes #1118
Closes #1273

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -49,6 +49,28 @@ type ErrorAssertionFunc func(TestingT, error, ...interface{}) bool
 // Comparison is a custom function that returns true on success and false on failure
 type Comparison func() (success bool)
 
+// List of basic types. This is specifically a list of types whose underlying value is not printed
+// when using the %#v format specifier to print a pointer to the value.
+var basicTypes = []reflect.Kind{
+	reflect.Int,
+	reflect.Int8,
+	reflect.Int16,
+	reflect.Int32,
+	reflect.Int64,
+	reflect.Uint,
+	reflect.Uint8,
+	reflect.Uint16,
+	reflect.Uint32,
+	reflect.Uint64,
+	reflect.Uintptr,
+	reflect.Float32,
+	reflect.Float64,
+	reflect.Complex64,
+	reflect.Complex128,
+	reflect.String,
+	reflect.Bool,
+}
+
 /*
 	Helper functions
 */
@@ -56,6 +78,7 @@ type Comparison func() (success bool)
 // ObjectsAreEqual determines if two objects are considered equal.
 //
 // This function does no assertion of any kind.
+
 func ObjectsAreEqual(expected, actual interface{}) bool {
 	if expected == nil || actual == nil {
 		return expected == actual
@@ -431,22 +454,37 @@ func samePointers(first, second interface{}) bool {
 // to a type conversion in the Go grammar.
 func formatUnequalValues(expected, actual interface{}) (e string, a string) {
 	if reflect.TypeOf(expected) != reflect.TypeOf(actual) {
-		return fmt.Sprintf("%T(%s)", expected, truncatingFormat(expected)),
-			fmt.Sprintf("%T(%s)", actual, truncatingFormat(actual))
+		return fmt.Sprintf("%T(%s)", expected, truncatingFormat(expected, false)),
+			fmt.Sprintf("%T(%s)", actual, truncatingFormat(actual, false))
 	}
 	switch expected.(type) {
 	case time.Duration:
 		return fmt.Sprintf("%v", expected), fmt.Sprintf("%v", actual)
 	}
-	return truncatingFormat(expected), truncatingFormat(actual)
+
+	return truncatingFormat(expected, true), truncatingFormat(actual, true)
 }
 
 // truncatingFormat formats the data and truncates it if it's too long.
 //
 // This helps keep formatted error messages lines from exceeding the
 // bufio.MaxScanTokenSize max line length that the go testing framework imposes.
-func truncatingFormat(data interface{}) string {
+//
+// The `printValueOfBasicTypes` flag controls whether or not to print the underlying
+// value of pointers of basic types in addition to the pointer address.
+func truncatingFormat(data interface{}, printValueOfBasicTypes bool) string {
 	value := fmt.Sprintf("%#v", data)
+
+	if printValueOfBasicTypes && reflect.ValueOf(data).Kind() == reflect.Pointer {
+		v := reflect.ValueOf(data).Elem()
+		for _, t := range basicTypes {
+			if v.Kind() == t {
+				value = fmt.Sprintf("%#v %v", data, v.Interface())
+				break
+			}
+		}
+	}
+
 	max := bufio.MaxScanTokenSize - 100 // Give us some space the type info too if needed.
 	if len(value) > max {
 		value = value[0:max] + "<... truncated>"
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -2475,15 +2475,38 @@ func Test_validateEqualArgs(t *testing.T) {
 func Test_truncatingFormat(t *testing.T) {
 
 	original := strings.Repeat("a", bufio.MaxScanTokenSize-102)
-	result := truncatingFormat(original)
+	result := truncatingFormat(original, false)
 	Equal(t, fmt.Sprintf("%#v", original), result, "string should not be truncated")
 
 	original = original + "x"
-	result = truncatingFormat(original)
+	result = truncatingFormat(original, false)
 	NotEqual(t, fmt.Sprintf("%#v", original), result, "string should have been truncated.")
 
 	if !strings.HasSuffix(result, "<... truncated>") {
 		t.Error("truncated string should have <... truncated> suffix")
+	}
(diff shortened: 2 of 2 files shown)


boyan-soubachov (approved):

boyan-soubachov:
Thank you for your contribution :)

assert/assertions.go
```diff
@@ -431,22 +454,37 @@ func samePointers(first, second interface{}) bool {
...
 // truncatingFormat formats the data and truncates it if it's too long.
 //
 // This helps keep formatted error messages lines from exceeding the
 // bufio.MaxScanTokenSize max line length that the go testing framework imposes.
-func truncatingFormat(data interface{}) string {
+//
+// The `printValueOfBasicTypes` flag controls whether or not to print the underlying
+// value of pointers of basic types in addition to the pointer address.
+func truncatingFormat(data interface{}, printValueOfBasicTypes bool) string {
 	value := fmt.Sprintf("%#v", data)
```
boyan-soubachov, line 476: Shouldn't this _not_ be executed if `printValueOfBasicTypes == true`?

mchlp:
@boyan-soubachov Would be great if you could take a look at thi and merge if it looks good!

assert/assertions.go
```diff
@@ -431,22 +454,40 @@ func samePointers(first, second interface{}) bool {
...
 // truncatingFormat formats the data and truncates it if it's too long.
 //
 // This helps keep formatted error messages lines from exceeding the
 // bufio.MaxScanTokenSize max line length that the go testing framework imposes.
-func truncatingFormat(data interface{}) string {
-	value := fmt.Sprintf("%#v", data)
+//
+// The `printValueOfBasicTypes` flag controls whether or not to print the underlying
+// value of pointers of basic types in addition to the pointer address.
+func truncatingFormat(data interface{}, printValueOfBasicTypes bool) string {
```
boyan-soubachov, line 475: Would it not be neater if we just moved the logic for checking whether to print the value into this function rather than passing it in as an argument at every call?

boyan-soubachov (commented):
Please excuse the late review, I have just one question above.

assert/assertions.go
mchlp: The logic for checking whether to print the value is already inside the `truncatingFormat` function (when it loops through the `basicTypes` list and checks if the type of `data` matches any of the types in the `basicTypes` list).

The use of the `printValueOfBasicTypes` is an override, where if it is set to `false`, only the pointer address (and not the underlying value) is returned, regardless of whether it is a basic type or not. The use case of this is in the `formatUnequalValues` function where two pointers of different types are being compared. In this case, we only want the pointer addresses to be printed and not the underlying values (even if the underlying values are basic types). This emphasizes the fact that the pointers are different types (which is why the comparison failed) and that the underlying value of the pointers do not matter for the comparison.

Also updated the comments for the `truncatingFormat` function to better explain the use of the `printValueOfBasicTypes`

mchlp:
@boyan-soubachov Would be great if you could take a look and merge when you get the chance!

dropdevrahul:
The MR looks good @boyan-soubachov

boindil:
@boyan-soubachov please have a look

mchlp:
Rebased on top of what is currently on `master` and builds are passing on my local branch (https://github.com/mchlp/testify/actions/runs/7200906284), but are failing here.

@dolmen Do you know why this might be happening? The build failures seem to be very inconsistent (they will pass some runs, but will fail some other runs with the exact same commit)

assert/assertions.go
```diff
@@ -512,22 +534,44 @@ func samePointers(first, second interface{}) bool {
...
+	var value string
+	if printValueOfBasicTypes && reflect.ValueOf(data).Kind() == reflect.Ptr {
+		v := reflect.ValueOf(data).Elem()
+		for _, t := range basicTypes {
+			if v.Kind() == t {
+				value = fmt.Sprintf("%#v %v", data, v.Interface())
+				break
+			}
+		}
```
brackendawson, line 568: I'm not against your `basicTypes` but this could be written as:
```suggestion
	if isBasicType(v.Kind()) {
		value = fmt.Sprintf("%#v %v", data, v.Interface())
	}
```

With:
```go
func isBasicType(k reflect.Kind) bool {
	return k == reflect.String || (k >= reflect.Bool && k <= reflect.Complex128)
}
```

Rather than iterating.

brackendawson (changes requested):

PR #1290, opened 2022-10-31
assert: Fail `Subset`/`NotSubset` when a map superset is missing elements

### LINKED ISSUE #1270: `assert.Subset` doesn't fail tests
## Story

We were writing some unit tests, and I was surprised that they passed. Here is a re-creation:

```go
func TestSubsetSlices(t *testing.T) {
	a := []string{`a`}
	ab := []string{`a`, `b`}

	assert.Subset(t, ab, a)
	assert.NotSubset(t, a, ab)

	// The following asserts should fail this test
	assert.Subset(t, a, ab)
	assert.NotSubset(t, ab, a)
}

func TestSubsetMapsWithValues(t *testing.T) {
	a := map[string]int{
		`a`: 1,
	}
	ab := map[string]int{
		`a`: 1,
		`b`: 2,
	}

	assert.Subset(t, ab, a)
	assert.NotSubset(t, a, ab)

	// The following asserts should fail this test
	assert.Subset(t, a, ab)
	assert.NotSubset(t, ab, a)
}

func TestSubsetMapsWithEmptyStructs(t *testing.T) {
	a := map[string]struct{}{
		`a`: {},
	}
	ab := map[string]struct{}{
		`a`: {},
		`b`: {},
	}

	assert.Subset(t, ab, a)
	assert.NotSubset(t, a, ab)

	// The following asserts should fail this test
	assert.Subset(t, a, ab)
	assert.NotSubset(t, ab, a)
}
```

## Problem

I think the problem is here:

https://github.com/stretchr/testify/blob/181cea6eab8b2de7071383eca4be32a424db38dd/assert/assertions.go#L821-L825

https://github.com/stretchr/testify/blob/181cea6eab8b2de7071383eca4be32a424db38dd/assert/assertions.go#L882-L886

It returns a false value, but it doesn't fail the `*testing.T`.

antichris:
## Summary
Avoid panics in `assert.Subset`/`assert.NotSubset` and properly fail the test when a superset map is missing elements from the subset map.

## Changes
- Check whether the value reflected at the `subsetKey` of the superset reflection `listValue` `.isValid`, before attempting to get its interface value for use in `ObjectsAreEqual` assertion.
- Align the `Subset` failure message with the corresponding `NotSubset` failure message.

## Motivation
When a subset map contains an item with a key that is not present in the superset map, attempts to reflect the interface of the value at the absent key cause panics, which return `false` on recovery but do not properly fail the test with a message.

## Related issues
Closes #1270

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -843,10 +843,10 @@ func Subset(t TestingT, list, subset interface{}, msgAndArgs ...interface{}) (ok
 		for i := 0; i < len(subsetKeys); i++ {
 			subsetKey := subsetKeys[i]
 			subsetElement := subsetValue.MapIndex(subsetKey).Interface()
-			listElement := listValue.MapIndex(subsetKey).Interface()
+			listEntry := listValue.MapIndex(subsetKey)
 
-			if !ObjectsAreEqual(subsetElement, listElement) {
-				return Fail(t, fmt.Sprintf("\"%s\" does not contain \"%s\"", list, subsetElement), msgAndArgs...)
+			if !listEntry.IsValid() || !ObjectsAreEqual(subsetElement, listEntry.Interface()) {
+				return Fail(t, fmt.Sprintf("%q is a not subset of %q", subset, list), msgAndArgs...)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


st3penta:
Duplicate of https://github.com/stretchr/testify/pull/1261

brackendawson:
Thank you for your efforts, I can confirm this was fixed in the other PR, it's no longer a problem on v1.8.4 (latest) or on master.

PR #1293, opened 2022-11-02
Recover from the panic in suite subtest

### LINKED ISSUE #764: recover in subtests
```
package main

import (
	"fmt"
	"testing"

)

type Suite struct {
	suite.Suite
}

func TestXX(t *testing.T) {
	suite.Run(t, new(Suite))
}

func (s *Suite) SetupSuite() {
	fmt.Println("In SetupSuite")
}

func (s *Suite) TearDownSuite() {
	fmt.Println("In TearDownSuite")
}

func (s *Suite) TestXX_YY() {
	s.Run("test xx yy", func() {
		panic("test")
	})
}
```

For the above snippet, it panics in subtest and does not enter the `TearDownSuite`, should we add `failOnPanic` in the subtests? like this: 

```
func (suite *Suite) Run(name string, subtest func()) bool {
	oldT := suite.T()
	defer suite.SetT(oldT)
	return oldT.Run(name, func(t *testing.T) {
+		defer failOnPanic(t)
		suite.SetT(t)
		subtest()
	})
}
```

xwjdsh:
## Summary

Example Playground: https://go.dev/play/p/MOJuXPikRMU
Recover from the panic in the suite subtest.

## Changes

Add `defer recoverAndFailOnPanic(t)` into the implement of `func (suite *Suite) Run`.
## Motivation

If the subtest panicked and it's not recovered, then the `TearDownTest` and `TearDownSuite` will be skipped, we should make sure that the TearDown methods are executed.

## Related issues

Close #764

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -108,6 +108,7 @@ func (suite *Suite) Run(name string, subtest func()) bool {
 	}()
 
 	return oldT.Run(name, func(t *testing.T) {
+		defer recoverAndFailOnPanic(t)
 		suite.SetT(t)
 		subtest()
 	})
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -617,3 +617,39 @@ func (s *FailfastSuite) Test_B_Passes() {
 	s.call("Test B Passes")
 	s.Require().True(true)
 }
+
+type subtestPanicSuite struct {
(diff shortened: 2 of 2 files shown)


xwjdsh:
Close it as it will be fixed in #1471.

PR #1296, opened 2022-11-08
Use reflect.Value.Pointer() to compare pointers

### LINKED ISSUE #1076: Assert equal pointers

AlexanderYastrebov:
## Summary

Use `reflect.Value.Pointer()` to compare pointers

## Changes

Updates `samePointers` implementation and adds test cases

## Motivation

This is needed to check if map, slice and other built-in types point to the same value

## Related issues

Fixes #1076

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -491,12 +491,12 @@ func validateEqualArgs(expected, actual interface{}) error {
 	return nil
 }
 
-// Same asserts that two pointers reference the same object.
+// Same asserts that two arguments reference the same object.
 //
-//	assert.Same(t, ptr1, ptr2)
+//	assert.Same(t, arg1, arg2)
 //
-// Both arguments must be pointer variables. Pointer variable sameness is
-// determined based on the equality of both type and value.
+// Both arguments can be pointers, channels, functions, maps, slices or strings.
+// Argument sameness is determined based on the equality of both type and value.
 func Same(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
@@ -511,12 +511,12 @@ func Same(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) b
 	return true
 }
 
-// NotSame asserts that two pointers do not reference the same object.
+// NotSame asserts that two arguments do not reference the same object.
 //
-//	assert.NotSame(t, ptr1, ptr2)
+//	assert.NotSame(t, arg1, arg2)
 //
-// Both arguments must be pointer variables. Pointer variable sameness is
-// determined based on the equality of both type and value.
+// Both arguments can be pointers, channels, functions, maps, slices or strings.
+// Argument sameness is determined based on the equality of both type and value.
 func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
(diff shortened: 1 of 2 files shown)


stevenh (approved):
LGTM

assert/assertions.go
```diff
@@ -410,17 +410,16 @@ func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
...
-	if firstPtr.Kind() != reflect.Ptr || secondPtr.Kind() != reflect.Ptr {
+	if firstPtr.Kind() != secondPtr.Kind() {
 		return false
 	}
 
-	firstType, secondType := reflect.TypeOf(first), reflect.TypeOf(second)
-	if firstType != secondType {
+	switch firstPtr.Kind() {
+	case reflect.Chan, reflect.Func, reflect.Map, reflect.Ptr, reflect.Slice:
+		return firstPtr.Pointer() == secondPtr.Pointer()
```
dolmen, line 419:
```suggestion
		return firstPtr.Kind() == secondPtr.Kind() && firstPtr.Pointer() == secondPtr.Pointer()
```

assert/assertions.go
```diff
@@ -410,17 +410,16 @@ func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
 // they point to the same object
 func samePointers(first, second interface{}) bool {
 	firstPtr, secondPtr := reflect.ValueOf(first), reflect.ValueOf(second)
-	if firstPtr.Kind() != reflect.Ptr || secondPtr.Kind() != reflect.Ptr {
+	if firstPtr.Kind() != secondPtr.Kind() {
 		return false
 	}
```
dolmen, line 415: This block can be moved after the switch on kind.

dolmen (changes requested):

assert/assertions.go
AlexanderYastrebov: Fixed

stevenh:
@dolmen quick ping as @AlexanderYastrebov has updated based on your feedback

brackendawson (changes requested):
The docstring for `Same` would need to be changed:
> ```
> Same asserts that two pointers reference the same object.
>
>	assert.Same(t, ptr1, ptr2)
>
> Both arguments must be pointer variables. Pointer variable sameness is
> determined based on the equality of both type and value.
> ```

* This now works for more than pointer variables.
* This now works for types which are convertable to the same type.
* It needs to be clear that only slices pointing to the **start** of the same array are the same, and not slices pointing to different indices of the same array:
```go
var (
	a   [8]int
	sa0 = a[:]
	sa3 = a[3:]
)
assert.Same(t, sa0, sa3)
```

assert/assertions.go
```diff
@@ -491,12 +491,12 @@ func validateEqualArgs(expected, actual interface{}) error {
 	return nil
 }
 
-// Same asserts that two pointers reference the same object.
+// Same asserts that two arguments reference the same object.
```
brackendawson, line 494: This is ambiguous for slices in this case:
```go
var (
	a   [8]int
	sa0 = a[:]
	sa3 = a[3:]
)
assert.Same(t, sa0, sa3)
```
sa0 and sa3 reference the same _object_ if you take object to mean _array_, but not if you mean _contiguous segment of an underlying array_. The language specification uses both terms at various points. We should make it clear that `Same` only considers slices to be the same if they reference the same _contiguous segment of an underlying array_.

brackendawson (changes requested):
This is a change in behaviour and we must decide if it's safe to do in v1.

No invocation of Same could previously pass with any non-pointer types.

Previously NotSame will always pass if non-pointer type/s are passed it, it can now fail of those reference the same object. In this case I think it would fix a test the author would have intended to fail.

I'd like some more opinions before merging this.

assert/assertions.go
AlexanderYastrebov: IMO reverse is surprising
```go
var (
	a   [8]int
	sa0 = a[:]
	sa1 = a[:1]
)
assert.Same(t, sa0, sa1) // passes
```

PR #1300, opened 2022-11-12
fix(generate): go formatting

stevenh:
## Summary
Updated to use golang.org/x/tools/imports so that generated code is formatted correctly.

## Changes
* Split import blocks.
* Corrected code comment block indents.
* Updated the generated header to use the standard format.

## Motivation
Contributors using modern tool chain don't have to manually reformat their code before putting in a PR.
Tools can correctly detect the generated files as defined by [cmd/go](https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source)


stevenh:
Seems goimports and gofmt don't format comment blocks the same so closing this.

PR #1252, opened 2022-08-16
Add 1.19 to go_version matrix to run tests against latest go versions.

lisitsky:
## Summary
Test testify against actual golang version.

## Changes
```yaml
matrix:
        go_version: ["1.19", ...]
``` 

## Motivation
To ensure testify is ready for actuality and future.

PR #1301, opened 2022-11-13
fix: ci for go 1.19 and standard generated header 

stevenh:
## Summary
Fix continuous integration with go 1.19 and use the standard generated header for golang files.

## Changes
* Use the standard generated header golang files.
* Fix ci so it works with go 1.19, only checking format and generate for this.
* Centralise ci in the workflow file.
* Run ci on the supported versions of golang 1.18 and 1.19.
* Run go mod tidy on _codegen module.

## Motivation
Allows tools which comply with the [recommended header](https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source) to correctly detect the files are generated.

Allow users on go 1.19+ to contribute without having to install old versions of golang.

Centralise the ci in the github workflow file so its easier to maintain.


lisitsky:
Hi, @stevenh ! 
I've got the same problem with CI for different go versions simultaneously.
Look like code gets  generated code in different ways with space characters, so it cannot fit all versions. Or may be checks should be different for different versions.
https://github.com/stretchr/testify/pull/1252

stevenh:
Yer my take as just to go with the latest formatting 1.19+, and skip tests on earlier versions.

dolmen:
Too many unrelated changes in a single MR. This is too much work to review and accept/reject individual changes.

Please resubmit changes in separate MR well contained if they still apply.

dolmen:
About _Centralise the ci in the github workflow file so its easier to maintain_., I disagree because it make harder to debug the CI build steps on a developer's machine. You should at least explain _why_ you think it's easier to maintain.

stevenh:
Have split this up into the 2 separate PRs with the two key change sets
* https://github.com/stretchr/testify/pull/1458 - Atandard generated header
* https://github.com/stretchr/testify/pull/1460 - Improve workflows

For the [workflow change](https://github.com/stretchr/testify/pull/1301/commits/e93b9f35664d00ac241b3d758e9e59869524d3cd) if you want to test locally I would recommend [act](https://github.com/nektos/act) for that. However as the flows themselves are very basic, they are unlikely to need any debugging.

Two clean up steps:
* `go mod tidy` - Only needed if dependencies are changed, something developers following best practices are used to running
* `go generate ./...` - Only needed if templates are changes, so something people should know to run.

Two validation steps, which are both handled by editors so should be no-op for the developer:
* `gofmt -w .`
* `go vet ./...`

Hope this helps.

stevenh:
Looks like the standard headers have already been done, didn't spot that as this was still pending, so just workflow improvements left.

PR #1302, opened 2022-11-13
fix: test failures

stevenh:
## Summary
Fix random test failure and compatibility with `go test -count=X` where `X > 1`.

## Changes
* Fix random failure TestEventuallyIssue805 due to timing dependency.
* Fix `TestFailfastSuite` when run with `go test` flag `-count=X` where `X > 1`.

## Motivation
So we don't get false test failures causing confusion.

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -2453,7 +2453,7 @@ func TestEventuallyIssue805(t *testing.T) {
 	mockT := new(testing.T)
 
 	NotPanics(t, func() {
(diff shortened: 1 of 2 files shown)


dolmen:
Two unrelated changes in a single commit. This can't be merged as is. Commits across multiple packages should have a good reason, and this is not a good case.

I also have independently written a fix for TestEventuallyIssu805 but not just by increasing delay. Could you review #1412?

Please submit the change about the `suite` package as a separate MR.

PR #1308, opened 2022-11-27
Refactor assertions to use generics

TheAndrew2115:

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -56,17 +56,18 @@ type Comparison func() (success bool)
 // ObjectsAreEqual determines if two objects are considered equal.
 //
 // This function does no assertion of any kind.
-func ObjectsAreEqual(expected, actual interface{}) bool {
-	if expected == nil || actual == nil {
-		return expected == actual
+func ObjectsAreEqual[T any](expected, actual T) bool {
+	anyExp, anyAct := any(expected), any(actual)
+	if anyExp == nil || anyAct == nil {
+		return anyExp == anyAct
 	}
 
-	exp, ok := expected.([]byte)
+	exp, ok := anyExp.([]byte)
 	if !ok {
 		return reflect.DeepEqual(expected, actual)
 	}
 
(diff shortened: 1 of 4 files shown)


assert/assertions.go
```diff
@@ -1167,58 +1160,25 @@ func WithinRange(t TestingT, actual, start, end time.Time, msgAndArgs ...interfa
...
+	return float64(x)
 }
 
+
 // InDelta asserts that the two numerals are within delta of each other.
 //
-// 	 assert.InDelta(t, math.Pi, 22/7.0, 0.01)
-func InDelta(t TestingT, expected, actual interface{}, delta float64, msgAndArgs ...interface{}) bool {
+//	assert.InDelta(t, math.Pi, 22/7.0, 0.01)
+func InDelta[T ConvertibleToFloat64](t TestingT, expected, actual T, delta float64, msgAndArgs ...interface{}) bool {
```
dolmen, line 1177: InDelta should only take `~float32 | ~float64`. `ConvertibleToFloat64` is too wide.

assert/assertions.go
```diff
@@ -1507,22 +1445,26 @@ func NotRegexp(t TestingT, rx interface{}, str interface{}, msgAndArgs ...interf
...
+	var zeroVal T
+	if !reflect.DeepEqual(i, zeroVal) {
 		return Fail(t, fmt.Sprintf("Should be zero, but was %v", i), msgAndArgs...)
 	}
 	return true
 }
 
 // NotZero asserts that i is not the zero value for its type.
-func NotZero(t TestingT, i interface{}, msgAndArgs ...interface{}) bool {
+func NotZero[T any](t TestingT, i T, msgAndArgs ...interface{}) bool {
```
dolmen: Inconsistency with `Zero` where `i` is ìnterface{}`.

dolmen (commented):
This implementation puts much more compile-time constraints on types. I expect this will break existing code.

PR #1317, opened 2022-12-17
Upgrade go module version to the minimum go version supported by the Go team

### LINKED ISSUE #1313: Consider updating declared Go version in go.mod
The go.mod file declares go 1.13 right now. [Module graph pruning](https://tip.golang.org/ref/mod#graph-pruning) was added in Go 1.17, meaning that other modules can import your module and only pick up the transitive dependencies that are actually reachable based on how they use your code. Right now, importing testify means importing a yaml parser because the declared go version predates go 1.17.

The tidiest dependency graph will come from using go 1.18, as:

> The [go.sum file](https://tip.golang.org/ref/mod#go-sum-files) recorded by [go mod tidy](https://tip.golang.org/ref/mod#go-mod-tidy) for a module by default includes checksums needed by the Go version one below the version specified in its [go directive](https://tip.golang.org/ref/mod#go-mod-file-go). So a go 1.17 module includes checksums needed for the full module graph loaded by Go 1.16, but a go 1.18 module will include only the checksums needed for the pruned module graph loaded by Go 1.17. The -compat flag can be used to override the default version (for example, to prune the go.sum file more aggressively in a go 1.17 module).

Go 1.18 is also the oldest Go release still supported by the Go team, so it's a reasonable choice of language version to declare.

kishaningithub:
## Summary

Fixes #1313 

## Changes
- Upgrading go module version to 1.18 which is the least supported version by the golang team

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,11 +6,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go_version: ["1.18.1", "1.17.6", "1.16.5"]
+        go_version: ["1.19", "1.18"]
     steps:
       - uses: actions/checkout@v3
       - name: Setup Go
-        uses: actions/setup-go@v3.2.0
+        uses: actions/setup-go@v3.5.0
         with:
           go-version: ${{ matrix.go_version }}
       - run: ./.ci.gogenerate.sh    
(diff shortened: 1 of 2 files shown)


kishaningithub:
Depends on https://github.com/stretchr/testify/pull/1301

dolmen:
No. I want to support old projects as long as it is not too painful.
At this point I don't see a good reason for not allowing users of go 1.16 to upgrade to the latest release of testify.
Of course we will revise this when we'll use generics in the API.

PR #1320, opened 2022-12-22
Fix adding ? when no values passed

lesichkovm:
## Summary
Fix adding ? when no values passed

## Changes
Small fix of incorrect URL generation

## Motivation
When no values pased to the HTTPBody method, a ? is incorrectly added, when not essentially required


dolmen (approved):
LGTM

PR #1330, opened 2023-01-27
Fix: assert.Subset for maps

Agurato:
## Summary

This PR aims to fix the assert.Subset (and assert.NotSubset) functions for maps

## Changes

- Updated the `NotSubset(…)` function to check the key's existence in the containing map. Before the fix, if the key did not exist in the containing map, a panic would occur and the function would return false.

- Updated the `Subset(…)` function to do the same and avoid a panic. This panic was well intercepted and the test returned the correct result, but I figured that this was a better way.

## Motivation

I noticed that the NotSubset function would not work properly for intersecting maps.

For example, I wanted to check that `{ "a": "x", "d": "z"}` is indeed not a subset of `{ "a": "x", "b": "y", "c": "z"}`

## Related issues

#1173


dolmen (approved):

dolmen:
Please rebase your branch on top of master and fix conflicts.

Agurato:
This has been fixed in #1261

Issue #1329, opened 2023-01-26
AssertExpectations should not fail if the test was skipped


ianrose14:
If you set up a test with something like `defer myval.AssertExpectations(t)` and then later skip the test (`t.Skip()`), the `AssertExpectations` check can still fail your test.  I propose that `AssertExpectations` should check whether the test has already been marked as "skipped" and, if so, return nil (success).  Here's the change I'm envisioning:

```go
type tSkipped interface {
	Skipped() bool
}

// AssertExpectations asserts that everything specified with On and Return was
// in fact called as expected.  Calls may have occurred in any order.
func (m *Mock) AssertExpectations(t TestingT) bool {
	if s, ok := t.(tSkipped); ok {     // <--- START OF PROPOSED NEW CODE
		if s.Skipped() {
			return true
		}
	}                                  // <--- END OF PROPOSED NEW CODE
	if h, ok := t.(tHelper); ok {
		h.Helper()
	}
	m.mutex.Lock()
	defer m.mutex.Unlock()
	[...]
}
```

If y'all are in agreement with this, I'm very happy to create a PR.

ianrose14:
I went ahead and made a PR here: https://github.com/stretchr/testify/pull/1331

PR #1331, opened 2023-01-27
Ensure AssertExpectations does not fail in skipped tests

ianrose14:
## Summary

Prevent `(*Mock).AssertExpectations` from failing if test expectations are not met but the test has already been marked as "Skipped".

## Changes
* Add check to `(*Mock).AssertExpectations` such that if the `*testing.T` is already marked skipped, we just return true immediately.
* Add a unit test (`Test_Mock_AssertExpectations_Skipped_Test`) for the new behavior.

## Motivation
Normally the decision to skip a test is made early in the test function, but on occasion it might be the case that you have already set up some mock expectations _before_ you call `t.Skip`.  If that's the case, then a call to `AssertExpectations` will fail, which is not intended.

```go
	val := new(MockedValue)
	val.On("DoThing", 1, 2).Return(3, 4)
	defer val.AssertExpectations(t)

	if os.GetEnv("foo") == "" {
		// This will cause the test to exit, which will run the deferred AssertExpections call above
		t.Skip("nevermind")
	}
	...
}
```

## Related issues
Closes https://github.com/stretchr/testify/issues/1329

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -594,6 +594,9 @@ func AssertExpectationsForObjects(t TestingT, testObjects ...interface{}) bool {
 // AssertExpectations asserts that everything specified with On and Return was
 // in fact called as expected.  Calls may have occurred in any order.
 func (m *Mock) AssertExpectations(t TestingT) bool {
+	if s, ok := t.(interface{ Skipped() bool }); ok && s.Skipped() {
+		return true
+	}
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -1493,6 +1493,16 @@ func Test_Mock_AssertExpectations_With_Repeatability(t *testing.T) {
 
 }
 
+func Test_Mock_AssertExpectations_Skipped_Test(t *testing.T) {
+
+	var mockedService = new(TestExampleImplementation)
+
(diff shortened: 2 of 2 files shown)


brackendawson:
Thanks for the PR, I'd love to see this merged.

dolmen (approved):

dolmen:
@ianrose14 Could you rebase your branch on top of master (and fix conflicts)?

ianrose14:
> @ianrose14 Could you rebase your branch on top of master (and fix conflicts)?

done!

mock/mock.go
```diff
@@ -592,6 +592,11 @@ func AssertExpectationsForObjects(t TestingT, testObjects ...interface{}) bool {
 // AssertExpectations asserts that everything specified with On and Return was
 // in fact called as expected.  Calls may have occurred in any order.
 func (m *Mock) AssertExpectations(t TestingT) bool {
+	if s, ok := t.(tSkipped); ok {
+		if s.Skipped() {
```
dolmen, line 596: Merge this condition in the check above:

```go
ok && s.Skipped()
```

dolmen (changes requested):

dolmen:
Please squash commits.

ianrose14:
> Code looks good.
> 
> Please squash commits.

@dolmen done!

dolmen (approved):

brackendawson (approved):
Thank you for your contribution.

Issue #1332, opened 2023-01-28
Add `Cap` and `Capf` assertions


moeen:
To asserts capacity of `Array`, `Slice` and `Channel` types, add `Cap` and `Capf` assertions.

dolmen:
That would be just a convenience for a niche use case. I don't think that it is worth extending the API surface at this point.

nikitacrit:
I would like to have this method in the library, I was surprised not to find it when I needed it, which is what brought me here.

PR #1333, opened 2023-01-28
assert: add `Cap` and `Capf` assertions

moeen:
## Summary
Add `Cap` and `Capf` assertions to assert capacity of `Array`, `Slice` and `Channel` types.

## Motivation
There is no assertions for capacities in Go in this library. It is specially useful to assert capacity of buffered channels.
Currently you should use `assert.Equal` and use the `cap` function.

## Related issues
#1332


dolmen:
That would be just a convenience for a niche use case. I don't think that it is worth extending the API surface at this point.

dolmen (changes requested):
`assertion_forward.go` is generated code. It must not be edited by hand.

PR #1336, opened 2023-02-01
Adds AnythingOfTypeT wrapper function

chrsalx:
## Summary

Adds `AnythingOfTypeT` function as a wrapper that provides generics support to `AnythingOfType`.

## Changes

* bumps go.mod version to 1.19.
* Adds `AnythingOfTypeT` function.
* Adds tests for `AnythingOfTypeT`.

## Motivation

The `AnythingOfType` function allows us to configure mocking behavior for any value for the provided type. The caveat to this is that `AnythingOfType` expects a string, this requires the user to know beforehand the full name of the type as it will be defined by Go at runtime.

This can lead to issues during the development of the tests caused by typos or by not naming the desired type properly. 

Using generics, Go can provide the name of the type without breaking the existing API.

```go
c := mockedService.
	On("TestMethod", AnythingOfTypeT[int]()).
	Return(nil)

assert.Nil(t, c.TestMethod(5))
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go_version: ["1.18.1", "1.17.6", "1.16.5"]
+        go_version: ["1.19.5"]
     steps:
       - uses: actions/checkout@v3
       - name: Setup Go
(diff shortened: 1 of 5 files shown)


dolmen:
The `testify` module doesn't yet rely on a Go version that has generics. `testify` is now a major point in the Go ecosystem, so it would not be wise yet to block users from upgrading `testify` (to benefit from bug fixes) if they are not able to upgrade their Go toolchain (example: support of a GOOS or GOARCH being removed in recent versions).

So that new code that use generics must be put in a separate file `mock_generics.go` with a `//go:build go1.18` guard.

dolmen:
We have `mock.IsType`: this is the function that should have a generic version. So prefix should be IsType, not AnythingOfType.

dolmen:
See also #1442.

PR #1347, opened 2023-02-11
Add mock.ElementsMatch argument matcher

### LINKED ISSUE #1117: Feature Request : Support non-deterministic ordering of slice args in `Mock.On(...)`
Assume that a function that we're mocking accepts a `slice[]` of `strings`, and it returns a response `friendData` thereafter.

A typical mock for this would be something like below - 

`serviceMock.On("GetFriends", friendIDs).Return(friendData, err)`

where, `friendIDs = []string{"abc", "def", "xyz"}`

The mock will work if (and only if)`GetFriends` is definitely called with `friendIDs = []string{"abc", "def", "xyz"}`. However, if it's called with say, `friendIDs = []string{"xyz", "abc", "def"}`, the mock will `panic` with `mock: Unexpected Method Call`.

Something like below - 

```
panic: 

mock: Unexpected Method Call
-----------------------------

GetFriends([]string)
                0: []string{"xyz", "abc", "def"}

The closest call I have is: 

GetFriends([]string)
                0: []string{"abc", "def", "xyz"}

Difference found in argument 0:

--- Expected
+++ Actual
@@ -1,4 +1,4 @@
 ([]string) (len=2) {
- (string) (len=3) "abc",
- (string) (len=3) "def",
- (string) (len=3) "xyz",
- (string) (len=3) "xyz",
- (string) (len=3) "abc",
- (string) (len=3) "def",
 }

Diff: 
        0: FAIL:  ([]string{"xyz", "abc", "def"}) != ([]string{"abc", "def", "xyz"}) [recovered]
        panic: 

```

It could be that the `friendIDs` are returned by another service (or a DB). Using `mock.Anything` (although it works) seems a bit _misleading_ - it's not as if we don't care as to what argument was used to call the function, just that the ordering of individual elements inside that slice is immaterial.

Is there an intention to support `Any` ordering of elements inside a `slice`, so long as all the elements (and only those elements) are there present in the argument?

### LINKED ISSUE #1180: Deprecate ObjectsAreEqual and ObjectsAreEqualValues from the assert package and move them in a separate package for reuse

brackendawson:
## Summary
Add an argument marcher to mock which matches if array/slice elements match.

## Changes
- Add `mock.ElementsMatch()` argument matcher.
- Refactor `mock.argumentMatcher` to be an interface, it pretty much already was.
- Move common code used by both assert and mock into an internal package.

## Motivation
There is already `mock.MatchedBy` which allows users to write this themselves, but it's not trivial and a lot of people seem to need this specific matcher.

```go
func TestIfy(t *testing.T) {
	m := &myMock{}
	m.On("Do", mock.ElementsMatch([]int{1, 2, 3})).Return()
	m.Do([]int{2, 3, 1})
}
```

## Related issues
Closes #1117 
Closes #1180

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -17,8 +17,8 @@ import (
 	"unicode"
 	"unicode/utf8"
 
-	"github.com/davecgh/go-spew/spew"
 	"github.com/pmezard/go-difflib/difflib"
+	"github.com/stretchr/testify/internal/check"
 	yaml "gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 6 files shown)


dolmen:
This MR includes changes to use `internal/check`. They are unrelated to the description of this issue. Please makes those changes available in a separate MR.

brackendawson:
@dolmen done in #1431, this will need rebasing somewhat after that is merged.

dolmen:
We must first cleanup the argument matching API. See my proposal in #1578.

It is also too early to merge #1431: we have plenty of pending patches in the queue that would be broken by #1431, so I would like that we review everything before such a major refactor.

In light of this, I think that this implementation is not useful at this point, so I'm closing the issue. Could you instead open an issue with an API proposal that would fit #1578?

Issue #1350, opened 2023-02-27
assert: add PanicsWithErrorIs


jannisbaudisch:
I have some code that wraps a database error with fmt.Errorf() and panics it. Now I want to assert that this error is paniced.

The current method PanicsWithError only compares the error message, but does not use a error chain comparison.

Example:
```
func foo(source err) {
    panic(fmt.Errorf("Wrapping error %w", source))
}

func Test(t *testing.T) {
    expectedError := errors.New("some error")
    require.PanicsWithErrorIs(t, expectedError, func() { foo(expectedError) }) // This functionality is missing
}
```

jannisbaudisch:
Implemented in pull request #1353

dolmen:
@brackendawson You voted this issue down. Do you have a comment?

jannisbaudisch:
The assertion function ErrorIs is already implemented. So i guess if there was ever a discussion. Someone decided to implement it.

Ad hoc there is no standard library that panics with an error. But at least panicking as well es ErrorIs as well as error wrapping is all part of the standard library.

mail2nnv:
I'll add my five cents. Besides the specified `PanicsWithErrorIs` method, I lack the `PanicsWithErrorContains()`.

https://go.dev/play/p/-ePsCuCxyW1

```go
import (
	"errors"
	"fmt"
	"testing"

	"github.com/stretchr/testify/require"
)

var ErrGoCrazy = errors.New("go crazy")
var GoCrazy = func() { panic(fmt.Errorf("🔫: %w", ErrGoCrazy)) }

func Test_SomePanicWithError(t *testing.T) {
	require := require.New(t)

	t.Run("must be ok to test panics wrapped errors", func(t *testing.T) {
		require.PanicsWithErrorIs( // <— error here ⛔
			func() { GoCrazy() },
			ErrGoCrazy,
		)
	})

	t.Run("must be ok to test panics error contains string", func(t *testing.T) {
		require.PanicsWithErrorContains( // <— error here ⛔
			func() { GoCrazy() },
			"crazy",
		)
	})
}
```

PR #1351, opened 2023-02-27
Add PanicWithErrorIs assertions (#1350)

jannisbaudisch:
## Summary
Adds new assertion method to test if a paniced error has a specific error in its chain

## Changes

- Add PanicWithErrorIs functions/methods
- Add tests for base assert function

## Motivation
See my `Issue #1350 `

I have some code that wraps a database error with fmt.Errorf() and panics it. Now I want to assert that this error is paniced.

The current method PanicsWithError only compares the error message, but does not use a error chain comparison.

Example:

func foo(source err) {
    panic(fmt.Errorf("Wrapping error %w", source))
}

func Test(t *testing.T) {
    expectedError := errors.New("some error")
    require.PanicsWithErrorIs(t, expectedError, func() { foo(expectedError) }) // This functionality is missing
}

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -659,6 +659,18 @@ func PanicsWithErrorf(t TestingT, errString string, f PanicTestFunc, msg string,
 	return PanicsWithError(t, errString, f, append([]interface{}{msg}, args...)...)
 }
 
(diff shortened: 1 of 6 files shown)


jannisbaudisch:
Oh man... I did not know there was code generation. I will create a new merge pull request. My bad

PR #1354, opened 2023-02-28
Create GitHub release when new release tag is pushed

### LINKED ISSUE #1295: Github releases page does not contain newer versions
Please consider changing testify release process so new releases will be visible from Github release page. Seeing latest version date at the right side of the repository landing page in Github is useful.

https://github.com/stretchr/testify/releases
https://github.com/stretchr/testify/tags

Currently these pages are quite different. If you are used to check releases page the current version then what you would see is quite different what tags page shows. 

It does not need to contain fancy changelog or changelog at all.

aldas:
This GitHub workflow addition will create new GitHub [release](https://github.com/stretchr/testify/releases) for library when new [tag](https://github.com/stretchr/testify/tags), with matching pattern is pushed.

It does not change current "release process" - developer needs to create tag and push it.

Closes #1295

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,19 @@
+name: Create release from new tag
+
+# this flow will be run only when new tags are pushed that match our pattern
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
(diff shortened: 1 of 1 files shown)


HaraldNordgren:
Sounds very good! Currently it looks in the releases page like this excellent repo is dead, which it is not!

HaraldNordgren:
@boyan-soubachov I would suggest we merge this, and also make a new release with the `EqualExportedValues` stuff. Sounds ok?

aviscode (approved):

boyan-soubachov:
Let's test this out :)

Fernandosilvas35 (approved):

PR #1356, opened 2023-03-02
Updated didPanic to work with Go tip.

bwplotka:
I know it's not released yet, but new Go version will have new behaviour on `panic(nil)`. Fixing to make it work, if you don't want to merge until current tip is released, feel free to close.

See behaviour by changing "dev" to "1.20":  https://go.dev/play/p/Cevh9AJdhC6?v=gotip


dolmen:
The code should be more strict and instead check if `errors.As(err, new(runtime.PanicNilError))`. However that will require go1.21, so this check must be wrapped in a function `func IsPanicNilError(error) bool` that must be declared in a separate Go source with a `//go:build go1.21` tag.

PR #1357, opened 2023-03-02
fix bad interface{} string formats

### LINKED ISSUE #1355: Wrong format string for objects in error messages
It appears that some functions are using the wrong format string parameters when generating their error logs. For example `assert.Len()` has: https://github.com/stretchr/testify/blob/f36bfe3c337aa95c86f04c721acdbafb5ffb1611/assert/assertions.go#L645

The `%s` is only applicable for strings, while the function takes any object. When used on something other than a string, the message contains ugly `%!s` indicators scattered all over the place, making the message harder to read.

The format string should be using `%v`.

I noticed other functions where this applies as well. So it's not just `assert.Len()`. But I have not done a full review.

nomionz:
## Summary

fix the bad interface{] objects string formats
## Motivation

when you call a %s or %q on a not string slice it will produce a weird output, testify is testing lib so it should output meaningful error messages

## Related issues

'Fixes #1355 '

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -271,7 +271,7 @@ type labeledContent struct {
 
 // labeledOutput returns a string consisting of the provided labeledContent. Each labeled output is appended in the following manner:
 //
-//   \t{{label}}:{{align_spaces}}\t{{content}}\n
+//	\t{{label}}:{{align_spaces}}\t{{content}}\n
 //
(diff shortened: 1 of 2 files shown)


nomionz:
I'll fix it so that the builds go through. Sorry, first time trying to contribute to an open-source project)

PR #1361, opened 2023-03-11
Improve InEpsilon assertion error message

jgannonjr:
## Summary
I propose adding the actual/expected values o the error message when the InEpsilon assertion fails.  This information is very useful when interpreting/debugging test failures.

## Changes
Changing the output from the original message:
```
        	Error Trace:	math_test.go:50
        	Error:      	Relative error is too high: 1e-09 (expected)
        	            	        < 0.00024990271071290897 (actual)
        	Test:       	TestMath
```

to the improved message:
```
        	Error Trace:	math_test.go:50
        	Error:      	Relative error is too high: 1e-09 (expected)
        	            	        < 0.00024990271071290897 (actual)
        	            	Values:
        	            	    expected: 7.703935578834133
        	            	    actual  : 7.702010344449825
        	Test:       	TestMath
```

## Motivation

The information is extremely useful when interpreting/debugging the failed assertion.  It's difficult to interpret the relative error value on its own.

It's possible to log the expected/actual values in the messages field, but that requires storing every expected value in a variable so you can add it to both the expected param and in the message format string.  It makes more sense for this to be included by default in the error message.

n/a -- the usage is the same
```
assert.InEpsilon(t, expected, actual, epsilon)
```

## Related issues
n/a

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1331,8 +1331,15 @@ func InEpsilon(t TestingT, expected, actual interface{}, epsilon float64, msgAnd
 		return Fail(t, err.Error(), msgAndArgs...)
 	}
 	if actualEpsilon > epsilon {
-		return Fail(t, fmt.Sprintf("Relative error is too high: %#v (expected)\n"+
-			"        < %#v (actual)", epsilon, actualEpsilon), msgAndArgs...)
+		diff := diff(expected, actual)
+		expected, actual = formatUnequalValues(expected, actual)
+		return Fail(t, fmt.Sprintf(
+			"Relative error is too high: %#v (expected)\n"+
+				"        < %#v (actual)\n"+
+				"Values:\n"+
+				"    expected: %s\n"+
+				"    actual  : %s%s",
+			epsilon, actualEpsilon, expected, actual, diff), msgAndArgs...)
 	}
 
 	return true


aviscode (approved):

dolmen (changes requested):
Missing coverage in the test suite.

dolmen:
Please provide a test that shows the issue running on go.dev/play.

jgannonjr:
@dolmen this is an example of the code prior to the change:  https://go.dev/play/p/AP7491IOSvO

This tests the (failing) assertion:
```
assert.InEpsilon(t, float64(100), float64(101), 1e-3)
```
where 100 is the expected value, 101 is the actual value, and 1e-3 is the epislon.

Notice the error message only shows the expected epsilon value (1e-3) is less than the actual (1e-2).  It doesn't show the actual and expect values 100 and 101:
```
        	Error:      	Relative error is too high: 0.001 (expected)
        	            	        < 0.01 (actual)
        	Test:       	TestInEpsilonErrorMessage
```

My change updates the message to include 100 and 101:
```
        	Error:      	Relative error is too high: 0.001 (expected)
        	            	        < 0.01 (actual)
        	            	Values:
        	            	    expected: 100
        	            	    actual  : 101
        	Test:       	TestInEpsilonErrorMessage
```

It's helpful to see these values in the error message.

NOTE:  This PR doesn't change any logic in the assertions, it only adds extra text to the error message.

dolmen (changes requested):
There was no coverage for this check. We must add the coverage with this PR.

https://github.com/stretchr/testify/pull/1361#pullrequestreview-1512374195

PR #1309, opened 2022-11-28
Compare public elements of struct

mchlp:
## Summary
Adds a method to check for equality of two structs of the same type by checking ONLY the exported (public) fields.

## Changes
- Added `ObjectsExportedFieldsAreEqual` method in `assert/assertions.go` that contains core code for the functionality (loops through fields of the struct using the `reflect` library and only compares fields that are exported
- Added `EqualExportedValues` method in `assert/assertions.go` that calls the `ObjectsExportedFieldsAreEqual` method and contains assertion related code that logs the appropriate message on failure
- Added corresponding unit test `TestObjectsExportedFieldsAreEqual` in `assert/assertions_test.go`

## Motivation
There are use cases where users want to compare only the public fields of two struct objects (when they only care about the public interface of the object and the internal private representation does not matter). This added method provides users the ability to do this with the `testify` library.

Example Usage:
```go
type S struct {
	Exported int
        notExported int
}

ObjectsExportedFieldsAreEqual(S{1, 2}, S{1, 2}) => true
ObjectsExportedFieldsAreEqual(S{1, 2}, S{1, 3}) => true
ObjectsExportedFieldsAreEqual(S{1, 2}, S{2, 3}) => false
```
The method also works with nested structures, where `ObjectsExportedFieldsAreEqual` is recursively called for nested fields that are of type struct, or `ObjectsAreEqualValues` are called on non-struct type fields.
```go
type Nested struct {
	Exported    interface{}
	notExported interface{}
}

type S struct {
	Exported1    interface{}
	Exported2    Nested
	notExported1 interface{}
	notExported2 Nested
}

ObjectsExportedFieldsAreEqual(S{1, Nested{2, 3}, 4, Nested{5, 6}}, S{1, Nested{2, "a"}, 4, Nested{5, 6}}) => true
ObjectsExportedFieldsAreEqual(S{1, Nested{2, 3}, 4, Nested{5, 6}}, S{"a", Nested{2, 3}, 4, Nested{5, 6}}) => false
```

## Related issues
Closes #1033

PR #1373, opened 2023-04-11
EqualExportedValues: Support nested structs and time.Time fields

### LINKED ISSUE #984: Struct containing a time.Time can't be compared correctly.

feidtmb:
## Summary

Updates the new `EqualExportedValues` function to support comparing nested struct fields and `time.Time` values. The function explicitly no longer supports recursive data structures (see [this since removed test case](https://github.com/stretchr/testify/pull/1373/commits/96a3ee0f09fbe55b8802ea351cd20979d0742b54) for example).

## Changes

#1309 introduced `EqualExportedValues`. I was looking to add support for comparing structs containing `time.Time` fields, and this seemed like an ideal function to adjust for that since:
1. The explicit comparison of only exported values suggests the caller doesn't want a typical deep equality check, so comparing times using `time.Time.Equal` is likely more in line with the desired behavior.
2. `EqualExportedValues` is not yet in a tagged release, and so this behavior change should not affect many real users. 
3. Past PRs have tried to add time.Time equality support to functions like `assert.Equal`, but that task was found to be more complicated (for example, see #985). Adding this support to only `EqualExportedValues` is more easily achieved.

## Motivation

There is currently no good method for comparing structs that contain `time.Time` fields that don't have matching timezones or monotonic clock readings.

Example Usage:

```go
type Q struct {
	TS *time.Time
}
type R struct {
	Nested Q
}
type S struct {
	Nested R
	CreatedAt time.Time
}

t1 := time.Date(2023, time.April, 7, 12, 56, 32, 0, time.UTC)
t2 := time.Date(2023, time.April, 7, 7, 56, 32, 0, time.FixedZone("UTC-5", -5*60*60))

s1 := S{Nested: R{Nested: Q{TS: &t1}}, CreatedAt: t1}
s2 := S{Nested: R{Nested: Q{TS: &t2}}, CreatedAt: t2}

ObjectsExportedFieldsAreEqual(S1, S2) => true
```

## Related issues

Potentially closes #984, assuming we're okay with requiring use of `EqualExportedValues` to compare structs with `time.Time` fields and don't expect `assert.Equal` to handle that.

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -93,6 +93,7 @@ func EqualErrorf(t TestingT, theError error, errString string, msg string, args 
 // EqualExportedValuesf asserts that the types of two objects are equal and their public
 // fields are also equal. This is useful for comparing structs that have private fields
 // that could potentially differ. Types of time.Time are compared using time.Time.Equal.
+// Recursive data structures are not supported and may result in an infinite loop.
 //
 // 	 type S struct {
 // 		Exported     	int
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -92,7 +92,7 @@ func EqualErrorf(t TestingT, theError error, errString string, msg string, args 
 
 // EqualExportedValuesf asserts that the types of two objects are equal and their public
 // fields are also equal. This is useful for comparing structs that have private fields
-// that could potentially differ.
+// that could potentially differ. Types of time.Time are compared using time.Time.Equal.
 //
 // 	 type S struct {
 // 		Exported     	int
--- a/assert/assertion_forward.go
+++ b/assert/assertion_forward.go
@@ -158,6 +158,7 @@ func (a *Assertions) EqualErrorf(theError error, errString string, msg string, a
 // EqualExportedValues asserts that the types of two objects are equal and their public
 // fields are also equal. This is useful for comparing structs that have private fields
 // that could potentially differ. Types of time.Time are compared using time.Time.Equal.
+// Recursive data structures are not supported and may result in an infinite loop.
 //
 // 	 type S struct {
 // 		Exported     	int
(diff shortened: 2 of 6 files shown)


feidtmb:
Of course right after I undraft this I realize recursive structs need to be handled...

Looking into that and will undraft again when ready.

HaraldNordgren:
Hi @feidtmb, I also looking into the issues of the recursive structs (my entry point was the issues with pointers and slices). Here is a fix I proposed, if you want to take a look: https://github.com/stretchr/testify/pull/1379

If you think it makes sense, I we could move the `time.Time` logic in there too another case in the `switch` logic 🤗

mchlp:
I think `EqualExportedValues` is the right place to put the time comparisons with `time.Time.Equal`. However, maybe `EqualExportedValues` would no longer be an accurate name for the function since we're doing more than just comparing the exported values. Maybe `EqualPublicInterface`?

feidtmb:
@HaraldNordgren I left a review on your PR. It looks good to me, thanks for tackling the recursive structures problem!

I've removed the expectation for this PR to handle recursive data structures, and will leave it as a draft for now with the assumption that your PR will be merged instead. Once that's done I'll either close this or adjust it to handle only `time.Time` equality like you mentioned.

feidtmb:
@HaraldNordgren Looking over your PR again, while I do think your approach is more complete than this one since it handles maps and arrays, I don't think it's handling recursive data structures. (See [this test case](https://github.com/stretchr/testify/pull/1373/commits/96a3ee0f09fbe55b8802ea351cd20979d0742b54).)

Just wanted to call that out since I was mistaken in my comment above. Still happy to proceed with your solution though.

HaraldNordgren:
Hi @feidtmb!

In a real scenario in don’t see a case where there are objects referencing themselves. I don’t think I have ever seen them in production code. This I why I wouldn't want to add a comment explaining the lack of this. But maybe I’m wrong?

Regarding a possible solution to it: One level of recursion should be simple to solve. But if we have multiple levels of data before the structure refers back to top level, then it becomes difficult to solve, where we have to keep track of all previous levels. Probably not impossible, but much more complex to write.

If we want to do it, maybe all of this is better suited for a follow-up PR. What do you think? 🤗

PR #1379, opened 2023-04-24
EqualExportedValues: Handle nested pointer, slice and map fields

HaraldNordgren:
The current implementation of `ObjectsExportedFieldsAreEqual` does not handle (nested) pointer or slices fields. This means it doesn't work for comparing e.g. any Protobuf struct if there is any of these on the Protobuf message.

Here, I am hoisting the reflect.Kind() checking from inside the loop to earlier in the logic. This allows me to take better advantage of the recursiveness of the function for checking equality for different types of structs. All while still maintaining the original functionality, as seen by the fact that the original tests are not altered.

I added a lot of tests cove different cases. It may seems like a lot of tests for this feature, but I used them when developing and since the reflection logic is pretty brittle in general, they were quite necessary for me to be able to refactor with any kind of confidence.

Thanks a bunch to @mchlp for the original implementation, this function is very needed for using Testify together with Protobuf! 🙌🏻

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,14 +6,14 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go_version: ["1.18.1", "1.17.6", "1.16.5"]
+        go_version: ["1.20", "1.19"]
     steps:
       - uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v3.2.0
         with:
           go-version: ${{ matrix.go_version }}
-      - run: ./.ci.gogenerate.sh    
+      - run: ./.ci.gogenerate.sh
       - run: ./.ci.gofmt.sh
       - run: ./.ci.govet.sh
       - run: go test -v -race ./...
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -75,46 +75,65 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
 	return bytes.Equal(exp, act)
 }
 
+// copyExportedFields iterates downward through nested data structures and creates a copy
+// that only contains the exported struct fields.
+func copyExportedFields(expected interface{}) interface{} {
+	if isNil(expected) {
+		return expected
+	}
+
+	expectedType := reflect.TypeOf(expected)
+	expectedKind := expectedType.Kind()
+	expectedValue := reflect.ValueOf(expected)
+
+	switch expectedKind {
+	case reflect.Struct:
+		result := reflect.New(expectedType).Elem()
+		for i := 0; i < expectedType.NumField(); i++ {
+			field := expectedType.Field(i)
+			isExported := field.PkgPath == "" // should use field.IsExported() but it's not available in Go 1.16.5
+			if isExported {
+				fieldValue := expectedValue.Field(i)
+				if isNil(fieldValue) || isNil(fieldValue.Interface()) {
+					continue
+				}
+				newValue := copyExportedFields(fieldValue.Interface())
+				result.Field(i).Set(reflect.ValueOf(newValue))
+			}
+		}
+		return result.Interface()
+
+	case reflect.Ptr:
+		result := reflect.New(expectedType.Elem())
+		unexportedRemoved := copyExportedFields(expectedValue.Elem().Interface())
+		result.Elem().Set(reflect.ValueOf(unexportedRemoved))
+		return result.Interface()
+
+	case reflect.Array, reflect.Slice:
+		result := reflect.MakeSlice(expectedType, expectedValue.Len(), expectedValue.Len())
+		for i := 0; i < expectedValue.Len(); i++ {
+			index := expectedValue.Index(i)
+			if isNil(index) {
+				continue
+			}
+			unexportedRemoved := copyExportedFields(index.Interface())
+			result.Index(i).Set(reflect.ValueOf(unexportedRemoved))
+		}
+		return result.Interface()
+
+	default:
+		return expected
+	}
+}
+
 // ObjectsExportedFieldsAreEqual determines if the exported (public) fields of two structs are considered equal.
 // If the two objects are not of the same type, or if either of them are not a struct, they are not considered equal.
 //
 // This function does no assertion of any kind.
 func ObjectsExportedFieldsAreEqual(expected, actual interface{}) bool {
-	if expected == nil || actual == nil {
-		return expected == actual
-	}
-
-	expectedType := reflect.TypeOf(expected)
-	actualType := reflect.TypeOf(actual)
-
-	if expectedType != actualType {
-		return false
-	}
-
-	if expectedType.Kind() != reflect.Struct || actualType.Kind() != reflect.Struct {
-		return false
-	}
-
-	expectedValue := reflect.ValueOf(expected)
-	actualValue := reflect.ValueOf(actual)
-
-	for i := 0; i < expectedType.NumField(); i++ {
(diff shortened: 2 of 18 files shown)


HaraldNordgren:
Ping @boyan-soubachov @mchlp @feidtmb

assert/assertions.go
```diff
@@ -75,46 +75,65 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
...
 		}
+		return result.Interface()
+
+	default:
+		return expected
 	}
-	return true
+}
+
+// ObjectsExportedFieldsAreEqual determines if the exported (public) fields of two structs are considered equal.
```
mchlp, line 129:
```suggestion
// ObjectsExportedFieldsAreEqual determines if the exported (public) fields of two structs are considered equal. This comparison of only exported fields is applied recursively to nested structs.
```

mchlp (commented):
Just a question about handling of `map`. Looks good otherwise!

assert/assertions.go
```diff
@@ -75,46 +75,65 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
...
+			index := expectedValue.Index(i)
+			if isNil(index) {
+				continue
 			}
+			unexportedRemoved := copyExportedFields(index.Interface())
+			result.Index(i).Set(reflect.ValueOf(unexportedRemoved))
 		}
+		return result.Interface()
+
+	default:
```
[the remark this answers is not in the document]
HaraldNordgren: @mchlp I skipped map for now, but I think we should definitely implement it in the future!

feidtmb (commented):
I like this approach! I left a few nits but it looks good to me otherwise.

assert/assertions.go
```diff
@@ -75,46 +75,65 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
...
+		return expected
 	}
-	return true
+}
+
+// ObjectsExportedFieldsAreEqual determines if the exported (public) fields of two structs are considered equal. This comparison of only exported fields is applied recursively to nested structs.
+// If the two objects are not of the same type, or if either of them are not a struct, they are not considered equal.
+//
+// This function does no assertion of any kind.
+func ObjectsExportedFieldsAreEqual(expected, actual interface{}) bool {
```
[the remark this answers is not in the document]
HaraldNordgren: Thanks, updated it!

assert/assertions.go
```diff
@@ -545,6 +564,9 @@ func EqualExportedValues(t TestingT, expected, actual interface{}, msgAndArgs ..
 		return Fail(t, fmt.Sprintf("Types expected to both be struct \n\t%v != %v", bType.Kind(), reflect.Struct), msgAndArgs...)
 	}
 
+	expected = copyExportedFields(expected)
+	actual = copyExportedFields(actual)
+
 	if !ObjectsExportedFieldsAreEqual(expected, actual) {
```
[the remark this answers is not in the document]
HaraldNordgren: Good point! I updated it now!

HaraldNordgren:
@mchlp @feidtmb I added support for maps now, and I think I address all of your comments as well! 🤗

feidtmb (approved):
It doesn't look like `ObjectsExportedFieldsAreEqual` is actually used anymore beyond tests, so we might want to remove it. Other than that though ✅ . (And thanks for adding map support!)

_edit:_ I think a data structure that references itself would cause an infinite loop. IMO this is acceptable for a test helper package like `testify`, but I think it would be worth noting in the function comment on `EqualExportedValues`.

mchlp (dismissed):
Looks good! Thanks for adding support for maps - I think this covers all of the composite data types.

HaraldNordgren:
Thanks again!

@feidtmb, I added a comment about infinite recursion to your PR: https://github.com/stretchr/testify/pull/1373

HaraldNordgren:
Hi again @boyan-soubachov, would you have some time to look at this?

assert/assertions.go
```diff
@@ -75,46 +75,75 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
...
-				equal = ObjectsExportedFieldsAreEqual(expectedValue.Field(i).Interface(), actualValue.Field(i).Interface())
-			} else {
-				equal = ObjectsAreEqualValues(expectedValue.Field(i).Interface(), actualValue.Field(i).Interface())
+
+	switch expectedKind {
+	case reflect.Struct:
+		result := reflect.New(expectedType).Elem()
+		for i := 0; i < expectedType.NumField(); i++ {
+			field := expectedType.Field(i)
+			isExported := field.PkgPath == "" // should use field.IsExported() but it's not available in Go 1.16.5
```
boyan-soubachov, line 94: We probably shouldn't support Go 1.16 any more since it's not officially supported anyways.

boyan-soubachov: Feel free to change the build to only support 1.20, 1.19, and 1.18 -- you can then change this code to not consider Go 1.16

HaraldNordgren: @boyan-soubachov Updated!

Be aware the the formatting the auto-formatting rules seem to have changed for versions >= 1.19. So we are now skipping that part of the build job for 1.18.

.github/workflows/main.yml
```diff
@@ -6,14 +6,18 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
       - name: Setup Go
         uses: actions/setup-go@v3.2.0
         with:
           go-version: ${{ matrix.go_version }}
-      - run: ./.ci.gogenerate.sh    
-      - run: ./.ci.gofmt.sh
+      - name: Go generate
+        if: ${{ matrix.go_version != 1.18 }}
```
boyan-soubachov, line 17: Why are we excluding `gogenerate.sh` and `gofmt.sh` from running if it's Go 1.18?

HaraldNordgren: It's because it failed. I wrote a comment about it here: https://github.com/stretchr/testify/pull/1379#discussion_r1189008882

If we don't want this extra logic, we could also drop support for 1.18, maybe it's old enough at this point?

boyan-soubachov: That's fair -> https://endoflife.date/go

We can drop 1.18 support.

HaraldNordgren: @boyan-soubachov Updated to drop 1.18!

boyan-soubachov (approved):
Thank you :)

HaraldNordgren:
Hi @nars1!

In general I would definitely advise against using the 'master' build of any library. You are better off pinning the version to testify@v1.8.2 and you would avoid this issue.

This repo has been updated to only support Go versions >=1.18. Be aware the the Go versions you are using are not officially supported anymore: https://endoflife.date/go

It would also be a good idea to upgrade your Go versions 🤗

nars1:
@HaraldNordgren : Will pin the version like you suggest and try to upgrade our Go versions. Thank you for the prompt response.

nars1:
@HaraldNordgren : How do I pin the version like you suggest. My project `go.mod` file looks as follows.

```go
module lang.yottadb.com/go/yottadb

go 1.13

require github.com/stretchr/testify v1.4.0
```

How do I say don't go more than 1.8.2? I tried searching for pinning but could not find any references. Hoping you can point me in the right direction. Thanks.

SuperQ:
This change has set the Go minimum version hint to 1.20 in `go.mod`, but continues to test with 1.19. This is confusing for downstream users.

HaraldNordgren:
Hi @SuperQ, it’s testing both for 1.19 and 1.20.

PR #1384, opened 2023-05-19
assert: Remove the struct type constraint on 'TestEqualExportedValues' to support pointer comparisons

HaraldNordgren:
Pointer comparisons are supported already, just as long as the are not the top-level element. Logically, if two item or sub-items have different unexported field, then they cannot have the same underlying address. With this change, the top-level will be treated no differently from the rest of the struct fields which also makes the logic simpler.

The normal use case here is when asserting any response data that has a pointer type. As the code currently stands on `master`, you have get the underlying values and compare those instead, otherwise it fails:

```go
require.NotNil(t, actual)
assert.EqualExportedValues(t, *expected, *actual)
```

I'm adding a test case that would have failed with the old logic.

I am also getting rid of the duplicated `ObjectsExportedFieldsAreEqual`. Since each assert function return a boolean result, we can achieve exactly the same without the extra function. This was discussed when adding the original code (https://github.com/stretchr/testify/pull/1309#discussion_r1063031396), which made sense then but now I think the time is right to remove it.

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -190,7 +190,7 @@ type S6 struct {
 	unexported string
 }
 
-func TestObjectsExportedFieldsAreEqual(t *testing.T) {
+func TestEqualExportedValues(t *testing.T) {
 
 	intValue := 1
 
@@ -225,6 +225,8 @@ func TestObjectsExportedFieldsAreEqual(t *testing.T) {
 		{Nested{&intValue, 2}, Nested{&intValue, 2}, true},
 		{Nested{&Nested{1, 2}, 3}, Nested{&Nested{1, "b"}, 3}, true},
 		{Nested{&Nested{1, 2}, 3}, Nested{nil, 3}, false},
+		{&Nested{1, 2}, &Nested{1, "b"}, true},
+		{&Nested{1, 2}, &Nested{"a", 2}, false},
 
 		{
 			Nested{map[interface{}]*Nested{nil: nil}, 2},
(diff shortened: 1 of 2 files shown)


HaraldNordgren:
Ping @boyan-soubachov

HaraldNordgren:
Ping @boyan-soubachov @mchlp @feidtmb

mchlp (approved):

HaraldNordgren:
Ping @boyan-soubachov 

This will improve the user experience for `TestEqualExportedValues`.

HaraldNordgren:
Ping @boyan-soubachov

Currently the library is usable, but it gives this error when comparing Protobuf objects because you have to access the underlying pointer, otherwise the comparison is false:

```
call of assert.EqualExportedValues copies lock value: github.com/dietdoctor/genproto/go/my-project/v1.ListItemsResponse contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutexcopylocks
```

dolmen (changes requested):
> I am also getting rid of the duplicated `ObjectsExportedFieldsAreEqual`

While this looks sane, this is not that easy. We have to preserve the API surface. Please keep the function and mark it as `// Deprecated:`.

assert/assertions_test.go
```diff
@@ -249,7 +251,8 @@ func TestObjectsExportedFieldsAreEqual(t *testing.T) {
 
 	for _, c := range cases {
 		t.Run(fmt.Sprintf("ObjectsExportedFieldsAreEqual(%#v, %#v)", c.expected, c.actual), func(t *testing.T) {
-			res := ObjectsExportedFieldsAreEqual(c.expected, c.actual)
+			mockT := new(testing.T)
+			res := EqualExportedValues(mockT, c.expected, c.actual)
 
 			if res != c.result {
 				t.Errorf("ObjectsExportedFieldsAreEqual(%#v, %#v) should return %#v", c.expected, c.actual, c.result)
```
dolmen, line 258: The message prefix must be fixed.

dolmen (changes requested):

HaraldNordgren:
@dolmen Done, I have update the code! 🤗

jufemaiz:
👀

HaraldNordgren:
@dolmen Hi, did you have time to take another look at this?

HaraldNordgren:
Hi @dolmen @boyan-soubachov, did you have a chance to take a look at this?

boyan-soubachov (dismissed):
LGTM, thank you

dolmen (changes requested):
Please open a separate MR to deprecate `ObjectsExportedFieldsAreEqual`, and move that change there.

Also cleanup commits, and rebase (do not merge master into your branch!).

HaraldNordgren:
@dolmen Done!

dolmen (approved):
Ready for v1.9.0: this is a behaviour change that require a minor release.

HaraldNordgren:
@dolmen Any update on when 1.9.0 is expected? Let's merge this in the meantime?

HaraldNordgren:
Ping @dolmen

brackendawson:
I think #1517 fully implemented this.

PR #1385, opened 2023-05-22
Add `NotImplements` and variants

### LINKED ISSUE #1065: NotImplements
A useful pattern is to have a function take a `thing` that doesn't implement an interface and return `anotherThing` that does. These glue functions work really well to connect different parts of an application, and create a useful abstraction across interfaces.

The way I test these functions is with `assert.Implements(t, interface, anotherThing)` on the result of the function, and `_, ok := thing.(interface); assert.False(ok)` on the input for good measure.

It would be _oh so convenient_ if there were an `assert.NotImplements(t, interface, thing)` I could run instead, and it would be fairly trivial to add to the API as it stands. If this isn't the kind of change that's likely to happen, is there a more semantically reasonable way of doing this?

hslatman:
## Summary
Add `assert.NotImplements` (and variants)

## Changes
Added `assert.NotImplements` (and variants)

## Motivation
There was no way to test an object did not implement a specific interface yet. Its counterpart does exist, so it was somewhat surprising to not be able to test the inverse.

```golang
import (
	"crypto/ecdsa"
	"crypto/elliptic"
	"crypto/rand"
	
)


func Test_NotASigner(t *testing.T) {
	e, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
        require.NoError(t, err)
        assert.NotImplements(t, (*crypto.Signer)(nil), e.Public()) // public key only isn't a signer
}

```

## Related issues
Closes #1065

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -613,6 +613,16 @@ func NotErrorIsf(t TestingT, err error, target error, msg string, args ...interf
 	return NotErrorIs(t, err, target, append([]interface{}{msg}, args...)...)
 }
 
+// NotImplementsf asserts that an object does not implement the specified interface.
+//
+//	assert.NotImplementsf(t, (*MyInterface)(nil), new(MyObject), "error message %s", "formatted")
+func NotImplementsf(t TestingT, interfaceObject interface{}, object interface{}, msg string, args ...interface{}) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	return NotImplements(t, interfaceObject, object, append([]interface{}{msg}, args...)...)
+}
+
 // NotNilf asserts that the specified object is not nil.
 //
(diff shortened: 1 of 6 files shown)


coditamar:
@CodiumAI-Agent please review

dolmen (approved):

dolmen:
Patch is good.

However I'm not sure we want to extend the API. At least the must wait for 1.9.0.

brackendawson (approved):
Personally I would be hesitant to include this kind of function in the API because it is somewhat a static assertion. I would prefer to encourage testing for state in the case used to justify this.

However in such cases I will defer to our predecessors, the previous acceptance of Implements and all the existing negative assertions makes this a good fit and the implementation looks sound.

At the time of writing there have been no functional changes since v1.8.5. Either we cut v1.8.6 soon and then merge this, or we merge this now and cut v1.9.0.

PR #1390, opened 2023-05-28
match function call prefer later mock.On

### LINKED ISSUE #1375: How can I retire mock.On condition to allow mock to match later mock.On?
My interface is like this
```go
type Meta interface {
	MultiPut(ctx context.Context, keys []string, values []string) error
}
```
I have multiple mock.On with mock.MatchedBy. As the function is called multiple times, I want later function call is matched by later mock.On
```go
		It("use matched by", func() {
			keys1 := make([]string, 1)
			keys1[0] = "called 0"
			keys2 := make([]string, 2)
			keys2[0] = "called 1"
			keys2[1] = "called 2"
			values := make([]string, 1)
			meta.On("MultiPut", mock.Anything, mock.MatchedBy(func(keys []string) bool {
				Expect(len(keys)).To(Equal(1))
				Expect(keys[0]).To(Equal("called 0"))
				return true
			}), mock.Anything).Once().Return(nil)
			Expect(meta.MultiPut(context.Background(), keys1, values)).To(BeNil())

			meta.On("MultiPut", mock.Anything, mock.MatchedBy(func(keys []string) bool {
				Expect(len(keys)).To(Equal(2))
				Expect(keys[0]).To(Equal("called 1"))
				Expect(keys[1]).To(Equal("called 2"))
				return true
			}), mock.Anything).Return(nil)
			Expect(meta.MultiPut(context.Background(), keys2, values)).To(BeNil())
		})


```
However, I got the following error. Clearly the first mock.On is used against the second function call.
```
 [FAILED] Expected
      <int>: 2
  to equal
      <int>: 1
  In [It] at: test.go
```

One work around is to build the exact arguments instead of using mock.MatchedBy
```go
		It("use direct value", func() {
			keys1 := make([]string, 1)
			keys1[0] = "called 0"
			keys1Dup := make([]string, len(keys1))
			copy(keys1Dup, keys1)
			keys2 := make([]string, 2)
			keys2[0] = "called 1"
			keys2[1] = "called 2"
			keys2Dup := make([]string, len(keys2))
			copy(keys2Dup, keys2)
			values := make([]string, 1)
			meta.On("MultiPut", mock.Anything, keys1Dup, mock.Anything).Return(nil)
			Expect(meta.MultiPut(context.Background(), keys1, values)).To(BeNil())

			meta.On("MultiPut", mock.Anything, keys2Dup, mock.Anything).Return(nil)
			Expect(meta.MultiPut(context.Background(), keys2, values)).To(BeNil())
		})

```
However, in some scenarios, exact arguments can be complicated to build or we may only want to match some parts of a Struct. Is it possible to retire a previous mock.On match so that later mock.On will be matched against?

I also tried to use call.Once, it does not seem to work
```go
			meta.On("MultiPut", mock.Anything, mock.MatchedBy(func(keys []string) bool {
				Expect(len(keys)).To(Equal(1))
				Expect(keys[0]).To(Equal("1"))
				return true
			}), mock.Anything).Once().Return(nil)
```

Thanks for any guidance.

egolearner:
## Summary

Match function call prefer later mock.On.

## Changes

Iterate backwardly in findExpectedCall/findClosestCall

## Motivation

If there are multiple mock.On statement, it's natural to prefer later mock.On to earlier mock.On.

## Related issues

Closes #1375

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -360,7 +360,9 @@ func (m *Mock) On(methodName string, arguments ...interface{}) *Call {
 func (m *Mock) findExpectedCall(method string, arguments ...interface{}) (int, *Call) {
 	var expectedCall *Call
 
-	for i, call := range m.ExpectedCalls {
+	for i := range m.ExpectedCalls {
+		i := len(m.ExpectedCalls) - i - 1 // iterate backwards
+		call := m.ExpectedCalls[i]
 		if call.Method == method {
 			_, diffCount := call.Arguments.Diff(arguments)
 			if diffCount == 0 {
@@ -405,7 +407,9 @@ func (c matchCandidate) isBetterMatchThan(other matchCandidate) bool {
 func (m *Mock) findClosestCall(method string, arguments ...interface{}) (*Call, string) {
 	var bestMatch matchCandidate
 
-	for _, call := range m.expectedCalls() {
+	for i := range m.expectedCalls() {
+		i := len(m.ExpectedCalls) - i - 1 // iterate backwards
+		call := m.ExpectedCalls[i]
 		if call.Method == method {
 
 			errInfo, tempDiffCount := call.Arguments.Diff(arguments)


egolearner:
From my point of view, it's natural to write the in the following way
```go
mock.On(...)
doSthWithInterface().Expect(mock arg or mock return)

mock.On(...)
anotherDoSthWithInterface().Expect(mock arg or mock return)
```
instead of 
```go
mock.On(...)
mock.On(...)

doSthWithInterface().Expect(mock arg or mock return)
anotherDoSthWithInterface().Expect(mock arg or mock return)
```

However, this modification breaks a lot of tests. I need to know if this is the right direction to go.

dolmen:
This is a breaking change. No way to have it accepted while maintaining compatibility.

PR #1393, opened 2023-06-01
Ensure that subtest name is set in SetupSubTest/TearDownSubTest

thekidder:
## Summary
Ensure that subtest name is set in SetupSubTest/TearDownSubTest

## Changes
- Ensures that suite.SetT() (with the subtest T) is called before SetupSubTest
- Ensures that the original T is restored after TearDownSubTest

## Motivation
One common usecase of SetupSubTest is to set up mocks and expectations for a table test. In this usecase, the current behavior does not set the test name until *after* SetupSubTest is run. This makes test failure reporting substantially less informative.

### DIFF
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -162,6 +162,9 @@ type SuiteTester struct {
 	SetupSubTestRunCount    int
 	TearDownSubTestRunCount int
 
+	SetupSubTestNames    []string
+	TearDownSubTestNames []string
+
 	SuiteNameBefore []string
 	TestNameBefore  []string
 
@@ -170,6 +173,8 @@ type SuiteTester struct {
 
 	TimeBefore []time.Time
 	TimeAfter  []time.Time
+
+	SuiteT *testing.T
 }
 
 // The SetupSuite method will be run by testify once, at the very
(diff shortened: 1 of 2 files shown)


suite/suite_test.go
```diff
@@ -245,23 +247,29 @@ func (suite *SuiteTester) TestSubtest() {
...
 		})
-		suite.Equal(suiteT, suite.T())
+		suite.Equal(suite.SuiteT, suite.T())
 	}
 }
 
 func (suite *SuiteTester) TearDownSubTest() {
+	subTestT := suite.T()
+	suite.NotEqual(subTestT, suite.SuiteT)
+
```
tomhutch, line 265: Hey, something like this would allow you to add your assertions alongside the others and call `assert.Contains` with the specific values we expect to see:
```suggestion
	suite.TearDownSubTestNames = append(suite.SetupSubTestNames, suite.T().Name())
```

thekidder:
Thanks for the review @tomhutch -- is this similar to what you were thinking?

thekidder:
Hey @tomhutch -- this is ready for re-review

dolmen:
This will be fixed by #1471.

@thekidder I'm sorry your own MR didn't get the attention it needed. Could you help the community effort by adding your review to #1471?

dolmen:
@tomhutch You had also reviewed this MR. Could you review #1471?

thekidder:
#1471 looks like a great fix, thanks for pointing me toward that. Will close this request.

Issue #1391, opened 2023-05-30
Data race when using `EventuallyWithT()`


pmalek:
When using [`EventuallyWithT()`](https://pkg.go.dev/github.com/stretchr/testify/assert?utm_source=godoc#EventuallyWithT) from 1.8.3 like so:

```
package main

import (
	"testing"
	"time"

	"github.com/stretchr/testify/assert"
)

func TestEventuallyWithT(t *testing.T) {
	assert.EventuallyWithT(t, func(c *assert.CollectT) {
		assert.NoError(c, assert.AnError)
	}, time.Millisecond, time.Nanosecond)
}
```

the following data race is triggered:

```
=== RUN   TestEventuallyWithT
==================
WARNING: DATA RACE
Write at 0x00c0000b42a0 by goroutine 26:
  github.com/stretchr/testify/assert.(*CollectT).Errorf()
      github.com/stretchr/testify@v1.8.3/assert/assertions.go:1869 +0x110
  github.com/stretchr/testify/assert.Fail()
      github.com/stretchr/testify@v1.8.3/assert/assertions.go:333 +0x358
  github.com/stretchr/testify/assert.NoError()
      github.com/stretchr/testify@v1.8.3/assert/assertions.go:1495 +0xe0
  eventuallwitht.TestEventuallyWithT.func1()
      eventuallwitht/main_test.go:12 +0x58
  github.com/stretchr/testify/assert.EventuallyWithT.func1()
      github.com/stretchr/testify@v1.8.3/assert/assertions.go:1933 +0x44

Previous read at 0x00c0000b42a0 by goroutine 6:
  github.com/stretchr/testify/assert.(*CollectT).Copy()
      github.com/stretchr/testify@v1.8.3/assert/assertions.go:1887 +0x58
  github.com/stretchr/testify/assert.EventuallyWithT()
      github.com/stretchr/testify@v1.8.3/assert/assertions.go:1927 +0x354
  eventuallwitht.TestEventuallyWithT()
      eventuallwitht/main_test.go:11 +0x50
  testing.tRunner()
      testing/testing.go:1576 +0x188
  testing.(*T).Run.func1()
      testing/testing.go:1629 +0x40

Goroutine 26 (running) created at:
  github.com/stretchr/testify/assert.EventuallyWithT()
      github.com/stretchr/testify@v1.8.3/assert/assertions.go:1932 +0x310
  eventuallwitht.TestEventuallyWithT()
[...]

PR #1395, opened 2023-06-02
fix: make EventuallyWithT concurrency safe

czeslavo:
## Summary

In `assert.EventuallyWithT`, `assert.CollectT` is used in a concurrent context which may lead to races as explained in https://github.com/stretchr/testify/issues/1391.

## Changes

We create a `CollectT` per every condition tick and send its errors over a channel. On the reader side of the channel, we cache the `lastTickErrs` so they can be copied over to t once the timeout is fired. The only access to `lastTickErrs` happens in the reader case and the timeout case which both run in the same goroutine.

## Motivation

We cannot use `EventuallyWithT` family of functions/methods in our codebase due to data races it causes.

## Related issues
Closes https://github.com/stretchr/testify/issues/1391.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1873,23 +1873,18 @@ func (c *CollectT) Errorf(format string, args ...interface{}) {
 }
 
 // FailNow panics.
-func (c *CollectT) FailNow() {
+func (*CollectT) FailNow() {
 	panic("Assertion failed")
 }
 
-// Reset clears the collected errors.
-func (c *CollectT) Reset() {
-	c.errors = nil
+// Deprecated: That was a method for internal usage that should not have been published. Now just panics.
+func (*CollectT) Reset() {
+	panic("Reset() is deprecated")
 }
 
-// Copy copies the collected errors to the supplied t.
-func (c *CollectT) Copy(t TestingT) {
-	if tt, ok := t.(tHelper); ok {
-		tt.Helper()
-	}
-	for _, err := range c.errors {
-		t.Errorf("%v", err)
-	}
+// Deprecated: That was a method for internal usage that should not have been published. Now just panics.
+func (*CollectT) Copy(TestingT) {
+	panic("Copy() is deprecated")
 }
 
 // EventuallyWithT asserts that given condition will be met in waitFor time,
@@ -1915,8 +1910,8 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 		h.Helper()
 	}
 
-	collect := new(CollectT)
-	ch := make(chan bool, 1)
+	var lastFinishedTickErrs []error
+	ch := make(chan []error, 1)
 
 	timer := time.NewTimer(waitFor)
 	defer timer.Stop()
@@ -1927,19 +1922,23 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 	for tick := ticker.C; ; {
 		select {
 		case <-timer.C:
-			collect.Copy(t)
+			for _, err := range lastFinishedTickErrs {
+				t.Errorf("%v", err)
+			}
 			return Fail(t, "Condition never satisfied", msgAndArgs...)
 		case <-tick:
 			tick = nil
-			collect.Reset()
 			go func() {
+				collect := new(CollectT)
 				condition(collect)
-				ch <- len(collect.errors) == 0
+				ch <- collect.errors
 			}()
-		case v := <-ch:
-			if v {
+		case errs := <-ch:
+			if len(errs) == 0 {
 				return true
 			}
+			// Keep the errors from the last ended condition, so that they can be copied to t if timeout is reached.
+			lastFinishedTickErrs = errs
 			tick = ticker.C
 		}
 	}
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1930,8 +1930,10 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 			tick = nil
 			go func() {
 				collect := new(CollectT)
+				defer func() {
+					ch <- collect.errors
+				}()
 				condition(collect)
-				ch <- collect.errors
 			}()
 		case errs := <-ch:
 			if len(errs) == 0 {
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -2766,11 +2766,22 @@ func TestEventuallyTrue(t *testing.T) {
 	True(t, Eventually(t, condition, 100*time.Millisecond, 20*time.Millisecond))
 }
 
+// errorsCapturingT is a mock implementation of TestingT that captures errors reported with Errorf.
+type errorsCapturingT struct {
+	errors []error
+}
+
+func (t *errorsCapturingT) Errorf(format string, args ...interface{}) {
+	t.errors = append(t.errors, fmt.Errorf(format, args...))
+}
+
+func (t *errorsCapturingT) Helper() {}
+
 func TestEventuallyWithTFalse(t *testing.T) {
-	mockT := new(CollectT)
+	mockT := new(errorsCapturingT)
 
 	condition := func(collect *CollectT) {
-		True(collect, false)
+		Fail(collect, "condition fixed failure")
 	}
 
 	False(t, EventuallyWithT(mockT, condition, 100*time.Millisecond, 20*time.Millisecond))
@@ -2778,7 +2789,7 @@ func TestEventuallyWithTFalse(t *testing.T) {
 }
 
 func TestEventuallyWithTTrue(t *testing.T) {
-	mockT := new(CollectT)
+	mockT := new(errorsCapturingT)
 
 	state := 0
 	condition := func(collect *CollectT) {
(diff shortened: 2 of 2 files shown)


brackendawson (approved):
I was thinking whether this can be fixed by enforcing that condition funcs can never overlap but the documentation is clear that it is checked every tick, so that would be a breaking change. The test relies on the race detector, I don't see a better way to do it.

brackendawson (approved):

assert/assertions.go
```diff
@@ -1884,11 +1890,20 @@ func (c *CollectT) Copy(t TestingT) {
...
 	}
+	c.mu.RLock()
+	defer c.mu.RUnlock()
 	for _, err := range c.errors {
 		t.Errorf("%v", err)
 	}
 }
 
+// HasErrors returns true if any errors were collected.
+func (c *CollectT) HasErrors() bool {
```
dolmen, line 1901: `HasErrors` should be private (`Reset` and `Copy` should have been made private too unfortunately, but they are now part of the public API).

dolmen (changes requested):

dolmen:
So while this effectively serializes calls to CollectT methods (#1418), I'm not convinced this really fixes the core issue.

This implementation doesn't fix the real problem which that we may have multiple instances of the `condition` function running at the same time (overlapping) but we only use a single `CollectT` instance to collect the results, with calls to `Reset` which could alter the result of conditions.

I think we will merge this (with `HasErrors` made private), but work on `EventuallyWithT` must not be considered done.

czeslavo:
> So while this effectively serializes calls to CollectT methods (#1418), I'm not convinced this really fixes the core issue.
> 
> This implementation doesn't fix the real problem which that we may have multiple instances of the `condition` function running at the same time (overlapping) but we only use a single `CollectT` instance to collect the results, with calls to `Reset` which could alter the result of conditions.
> 
> I think we will merge this (with `HasErrors` made private), but work on `EventuallyWithT` must not be considered done.

@dolmen Thanks for the comment, you're right! I've taken a look into that again and came up with a solution that doesn't involve a mutex. It also doesn't allow a situation where we end up with an empty `CollectT` on timeout in case of timeout firing up just after `collect.Reset()` call. We create a `CollectT` per every condition tick and send its errors over a channel. On the reader side of the channel, we cache the `lastTickErrs` so they can be copied over to `t` once the timeout is fired. The only access to `lastTickErrs` happens in the reader case and the timeout case which both run in the same goroutine.

assert/assertions_test.go
```diff
@@ -2786,6 +2786,18 @@ func TestEventuallyWithTTrue(t *testing.T) {
 	Len(t, mockT.errors, 0)
 }
 
+func TestEventuallyWithT_ConcurrencySafe(t *testing.T) {
+	mockT := new(CollectT)
```
dolmen, line 2790: Don't use `CollectT` here.

dolmen (changes requested):
This looks very good!

This is the solution I want!

Also it will allow us to get rid of `CollectT` which was a design error of the original implementation (a separate work for which I am studying plans since a week).

dolmen:
As the `Reset` and `Copy` methods of `CollectT` are not needed anymore, and they should have been made private in the first place, please replace their implementation with a panic and add a `// Deprecated:` comment.

dolmen:
@czeslavo Note for the future, that completely replacing your original implementation on which at least 2 other people worked too (reviewers) with a completely different one, is not a good practice. It would have been better to open a separate PR and close this one.

Anyway, in this case, the original implementation can be forgotten.

assert/assertions_test.go
czeslavo: Done. Please note this prevents us from checking how many errors were reported in this test. Nevertheless we have others that do check that and this one was meant for triggering the race itself so I guess it can be like that.

dolmen: I didn't say to not capture errors.

Just don't use CollectT for that.

Because my aim is to deprecate CollectT.

dolmen (changes requested):

dolmen:
"address review comments" is never a good commit message. It is completely useless when looking at the Git log or with "git blame". A good commit message must give a summary of what changes, and, if necessary, why. A commit that ONLY says why is never good.

In the present case squashing the commits is the best thing to do.

assert/assertions_test.go
```diff
@@ -2786,6 +2786,17 @@ func TestEventuallyWithTTrue(t *testing.T) {
 	Len(t, mockT.errors, 0)
 }
 
+func TestEventuallyWithT_ConcurrencySafe(t *testing.T) {
```
cszczepaniak, line 2789: It may be beneficial to have an additional test which explicitly exposes the bug this PR fixes. [Here](https://github.com/stretchr/testify/pull/1438/files#diff-7190065248695df477329c4cada2d71f323ab7940ecffff2d3d5fa43caf79427R2766) is an example -- if the first check of the condition is allowed to complete quickly and the next takes longer than the timeout, it is very easy to expose the bug.

assert/assertions.go
```diff
@@ -1924,19 +1921,23 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
...
-				ch <- len(collect.errors) == 0
+				ch <- collect.errors
 			}()
-		case v := <-ch:
-			if v {
+		case errs := <-ch:
+			if len(errs) == 0 {
 				return true
 			}
+			// Keep the last tick's errors, so that they can be copied to t if the condition is not met on time.
```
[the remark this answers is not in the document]
czeslavo: Addressed.

assert/assertions_test.go
czeslavo: Thank you for the suggestion. Yes, this reproduces the issue of reporting no errors due to the `Reset()` call just before running another condition. Allowed myself to copy the test and supplement that with comments. 👍

assert/assertions_test.go
czeslavo: 👍 Added `errorsCapturingT` mock implementation so that it fits the `EventuallyWithT` needs.

assert/assertions_test.go
```diff
@@ -2786,6 +2795,36 @@ func TestEventuallyWithTTrue(t *testing.T) {
...
+}
+
+func TestEventuallyWithT_ReturnsTheLatestFinishedConditionErrors(t *testing.T) {
+	var calledOnce bool
+	condition := func(collect *CollectT) {
+		if calledOnce {
+			// Sleep to ensure that the second condition runs longer than timeout.
+			time.Sleep(time.Second)
+			return
+		}
```
dolmen, line 2817: To better protect against race condition (the implementation of EventuallyWithT doesn't run overlapping conditions for now, but we don't want the test to be sensible to such a change), it would be better to avoid using a variable shared by multiple runs of the condition.

So it would be wise to use a channel of bools instead of `calledOnce`.

```go
mustSleep := make(chan bool, 2)
mustSleep <- false
mustSleep <- true
close(mustSleep)

...

if <-mustSleep {
    time.Sleep(time.Second)
}
```

assert/assertions_test.go
```diff
@@ -2786,6 +2795,36 @@ func TestEventuallyWithTTrue(t *testing.T) {
...
+
+		// The first condition will fail. We expect to get this error as a result.
+		True(collect, false)
```
dolmen, line 2820: Use `Fail` instead. Or directly `collect.Errorf`.

```suggestion
		Fail(collect, "condition fixed failure")
```

assert/assertions_test.go
```diff
@@ -2786,6 +2795,36 @@ func TestEventuallyWithTTrue(t *testing.T) {
...
+		calledOnce = true
+	}
+
+	mockT := new(testing.T)
```
dolmen, line 2824: Use your `errorsCapturingT`.

dolmen (changes requested):

czeslavo:
@dolmen Is there anything else that I could do that's blocking this PR from merging?

dolmen (changes requested):

assert/assertions.go
```diff
@@ -1927,19 +1922,23 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
...
+			}
 			return Fail(t, "Condition never satisfied", msgAndArgs...)
 		case <-tick:
 			tick = nil
-			collect.Reset()
 			go func() {
+				collect := new(CollectT)
 				condition(collect)
-				ch <- len(collect.errors) == 0
+				ch <- collect.errors
```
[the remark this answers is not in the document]
marshall-lee: Yes this way is more safer and is mandatory if `FailNow()` will use `runtime.Goexit()` as in #1481.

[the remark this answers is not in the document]
czeslavo: Yeah, that makes sense. I've applied the suggestion. 

I'll _defer_ 🙂  implementing the test case ensuring errors are collected when the panic occurs to https://github.com/stretchr/testify/issues/1457 because that will depend on how we will handle it. As I understand, `runtime.Goexit()` in `CollectT.FailNow` will unblock using `require` assertions, but not ANY panic that could occur in conditions. To handle any panic, we could catch the panic in the goroutine running the condition, i.e.:

```go
collect := new(CollectT)
defer func() {
	if err := recover(); err != nil {
		collect.Errorf("Panic caught from condition: %v", err)
	}
	ch <- collect.errors
}()
condition(collect)
```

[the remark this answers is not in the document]
dolmen: @czeslavo You're right. I'll merge as is. I also notice that @marshall-lee has already implemented the defer in #1481 which is in the pipe.

dolmen (approved):

dolmen:
Thank you @czeslavo and all reviewers!

PR #1398, opened 2023-06-05
fix: update Empty()'s error message

flowck:
## Summary
The message returned by `Empty()` is a bit confusing because it states the following: `Should be empty, but was [1, 2].`

## Changes
* Update the error message thrown/panicked by `assert.Empty()`
* Update the error message thrown/panicked by `assert.NotEmpty()`

## Testcase
https://go.dev/play/p/Rx_KYuAQe2J

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -705,7 +705,7 @@ func Empty(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 		if h, ok := t.(tHelper); ok {
 			h.Helper()
 		}
-		Fail(t, fmt.Sprintf("Should be empty, but was %v", object), msgAndArgs...)
+		Fail(t, fmt.Sprintf("Should be empty, but isn't %v", object), msgAndArgs...)
 	}
 
 	return pass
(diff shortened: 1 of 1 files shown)


assert/assertions.go
```diff
@@ -705,7 +705,7 @@ func Empty(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 		if h, ok := t.(tHelper); ok {
 			h.Helper()
 		}
-		Fail(t, fmt.Sprintf("Should be empty, but was %v", object), msgAndArgs...)
+		Fail(t, fmt.Sprintf("Should be empty, but isn't %v", object), msgAndArgs...)
```
dolmen, line 708: The message shows the unexpected value. `is` is ok.
```suggestion
		Fail(t, fmt.Sprintf("Should be empty, but is %v", object), msgAndArgs...)
```

dolmen (changes requested):

dolmen:
The problem you see in the original message is not obvious to me. Could you explain with more details what you find confusing in that message and how you are solving the issue?

PR #1400, opened 2023-06-08
Fix #1399: Always report error message on PanicsWithError mismatch

### LINKED ISSUE #1399: PanicsWithError does not report error message for complex errors
When `PanicsWithError` fails with a message different from the expected one, the message might not indicate very well what the error was, and what the string mismatch is exactly.

It *does* work in simple cases:
```go
assert.PanicsWithError("expected message", func() {
	panic(errors.New("surprising message"))
})
```
Fails with: 
```
func (assert.PanicTestFunc)(0x84d480) should panic with error message:	"expected message"
        	       Panic value:	&errors.errorString{s:"surprising message"}
        	       Panic stack:	goroutine 19 [running]:
```

It *does not* indicate the actual error message in case the error message is not in the fields:
```go
assert.PanicsWithError("expected message", func() {
	panic(errors.Join(errors.New("surprising message")))
})
```
Fails with:
```
func (assert.PanicTestFunc)(0x84d4e0) should panic with error message:	"expected message"
           		Panic value:	&errors.joinError{errs:[]error{(*errors.errorString)(0xc000119570)}}
            		Panic stack:	goroutine 19 [running]:
```

A real world example here is https://pkg.go.dev/text/template#ExecError.

---

Since the assertion expects this to be an error and it checks on the error message only, it would suffice to replace the listing of entire "Panic value" with `panicErr.Error()`.

olivergondza:
## Summary
Always report error message on `PanicsWithError` error message mismatch.

## Changes
Add `Error message` field to the failure message, that reports `panicErr.Error()` IFF it panicked with an error its message is different from expected.

## Motivation
Better defect localization in case of nested errors causing panic.

Before:
```
Error: func (assert.PanicTestFunc)(0x627a80) should panic with error message:	"at the disco"
	Panic value:	&errors.joinError{errs:[]error{(*errors.errorString)(0xc0000a1540), (*errors.errorString)(0xc0000a1550)}}
	Panic stack:	goroutine 19 [running]:
	...
```
After:
```
Error: func (assert.PanicTestFunc)(0x627b20) should panic with error message:	"at the disco"
	Panic value:	&errors.joinError{errs:[]error{(*errors.errorString)(0xc000065550), (*errors.errorString)(0xc000065560)}}
	Error message:	"wrapped err msg\nother err msg"
	Panic stack:	goroutine 6 [running]:
	...
```

## Related issues
Closes https://github.com/stretchr/testify/issues/1399

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1212,7 +1212,15 @@ func PanicsWithError(t TestingT, errString string, f PanicTestFunc, msgAndArgs .
 	}
 	panicErr, ok := panicValue.(error)
 	if !ok || panicErr.Error() != errString {
-		return Fail(t, fmt.Sprintf("func %#v should panic with error message:\t%#v\n\tPanic value:\t%#v\n\tPanic stack:\t%s", f, errString, panicValue, panickedStack), msgAndArgs...)
+		errorMsg := ""
+		if ok {
+			errorMsg = fmt.Sprintf("\tError message:\t%#v\n", panicErr.Error())
+		}
+		msg := fmt.Sprintf(
+			"func %#v should panic with error message:\t%#v\n\tPanic value:\t%#v\n%s\tPanic stack:\t%s",
+			f, errString, panicValue, errorMsg, panickedStack,
+		)
+		return Fail(t, msg, msgAndArgs...)
 	}
 
 	return true
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -1339,29 +1339,56 @@ func TestPanicsWithValue(t *testing.T) {
 
 func TestPanicsWithError(t *testing.T) {
 
-	mockT := new(testing.T)
-
+	mockT := new(CollectT)
 	if !PanicsWithError(mockT, "panic", func() {
 		panic(errors.New("panic"))
 	}) {
 		t.Error("PanicsWithError should return true")
 	}
+	Len(t, mockT.errors, 0)
 
+	mockT = new(CollectT)
 	if PanicsWithError(mockT, "Panic!", func() {
 	}) {
 		t.Error("PanicsWithError should return false")
 	}
+	if Len(t, mockT.errors, 1) {
+		actual := mockT.errors[0].Error()
+		Contains(t, actual, "Panic value:	<nil>")
+	}
 
+	mockT = new(CollectT)
 	if PanicsWithError(mockT, "at the disco", func() {
-		panic(errors.New("panic"))
+		panic(errors.New("actual err msg"))
 	}) {
 		t.Error("PanicsWithError should return false")
 	}
+	if Len(t, mockT.errors, 1) {
+		actual := mockT.errors[0].Error()
+		Contains(t, actual, `Error message:	"actual err msg"`)
+	}
 
+	mockT = new(CollectT)
+	if PanicsWithError(mockT, "at the disco", func() {
+		panic(errors.Join(errors.New("wrapped err msg"), errors.New("other err msg")))
+	}) {
+		t.Error("PanicsWithError should return false")
+	}
+	if Len(t, mockT.errors, 1) {
+		actual := mockT.errors[0].Error()
+		Contains(t, actual, `Error message:	"wrapped err msg\nother err msg"`)
+	}
+
+	mockT = new(CollectT)
 	if PanicsWithError(mockT, "Panic!", func() {
 		panic("panic")
 	}) {
 		t.Error("PanicsWithError should return false")
+	}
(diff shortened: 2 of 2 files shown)


assert/assertions_test.go
```diff
@@ -1338,31 +1338,70 @@ func TestPanicsWithValue(t *testing.T) {
 }
 
 func TestPanicsWithError(t *testing.T) {
+	type NestedErr struct {
+		Err error
+	}
 
-	mockT := new(testing.T)
-
+	mockT := new(CollectT)
```
dolmen: Do not use CollectT for that purpose. `CollectT` is reserved for use with `EventuallyWithT`.

`captureTestingT` is probably what you need instead.

dolmen (changes requested):
The tests must be fixed to not use `assert.CollectT`.

olivergondza:
@dolmen, I have performed the needed refactor and rebased. PTAL.

The `captureTestingT` could not have been used in the most straightforward way possible, as some of the checks are negative, and test other "labels" than error...

assert/assertions_test.go
```diff
@@ -1576,30 +1576,43 @@ func TestPanicsWithValue(t *testing.T) {
 func TestPanicsWithError(t *testing.T) {
 	t.Parallel()
 
-	mockT := new(testing.T)
-
-	if !PanicsWithError(mockT, "panic", func() {
+	mckT := new(captureTestingT)
```
dolmen: Rename mckT to mockT for consistency with other tests.

dolmen (changes requested):

assert/assertions_test.go
olivergondza: Good catch, amended!

assert/assertions.go
```diff
@@ -1336,7 +1336,15 @@ func PanicsWithError(t TestingT, errString string, f PanicTestFunc, msgAndArgs .
...
 	if !ok || panicErr.Error() != errString {
-		return Fail(t, fmt.Sprintf("func %#v should panic with error message:\t%#v\n\tPanic value:\t%#v\n\tPanic stack:\t%s", f, errString, panicValue, panickedStack), msgAndArgs...)
+		errorMsg := ""
+		if ok {
+			errorMsg = fmt.Sprintf("\tError message:\t%#v\n", panicErr.Error())
+		}
+		msg := fmt.Sprintf(
+			"func %#v should panic with error message:\t%#v\n\tPanic value:\t%#v\n%s\tPanic stack:\t%s",
+			f, errString, panicValue, errorMsg, panickedStack,
+		)
```
brackendawson: There is a cleaner implementation here that doesn't require the errorMsg variable. Either using strings.Builder or by appending to msg we can remove errorMsg and significantly clean up the Sprintf format string, which was already quite messy.

olivergondza: Good point. Simplified.

assert/assertions.go
```diff
@@ -1343,9 +1343,15 @@ func PanicsWithError(t TestingT, errString string, f PanicTestFunc, msgAndArgs .
...
-	panicErr, ok := panicValue.(error)
-	if !ok || panicErr.Error() != errString {
-		return Fail(t, fmt.Sprintf("func %#v should panic with error message:\t%#v\n\tPanic value:\t%#v\n\tPanic stack:\t%s", f, errString, panicValue, panickedStack), msgAndArgs...)
+	panicErr, isError := panicValue.(error)
+	if !isError || panicErr.Error() != errString {
+		msg := fmt.Sprintf("func %#v should panic with error message:\t%#v\n", f, errString)
+		if isError {
+			msg += fmt.Sprintf("\tError message:\t%#v\n", panicErr.Error())
+		}
+		msg += fmt.Sprintf("\tPanic value:\t%#v\n", panicValue)
```
ccoVeille, line 1352: Is there a need for panic value when isError is satisfied?


```suggestion
			msg += fmt.Sprintf("\tError message:\t%#v\n", panicErr.Error())
		} else {
			msg += fmt.Sprintf("\tPanic value:\t%#v\n", panicValue)
		}
```

olivergondza: In theory, it should not be. Not sure how reliable it is in practice that that error's `.Error()` will always give all the info.

ccoVeille: Indeed.

Maybe tiu should add tests about this, an error providing more info. I'm unsure how it can happen.

olivergondza: On a second though, let's keep it. It contains the exact type information, not just string representation of the error (see PR description for what would be list if "Panic value" gets omited).

PR #1401, opened 2023-06-10
mock: fix map[string]struct{} map[string]interface{} Argument compare panic

lecw199:
## Summary

Panic in https://go.dev/play/p/CPpk9RyaCI1 (test case added by @dolmen)

## Changes
Change `AnythingOfType` matching by ignoring spaces in types names (description added by @dolmen)

## Motivation

the case in the mock.go file will case panic
```go
	Mock.On("Unmarshal", AnythingOfType("*map[string]interface{}")).Return().Run(func(args Arguments) {
		arg := args.Get(0).(*map[string]interface{})
		arg["foo"] = "bar"
	})
```

`reflect.TypeOf(actual).String()` will print `*map[string]interface {}`，not `*map[string]interface{}`, there will be an additional space  in the former.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -967,7 +967,7 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			switch expected := expected.(type) {
 			case anythingOfTypeArgument:
 				// type checking
-				if reflect.TypeOf(actual).Name() != string(expected) && reflect.TypeOf(actual).String() != string(expected) {
+				if reflect.TypeOf(actual).Name() != string(expected) && strings.ReplaceAll(reflect.TypeOf(actual).String(), " ", "") != strings.ReplaceAll(string(expected), " ", "") {
 					// not match
 					differences++
 					output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type %s - %s\n", output, i, expected, reflect.TypeOf(actual).Name(), actualFmt)
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -1739,6 +1739,28 @@ func Test_Arguments_Diff_WithAnythingOfTypeArgument(t *testing.T) {
 
 }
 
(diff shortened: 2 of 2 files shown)


dolmen:
Please take more time to edit the description of this issue. This project gets many request, so please spare the valuable maintainers' time.

Also adding a reproducible test case running on go.dev/play would be helpful.

lecw199:
I'm sorry,  show the code: https://go.dev/play/p/JL7B_CU8tiJ

lecw199:
Pay attention to this sentence， **reflect.TypeOf(actual).String() will print *map[string]interfac {}，not *map[string]interfac{} , there will be an additional space in the former.**

dolmen:
Here is a simplified and self-contained test case:

https://go.dev/play/p/CPpk9RyaCI1

```go
type X struct {
	mock.Mock
}

func (x *X) M(m *map[string]interface{}) {
	_ = x.Mock.Called(m)
}

func Test(t *testing.T) {
	var x X
	x.On("M", mock.AnythingOfType("*map[string]interface{}")).Return(nil).Run(func(args mock.Arguments) {
		arg := args.Get(0).(*map[string]interface{})
		(*arg)["foo"] = "bar"
	})

	m := make(map[string]interface{})
	x.M(&m)
}
```

dolmen:
@lecw199 The workaround for this case is to use `mock.IsType((*map[string]interface{})(nil))` instead of `mock.AnythingOfType("*map[string]interface{}")`. This is type safe and do not depend on stringification of types (which might change from one Go version to the other).

I am not convinced by the proposed patch which, I fear, could trigger other issues.

dolmen:
I'm marking this as "wontfix" as the right solution is to use `mock.IsType`.

(My personal opinion: in v2 we should just drop `mock.AnythingOfType(string)` and keep only `mock.IsType`).

PR #1406, opened 2023-07-03
codegen: use standard header for generated files

dolmen:
## Summary

Modify _codegen to use the standard header in generated Go sources.
See https://go.dev/s/generatedcode.

## Motivation

This enables editor support for blocking manual changes in generated sources.


dolmen:
#1338 already proposed that change. Mine also add references to the documentation of the standard.

brackendawson (approved):
Thanks!

PR #1407, opened 2023-07-05
chore: remove refs to deprecated io/ioutil

testwill:
## Summary

"io/ioutil" has been deprecated since Go 1.19: As of Go 1.16
## Motivation

remove refs to deprecated io/ioutil

### DIFF
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"flag"
-	"io/ioutil"
+	"io"
 	"math/rand"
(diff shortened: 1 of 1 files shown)


SuperQ (approved):

SuperQ:
We already have this problem by using Go modules (1.13), as well as specifying a minimum Go version in the `go.mod` as 1.19.

I haven't done an exhaustive search, but I'm guessing there are already features in use that make this unusable in older versions.

SuperQ:
The CI tests are updated, so a rebase should show this is safe.

Issue #865, opened 2020-01-09
Data race when calling assert.Eventually()


snebel29:
Hi,
Today I realize that one of my tests is flaky, working most of times but failing others with a data race being reported.

I'm using `github.com/stretchr/testify v1.4.0`

I have the impression that the following return causes the function to exit and close `checkPassed` channel.
https://github.com/stretchr/testify/blob/221dbe5ed46703ee255b1da0dec05086f5035f62/assert/assertions.go#L1486-L1487

Which is still being targeted by the following goroutine.
https://github.com/stretchr/testify/blob/221dbe5ed46703ee255b1da0dec05086f5035f62/assert/assertions.go#L1493-L1495

This would only happen when the goroutine with `condition()` is running but not finish yet, and the timer hits causing the return close.

Assuming I'm right I'm not sure yet what's the best way to fix it, may be decoupling those two into different select and using more channels, or may be there is some trick to not try to send if the channel was closed/deallocated, I would happily create a PR once/if in agreement...

Here is the data race trace.
```
WARNING: DATA RACE
Write at 0x00c00011a190 by goroutine 9:
  runtime.closechan()
      /usr/local/go/src/runtime/chan.go:334 +0x0
  github.com/stretchr/testify/assert.Eventually()
      /home/user/workspace/project/vendor/github.com/stretchr/testify/assert/assertions.go:1487 +0x3d5
  github.com/stretchr/testify/assert.Eventuallyf()
      /home/user/workspace/project/vendor/github.com/stretchr/testify/assert/assertion_format.go:124 +0x197
  example.com/group/project/pkg/configwatcher.TestFileWatcherReturnErrorOnTimeoutWhenFileNotExist_Start()
      /home/user/workspace/project/pkg/configwatcher/file_watcher_test.go:55 +0x3d7
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:909 +0x199

Previous read at 0x00c00011a190 by goroutine 25:
  runtime.chansend()
      /usr/local/go/src/runtime/chan.go:142 +0x0
  github.com/stretchr/testify/assert.Eventually.func1()
      /home/user/workspace/project/vendor/github.com/stretchr/testify/assert/assertions.go:1494 +0x54
Data 
Goroutine 9 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:960 +0x651
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:1202 +0xa6
[...]

PR #1412, opened 2023-07-06
assert: fix TestEventuallyTimeout

### LINKED ISSUE #1411: TestEventuallyIssue805: flaky test
We still have flaky tests. It looks like #865 did not fix everything related to #805 and #835.
```console
$ go test -race -failfast -count 100 -run TestEventuallyIssue805 ./assert 
--- FAIL: TestEventuallyIssue805 (0.00s)
    assertions_test.go:2815: 
        	Error Trace:	
        	Error:      	Should be false
        	Test:       	TestEventuallyIssue805
FAIL
FAIL	github.com/stretchr/testify/assert	0.451s
FAIL
```

Cc: @snebel29

dolmen:
## Summary
Fix TestEventuallyIssue805 which was flaky because of the use of a timer that has exactly the same duration as the Eventually timout. But time is not a synchronization primitive.

## Changes
* Now we use channels to ensure that the condition is still running and Eventually times out before checking its return value.
* The test is also renamed to TestEventuallyTimeout to more clearly show its purpose.

## Motivation
Test was flaky (see #1411). CI was unreliable.

## Related issues
Closes #1411.
Related to #805, #869, #865.


SuperQ (approved):
Nice find

SuperQ:
I'd also recommend changing the timer values for this to be similar to the other timer tests to be values larger than a single millisecond/microsecond. I've seen cases where CI tests be flaky because of too-close (+-1ms) a timer wait.

brackendawson:
The durations shouldn't need to be increased now because there are no longer two short timers running at the same time. Increasing them would only slow down the test. I'd approve this as is.

I've found similar cases where a potential allocation by `time.After` can slow down the creation of one timer after the other has started, this can be solved like so: https://github.com/stretchr/testify/pull/1124/commits/8cd90172afd36768960e21660b1ef09c4bfd6d46

assert/assertions_test.go
```diff
@@ -2807,12 +2807,31 @@ func TestNeverTrue(t *testing.T) {
...
-func TestEventuallyIssue805(t *testing.T) {
+// Check that a long running condition doesn't block Eventually.
+// See issue 805 (and its long tail of following issues)
+func TestEventuallyTimeout(t *testing.T) {
 	mockT := new(testing.T)
 
 	NotPanics(t, func() {
-		condition := func() bool { <-time.After(time.Millisecond); return true }
+		t.Log("start")
+		done, done2 := make(chan bool), make(chan bool)
```
stevenh, line 2817: It's idiomatic to use a `struct{}` instead of `bool` as it takes no space.
```suggestion
		done, done2 := make(chan struct{}), make(chan struct{})
```

stevenh (changes requested):
Nice approach to eliminate the timing totally.

I would use the idiomatic `chan struct{}` for the done channel's, but that's just a nit given it's a test.

Not sure if you intended to leave the logging in or not?

assert/assertions_test.go
```diff
@@ -2807,12 +2807,31 @@ func TestNeverTrue(t *testing.T) {
...
 
-func TestEventuallyIssue805(t *testing.T) {
+// Check that a long running condition doesn't block Eventually.
+// See issue 805 (and its long tail of following issues)
+func TestEventuallyTimeout(t *testing.T) {
 	mockT := new(testing.T)
 
 	NotPanics(t, func() {
-		condition := func() bool { <-time.After(time.Millisecond); return true }
+		t.Log("start")
```
brackendawson, line 2816: Good spot on the t.Log @stevenh. Looking at the rest of the package there is no other test which logs on success, so we should probably hold that convention.

brackendawson (changes requested):

assert/assertions_test.go
dolmen: Thanks for the catch.

Fixed (amending my commit).

assert/assertions_test.go
dolmen: Well, I think that the logs are helpful to show what happens. In fact the lack of logging was what made debugging the issue harder.

The sequence of events is really important in this test, and the log output allows to validate visually that everything happens in the right order. I think this will be useful for regression testing if the test has to be changed later.

This is the only test that has verbose log output, but I think that's is what verbose output is for. This is a case that is worth it.

The log messages could be improved though. Suggestions welcome.

stevenh: In my opinion good comments would be more valuable, you already have some.

The goal as you say is to help those who might have to debug it in the future, by making it clear why flow is how it is.

I don't think you need logging to achieve this, and the current output e.g. `ok1` and `ok2` doesn't add any value as it doesn't describe what is happening or why, leaving the user thinking what is `ok1` etc?

dolmen: Logs removed.

craig65535 (approved):

dolmen:
Following [my announcement on Slack](https://gophers.slack.com/archives/C92321HDJ/p1698873887431609), I'm pinging all other co-maintainers.

MovieStoreGuy (approved):

MovieStoreGuy:
I will be happy to merge this once @brackendawson has reviewed again after the request for changes.

stevenh (approved):

brackendawson (approved):
Looks great!

PR #1416, opened 2023-07-11
mock: optimize argument matching checks

dolmen:
## Summary
Two commits to refactor mock.Arguments.Diff to simplify matching by reducing use of `reflect`.

## Changes
* commit 1: replace multiple comparisons `reflect.TypeOf(actual) == reflect.TypeOf(constant)` with a type switch
* commit 2: refactor `mock.IsType` to take the type early

## Motivation
Reduce use of reflect for more speed and more compiler checks.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -960,52 +960,54 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 				differences++
 				output = fmt.Sprintf("%s\t%d: FAIL:  %s not matched by %s\n", output, i, actualFmt, matcher)
 			}
-		} else if reflect.TypeOf(expected) == reflect.TypeOf((*anythingOfTypeArgument)(nil)).Elem() {
-			// type checking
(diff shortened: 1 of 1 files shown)


dolmen:
Cc: @brackendawson @LandonTClipp @SuperQ for review.

LandonTClipp:
Do you have any performance comparisons between this and the old style? Also could you point me to the relevant tests that would be touching these methods (I'm fairly unfamiliar with the code base so just want to see what the tests are currently doing).

dolmen:
@LandonTClipp I am not more familiar than you with the codebase. I'm discovering it while reviewing issues/PRs. So I relied on existing code coverage (`cd mock; go test -coverprofile=.coverage.out && exec go tool cover -html=.coverage.out`) to check that the modified code was fully covered.

Besides that, I did not benchmark. But just be aware as a general rule that any use of reflect must be avoided. Type assertions and type switches are always faster. Also code is much more readable.

MovieStoreGuy (approved):

PR #1421, opened 2023-07-12
suite: fix test failures

stevenh:
## Summary
Fix TestFailfastSuite when run with go test flag -count=X where X greater than 1.

## Motivation
So we don't get false test failures causing confusion.

### DIFF
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -604,14 +604,44 @@ func TestFailfastSuite(t *testing.T) {
 		}},
 	)
 	assert.False(t, ok)
+	var expect []string
 	if failFast {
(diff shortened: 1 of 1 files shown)


stevenh:
@dolmen sorry missed the notification for this, conflict fixed.

brackendawson (approved):
I'm not super happy with `if failFast {` having an `else` but it was already there.

This looks to me like the most appropriate fix for the issue. Down to us running testing.RunTests.

We could avoid a lot of the acrobatics by replacing callOrder with a map or set of booleans to indicate whether test B was called, but then we lose the testing that setup and teardown weren't called for it.

Issue #1422, opened 2023-07-13
assert: add Helper() in TestingT interface


dolmen:
Since Go 1.9 The [`testing.T`](https://pkg.go.dev/testing/#T.Helper)/[`testing.B`](https://pkg.go.dev/testing/#B.Helper)/[`testing.F`](https://pkg.go.dev/testing/#F.Helper) and more generally the [`testing.TB`](https://pkg.go.dev/testing/#TB) interface all have an [`Helper()`]((https://pkg.go.dev/testing/#T.Helper)) method that must be used by test helpers to hide them in the caller stack in order to properly report test errors at the location where the user cares.

In testify MRs #554, #1026 added the call to `Helper()` in every testify function. Here is the code:
```go
 if h, ok := t.(tHelper); ok {
	h.Helper()
}
```

This works fine, but:
* this code is verbose
* every standard use of testify is done with a testing context (`testing.TB`) that supports `Helper()`, so the guard 
* the interface guard make every testify function slower than necessary

It would be easier if we could just do:
```go
t.Helper()
```

However our current definition of `assert.TestingT` is:
```go
// TestingT is an interface wrapper around *testing.T                                                                                                                                                
type TestingT interface {                                                                                                                                                                            
        Errorf(format string, args ...interface{})                                                                                                                                                   
}
```

I propose to add the `Helper()` method to `assert.TestingT`.

This can be considered a breaking change of the API contract. However:
* this will not affect any user code that calls the testify methods with values implementing `testing.TB`.
* there is a risk that this could affect helpers built around testify functions, however those wrappers should also call `Helper()` and if they didn't we can safely consider this is a good thing if they have to be fixed when the user will upgrade `testify`.
* some mocks of `assert.Testing.T` in user code may not implement `Helper()`. I think this is reasonable to force them to be fixed.
* this is quite similar as when interface `testing.TB` has been extended to include `Helper()`
 
Steps required to accomplish this:
1. Fix our internal mocks to implement `Helper()` (not a breaking change)
2. Implement `Helper()` on type [`assert.CollectT`](https://pkg.go.dev/github.com/stretchr/testify/assert/#CollectT) (not a breaking change)
3. Extend `assert.TestingT`, `require.TestingT`, `mock.TestingT` to add `Helper()` (limited breaking change)
4. Replace the calls to `Helper()`

Cc: @fefe982 @boyan-soubachov @brackendawson

dolmen:
[...]

PR #1423, opened 2023-07-13
Add helper method in tests

dolmen:
## Summary
Start to extend our definition of `testing.TB` to include its `Helper()` method. See #1422 for details.

## Changes
* Add an `Helper()` method on all internal mocks of `testing.T`.
* Add an `Helper()` method on type `assert.CollectT`.

## Motivation

These are steps 1 and 2 of #1422.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1864,6 +1864,9 @@ type CollectT struct {
 	errors []error
 }
 
+// Helper is like [testing.T.Helper] but does nothing.
+func (CollectT) Helper() {}
+
 // Errorf collects the error.
 func (c *CollectT) Errorf(format string, args ...interface{}) {
 	c.errors = append(c.errors, fmt.Errorf(format, args...))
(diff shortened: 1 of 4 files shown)


assert/assertions.go
```diff
@@ -1874,6 +1874,9 @@ type CollectT struct {
 	errors []error
 }
 
+// Helper is like [testing.T.Helper] but does nothing.
+func (CollectT) Helper() {}
```
craig65535, line 1878: Should the newly merged `captureTestingT` also get a `Helper()` method? (Or, should `captureTestingT` be replaced with `CollectT`?)

craig65535 (approved):

brackendawson (dismissed):
Given this doesn't change the interface I don't want to block progress, so approving.

But I think @craig65535 is right, ideally re-base this and add consolidate the new internal type using one of those approaches.

PR #1425, opened 2023-07-13
Fix calling Unset on call with argumentMatcher.

jslang:
## Summary

Fix panic when calling Unset on call with argumentMatcher argument.

## Changes
When calling Unset on an call that contains an argumentMatcher as an expected value, the call fails with panic, fails the test and any remaining Expectations remain in place causing the test to fail with messaging about failed expectations. To work around this, during Diff check, compare the matcher to the incoming actualMatcher if it is an argumentMatcher instead of attempting to match it as if it was a received argument.

## Motivation
This bug causes tests that rely on `mock.MatchedBy` or anything else that returns an `argumentMatcher` to fail if they use `Unset` to remove expected calls.

See test case added for example panic.

## Related issues
- N/A


dolmen:
Please add to the `Example usage` section a simple reproducer that can run on `go.dev/play` (you'll find examples on recent opened issues).

jslang:
> Please add to the Example usage section a simple reproducer

Done.

jslang:
Anything I can do to help get this fix in? It keeps popping up for us.

skantay (commented):
lgtm

canni:
Hey team, is there any thing we can do to unblock merging this change?

brackendawson (changes requested):
This has the interesting side effect of making the argument matcher itself an acceptable argument for the call as I have [mentioned in similar issues](https://github.com/stretchr/testify/issues/1599#issuecomment-2395131710). But this is quite an unrealistic problem.

<details>

<summary>click to expand example</summary>

```
package kata_test

import (
	"testing"

)

type My struct {
	mock.Mock
}

func (m *My) Do(i any) {
	m.Called(i)
}

func TestIfy(t *testing.T) {
	m := &My{}
	m.Test(t)
	defer m.AssertExpectations(t)

	matcher := mock.MatchedBy(func(i any) bool {
		i, ok := i.(bool)
		if !ok {
			return false
		}
		return i == 1
	})

	m.On("Do", matcher).Return().Once()

	m.Do(matcher)
}
```

</details>

The original bug still exists for `mock.AnythingOfType` and `mock.IsType` matchers. That should not be a blocker either, just an observation.

However a more serious problem is that this doesn't just unset the Call it was called upon, but all matching Calls, and any calls that match due to using mock.Anything:

```go
package kata_test

import (
	"testing"

)

type My struct {
	mock.Mock
}

func (m *My) Do(i int) {
	m.Called(i)
}

func TestDuplicate(t *testing.T) {
	m := &My{}
	m.Test(t)
	defer m.AssertExpectations(t)

	m.On("Do", mock.MatchedBy(func(i int) bool { return i == 2 })).Return().Once()
	call := m.On("Do", mock.MatchedBy(func(i int) bool { return i == 2 })).Return().Once()

	call.Unset()

	m.Do(1)
}

func TestAnything(t *testing.T) {
	m := &My{}
	m.Test(t)
	defer m.AssertExpectations(t)

	m.On("Do", mock.Anything).Return().Once()
	call := m.On("Do", mock.MatchedBy(func(i int) bool { return i == 2 })).Return().Once()

	call.Unset()

	m.Do(1)
}
```

```
--- FAIL: TestDuplicate (0.00s)
    /Users/bracken/src/githib.com/brackendawson/kata/mock.go:337: 
        
        mock: Unexpected Method Call
        -----------------------------
        
        Do(int)
        		0: 1
        
        The closest call I have is: 
        
        Do(mock.argumentMatcher)
        		0: mock.argumentMatcher{fn:reflect.Value{typ_:(*abi.Type)(0x10472d500), ptr:(unsafe.Pointer)(0x1047810a0), flag:0x13}}
        
        
        Diff: 0: FAIL:  (int=1) not matched by func(int) bool
    /Users/bracken/src/githib.com/brackendawson/kata/panic.go:636: FAIL:	Do(mock.argumentMatcher)
        		at: [/Users/bracken/src/githib.com/brackendawson/kata/kata_test.go:22]
    /Users/bracken/src/githib.com/brackendawson/kata/panic.go:636: FAIL: 0 out of 1 expectation(s) were met.
        	at: [/usr/local/go/src/runtime/panic.go:636 /usr/local/go/src/testing/testing.go:1041 /Users/bracken/src/githib.com/brackendawson/kata/kata_test.go:14 /Users/bracken/src/githib.com/brackendawson/kata/kata_test.go:27]
--- FAIL: TestAnything (0.00s)
    /Users/bracken/src/githib.com/brackendawson/kata/mock.go:337: 
        assert: mock: I don't know what to return because the method call was unexpected.
        	Either do Mock.On("Do").Return(...) first, or remove the Do() call.
        	This method was unexpected:
        		Do(int)
        		0: 1
        	at: [/Users/bracken/src/githib.com/brackendawson/kata/kata_test.go:14 /Users/bracken/src/githib.com/brackendawson/kata/kata_test.go:40]
FAIL
coverage: [no statements]
FAIL	github.com/brackendawson/kata	0.249s
FAIL
```

It must only unset the exact call on which it was called.

brackendawson:
On further consideration this might not be a regression, see #1707. I still think Unset should not be re-using Diff and I have proposed an alternative solution in that issue which might more cleanly solve the original issue here.

brackendawson (changes requested):
Disregard my previous review, it's based on the incorrect assumption that Unset *"must only unset the exact call on which it was called"*, this is not the case as I've [documented in #1621](https://github.com/stretchr/testify/issues/1621#issuecomment-2731043626).

It's not an elegant design at all, but the intention of Unset is that you can create a new Call with args that shouldn't be accepted any more. Being more specific for this issue, this test should pass:

```go
package kata_test

import (
	"testing"

)

type My struct {
	mock.Mock
}

func (m *My) Do(i int) {
	m.Called(i)
}

func TestIfy(t *testing.T) {
	m := &My{}
	m.Test(t)

	m.On("Do", mock.MatchedBy(func(i int) bool { return i == 1 })).Return().Once()
	m.On("Do", mock.MatchedBy(func(i int) bool { return i == 1 })).Return().Once().Unset()

	m.AssertExpectations(t)
}
```

Issue #1424, opened 2023-07-13
Succeed quickly in `Eventually`, `EventuallyWithT`


cszczepaniak:
`assert.Eventually` does not check initially whether the condition is satisfied. This means it must [wait a minimum of the tick duration](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1850) before checking the condition for the first time. It would be a nice optimization to early-out if the condition is met already when the function is called.

dolmen:
I think this is a good behavior change.

We must involve people who designed it.

cszczepaniak:
This would be addressed by the proposal in https://github.com/stretchr/testify/issues/1439 if that were to move forward.

pgimalac:
👋 Hey @dolmen @cszczepaniak, any update on this ? 
Would you consider a PR fixing this specifically ? (without necessarily going with the bigger https://github.com/stretchr/testify/issues/1439)

I work on a project where `Eventually` is used pretty extensively in e2e tests where the tick is often in the order of several seconds, so this would be a great improvement duration-wise !

pgimalac:
Works for me, as long as this is fixed !
I didn't notice you already had a PR opened by the way 😄

cszczepaniak:
@dolmen @pgimalac https://github.com/stretchr/testify/pull/1427 is now rebased and cleaned up. Ready for review again!

dolmen:
@cszczepaniak I just merged #1481 which uses `runtime.Goexit`. The single goroutine approach seems incompatible.

cszczepaniak:
@dolmen commented on the PR as well, but checking the condition first doesn't require the single goroutine change and still seems worth pursuing.

ccoVeille:
Isn't this one fixed by #1427 ?

- #1427

PR #1427, opened 2023-07-15
assert: check early in Eventually, EventuallyWithT, and Never

cszczepaniak:
## Summary
Addresses #1424.

`Eventually` and `EventuallyWithT` current must always wait at least the polling duration before they can succeed. This PR starts checking the condition immediately. The assertion still fails if the initial check of the condition takes longer than the configured timeout.

## Motivation
This will provide a small optimization to callers, because tests can complete more quickly than they did before if conditions are met already.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1930,11 +1930,8 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 		h.Helper()
 	}
 
-	if condition() {
-		return true
-	}
-
 	ch := make(chan bool, 1)
+	checkCond := func() { ch <- condition() }
 
 	timer := time.NewTimer(waitFor)
 	defer timer.Stop()
@@ -1942,18 +1939,23 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 	ticker := time.NewTicker(tick)
 	defer ticker.Stop()
 
-	for tick := ticker.C; ; {
+	var tickC <-chan time.Time
+
+	// Check the condition once first on the initial call.
+	go checkCond()
+
+	for {
 		select {
 		case <-timer.C:
 			return Fail(t, "Condition never satisfied", msgAndArgs...)
-		case <-tick:
-			tick = nil
-			go func() { ch <- condition() }()
+		case <-tickC:
+			tickC = nil
+			go checkCond()
 		case v := <-ch:
 			if v {
 				return true
 			}
-			tick = ticker.C
+			tickC = ticker.C
 		}
 	}
 }
@@ -2023,35 +2025,42 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 	var lastFinishedTickErrs []error
 	ch := make(chan *CollectT, 1)
 
+	checkCond := func() {
+		collect := new(CollectT)
+		defer func() {
+			ch <- collect
+		}()
+		condition(collect)
+	}
+
 	timer := time.NewTimer(waitFor)
 	defer timer.Stop()
 
 	ticker := time.NewTicker(tick)
 	defer ticker.Stop()
 
-	for tick := ticker.C; ; {
+	var tickC <-chan time.Time
+
+	// Check the condition once first on the initial call.
+	go checkCond()
+
+	for {
 		select {
 		case <-timer.C:
 			for _, err := range lastFinishedTickErrs {
 				t.Errorf("%v", err)
 			}
 			return Fail(t, "Condition never satisfied", msgAndArgs...)
-		case <-tick:
-			tick = nil
-			go func() {
-				collect := new(CollectT)
-				defer func() {
-					ch <- collect
-				}()
-				condition(collect)
-			}()
+		case <-tickC:
+			tickC = nil
+			go checkCond()
 		case collect := <-ch:
 			if !collect.failed() {
 				return true
 			}
 			// Keep the errors from the last ended condition, so that they can be copied to t if timeout is reached.
 			lastFinishedTickErrs = collect.errors
-			tick = ticker.C
+			tickC = ticker.C
 		}
 	}
 }
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -2075,6 +2075,7 @@ func Never(t TestingT, condition func() bool, waitFor time.Duration, tick time.D
 	}
 
 	ch := make(chan bool, 1)
+	checkCond := func() { ch <- condition() }
 
 	timer := time.NewTimer(waitFor)
 	defer timer.Stop()
@@ -2082,18 +2083,23 @@ func Never(t TestingT, condition func() bool, waitFor time.Duration, tick time.D
 	ticker := time.NewTicker(tick)
 	defer ticker.Stop()
 
-	for tick := ticker.C; ; {
+	var tickC <-chan time.Time
+
+	// Check the condition once first on the initial call.
+	go checkCond()
+
+	for {
 		select {
 		case <-timer.C:
 			return true
-		case <-tick:
-			tick = nil
-			go func() { ch <- condition() }()
+		case <-tickC:
+			tickC = nil
+			go checkCond()
 		case v := <-ch:
 			if v {
 				return Fail(t, "Condition satisfied", msgAndArgs...)
 			}
-			tick = ticker.C
+			tickC = ticker.C
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


dolmen:
This PR changes more than what it claims.

Please keep it focused on #1424 and move the changes related to the tick separately. Because it seems that runs of the condition will not overlap anymore.

connorszczepaniak-wk:
@dolmen What extra changes are you referring to? The changes I made here were all to facilitate the addressing of #1424. How would you do it differently?

cszczepaniak:
@dolmen What more does it change? The changes I made here are to facilitate addressing #1424. How would you do it differently?

assert/assertions.go
```diff
@@ -1912,33 +1920,43 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
...
 			}
-			tick = ticker.C
 		}
+	}()
+
+	select {
+	case <-ch:
+		return true
+	case <-ctx.Done():
+		wg.Wait()
```
cszczepaniak, line 1957: This highlights a race condition that currently exists in `EventuallyWithT`: if the timer expires, we copy the CollectT's errors to the TestingT. This could happen before the goroutine executing the condition finishes, and we would not get the last tick's errors.

I think we should address this bug (#1437) prior to this PR moving forward.

cszczepaniak:
#1439 could also address this issue. @dolmen what are your thoughts on this PR's approach vs. addressing #1439 instead?

dolmen:
I intend to merge #1395 first. Could you help with your own review now, and we will work on your own changes after the merge?

cszczepaniak:
@dolmen After #1395, I think we should pursue either the approach in this PR or the approach outlined in #1439, but not both. I will wait until we have the input needed to decide on #1439

assert/assertions_test.go
```diff
@@ -2816,6 +2816,42 @@ func TestEventuallyIssue805(t *testing.T) {
...
+	go func() {
+		defer close(done)
+		True(t, EventuallyWithT(mockT, condition, 1000*time.Millisecond, 100*time.Millisecond))
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(10 * time.Millisecond):
+		Fail(t, `condition not satisfied quickly enough`)
+	}
```
pgimalac, line 2852: I think you could simplify the test by just checking with `tick` longer than `waitFor`:
```suggestion
	condition := func(t *CollectT) {}

	True(t, EventuallyWithT(mockT, condition, 100*time.Millisecond, time.Second))
```

assert/assertions_test.go
```diff
@@ -2816,6 +2816,42 @@ func TestEventuallyIssue805(t *testing.T) {
...
+	go func() {
+		defer close(done)
+		True(t, Eventually(mockT, condition, 1000*time.Millisecond, 100*time.Millisecond))
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(10 * time.Millisecond):
+		Fail(t, `condition not satisfied quickly enough`)
+	}
```
pgimalac, line 2834: Similarly:
```suggestion
	condition := func() bool { return true }

	True(t, Eventually(mockT, condition, 100*time.Millisecond, 1*time.Second))
```

assert/assertions.go
pgimalac, line 1: Should the `Never` function be updated similarly to stay coherent ?
It currently has the same issue as `Eventually` in that it doesn't run the condition before one tick

assert/assertions_test.go
cszczepaniak: Great idea!

assert/assertions_test.go
```diff
@@ -2970,6 +2970,49 @@ func TestEventuallyWithT_ReturnsTheLatestFinishedConditionErrors(t *testing.T) {
 	Len(t, mockT.errors, 2)
 }
 
+// Check that a long running condition doesn't block Eventually.
+// See issue 805 (and its long tail of following issues)
+func TestEventuallyTimeout(t *testing.T) {
```
cszczepaniak, line 2975: This test isn't new; I just moved it so that all of the `Eventually*` tests are grouped together rather than having the `Never*` tests in between them

dolmen:
@cszczepaniak I just merged #1481 which uses `runtime.Goexit`. The single goroutine approach is incompatible.

cszczepaniak:
@dolmen sounds good. I don't think that affects this PR, though.

cszczepaniak:
@dolmen Thanks, I've rebased now.

pgimalac:
👋 Hey @dolmen @cszczepaniak, any update on this PR ? 
I think it's a really nice improvement of `Eventually` so it would be nice if it was merged

cszczepaniak:
@pgimalac I'd like it to merge. Just waiting on final review from @dolmen

pgimalac:
👋 Hey @dolmen, any chance you could give this PR another review ?

brackendawson (approved):
I think the code in this PR is perfectly fine and implements the feature cleanly. I especially like that it removes the rather implicit assignment of `ticker.C` to `tick` in the `for` statement.

@dolmen's concern that this removes any overlapping runs of the condition isn't an issue. The condition already cannot overlap because of `tick` being set to `null` before calling the condition function.

Because of the way tickers work this cannot alter the timing of when the last condition function call happens so long as the condition function function always has a shorter duration than tick. If the condition function can take longer than tick then this may alter the timing. In all cases where the duration of the condition function is unrelated to when it is called this change can add at most 1 additional call of the condition function. It is possible that poorly written code with race conditions may now panic, deadlock, or any number of other things that make tests fail.

It's unlikely that even very fast asynchronous code will satisfy the new first call to the condition function if Eventually is called soon after the asynchronous code starts. But this means it's more likely that more condition functions will be exercised in their failing state. I don't think this will speed up many tests at all.

I think these factors all stand to improve the condition of code tested by Testify, but I'm sure there will be tests which previously passed which now fail in poorly written code-bases. We will need to be ready to field support requests.

I'm going to leave my approve but give other maintainers a couple of weeks to consider what I've written here before merging.

pgimalac:
@brackendawson can we consider getting this merged now ?
Thanks !

brackendawson:
Thank you all for your contributions.

pgimalac:
@brackendawson any idea when the next release will happen ?

brackendawson:
We've had a cadence of about 6 months, so it's about the right time now. I'll look at preparing a release and get the other maintainers to review.

pgimalac:
@brackendawson sorry for pinging you again but do you know when the release will happen ?
Thanks !

brackendawson:
I don't mind at all, thank you for the nudge, v1.11.0 includes this change.

assert/assertions.go
```diff
@@ -1931,25 +1931,31 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
...
 	defer timer.Stop()
 
 	ticker := time.NewTicker(tick)
 	defer ticker.Stop()
 
-	for tick := ticker.C; ; {
+	var tickC <-chan time.Time
+
+	// Check the condition once first on the initial call.
+	go checkCond()
```
clementnuss, line 1945: @brackendawson I just got impacted in some of our test cases, because the behaviour changes a little here: the `checkCond` function was only evaluated *after* the first tick interval, but that is no longer the case. not sure if that is worth opening an issue though, it also makes sense to check immediately.

(in our case your change pinpointed a race condition in some kubernetes ingress / contour / envoy reconfiguration that we were not aware of 😅 )

OGoodness: This PR is causing issues with our eventually check as well. The full time isn't being respected. We have a tick of 100ms and a time of 1 second, but only 100 ms is being used.  
The same test works on the v1.10.0 major version but fails on the v1.11.0 update, supposedly due to this PR.


Alternatively, this could be an issue with the assert function, so this could be a red herring.

pgimalac: > the behaviour changes a little here: the checkCond function was only evaluated after the first tick interval, but that is no longer the case

That's precisely what the PR implements, so it's very much on purpose, it helps speed up tests in the happy path
I was pushing for this change while expecting it to break some of my tests (which it did), but it only revealed some logic errors or races, so it's beneficial anyway

> The full time isn't being respected. We have a tick of 100ms and a time of 1 second, but only 100 ms is being used.

Do you mean the check function only runs once and not every 100ms ? Is it blocking maybe ?

PR #1430, opened 2023-07-20
assert: IsIncreasing et al can return false w/out failing

### LINKED ISSUE #1419: assert.IsIncreasing can return false without failing the test
[go.dev/play](https://go.dev/play/p/ESqIZMWW_Wa)
```go
package main

import (
	"testing"

)

func Test(t *testing.T) {
	if !assert.IsIncreasing(t, 0) {
		t.Log("assertion is false, test should fail")
	}
}
```

yields
```
=== RUN   Test
    prog_test.go:11: assertion is false, test should fail
--- PASS: Test (0.00s)
PASS
```

#1124 would fix this.

brackendawson:
## Summary
If you passed a non-collection to IsIncreasing or any of its compatriots then the assertion would return false without failing the test.

## Changes
* Make `isOrdered()` return `Fail()` rather than `false` in validation so that the test will fail.
* Test the above.

## Motivation
The following invalid usage would previously pass:
```go
var nonList int
assert.IsIncreasing(t, nonList)
```

## Related issues
Closes #1419

### DIFF
--- a/assert/assertion_order.go
+++ b/assert/assertion_order.go
@@ -9,7 +9,7 @@ import (
 func isOrdered(t TestingT, object interface{}, allowedComparesResults []CompareType, failMessage string, msgAndArgs ...interface{}) bool {
 	objKind := reflect.TypeOf(object).Kind()
 	if objKind != reflect.Slice && objKind != reflect.Array {
-		return false
+		return Fail(t, fmt.Sprintf("object %T is not a collection", object), msgAndArgs...)
 	}
 
 	objValue := reflect.ValueOf(object)
--- a/assert/assertion_order.go
+++ b/assert/assertion_order.go
@@ -9,7 +9,7 @@ import (
 func isOrdered(t TestingT, object interface{}, allowedComparesResults []CompareType, failMessage string, msgAndArgs ...interface{}) bool {
 	objKind := reflect.TypeOf(object).Kind()
 	if objKind != reflect.Slice && objKind != reflect.Array {
-		return Fail(t, fmt.Sprintf("object %T is not a collection", object), msgAndArgs...)
+		return Fail(t, fmt.Sprintf("object %T is not an ordered collection", object), msgAndArgs...)
 	}
 
 	objValue := reflect.ValueOf(object)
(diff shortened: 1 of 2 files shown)


assert/assertion_order.go
```diff
@@ -9,7 +9,7 @@ import (
 func isOrdered(t TestingT, object interface{}, allowedComparesResults []CompareType, failMessage string, msgAndArgs ...interface{}) bool {
 	objKind := reflect.TypeOf(object).Kind()
 	if objKind != reflect.Slice && objKind != reflect.Array {
-		return false
+		return Fail(t, fmt.Sprintf("object %T is not a collection", object), msgAndArgs...)
```
dolmen, line 12: Because maps are also collections:

```suggestion
		return Fail(t, fmt.Sprintf("object %T is not an ordered collection", object), msgAndArgs...)
```

dolmen (changes requested):

assert/assertion_order.go
```diff
@@ -9,7 +9,7 @@ import (
 func isOrdered(t TestingT, object interface{}, allowedComparesResults []CompareType, failMessage string, msgAndArgs ...interface{}) bool {
```
dolmen, line 9: `isOrdered` is now an `Helper`. It must have the helper block before calling Fail.

brackendawson: The call to Helper must be inside `IsDecreasing()` et al, not inside `isOrdered()`, this is not a regression as the functions are not currently helpers. We should have a separate issue to add `Helper()` calls to exported functions which need them.

dolmen: > The call to Helper must be inside IsDecreasing() et al, not inside isOrdered()

`isOrdered` now calls `assert.Fail` so errors will be reported there. But that not what we want. It is now an `Helper` while it wasn't the case before.

> this is not a regression as the functions are not currently helpers.

But you are changing the behaviour.

> We should have a separate issue to add Helper() calls to exported functions which need them.

This is assuming that the Testify project had ressources to do such a task. In the current state I'm the only active maintainer and I'm not allowed to merge my own code. So I can't wait for a community effort to do it.

So I'm asking again to fix this.

brackendawson:
I've moved the branch for this PR to my personal fork and opened #1787

PR #1431, opened 2023-07-20
assert: deprecate and move exported internal comparison functions

### LINKED ISSUE #1180: Deprecate ObjectsAreEqual and ObjectsAreEqualValues from the assert package and move them in a separate package for reuse

brackendawson:
## Summary
These functions should not have been exported, but we must maintain the compatibility promise until v2.

## Changes
* Depreacate all the non-assertion functions exported by assert.
* Move the above functions to a new internal package.

## Motivation
These functions look like assertions but are not, they can never cause any test to fail.

## Related issues
Closes #1180

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -19,8 +19,9 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/pmezard/go-difflib/difflib"
-	"github.com/stretchr/testify/internal/check"
 	"gopkg.in/yaml.v3"
+
+	"github.com/stretchr/testify/internal/check"
 )
 
 //go:generate sh -c "cd ../_codegen && go build && cd - && ../_codegen/_codegen -output-package=assert -template=assertion_format.go.tmpl"
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -56,7 +56,7 @@ type Comparison func() (success bool)
 
 // ObjectsAreEqual determines if two objects are considered equal.
 //
-// Deprecated: This function does no assertion of any kind. Use Equal instead.
+// Deprecated: This function does no assertion of any kind. Use [Equal] instead.
 func ObjectsAreEqual(expected, actual interface{}) bool {
 	return check.ObjectsAreEqual(expected, actual)
 }
@@ -66,7 +66,7 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
 // structures.
 //
 // Deprecated: This function does no assertion of any kind. Use
-// EqualExportedValues instead.
+// [EqualExportedValues] instead.
 func ObjectsExportedFieldsAreEqual(expected, actual interface{}) bool {
 	return check.ObjectsExportedFieldsAreEqual(expected, actual)
 }
@@ -74,7 +74,7 @@ func ObjectsExportedFieldsAreEqual(expected, actual interface{}) bool {
 // ObjectsAreEqualValues gets whether two objects are equal, or if their
 // values are equal.
 //
-// Deprecated: This function does no assertion of any kind. Use EqualValues
+// Deprecated: This function does no assertion of any kind. Use [EqualValues]
 // instead.
 func ObjectsAreEqualValues(expected, actual interface{}) bool {
 	return check.ObjectsAreEqualValues(expected, actual)
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/pmezard/go-difflib/difflib"
+	"github.com/stretchr/testify/internal/check"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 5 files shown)


assert/assertions.go
```diff
@@ -19,6 +19,7 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/pmezard/go-difflib/difflib"
+	"github.com/stretchr/testify/internal/check"
```
dolmen, line 22: Move the import in a separate block below to separate internal dependencies and external ones.

assert/assertions.go
```diff
@@ -54,116 +55,28 @@ type Comparison func() (success bool)
 
 // ObjectsAreEqual determines if two objects are considered equal.
 //
-// This function does no assertion of any kind.
+// Deprecated: This function does no assertion of any kind. Use Equal instead.
```
dolmen, line 58: `ObjectsAreEqual` might have been used not as a mistake (vs assertions) but as a conveniency (to avoid `reflect.Equal` or a dependency on goolgle/go-cmp). (I admit to have done that myself) so the deprecation message should also handle that case.

Also, use [godoc links](https://go.dev/doc/comment#doclinks).

assert/assertions.go
```diff
@@ -54,116 +55,28 @@ type Comparison func() (success bool)
...
 }
 
 // ObjectsExportedFieldsAreEqual determines if the exported (public) fields of two objects are
 // considered equal. This comparison of only exported fields is applied recursively to nested data
 // structures.
 //
-// This function does no assertion of any kind.
+// Deprecated: This function does no assertion of any kind. Use
+// EqualExportedValues instead.
 func ObjectsExportedFieldsAreEqual(expected, actual interface{}) bool {
```
dolmen, line 69: The deprecation message must suggest a full alternative to ease migrating away.

assert/assertions_test.go
```diff
@@ -175,173 +125,6 @@ type S4 struct {
...
-		t.Run("", func(t *testing.T) {
-			output := copyExportedFields(c.input)
-			if !ObjectsAreEqualValues(c.expected, output) {
-				t.Errorf("%#v, %#v should be equal", c.expected, output)
-			}
-		})
-	}
-}
-
 func TestEqualExportedValues(t *testing.T) {
```
dolmen, line 128: This test should also move to `internal/check`.

assert/assertions_test.go
```diff
@@ -2919,7 +2702,7 @@ func TestErrorAs(t *testing.T) {
 
 func TestIsNil(t *testing.T) {
```
dolmen, line 2703: This test should also move to `internal/check`.

dolmen (changes requested):
There are valid use cases for using `ObjectsAreEqual` as shown by the usage in the `mock` package. So the deprecation message should provide alternatives for that use case too.

assert/assertions.go
brackendawson: I think in #1279 we established that testify is not a utility library, I think that includes our docs.

Godoc links added.

assert/assertions.go
brackendawson: The signature shares the same arguments, no?

assert/assertions_test.go
brackendawson: That tests an assertion, are you sure?

assert/assertions_test.go
brackendawson: Turns out this can actually just be deleted.

assert/assertions.go
dolmen: In #1279 you hinted that people should use go-cmp instead. However the testify functions have a particular definition of equality as `nil` values in the *expected* value mean *ignore this field*.

I think that we have the responsibility of helping users to find that path if we are deprecating the functions.

So, do we have a path forward for all use cases? I think we must provide that detailed information in the documentation. If not, we will be get loads of support questions.

assert/assertions_test.go
dolmen: OK, never mind.

dolmen:
@brackendawson I would like to merge #1436 first and wait for the next minor release to merge this.

assert/assertions.go
brackendawson: Making a specific recommendation, such as go-cmp, is something we'd have to maintain though. What if go-cmp declares itself EOL tomorrow? I'd be happy to clarify that one should use EqualExportedValues if you intend to make an assertion and you should use another module if you intend to make a comparison. But I'm not happy to make any specific recommendation.

dolmen: If we have no safe migration to suggest we must not deprecate.

brackendawson (commented):
I think CallerInfo is another function that should receive this treatment.

assert/assertions.go
brackendawson: I think removing functions which make no assertion from Assert, so that people using testify to test don't have hidden bugs is more important than providing a forward path for people using testify as a utility library. The 4th paragraph of [Philosophy](https://github.com/stretchr/testify/blob/master/CONTRIBUTING.md#philosophy) supports this.

assert/assertions.go
dolmen: Testify should not have been used as a utility library, but the fact is it has been used like that, at least in testing frameworks/libraries.

See https://github.com/search?q=assert.ObjectsAreEqual+language%3AGo+NOT+repo%3Astretchr%2Ftestify&type=code

assert/assertions.go
dolmen: Some development teams have policies which disallow using deprecated functions/methods. If we don't provide a path forward some users might be tempted to stop upgrading Testify to avoid the deprecation warning instead of dropping the use of `ObjectsAreEquals`.

I don't think that this change is worth a fight with our users.

brackendawson:
OK, I concede, this cannot progress.

Issue #1404, opened 2023-06-21
EqualExportedValues panic when struct contains array 


olitez:
Hi

```golang
package main

import (
	"github.com/stretchr/testify/assert"
)

type Foo struct {
	Array [2]int
}

func main() {
	a := Foo{[2]int{1, 2}}
	b := Foo{[2]int{2, 1}}
	assert.EqualExportedValues(nil, a, b)
}
```

Will panic with "panic: reflect.MakeSlice of non-slice type"

Reason is there in assertion.go:113 :
```golang
case reflect.Array, reflect.Slice:
	result := reflect.MakeSlice(expectedType, expectedValue.Len(), expectedValue.Len())
```

reflect.MakeSlice only work with slice, not with Array. Fix could be follow:

```golang
case reflect.Array, reflect.Slice:
	result := reflect.New(expectedType).Elem()
	if expectedKind == reflect.Slice {
		result = reflect.MakeSlice(expectedType, expectedValue.Len(), expectedValue.Len())
	}
```

Or even carry reflect.Array to different switch case

dolmen:
Confirmed.
https://go.dev/play/p/0ZWBaOHwqjZ

PR #1436, opened 2023-07-29
assert: handle array case in copyExportedFields #1404

qerdcv:
## Summary
Implemented reflect.Array handling in the copyExportedFields to prevent panic when trying to make a slice from an array

## Changes
Creating new array and filling it with exported values following the example from reflect.Slice.

I've decide to copy paste some code from the reflect.Slice case to make the code more readable, straightforward and avoid mutating the existing case for reflect.Slice

## Motivation
Fixing a bug described in the issue #1404

## Related issues
#1404

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -109,7 +109,19 @@ func copyExportedFields(expected interface{}) interface{} {
 		result.Elem().Set(reflect.ValueOf(unexportedRemoved))
 		return result.Interface()
 
-	case reflect.Array, reflect.Slice:
+	case reflect.Array:
+		result := reflect.New(expectedType).Elem()
+		for i := 0; i < expectedValue.Len(); i++ {
+			index := expectedValue.Index(i)
+			if isNil(index.Interface()) {
+				continue
+			}
+			unexportedRemoved := copyExportedFields(index.Interface())
+			result.Index(i).Set(reflect.ValueOf(unexportedRemoved))
+		}
+		return result.Interface()
+
+	case reflect.Slice:
 		result := reflect.MakeSlice(expectedType, expectedValue.Len(), expectedValue.Len())
 		for i := 0; i < expectedValue.Len(); i++ {
 			index := expectedValue.Index(i)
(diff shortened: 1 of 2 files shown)


qerdcv:
@glesica 
@boyan-soubachov

Сheck it please, when you have time and inspiration

assert/assertions.go
```diff
@@ -109,7 +109,19 @@ func copyExportedFields(expected interface{}) interface{} {
 		result.Elem().Set(reflect.ValueOf(unexportedRemoved))
 		return result.Interface()
 
-	case reflect.Array, reflect.Slice:
+	case reflect.Array:
+		result := reflect.New(reflect.ArrayOf(expectedValue.Len(), expectedType.Elem())).Elem()
+		for i := 0; i < expectedValue.Len(); i++ {
+			index := expectedValue.Index(i)
+			if isNil(index) {
+				continue
```
dolmen, line 117: This case is not covered by the test suite.

dolmen (changes requested):

assert/assertions.go
qerdcv: Thanks!

I added one more test for nil-value, but it failed. I fixed it by passing an actual `interface{}` to the isNil function.

assert/assertions_test.go
```diff
@@ -402,6 +407,72 @@ func TestEqualExportedValues(t *testing.T) {
...
+	            	+   (int) 3
+	            	   },`,
+		},
+		{
+			value1: S7{
+				Nested{[2]interface{}{nil, 2}, [2]int{1, 2}},
+				Nested{[2]int{1, 2}, [2]int{3, 4}},
+			},
+			value2: S7{
+				Nested{[2]interface{}{nil, 2}, [2]int{3, 4}},
```
dolmen, line 460: As `nil` is a special value which means *anything matches*, it would be better to use a concrete value here. Or better: add more test cases to validate that anything matches, not just that `nil` matches *anything*.

dolmen (changes requested):

dolmen:
Issue #1404 has been fixed by #1473.

PR #1438, opened 2023-07-30
Report errors properly in EventuallyWithT

### LINKED ISSUE #1437: EventuallyWithT does not always copy the last errors
The documentation for EventuallyWithT states:

```go
// If the condition is not met before waitFor, the collected errors of
// the last tick are copied to t.
```

However, it is possible to show that this is not the case with the following test:

```go
func TestEventuallyWithTFalse(t *testing.T) {
	mockT := new(CollectT)

	var calledOnce bool
	condition := func(collect *CollectT) {
		if calledOnce {
			time.Sleep(time.Second)
		}
		True(collect, false)
		calledOnce = true
	}

	False(t, EventuallyWithT(mockT, condition, 100*time.Millisecond, 20*time.Millisecond))
	Len(t, mockT.errors, 2)
}
```

This test fails because the errors do not get copied; there is a race condition condition between the condition-checking goroutine and the assertion's goroutine.

cszczepaniak:
## Summary
EventuallyWithT can incorrectly report no errors.

## Changes
Keep track of the last errors in EventuallyWithT to use them for reporting.

## Motivation
Running `TestEventuallyWithTFalse` as changed in this PR against master will fail a shown below. This PR fixes the bug.

```
=== RUN   TestEventuallyWithTFalse
    /testify/assert/assertions_test.go:2776:
        	Error Trace:
        	Error:      	"[
        	            		Error Trace:
        	            		Error:      	Condition never satisfied
        	            	]" should have 2 item(s), but has 1
        	Test:       	TestEventuallyWithTFalse
--- FAIL: TestEventuallyWithTFalse (0.10s)
FAIL
FAIL	github.com/stretchr/testify/assert	0.104s
```

## Related issues
Closes #1437

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1921,13 +1921,22 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 	ticker := time.NewTicker(tick)
 	defer ticker.Stop()
 
+	var lastErrors []error
+
+	copyErrors := func(t TestingT) {
+		for _, err := range lastErrors {
+			t.Errorf("%v", err)
+		}
+	}
+
 	for tick := ticker.C; ; {
 		select {
 		case <-timer.C:
-			collect.Copy(t)
+			copyErrors(t)
 			return Fail(t, "Condition never satisfied", msgAndArgs...)
 		case <-tick:
 			tick = nil
+			lastErrors = collect.errors
 			collect.Reset()
 			go func() {
 				condition(collect)
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1879,6 +1879,7 @@ func (c *CollectT) Reset() {
 	c.errors = nil
 }
 
+// Deprecated: Copy should not be used.
 // Copy copies the collected errors to the supplied t.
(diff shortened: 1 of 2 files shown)


assert/assertions.go
```diff
@@ -1879,6 +1879,7 @@ func (c *CollectT) Reset() {
 	c.errors = nil
 }
 
+// Deprecated: Copy should not be used.
```
cszczepaniak, line 1882: **Overview**
Copy was only called internally in one place, and this PR removes the call. There is no reason for an integrator of testify to call it directly, so we could deprecate and remove it.

If this change is not acceptable, I will revert it.

cszczepaniak:
https://github.com/stretchr/testify/pull/1395 is a duplicate, so I'm closing this. However, that PR might benefit from a test similar to the one in this PR that explicitly exposes the problem.

dolmen:
@cszczepaniak Please post your suggestions on #1395.

assert/assertions.go
dolmen: Deprecating is totally fine. And we must delete the method. But my ultimate aim is to just get rid of `CollectT` that should have been private.

Issue #1432, opened 2023-07-21
Support structs in assert.Contains()


qualidafial:
Often large objects cannot be compared using `assert.Equal()`, because they can contain extra data that cannot be predicted in a test. For example, servers may assign randomly generated UUIDs to created objects, or set `CreatedAt` or `UpdatedAt` fields at the server.

```go
toCreate := Contact{
  Name: "Jenny",
  Numbers []Number{
    {
      Number: "8675309",
      Type: "Mobile",
    },
  },
  Addresses []Address{
    // etc
  }
}

beforeCreate := time.Now()
created, err := client.CreateContact(ctx, toCreate)
afterCreate := time.Now()
require.NoError(t, err)

assert.NotEmpty(t, created.ID)
assert.WithinRange(t, created.CreatedAt, beforeCreate, afterCreate)
assert.WithinRange(t, created.UpdatedAt, beforeCreate, afterCreate)

assert.Equal(t, toCreate.Name, created.Name)
assert.Len(t, created.Numbers, 1)
assert.Equal(t, "8675309", created.Numbers[0].Number)
assert.Equal(t, "Mobile", created.Numbers[0].Type)
// etc
```

When objects contain this extra data, often the test must compare each field one at a time. The `assert.Equal()` calls above become tedious to write when there are a lot of fields. They are also vulnerable to copy-paste errors, at least in my own experience.

It would be helpful (and easier to read) if we could check all the predictable fields in one shot:

```
assert.Contains(t, created, toCreate)
```

The way I envision this working is that when the object is a struct:
* Assert the structs are the same type
* For each field in the expected struct with a non-zero value:
  * Slice fields: assert the slices are the same length, and perform a Contains comparison on corresponding elements.
  * Map fields: for each key in the expected map, assert that the actual map contains that key, and perform a Contains comparison on corresponding values.
  * Struct or struct pointer fields: assert the struct types are the same, and perform a Contains comparison of the corresponding struct fields
  * All other field types (including strings): assert the fields values are equal.

qualidafial:
I would be willing to contribute the work for this, if there is interest from the testify team.

dolmen:
I don't understand how your specification would fit your use case. UUID would be of string type and create/update timestamp would be `time.Time` so they would be compared with strict equality.

brackendawson:
I feel like this is a risky way to sole the problem. The addition of fields to the struct in the future of the package could nullify the assertion. Given you know what field it is going to be in, you should test that field.

[...]

PR #1440, opened 2023-07-30
assert: add `NotZeroThenSetZero` and `NotNilThenSetNil`

HaraldNordgren:
I have used similar versions in my own code for many years. It helps to do assertions of API responses, like this:

```go
clientResponse := someApiRequest(request)

assert.NotZeroThenSetZero(t, &clientResponse.CreateTime)
assert.NotZeroThenSetZero(t, &clientResponse.UpdateTime)

assert.Equal(t, expected, clientResponse)
```

Could resolve the issue brought up in https://github.com/stretchr/testify/issues/1432.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,6 @@ jobs:
     strategy:
       matrix:
         go_version:
-          - "1.17"
           - "1.18"
           - "1.19"
           - "1.20"
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -628,28 +628,12 @@ func NotZeroThenSetZero[X any](t TestingT, x *X) bool {
 		h.Helper()
 	}
 
-	if ok := NotZero(t, *x); !ok {
+	if !NotZero(t, *x) {
 		return false
 	}
 
 	var zeroValue X
 	*x = zeroValue
-	return true
-}
-
-// NotNilThenSetNil asserts that the specified pointer is not nil and then sets it to nil.
-//
-//	assert.NotNilThenSetNil(t, &x)
-func NotNilThenSetNil[X any](t TestingT, x **X) bool {
-	if h, ok := t.(tHelper); ok {
-		h.Helper()
-	}
-
-	if ok := NotNil(t, *x); !ok {
-		return false
-	}
-
-	*x = nil
 	return true
 }
 
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -620,6 +620,39 @@ func NotNil(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 	return Fail(t, "Expected value not to be nil.", msgAndArgs...)
 }
 
+// NotNilThenSetNil asserts that the specified object is not zero and then sets it to zero.
+//
+//	assert.NotZeroThenSetZero(t, &x)
+func NotZeroThenSetZero[X any](t TestingT, x *X) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+
+	if ok := NotZero(t, *x); !ok {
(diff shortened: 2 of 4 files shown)


dolmen:
1. I don't understand the use case. If this API usage is not obvious, then it's place is probably not the core testify.
2. Implantation requires generics. This is a blocking for now as we still aim to support Go versions below 1.18.

.github/workflows/main.yml
```diff
@@ -24,7 +24,6 @@ jobs:
     strategy:
       matrix:
         go_version:
-          - "1.17"
```
dolmen, line 27: Don't change this. This is out of scope.

assert/assertions.go
```diff
@@ -620,6 +620,39 @@ func NotNil(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
...
 
+// NotNilThenSetNil asserts that the specified object is not zero and then sets it to zero.
+//
+//	assert.NotZeroThenSetZero(t, &x)
+func NotZeroThenSetZero[X any](t TestingT, x *X) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+
+	if ok := NotZero(t, *x); !ok {
```
dolmen, line 631: The ok variable can be removed.

assert/assertions.go
```diff
@@ -620,6 +620,39 @@ func NotNil(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
...
+
+	var zeroValue X
+	*x = zeroValue
+	return true
+}
+
+// NotNilThenSetNil asserts that the specified pointer is not nil and then sets it to nil.
+//
+//	assert.NotNilThenSetNil(t, &x)
+func NotNilThenSetNil[X any](t TestingT, x **X) bool {
```
dolmen, line 643: This func is redundant with NotZeroThenSetZero.

1. Why 2 funcs?
2. Why duplicate the implementation instead of wrapping one with the other?

dolmen (changes requested):
I do not plan to merge this.

But anyway I choose to give you feedback that might be valuable for another contribution 

Also you didn't run "go generate ./.. ".

.github/workflows/main.yml
HaraldNordgren: I changed the minimum version to be >=1.18 in discussion with Boyan Soubachov in this PR: https://github.com/stretchr/testify/pull/1379#discussion_r1188456518. Maybe that was a bad change?

HaraldNordgren:
I'm sorry you you feel you don't want to merge this.

> Also you didn't run "go generate ./.. ".

Yes, I noticed that the cannot run now. Seems to be because of generics as well. Should I close the PR?

dolmen:
From your description of the use case:

```go
clientResponse := someApiRequest(request)

assert.NotZeroThenSetZero(t, &clientResponse.CreateTime)
assert.NotZeroThenSetZero(t, &clientResponse.UpdateTime)

assert.Equal(t, expected, clientResponse)
```

This can be rewritten as:
```go
clientResponse := someApiRequest(request)

assert.NotZero(t, clientResponse.CreateTime)
assert.NotZero(t, clientResponse.UpdateTime)
var zeroTime time.Time
clientResponse.CreateTime = zeroTime
clientResponse.UpdateTime = zeroTime

assert.Equal(t, expected, clientResponse)
```

And I think that is would be more clear that the proposed API.

I think that if you want than feature it is best to copy your implementation in the test package where you use it. That way, the source is immediately available to the reader who may not be familiar with the full `testify` API. We don't need more niche features.

dolmen:
About the code generation not working with generics code, this is a problem that we must fix before introducing any generics-based API.

PR #1441, opened 2023-07-31
mock: deprecate AnythingOfTypeArgument (#1434)

### LINKED ISSUE #1434: mock: deprecate AnythingOfTypeArgument
[AnythingOfTypeArgument](https://pkg.go.dev/github.com/stretchr/testify/mock#AnythingOfTypeArgument) should never have been exposed in the API. Users must use `AnythingOfType` instead.

We must tag this type with `// Deprecate`.

Note: I think that we should go further and fix the `AnythingWithType` func signature, but we must release that deprecation first.

dolmen:
## Summary
Deprecate [`mock.AnythingOfTypeArgument`](https://pkg.go.dev/github.com/stretchr/testify/mock#AnythingOfTypeArgument) which should never have been publicly exposed.

`AnythingOfTypeArgument` is now a type alias to the private `anythingOfTypeArgument` and is marked as Deprecated in godoc. The [`AnythingOfType`](https://pkg.go.dev/github.com/stretchr/testify/mock#AnythingOfType) constructor is still available.

The aim is to completely remove that alias.

## Changes
* Rename `AnythingOfTypeArgument` to `anythingOfTypeArgument`.
* Add back `AnythingOfTypeArgument` as a type alias of `anythingOfTypeArgument`.
* Mark `AnythingOfTypeArgument` as Deprecated in godoc.
* Documentation is improved to hide implementation details.

## Motivation
API cleanup.

## Related issues
Closes #1434

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -760,18 +760,26 @@ const (
 	Anything = "mock.Anything"
 )
 
-// AnythingOfTypeArgument is a string that contains the type of an argument
+// AnythingOfTypeArgument contains the type of an argument
 // for use when type checking.  Used in Diff and Assert.
-type AnythingOfTypeArgument string
+//
+// Deprecated: this is an implementation detail that must not be used. Use [AnythingOfType] instead.
+type AnythingOfTypeArgument = anythingOfTypeArgument
 
-// AnythingOfType returns an AnythingOfTypeArgument object containing the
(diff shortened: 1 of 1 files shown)


brackendawson (approved):
This is another detail that has always bugged me, I'm in favuor of assisting consumers by marking it as deprecated. We would have to wait for a major version bump to remove it.

Issue #1139, opened 2021-12-28
suite.Run() triggers data race errors with parallel tests


bstpierre:
Running suite.Run() when test cases have called `suite.T().Parallel()` triggers a data race warning when using `go test -race`.

==> go.mod <==
```
module testify-race

go 1.17

require (
        github.com/davecgh/go-spew v1.1.0 // indirect
        github.com/pmezard/go-difflib v1.0.0 // indirect
        github.com/stretchr/objx v0.1.0 // indirect
        github.com/stretchr/testify v1.7.0 // indirect
        gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c // indirect
)
```

==> main.go <==
```
package main

import "fmt"

func add(x, y int) int {
        return x + y
}

func main() {
        fmt.Println("hello")
}
```

==> main_test.go <==
```
package main

import (
        "testing"

        "github.com/stretchr/testify/suite"
)

type MainSuite struct {
        suite.Suite
}

func (s *MainSuite) TestA() {
        s.T().Parallel()
        for x := 0; x < 100; x++ {
                s.Equal(x, add(x, 0))
        }
}

func (s *MainSuite) TestB() {
        s.T().Parallel()
        for x := 0; x < 100; x++ {
                s.Equal(x+1, add(x, 1))
        }
}

func Test_Main(t *testing.T) {
        suite.Run(t, new(MainSuite))
}
```

Running the test with the race detector:

```
$ go test -race .
==================                                                                                              
WARNING: DATA RACE        
Read at 0x00c0000b6138 by goroutine 9:            
  testify-race.(*MainSuite).TestB()
      /tmp/testify-bugreport/main_test.go:23 +0x73                                                              
  runtime.call16()                                                                                              
      /home/brian/sdk/go1.17.2/src/runtime/asm_amd64.s:625 +0x48                                     
  reflect.Value.Call()             
      /home/brian/sdk/go1.17.2/src/reflect/value.go:339 +0xd7
  github.com/stretchr/testify/suite.Run.func1()
      /home/brian/go/pkg/mod/github.com/stretchr/testify@v1.7.0/suite/suite.go:158 +0x71c
  testing.tRunner()   
      /home/brian/sdk/go1.17.2/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()                  
      /home/brian/sdk/go1.17.2/src/testing/testing.go:1306 +0x47                                                                                                                                                                 

[...]

PR #1447, opened 2023-08-01
Use WaitGroup to ensure suite.Run() doesn't exit until all tests are done

benvernier-sc:
## Summary
Add a `WaitGroup` to `suite`'s `Run` function to better support parallel tests 

## Changes
- Added a `WaitGroup` in `suite`'s `Run` function
- Add to the `WaitGroup` for each test in the suite
- Use `defer wg.Done()` as the very last defer of each test to appropriately mark it as done
- Add a call to `wg.Wait()` before returning from `Run` to ensure that `Run`'s `defer` statement are only executed after all tests have completed 

## Motivation
When using a suite with multiple tests that are defined as parallel tests (`t.Parallel()` called), the suite's stats reporting is inaccurate because the `WithStats` method is called from a `defer` in the `Run` method, which exits before all tests have completed.

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -216,7 +216,7 @@ func Run(t *testing.T, suite TestingSuite) {
 		}()
 	}
 
-	go runTests(t, tests)
+	runTests(t, tests)
 
 	// Wait until all tests are finished before returning, so that the stats handling in the `defer` statement above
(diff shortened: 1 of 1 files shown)


dolmen:
Related to #1139.

PR #1448, opened 2023-08-02
Print caller info on unexpected method call

Lochlanna:
## Summary
Print caller info when an unexpected method call is made on a mock and a "closest call" can be found

## Motivation
Both other call failures print this info so this change makes it consistent across the three failures.
The caller info is useful information when using the mock to develop tests.

## Related issues
None that I could find


Lochlanna:
@CodiumAI-Agent /review

reza-ryte-club (approved):

reza-ryte-club (approved):

reza-ryte-club (commented):
I found that printing callerInfo would be a neat thing to do.  LGTM.

dolmen:
Please provide a test running on `go.dev/play` for which the output would be improved by the proposed change.

Here is a template: https://go.dev/play/p/yobgjBYsyWz

Lochlanna:
> 

Here is an example which shows a test that would benefit from the extra information. In this test it doesn't tell you which of the calls to DoSomething failed. With the fix in this MR the location of the call that wasn't matched is printed out so the developer immediately knows which call is not being called with the correct input/what is missing in the mocks.

https://go.dev/play/p/hZulK4Fc_1_I

dolmen:
Reproducer, simplified: https://go.dev/play/p/LqwLNJpzo9e

The information is in the stacktrace, but having it directly in the message would be helpful.

mock/mock.go
```diff
@@ -491,11 +491,12 @@ func (m *Mock) MethodCalled(methodName string, arguments ...interface{}) Argumen
 		m.mutex.Unlock()
 
 		if closestCall != nil {
-			m.fail("\n\nmock: Unexpected Method Call\n-----------------------------\n\n%s\n\nThe closest call I have is: \n\n%s\n\n%s\nDiff: %s",
+			m.fail("\n\nmock: Unexpected Method Call\n-----------------------------\n\n%s\n\nThe closest call I have is: \n\n%s\n\n%s\nDiff: %s\n\nat: %s",
```
dolmen, line 494: Fix formatting to match the style of the other message in the other branch of the if.

```suggestion
			m.fail("\n\nmock: Unexpected Method Call\n-----------------------------\n\n%s\n\nThe closest call I have is: \n\n%s\n\n%s\n\tat: %s\nDiff: %s",
```

mock/mock.go
```diff
@@ -491,11 +491,12 @@ func (m *Mock) MethodCalled(methodName string, arguments ...interface{}) Argumen
...
 				callString(methodName, arguments, true),
 				callString(methodName, closestCall.Arguments, true),
 				diffArguments(closestCall.Arguments, arguments),
 				strings.TrimSpace(mismatch),
+				assert.CallerInfo(),
```
dolmen, line 499: Move the caller info nearer to the function call, above the diff.

dolmen (changes requested):

mock/mock.go
Lochlanna: I've fixed it up so that it uses `\n\t` but I feel like the caller info should go last as it does for both of the other scenarios. I'm not too fussed either way though if you're keen for diff to be last

Issue #1231, opened 2022-07-17
Misplaced "expected" and "actual" Arguments Inside InEpsilonSlice


vaguecoder:
Both `InEpsilon` and `InEpsilonSlice` has parameter `expected` before `actual` 
```go
func InEpsilon(t TestingT, expected, actual interface{}, epsilon float64, msgAndArgs ...interface{}) bool
func InEpsilonSlice(t TestingT, expected, actual interface{}, epsilon float64, msgAndArgs ...interface{}) bool
```

However, `InEpsilon`'s call inside  `InEpsilonSlice` has the arguments misplaced:
```go
result := InEpsilon(t, actualSlice.Index(i).Interface(), expectedSlice.Index(i).Interface(), epsilon)
```

premkirank:
@vaguecoder can you assign this to me? I would like to work on this if this is still open.

dolmen:
@vaguecoder @premkirank Please review #1483.

PR #1454, opened 2023-08-09
fix(InEpsilonSlice): Include msgAndArgs in InEpsilon for improved error visibility on failed assertion

### LINKED ISSUE #1324: InEpsilonSlice doesn't append custom error message
### Bug summary

When `InEpsilonSlice` takes a custom error message and the assertion fails, the custom message isn't displayed.

The issue is that in `InEpsilonSlice` implementation, `msgAndArgs` isn't passed to `InEpsilon` when iterating over slice elements:

https://github.com/stretchr/testify/blob/0ab3ce1249292a7221058b9e370472bca8f04813/assert/assertions.go#L1366

It should be:
```go
result := InEpsilon(t, actualSlice.Index(i).Interface(), expectedSlice.Index(i).Interface(), epsilon, msgAndArgs...)
```

### To reproduce
Write any test where `InEpsilonSlice` fails and takes a custom message. The message wouldn't be displayed.

VladPetriv:
## Summary

Added the `msgAndArgs` parameter to the `InEpsilon` function for better error reporting in case of failed assertion.

## Related issues

Closes #1324

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1466,7 +1466,7 @@ func InEpsilonSlice(t TestingT, expected, actual interface{}, epsilon float64, m
 	expectedSlice := reflect.ValueOf(expected)
 
 	for i := 0; i < actualSlice.Len(); i++ {
-		result := InEpsilon(t, actualSlice.Index(i).Interface(), expectedSlice.Index(i).Interface(), epsilon)
+		result := InEpsilon(t, actualSlice.Index(i).Interface(), expectedSlice.Index(i).Interface(), epsilon, msgAndArgs...)
 		if !result {
 			return result
 		}


dolmen (changes requested):
We also need a regression test.

I expect that the test will also be a good base for further improvement (in separate PRs) in reporting errors such as the index of the mismatch.

VladPetriv:
Hi @dolmen, thank you for the quick review!

>We also need a regression test.

Can you provide more details about which regression tests you are referring to in the `testify` project? Thank you!

dolmen:
Related: #1231 (also about `InEpsilonSlice`)

dolmen:
This blocked by #1483 that must be merged first.

assert/assertions_test.go
```diff
@@ -1915,6 +1923,17 @@ func TestInEpsilonSlice(t *testing.T) {
 		0.04), "{2.2, 2.0} is not element-wise close to {2.1, 2.1} in epsilon=0.04")
 
 	False(t, InEpsilonSlice(mockT, "", nil, 1), "Expected non numeral slices to fail")
+
+	inEpsilonSliceMock := new(inEpsilonSliceMockT)
```
dolmen, line 1927: Why define a new type instead of using the existing `mockT`?
Instead of checking `inEpsilonSliceMock.msgAndArgs` you could use `mockT.errorString()`.

dolmen (changes requested):

assert/assertions_test.go
VladPetriv: >Why define a new type instead of using the existing mockT?

Didn't know that `mockTestingT` exist) 

Thank you for suggestion.

assert/assertions_test.go
```diff
@@ -1915,6 +1915,17 @@ func TestInEpsilonSlice(t *testing.T) {
 		0.04), "{2.2, 2.0} is not element-wise close to {2.1, 2.1} in epsilon=0.04")
 
 	False(t, InEpsilonSlice(mockT, "", nil, 1), "Expected non numeral slices to fail")
+
+	inEpsilonSliceMockT := new(mockTestingT)
```
dolmen, line 1919: Rename the variable to `mockT` for consistency.

dolmen (changes requested):

justenstall:
Just ran into this bug, can this PR be picked back up since #1483 is merged?

glesica:
The merge conflicts are small, but I'm hesitant to touch them since I'm not in this codebase much any more.

ccoVeille:
I can rebase and fix this.

ccoVeille:
It's done, if someone can review and tell me

PR #1459, opened 2023-08-20
ci: update go versions

stevenh:
## Summary
Update go versions to 1.20 and 1.21.

## Changes
Update go versions to 1.20 and 1.21, which are the supported versions.

## Motivation
This fixes the issue with 1.18 generating files in a different way than later versions causing the CI tests to fail.


stevenh:
No needed already update.

PR #1464, opened 2023-08-31
assert: ObjectsAreEqual: use time.Equal for time.Time types

### LINKED ISSUE #1461: Same time values in same timezone are not equal
https://go.dev/play/p/hHHtLV8pnt0?v=goprev

```go
func TestAAAA(t *testing.T) {
	t1 := time.Now()
	// Create a copy of the same time using the same local timezone.
	t2 := t1.In(time.Local)
	assert.Equal(t, t1, t2)
}
```
```bash
=== RUN   TestAAAA
    prog_test.go:16: 
        	Error Trace:	/tmp/sandbox2123072853/prog_test.go:16
        	Error:      	Not equal: 
        	            	expected: time.Date(2009, time.November, 10, 23, 0, 0, 0, time.Local)
        	            	actual  : time.Date(2009, time.November, 10, 23, 0, 0, 0, time.Local)
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1,2 @@
        	            	-(time.Time) 2009-11-10 23:00:00 +0000 UTC m=+0.000000001
        	            	+(time.Time) 2009-11-10 23:00:00 +0000 UTC
        	            	 
        	Test:       	TestAAAA
```

In addition to this, I think it would be good if same time instants with **different** timezones would be equal. In my use case, it would help if time was compared using [time.Equal](https://pkg.go.dev/time#Time.Equal).

tscales:
## Summary
`reflect.DeepEqual` fails for `time.Time` instances where `time.Equal` would succeed. This is because `reflect.DeepEqual` is checking against some unexported fields, specifically `wall` and `ext`, that change when you add a time zone.  `time.Equal` accounts for this with some bitwise operations as far as I can tell.  

## Changes
Adds a type switch statement to `ObjectsAreEqual` so special cases can be checked. Default is to use `reflect.DeepEqual`

## Motivation
make `assert.Equal`  match `time.Equal` response.

## Related issues
Closes #1461

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -59,20 +59,25 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
 	if expected == nil || actual == nil {
 		return expected == actual
 	}
-
-	exp, ok := expected.([]byte)
-	if !ok {
+	switch exp := expected.(type) {
+	case []byte:
+		act, ok := actual.([]byte)
+		if !ok {
+			return false
+		}
+		if exp == nil || act == nil {
+			return exp == nil && act == nil
+		}
+		return bytes.Equal(exp, act)
+	case time.Time:
+		act, ok := actual.(time.Time)
+		if !ok {
+			return false
+		}
+		return exp.Equal(act)
+	default:
 		return reflect.DeepEqual(expected, actual)
 	}
-
-	act, ok := actual.([]byte)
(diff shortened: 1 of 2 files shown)


brackendawson:
This was reverted by #1537

PR #1468, opened 2023-09-12
mock: make (mock.Call).Times(int) add expectations for each expected call

### LINKED ISSUE #1383: AssertExpectations error message is not ideal
While testing out some additional AssertExpectations calls in some of our unit tests, I encountered some not great error messages. Specifically, https://github.com/stretchr/testify/blob/master/mock/mock.go#L614 seems to be focused on the number of expectations, rather than a mock's Repeatability. 

For reproducing, I took an existing mock that's calling `.Once()`, and I changed it to `.Times(5)`. Since the existing `mock.AssertExpectations(t)` was passing, changing this to expect 5 calls would then cause the test to fail with a message like:
```
FAIL:	UpdateAccount(string,string)
        		at: [run_test.go:72 run_test.go:83]
FAIL: 0 out of 1 expectation(s) were met.
        	at: [mock_client.go:418 testing.go:1150 testing.go:1328 testing.go:1570]
```

IMO the error message should indicate that it got 1 out of the expected 5 calls.

Ullaakut:
## Summary

See this [playground example that highlights the issue](https://go.dev/play/p/qFs9z0T2LLb).

The current behavior of `Times(int)` is that if the expectation of X calls being made is not met, the error message that gets written is `The code you are testing needs to make 1 more call(s).` which is very confusing, because there might be more than 1 call missing.

For example if you expect 50 calls and that only 5 are made, one would expect the message to be `The code you are testing needs to make 45 more call(s).`.

This pull request addresses that, by making the `Times(int)` method append one expectation per call, on the mock.

## Changes

* Makes the (mock.Call).Times() method append expectations for each expected call, on the call's parent.
* Updates unit tests accordingly.

## Motivation

The current mock error output is misleading.

## Related issues

Closes #1383.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -148,6 +148,12 @@ func (c *Call) Times(i int) *Call {
 	c.lock()
 	defer c.unlock()
 	c.Repeatability = i
+
+	// Add expected calls to the parent mock.
+	for j := 1; j < i; j++ {
+		c.Parent.ExpectedCalls = append(c.Parent.ExpectedCalls, c)
+	}
+
 	return c
 }
 
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -857,7 +857,7 @@ func Test_Mock_Return_Twice(t *testing.T) {
 		Return(1, "two", true).
 		Twice()
 
-	require.Equal(t, []*Call{c}, mockedService.ExpectedCalls)
+	require.Equal(t, []*Call{c, c}, mockedService.ExpectedCalls)
 
 	call := mockedService.ExpectedCalls[0]
 
@@ -882,7 +882,7 @@ func Test_Mock_Return_Times(t *testing.T) {
 		Return(1, "two", true).
 		Times(5)
 
-	require.Equal(t, []*Call{c}, mockedService.ExpectedCalls)
+	require.Equal(t, []*Call{c, c, c, c, c}, mockedService.ExpectedCalls)
 
 	call := mockedService.ExpectedCalls[0]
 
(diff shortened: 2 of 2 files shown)


mock/mock.go
```diff
@@ -148,6 +148,12 @@ func (c *Call) Times(i int) *Call {
 	c.lock()
 	defer c.unlock()
 	c.Repeatability = i
+
+	// Add expected calls to the parent mock.
+	for j := 1; j < i; j++ {
+		c.Parent.ExpectedCalls = append(c.Parent.ExpectedCalls, c)
```
brackendawson, line 154: `c` is a `*Call` (pointer to Call), so you actually casue the same call to be added to the Mock's expectations `i` times.

This is working the same way it used to because we are decrementing `c.Repeatabilty` on all the expectations at once, since they are the same expectation. The only difference is that the error now says you need to make `i` more calls, rather than "1" more calls, regarless of how many more you actually need to make.

brackendawson (changes requested):
I do like the idea of adding multiple expectations rather than having to count with `mock.Call.Repeatability`. Though it would make it even harder to fix `.Times(0)`, which is currently broken too. And it would also break `mock.Call.NotBefore`.

So we must go with the alternateive and smaller approach, which is to make the error message aware of `Repeatability` instance variable.

The approach in this PR is ineffective, it just changes the message to a different fixed number.

githoober:
@Ullaakut thanks for the PR. Are you planning on fixing it?

Ullaakut:
> @Ullaakut thanks for the PR. Are you planning on fixing it?

Oh sorry, forgot about this PR 😁 I would love to but I don't have time at the moment. If anyone else can pick it up, please do, otherwise feel free to close it and I'll reopen with an alternate fix if I find the time for it 👍

PR #1471, opened 2023-09-15
suite: fix SetupSubTest and TearDownSubTest execution order

### LINKED ISSUE #1465: Test name refers to parent test name in TearDownSubTest
When having tests with subtests, the calling of the `s.T().Name()` in the `TearDownSubTest` will return the parent test name and not the subtest that has been executed
Having this test executed will print the correct sub-test names:
```go
func (s *TestingSuite) TestWithSubtests() {
  s.Run("my subtest", func() {
    println("my subtest -> " + s.T().Name())
    s.Run("my sub subtest", func() {
      println("my sub subtest -> " + s.T().Name())
    })
  })
}
```
Output:
```
=== RUN   TestingSuite/TestWithSubtests
=== RUN   TestingSuite/TestWithSubtests/my_subtest
my subtest -> TestSuite/TestHazelcastExternalSmartClientLoadBalancer/my_subtest
=== RUN   TestingSuite/TestWithSubtests/my_subtest/my_sub_subtest
my sub subtest -> TestSuite/TestHazelcastExternalSmartClientLoadBalancer/my_subtest/my_sub_subtest
```

And I would expect that in the `TearDownSubTest`, the `s.T().Name()` should be consistent with these names. However, it always refers to the parent test name, i.e., `TestingSuite/TestWithSubtests` for the `TestingSuite/TestWithSubtests/my_subtest`, and `TestingSuite/TestWithSubtests/my_subtest` for the `TestingSuite/TestWithSubtests/my_subtest/my_sub_subtest`.

```go
func (s *TestingSuite) TearDownSubTest() {
  println("TearDownSubTest -> " + s.T().Name())
}
```

Output:
```
TearDownSubTest -> TestSuite/TestHazelcastExternalSmartClientLoadBalancer/my_subtest
TearDownSubTest -> TestSuite/TestHazelcastExternalSmartClientLoadBalancer
```

Environment:
Go 1.19
Testify 1.8.4

linusbarth:
## Summary
The order of execution of SetupSubTest and TearDownSubTest methods was changed so that the testing context inside these methods is correct.

## Changes
* The SetupSubTest was moved to after the SetT that set the child testing context
* The TearDownSubTest was moved to before the deferred SetT that reset the testing context (execution wise)
* The TearDownSubTest and SetT were deferred in one block. This was fine beforehand because a panic in the TearDownSubTest would have had no influence on the "suite.SetT(oldT)". Now since a panic in the TearDownSubTest would lead to omitting the "suite.SetT(oldT)" this defer was split into two separate defers.

## Motivation
There were two problems with the order of execution in the Suite.Run() method:
* One could not access the correct testing context ("s.T()") inside the SetupSubTest and TearDownSubTest methods. If the testing context was used for e.g. assertions of mocks in the TearDownSubTest, the results would not be "attached" to the correct test in the test output.
* The behavior was different to the order of execution for "root" tests regarding the SetupTest and TearDownTest methods (see lines 167-201). This could confuse users of the library.

## Related issues
Closes #1465

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -96,19 +96,23 @@ func failOnPanic(t *testing.T, r interface{}) {
 func (suite *Suite) Run(name string, subtest func()) bool {
 	oldT := suite.T()
 
-	if setupSubTest, ok := suite.s.(SetupSubTest); ok {
-		setupSubTest.SetupSubTest()
-	}
-
-	defer func() {
-		suite.SetT(oldT)
-		if tearDownSubTest, ok := suite.s.(TearDownSubTest); ok {
-			tearDownSubTest.TearDownSubTest()
-		}
-	}()
-
 	return oldT.Run(name, func(t *testing.T) {
 		suite.SetT(t)
+
+		defer func() {
+			suite.SetT(oldT)
+		}()
+
+		if setupSubTest, ok := suite.s.(SetupSubTest); ok {
+			setupSubTest.SetupSubTest()
+		}
+
+		defer func() {
+			if tearDownSubTest, ok := suite.s.(TearDownSubTest); ok {
+				tearDownSubTest.TearDownSubTest()
+			}
+		}()
+
 		subtest()
 	})
 }
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -98,8 +98,9 @@ func (suite *Suite) Run(name string, subtest func()) bool {
 
 	return oldT.Run(name, func(t *testing.T) {
 		suite.SetT(t)
-
 		defer suite.SetT(oldT)
+
+		defer recoverAndFailOnPanic(t)
 
 		if setupSubTest, ok := suite.s.(SetupSubTest); ok {
 			setupSubTest.SetupSubTest()
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -99,19 +99,15 @@ func (suite *Suite) Run(name string, subtest func()) bool {
 	return oldT.Run(name, func(t *testing.T) {
 		suite.SetT(t)
 
-		defer func() {
-			suite.SetT(oldT)
-		}()
+		defer suite.SetT(oldT)
 
 		if setupSubTest, ok := suite.s.(SetupSubTest); ok {
 			setupSubTest.SetupSubTest()
 		}
 
-		defer func() {
-			if tearDownSubTest, ok := suite.s.(TearDownSubTest); ok {
-				tearDownSubTest.TearDownSubTest()
-			}
-		}()
+		if tearDownSubTest, ok := suite.s.(TearDownSubTest); ok {
+			defer tearDownSubTest.TearDownSubTest()
+		}
 
 		subtest()
 	})
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -242,8 +242,8 @@ func (suite *SuiteTester) TestSubtest() {
 	for _, t := range []struct {
 		testName string
 	}{
-		{"first"},
-		{"second"},
+		{"first-subtest"},
+		{"second-subtest"},
 	} {
 		suiteT := suite.T()
 		suite.Run(t.testName, func() {
@@ -259,10 +259,14 @@ func (suite *SuiteTester) TestSubtest() {
 
 func (suite *SuiteTester) TearDownSubTest() {
 	suite.TearDownSubTestRunCount++
+	// We should get the *testing.T for the test that is to be torn down
+	suite.Contains(suite.T().Name(), "subtest")
 }
 
 func (suite *SuiteTester) SetupSubTest() {
 	suite.SetupSubTestRunCount++
+	// We should get the *testing.T for the test that is to be set up
+	suite.Contains(suite.T().Name(), "subtest")
 }
 
 type SuiteSkipTester struct {
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -162,6 +162,9 @@ type SuiteTester struct {
 	SetupSubTestRunCount    int
 	TearDownSubTestRunCount int
 
+	SetupSubTestNames    []string
+	TearDownSubTestNames []string
+
 	SuiteNameBefore []string
 	TestNameBefore  []string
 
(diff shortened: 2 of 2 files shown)


suite/suite.go
```diff
@@ -96,19 +96,23 @@ func failOnPanic(t *testing.T, r interface{}) {
...
 		}
-	}()
 
-	return oldT.Run(name, func(t *testing.T) {
-		suite.SetT(t)
+		defer func() {
+			if tearDownSubTest, ok := suite.s.(TearDownSubTest); ok {
+				tearDownSubTest.TearDownSubTest()
+			}
+		}()
```
dolmen, line 114: Move the if outside of the `defer`:
```go
		if tearDownSubTest, ok := suite.s.(TearDownSubTest); ok {
			defer tearDownSubTest.TearDownSubTest()
		}
```

dolmen (changes requested):

dolmen (changes requested):
We also need test that enforces the order. That test should fail before your fix and succeed after.

suite/suite.go
linusbarth: Done. Also I took the liberty to extract the defer-function in line 102. Please note if I should revert that. I think it should not produce noticeable change.

linusbarth:
> We also need test that enforces the order. That test should fail before your fix and succeed after.

done

dolmen:
@linusbarth Please drop the last commit (merge commit). Use `git rebase` instead.

suite/suite_test.go
```diff
@@ -259,10 +259,14 @@ func (suite *SuiteTester) TestSubtest() {
 
 func (suite *SuiteTester) TearDownSubTest() {
 	suite.TearDownSubTestRunCount++
+	// We should get the *testing.T for the test that is to be torn down
```
dolmen, line 262: Move this comment as 3rd argument of Contains below.

suite/suite_test.go
```diff
@@ -259,10 +259,14 @@ func (suite *SuiteTester) TestSubtest() {
...
+	suite.Contains(suite.T().Name(), "subtest")
 }
 
 func (suite *SuiteTester) SetupSubTest() {
 	suite.SetupSubTestRunCount++
+	// We should get the *testing.T for the test that is to be set up
```
dolmen, line 268: Move this comment as 3rd argument of Contains below.

dolmen (changes requested):

suite/suite_test.go
linusbarth: done

suite/suite_test.go
linusbarth: done

thekidder (approved):

dolmen (dismissed):

suite/suite_test.go
```diff
@@ -259,10 +259,12 @@ func (suite *SuiteTester) TestSubtest() {
 
 func (suite *SuiteTester) TearDownSubTest() {
 	suite.TearDownSubTestRunCount++
+	suite.Contains(suite.T().Name(), "subtest", "We should get the *testing.T for the test that is to be torn down")
```
tomhutch, line 262: Updating the test suite attributes here and performing assertions alongside the others in the main test body will be more consistent with the current testing setup:
https://github.com/stretchr/testify/blob/ac5cd69d4baa639ccd1a95c52d627827e4ed242d/suite/suite_test.go#L154-L172

See this for reference:
https://github.com/stretchr/testify/pull/1393/files#diff-ca4ffdb7050ad570420ac5682fefad8481384709c2884051cd57071283ce9ca4

tomhutch (commented):
Nice, I'd be really happy to see this fix get in.

suite/suite.go
```diff
@@ -96,19 +96,19 @@ func failOnPanic(t *testing.T, r interface{}) {
 func (suite *Suite) Run(name string, subtest func()) bool {
 	oldT := suite.T()
 
-	if setupSubTest, ok := suite.s.(SetupSubTest); ok {
-		setupSubTest.SetupSubTest()
-	}
+	return oldT.Run(name, func(t *testing.T) {
+		suite.SetT(t)
```
xwjdsh, line 100: Should we add `defer receiveAndFailOnPanic(t)` here to ensure that `TearDownSubTest` and `TearDownSuite` will be executed even if a panic occurs in a sub-test?

linusbarth: I assume you mean `recoverAndFailOnPanic(t)`.
Since the `TearDownSubTest` is deferred, it will be called even if a panic occurs in a subtest.
The same for the `TearDownSuite`: It is deferred further up in the call stack.

But your suggestion might still provide advantages: If a `SetupSubTest` or a `TearDownSubTest` panics, the test will fail "on the subtest it belongs to" and not on a parent test. Also the next subtests will still be able to execute (if that is the wanted behavior).
I would just add it for now and if there are any concerns please let me know.

xwjdsh: > The same for the TearDownSuite: It is deferred further up in the call stack.

To be clear, I tested against your branch and if a panic occurs in a subtest, the `TearDownSuite` will not be executed. (no `recoverAndFailOnPanic(t)`), my test code: https://go.dev/play/p/UaGXEUZgEU2


> Since the TearDownSubTest is deferred, it will be called even if a panic occurs in a subtest.

As you said below, if a panic occurs in a subtest, it will also terminate the entire subtest process. I think executing the next subtest should be expected behavior because the suite-level tests have that behavior and there is no reason for the subtest to be inconsistent with it.

In fact, there is [a PR](https://github.com/stretchr/testify/pull/1293) trying to solve this problem. Since you made changes here, it would be great if you could fix it by the way. This PR has a test case, would you mind moving it over for completeness? It may need to be modified to be more complete.

linusbarth: Oh man! You're right! Thanks for your testing code, that helped me reproducing it. You might already know, but I found out that this probably has to do with subtests being run in goroutines which I did not know (see: https://github.com/golang/go/issues/20394).

I agree with the consistency of the panic-terminate-behavior for tests and subtests.

I moved your test code into this PR. Thanks for your help. Would be great if you have a look to check if I missed something.

xwjdsh: LGTM, well done.

xwjdsh (approved):

suite/suite.go
dolmen: @linusbarth @xwjdsh Well done, guys! Great collaboration!

dolmen (approved):

PR #1476, opened 2023-09-27
assert: early return recovering on panic in EventuallyWithT

### LINKED ISSUE #1396: Improve the behavior of `require` assertions when using `assert.EventuallyWithT`
Consider the following use of `EventuallyWithT`.
```go
assert.EventuallyWithT(t, func(c *assert.CollectT) {
        exist, err := myFunc()
        require.NoError(c, err)
        assert.True(c, exist)
}, 1*time.Second, 10*time.Millisecond, "Object was not created")
```

Here the intention is to fail `assert.EventuallyWithT` immediately if the `require.NoError` assertions fails. Instead this code will panic if `myFunc()` returns an error. This is because of how the `FailNow` method is implemented on the `CollectT` object:
https://github.com/stretchr/testify/blob/f97607b89807936ac4ff96748d766cf4b9711f78/assert/assertions.go#L1872-L1875

The panic ensures that the call to the `condition` function "exits" immediately. However it will immediately interrupt test execution.
It feels like the implementation could be improved, so that it immediately fails the `assert.EventuallyWithT` assertion, without crashing the test.

There are 2 possible implementations IMO:

1) change the `FailNow` implementation as follows:
```go
func (c *CollectT) FailNow() {
        // a new channel in CollectT, to indicate that execution of EventuallyWithT should stop immediately
        c.failed <- struct{}
        // terminate the Goroutine evaluating the condition
	runtime.Goexit()
}
```

The relevant code in `EventuallyWithT` could then look like this:
```go
	for tick := ticker.C; ; {
		select {
		case <-timer.C:
			collect.Copy(t)
			return Fail(t, "Condition never satisfied", msgAndArgs...)
		case <-tick:
			tick = nil
			collect.Reset()
			go func() { // this Goroutine will exit if FailNow is called
				condition(collect)
				ch <- len(collect.errors) == 0
			}()
		case v := <-ch:
			if v {
				return true
			}
			tick = ticker.C
		}
		case v := <-collect.failed: // new case
			collect.Copy(t)
			return Fail(t, "Require assertion failed", msgAndArgs...)
		}
	}
```

2) the other option is to use `recover` in `EventuallyWithT` to recover from the panic. I am not going into too many details here, as I think the other solution is better. Panics can be cause by user code in the `condition` function, and we probably don't want to recover from these.

If the new suggested behavior is not acceptable, I'd like to understand the rationale for the current implementation.

cc @tnqn

### LINKED ISSUE #1457: assert: EventuallyWithT must catch panics in condition such as CollectT.FailNow
In [EventuallyWithT](https://pkg.go.dev/github.com/stretchr/testify/assert#EventuallyWithT) the condition function may call [`CollectT.FailNow`](https://pkg.go.dev/github.com/stretchr/testify/assert#CollectT.FailNow) which panics.

The implementation of EventuallyWithT should catch this panic (as well as any other panic in the condition).

pducolin:
## Summary

Catch and recover from panic on `assert.EventuallyWithT`

## Changes
*  Add `failed` channel to handle early return failures
* Trigger `failed` when recovering from panic in `condition` routine
* Add test to assert early fail on calls to `collect.FailNow`, preserving errors trace 

## Motivation

While using `EventuallyWithT` we might want to early return on a never expected condition: in this case, we would call `require` within the `condition` and expect to early return. Currently, calling `require` from a `condition` function panics losing the requirement error trace. Catching and recovering from this `panic` allows for early failing the `EventuallyWithT` assertion and keeping the failed `require` error trace.

An example of such a use case is an assertion on a network request, where we want to early failed on http error, but we want to eventually assert on the body's properties
```golang
import (
	"errors"
	"testing"
	"time"

)

// change shoulMockNetworkError to true to reproduce panic in eventually
var state = 0
const shouldMockNetworkError = false

func mockHttpGet() (ok bool, err error) {
	defer func() {
		state += 1
	}()
	if state > 1 {
		return true, nil
	}
	if shouldMockNetworkError {
		return false, errors.New("network error")
	}
	return false, nil
}

func TestEventuallyHttp(t *testing.T) {
	ret := assert.EventuallyWithT(t, func(c *assert.CollectT) {
		isOK, err := mockHttpGet()
		require.NoError(c, err)
		assert.True(c, isOK)
	}, 5*time.Second, 200*time.Millisecond)
	require.True(t, ret)
}
```

Try it in [Golang Playground](https://goplay.tools/snippet/H7R8wFb98B3) 

## Related issues

Closes #1396
Closes #1457

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1912,6 +1912,7 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 
 	collect := new(CollectT)
 	ch := make(chan bool, 1)
+	failed := make(chan struct{}, 1)
 
 	timer := time.NewTimer(waitFor)
 	defer timer.Stop()
@@ -1928,6 +1929,11 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 			tick = nil
 			collect.Reset()
 			go func() {
+				defer func() {
+					if r := recover(); r != nil {
+						failed <- struct{}{}
+					}
+				}()
 				condition(collect)
 				ch <- len(collect.errors) == 0
 			}()
@@ -1936,6 +1942,9 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 				return true
 			}
 			tick = ticker.C
+		case <-failed:
+			collect.Copy(t)
+			return Fail(t, "Require assertion failed", msgAndArgs...)
 		}
 	}
 }
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -2786,6 +2786,26 @@ func TestEventuallyWithTTrue(t *testing.T) {
 	Len(t, mockT.errors, 0)
 }
 
+func TestEventuallyWithTFailNow(t *testing.T) {
+	mockT := new(CollectT)
+
+	state := 0
+	condition := func(collect *CollectT) {
+		defer func() {
+			state += 1
+		}()
+		if state == 2 {
+			collect.Errorf("early failed")
+			collect.FailNow()
+		}
+		True(collect, state == 2)
+	}
+
+	False(t, EventuallyWithT(mockT, condition, 100*time.Millisecond, 20*time.Millisecond))
+	Len(t, mockT.errors, 2)
+	Equal(t, "early failed", mockT.errors[0].Error())
+}
+
 func TestNeverFalse(t *testing.T) {
 	condition := func() bool {
 		return false


pducolin:
I see this PR conflicts with https://github.com/stretchr/testify/pull/1395/files, I can come with an alternative solution, passing collected errors to the `failed` channel

```golang
func (c *CollectT) FailNow() {
	panic(c.errors)
}
...
case <-tick:
        go func() {
        defer func() {
	        if r := recover(); r != nil {
		        failed <- r
	        }
        }()
        condition(collect)
        ch <- collect.errors
}()

case errs := <-failed:
        for _, err := range errs {
            t.Errorf("%v", err)
        }
	return Fail(t, "Require assertion failed", msgAndArgs...)
}
```

assert/assertions.go
```diff
@@ -1936,6 +1942,9 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 				return true
 			}
 			tick = ticker.C
+		case <-failed:
+			collect.Copy(t)
+			return Fail(t, "Require assertion failed", msgAndArgs...)
```
marshall-lee, line 1947: Hi @pducolin!

Unfortunately it doesn't fix the compatibility with `require` package. The return is too early so I cannot for example write `require.Equal(...)` which fails several but eventually is true.

I proposed an alternative solution https://github.com/stretchr/testify/pull/1481

pducolin: Interesting, I interpreted `require` as assertions that should never happen, like network errors. So I expect `require` to panic and stop the loop.

marshall-lee: I see `assert` and `require` as two different styles of writing tests with `testify`. I've seen projects that forbid a use of `assert` API and `require` is all they use.

As of network errors they also may be temporary. Suppose we dial a `127.0.0.1:30000` but the service is starting in background. So instead of:

```go
	require.EventuallyWithT(t, func(c *assert.CollectT) {
		conn, err := net.Dial("tcp", "127.0.0.1:30000")
		if assert.NoError(err) {
			return
		}
		n, err := conn.Read(buf)
		if assert.NoError(err) {
			return
		}
	}, 100*time.Millisecond, 10*time.Millisecond)
```

I would want just write:

```go
	require.EventuallyWithT(t, func(c *assert.CollectT) {
		conn, err := net.Dial("tcp", address)
		require.NoError(err)
		n, err := conn.Read(buf)
		require.NoError(err)
	}, 100*time.Millisecond, 10*time.Millisecond)
```

pducolin: I tend to use `require` and `assert` within the same context. I use `require` when I want the unexpected condition to stop the test, so here I would expect it to panic as it does normally. If I want a test to fail and retry, I would probably use the longer 
```golang
assert.EventuallyWithT(t, func(c *assert.CollectT) {
  conn, err := net.Dial("tcp", "127.0.0.1:30000")
  if assert.NoError(err) {
	  return
  }
  n, err := conn.Read(buf)
  if assert.NoError(err) {
	  return
  }
}, 100*time.Millisecond, 10*time.Millisecond)
```

pducolin: Closing this PR, thank you for the review !

Issue #1479, opened 2023-10-05
Colorize the diffs produced by assert.Equal


mitioshi:
When running tests in IDE/text editor, it's difficult to analyze the diffs produced by assert.Equal without somehow highlighting the different lines. This could be improved by making diff colorized. That is, make the "expected" lines red and make the "actual" lines green. The result could be similar to what https://github.com/mcmire/super_diff/tree/main provides

Teistify uses `difflib` that is no longer maintained and cannot be extended to modify the diff structure, so it might be better to switch to another library: https://github.com/sergi/go-diff. It provides structured diffs and can be used to colorize the output

PR #1480, opened 2023-10-05
Add the option to colorize diffs produced by assert.Equal

mitioshi:
This PR add an option to colorize diffs produced by assert.Equal

## Changes
In this PR, the library used for diffing was replaced to diffmatchpatch which is actively maintained and enables us to access structured diffs. 
Since testify is frequently used in CI/CD pipelines or piped to non-tty buffers, I made colorized diffs controlled by the TESTIFY_COLORED_DIFF environment variable. This way testify will not break any existing pipelines while letting the user control this behavior(e.g. via setting it in the go test template like this 
image)

## Motivation
When running tests in IDE/text editor, it's difficult to analyze the diffs produced by assert.Equal. This could be improved by making diff colorized.
difflib is no longer maintained and cannot be extended to modify the diff structure, so it might be better to switch to another library: https://github.com/sergi/go-diff

The resulting diff looks like this after the suggested changes
image

The relevant issue: https://github.com/stretchr/testify/issues/1479


dolmen (changes requested):
This MR contains 2 changes:
* switch the diff package
* use the colored diff of the new package.

I would prefer if you could post a separate MR that contains only the library change, and once merged, rebase this one.

assert/assertions.go
```diff
@@ -6,19 +6,20 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/sergi/go-diff/diffmatchpatch"
```
dolmen, line 9: This import should not be in the block with stdlib imports.

dolmen:
Related: #1467.

assert/assertions.go
```diff
@@ -410,9 +411,9 @@ func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{})
 		return Fail(t, fmt.Sprintf("Invalid operation: %#v == %#v (%s)",
 			expected, actual, err), msgAndArgs...)
 	}
-
+	coloredOutput, _ := strconv.Atoi(os.Getenv("TESTIFY_COLORED_DIFF"))
```
peymanmortazavi, line 414: it might be nice to store this value on the suite once after evaluation so we don't have to check on every `Equal` call.

Additionally, how do you feel about using a function to recognize `yes`|no and `true|false` as well.

assert/assertions.go
peymanmortazavi: curious if we have explored `go-cmp` package for diffs.

assert/assertions.go
```diff
@@ -1782,18 +1787,89 @@ func diff(expected interface{}, actual interface{}) string {
...
-		ToFile:   "Actual",
-		ToDate:   "",
-		Context:  1,
-	})
-
-	return "\n\nDiff:\n" + diff
+func prettyDiff(diffs []diffmatchpatch.Diff, useColoredOutput bool) string {
+	var diff strings.Builder
+	if useColoredOutput {
+		diff.WriteString("\\033[31m--- Expected\\033[0m\n\\033[32m+++ Actual\\033[0m\n")
```
peymanmortazavi, line 1828: nit: it may be nice to used constants here for the colors.

assert/assertions.go
dolmen: @peymanmortazavi I haven't.

mitioshi: @peymanmortazavi I considered using `go-cmp` initally. Unfortunately `go-cmp` cannot produce structured diff so we can't customize them. This is why I decided to use diffmatchpatch here

assert/assertions.go
mitioshi: What's the right place for such constants? Should we keep them in the assert package or move it somewhere else? (e.g. to a `colors` package)

hendrywiranto:
Hello @mitioshi any plan to continue your work?

Some suggestion:
Maybe reverse the red and green color since the green one usually means the good one (expected value)

Cheers

mitioshi:
> Hello @mitioshi any plan to continue your work?
> 
> Some suggestion: Maybe reverse the red and green color since the green one usually means the good one (expected value)
> 
> Cheers

Aw, I didn't see the initial comments and thought this PR had been overlooked. 
Yes, I'll look into the suggested changes this week

mitioshi:
@dolmen I submitted a new PR for the library change as you suggested -> https://github.com/stretchr/testify/pull/1546.
Could you please review it, so we can proceed with the current PR ?

PR #1481, opened 2023-10-08
assert: collect.FailNow() should not panic

marshall-lee:
## Summary
`collect.FailNow()` should exit goroutine without a panic to be usable with `require` package.

## Changes
`collect.FailNow()` just does `runtime.Goexit()` instead of `panic()`. For example `FailNow()` from `testing` package [behaves similarly](https://cs.opensource.google/go/go/+/refs/tags/go1.21.2:src/testing/testing.go;l=973).

## Motivation

I just want `require` package to be usable with `EventuallyWithT` e.g. I want this example to pass but it panics:

```go
package main

import (
	"testing"
	"time"

)

func TestRequireEventuallyWithT(t *testing.T) {
	state := 0
	require.EventuallyWithT(t, func(c *assert.CollectT) {
		defer func() { state += 1 }()
		require.True(c, state == 2)
	}, 100*time.Millisecond, 10*time.Millisecond)
}
```

See https://go.dev/play/p/Oqd95IT7qxQ

## Related issues
Fixes https://github.com/stretchr/testify/issues/1396
Fixes https://github.com/stretchr/testify/issues/1457

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1956,6 +1956,9 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 
 // CollectT implements the TestingT interface and collects all errors.
 type CollectT struct {
+	// A slice of errors. Non-nil slice denotes a failure.
+	// If it's non-nil but len(c.errors) == 0, this is also a failure
+	// obtained by direct c.FailNow() call.
 	errors []error
 }
 
@@ -1964,9 +1967,10 @@ func (c *CollectT) Errorf(format string, args ...interface{}) {
 	c.errors = append(c.errors, fmt.Errorf(format, args...))
 }
 
-// FailNow panics.
-func (*CollectT) FailNow() {
-	panic("Assertion failed")
+// FailNow stops execution by calling runtime.Goexit.
+func (c *CollectT) FailNow() {
+	c.fail()
+	runtime.Goexit()
 }
 
 // Deprecated: That was a method for internal usage that should not have been published. Now just panics.
@@ -1977,6 +1981,16 @@ func (*CollectT) Reset() {
 // Deprecated: That was a method for internal usage that should not have been published. Now just panics.
 func (*CollectT) Copy(TestingT) {
 	panic("Copy() is deprecated")
+}
+
+func (c *CollectT) fail() {
+	if !c.failed() {
+		c.errors = []error{} // Make it non-nil to mark a failure.
+	}
+}
+
+func (c *CollectT) failed() bool {
+	return c.errors != nil
 }
 
 // EventuallyWithT asserts that given condition will be met in waitFor time,
@@ -2003,7 +2017,7 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 	}
 
 	var lastFinishedTickErrs []error
-	ch := make(chan []error, 1)
+	ch := make(chan *CollectT, 1)
 
 	timer := time.NewTimer(waitFor)
 	defer timer.Stop()
@@ -2023,16 +2037,16 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 			go func() {
 				collect := new(CollectT)
 				defer func() {
-					ch <- collect.errors
+					ch <- collect
 				}()
 				condition(collect)
 			}()
-		case errs := <-ch:
-			if len(errs) == 0 {
+		case collect := <-ch:
+			if !collect.failed() {
 				return true
 			}
 			// Keep the errors from the last ended condition, so that they can be copied to t if timeout is reached.
-			lastFinishedTickErrs = errs
+			lastFinishedTickErrs = collect.errors
 			tick = ticker.C
 		}
 	}
--- a/require/requirements_test.go
+++ b/require/requirements_test.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/assert"
 )
 
 // AssertionTesterInterface defines an interface to be used for testing assertion methods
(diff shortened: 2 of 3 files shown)


dolmen:
@marshall-lee Your review of #1395 would be welcome.

dolmen:
#1395 is now merged.

@marshall-lee Please rebase.

@czeslavo Please review.

dolmen (dismissed):

czeslavo:
I'm wondering whether we could document this behavior somewhere. It would be cool to have an example of using `EventuallyWithT` with a series of `require` and `assert` checks with an explanation of what happens underneath (i.e. if one of the `require` checks is not satisfied, this single condition run is aborted, but `EventuallyWithT` will retry if the timeout hasn't passed yet, etc.)

dolmen:
@cszczepaniak Could you also help with a review of this MR?

assert/assertions.go
```diff
@@ -1864,17 +1864,20 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
...
+	c.failed = true
 	c.errors = append(c.errors, fmt.Errorf(format, args...))
 }
 
-// FailNow panics.
-func (*CollectT) FailNow() {
-	panic("Assertion failed")
+// FailNow stops the execution by calling runtime.Goexit.
+func (c *CollectT) FailNow() {
+	c.failed = true
```
dolmen, line 1879: Add `if c.errors == nil { c.errors = []errors{} }`. That will allow to make the `failed` boolean equivalent to `c.errors != nil` and allow to avoid adding `failed`. That will make the patch much smaller.

assert/assertions.go
```diff
@@ -1931,16 +1934,16 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
...
 				defer func() {
-					ch <- collect.errors
+					ch <- collect
 				}()
 				condition(collect)
 			}()
-		case errs := <-ch:
-			if len(errs) == 0 {
+		case collect := <-ch:
+			if !collect.failed {
```
dolmen, line 1942: This check can be replaced with `if err == nil` once `Collect.FailNow` always sets a non-nil slice.

dolmen (changes requested):
I think that the `CollectT.failed` field is not necessary to fix the issue.

If we ensure `c.errors` is a non nil slice when `c.FailNow` is called, the patch can be much smaller.

assert/assertions.go
```diff
@@ -1931,16 +1934,16 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 			go func() {
 				collect := new(CollectT)
 				defer func() {
-					ch <- collect.errors
```
cszczepaniak, line 1934: Previously, we would hit this send on the channel because we hit this `defer` after the panic.

Now, with `runtime.Goexit`, this `defer` will not get called at all. Does that cause problems?

marshall-lee: https://pkg.go.dev/runtime#Goexit

> Goexit terminates the goroutine that calls it. No other goroutine is affected. Goexit runs all deferred calls before terminating the goroutine.

Unless the `FailNow` test won't pass

cszczepaniak: Ah, I didn't know that bit about it running the deferred calls :+1:

assert/assertions.go
cszczepaniak: I don't mind there being an explicit flag for `failed` -- it makes it a bit clearer rather than using `nil` as an implicit sentinel value meaning "failed."

require/requirements_test.go
```diff
@@ -681,3 +683,33 @@ func TestErrorAssertionFunc(t *testing.T) {
...
+}
+
+func TestEventuallyWithTTrue(t *testing.T) {
+	mockT := new(MockT)
+
+	state := 0
+	condition := func(collect *assert.CollectT) {
+		defer func() {
+			state += 1
+		}()
```
cszczepaniak, line 707: I think the intent here is to have the delegate called a few times before succeeding. It would be a bit clearer to do the increment outside of a defer to accomplish the same effect
```suggestion
		state += 1
```

marshall-lee: Done

assert/assertions.go
```diff
@@ -1936,7 +1939,7 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
 				condition(collect)
 			}()
 		case errs := <-ch:
-			if len(errs) == 0 {
+			if errs == nil {
```
czeslavo, line 1942: How about adding private getter (`failed() bool`) and setter (`fail()`) to `CollectT`? There we could properly document the difference between `errs == nil` and `len(errs) == 0` and have it encapsulated in this struct's method.

dolmen: Well, we could even call them `Failed() bool` and `Fail()`, as [`testing.TB`](https://pkg.go.dev/testing#TB) has such methods.

But let's play it safe for now about API exposure, and make them private for now as you suggest.

:+1: for adding `failed() bool` and `fail()`.

require/requirements_test.go
```diff
@@ -681,3 +683,31 @@ func TestErrorAssertionFunc(t *testing.T) {
...
+	condition := func(collect *assert.CollectT) {
+		state += 1
+		True(collect, state == 2)
+	}
+
+	EventuallyWithT(mockT, condition, 100*time.Millisecond, 20*time.Millisecond)
+	if mockT.Failed {
+		t.Error("Check should pass")
+	}
+}
```
dolmen, line 713: Add a check of `state` to enforce that the condition is not called after the first success.

require/requirements_test.go
```diff
@@ -681,3 +683,31 @@ func TestErrorAssertionFunc(t *testing.T) {
...
+	EventuallyWithT(mockT, condition, 100*time.Millisecond, 20*time.Millisecond)
+	if !mockT.Failed {
+		t.Error("Check should fail")
+	}
+}
+
+func TestEventuallyWithTTrue(t *testing.T) {
+	mockT := new(MockT)
+
+	state := 0
```
dolmen, line 703: Rename as `counter` for more meaning.

dolmen (changes requested):

assert/assertions.go
dolmen: We will make it clear by adding `fail()` and `failed() bool` methods.

marshall-lee:
@czeslavo @cszczepaniak @dolmen Hey I tried to address all the previous review issues. Please review.

require/requirements_test.go
```diff
@@ -681,3 +683,30 @@ func TestErrorAssertionFunc(t *testing.T) {
...
+	condition := func(collect *assert.CollectT) {
+		defer func() {
+			counter += 1
+		}()
+		True(collect, counter == 1)
+	}
+
+	EventuallyWithT(mockT, condition, 100*time.Millisecond, 20*time.Millisecond)
+	False(t, mockT.Failed, "Check should pass")
+	Equalf(t, 2, counter, "Condition is expected to be called 2 times but got %v", counter)
```
czeslavo, line 711: Ditto.
```suggestion
	Equal(t, 2, counter, "Condition is expected to be called 2 times")
```

marshall-lee:
@czeslavo @cszczepaniak @dolmen Hi! Any plans on merging this?

czeslavo (approved):
LGTM, but I'm just an ordinary contributor here, @dolmen decides 🙂

dolmen (approved):
LGTM.

Rebase needed.

ccoVeille:
It's a good thing this issues had been addressed. Thank you guys

adilansari:
when will this be released?

HaraldNordgren:
@marshall-lee This seems to have broken the tests for 1.18.

PR #1485, opened 2023-10-11
assert: better formatting for Len() error

### LINKED ISSUE #1482: Error message on Len() assumes string
This error message in `assert.Len` assumes that whatever you are asserting a length about can be formatted with the `%s` format string. 

```go
	if l != length {
		return Fail(t, fmt.Sprintf("\"%s\" should have %d item(s), but has %d", object, length, l), msgAndArgs...)
	}
```

If you look at the failure in e.g. https://drone.grafana.net/grafana/dskit/3189/1/4, I believe this is a []uint32, which is being formatted as `"[%!s(uint32=2452015680) %!s(uint32=2476020736)]"`. 

I think the `%v` format string might be a better fit.

kevinburkesegment:
Previously, the use of %s with array objects meant you would get an error like this:

    "[%!s(int=1) %!s(int=2) %!s(int=3)]\" should have 4 item(s), but has 3

Use %v instead, which provides a much nicer error.

    "[1 2 3]" should have 4 item(s), but has 3

Fixes #1482.

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -1705,6 +1705,33 @@ func TestLen(t *testing.T) {
 	for _, c := range cases {
 		False(t, Len(mockT, c.v, c.l), "%#v have %d items", c.v, c.l)
 	}
+
+	formatCases := []struct {
+		in      interface{}
+		wantLen int
+		want    string
+	}{
+		{[]int{1, 2, 3}, 3, "[1 2 3]"},
+		{[...]int{1, 2, 3}, 3, "[1 2 3]"},
+		{"ABC", 3, "ABC"},
+		{map[int]int{1: 2, 2: 4, 3: 6}, 3, "map[1:2 2:4 3:6]"},
+
+		{[]int{}, 0, "[]"},
+		{map[int]int{}, 0, "map[]"},
+
+		{[]int(nil), 0, "[]"},
+		{map[int]int(nil), 0, "map[]"},
+		{(chan int)(nil), 0, "<nil>"},
+	}
+
+	t.Run("Len() error message formatting", func(t *testing.T) {
+		for _, tt := range formatCases {
+			msgMock := new(mockTestingT)
+			Len(msgMock, tt.in, 1234567)
+			want := fmt.Sprintf(`"%s" should have 1234567 item(s), but has %d`, tt.want, tt.wantLen)
+			Contains(t, msgMock.errorString(), want)
+		}
+	})
 }
 
 func TestWithinDuration(t *testing.T) {
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -1661,77 +1661,35 @@ func TestLen(t *testing.T) {
 	ch <- 3
 
 	cases := []struct {
-		v interface{}
-		l int
-	}{
-		{[]int{1, 2, 3}, 3},
-		{[...]int{1, 2, 3}, 3},
-		{"ABC", 3},
-		{map[int]int{1: 2, 2: 4, 3: 6}, 3},
-		{ch, 3},
-
-		{[]int{}, 0},
-		{map[int]int{}, 0},
-		{make(chan int), 0},
-
-		{[]int(nil), 0},
-		{map[int]int(nil), 0},
-		{(chan int)(nil), 0},
-	}
-
-	for _, c := range cases {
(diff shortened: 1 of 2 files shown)


dolmen (changes requested):
The test suite must cover many more types. I think this will show that `%#v` is more appropriate.

We should also handle the case where the len is large (ex: 1000 items). Displaying the full value is just pollution in that case. That could be handled in a separate MR, but at least the test suite must be structured to allow this future change.

kevinburkesegment:
> We should also handle the case where the len is large (ex: 1000 items). Displaying the full value is just pollution in that case.

To be clear I think the current "%s" formatting also does not do a great job in this case

kevinburkesegment:
I added the types from the other len() tests, except for the channel types, which show up as a pointer address, that will change every single time.

dolmen (changes requested):
We don't need 2 loops on the same test cases.
Let's merge the old tests on the `cases` into your loop.

kevinburkesegment:
> Let's merge the old tests on the cases into your loop.

How do you want me to handle the channel tests?

dolmen:
@kevinburkesegment Skip the format test if the `want` string is `""`.

dolmen:
I just noticed we also have #1405 which is an alternative.

kevinburkesegment:
I'll update the patch today

kevinburkesegment:
OK, updated. I also shortened the test cases

kevinburkesegment:
Hi, just following up - let me know if you'd like further changes here.

assert/assertions_test.go
```diff
@@ -1644,49 +1644,34 @@ func TestLen(t *testing.T) {
...
-
-	for _, c := range cases {
-		False(t, Len(mockT, c.v, c.l), "%#v have %d items", c.v, c.l)
+		False(t, Len(mockT, c.v, c.l+1), "%#v have %d items", c.v, c.l)
+		if c.format != "" {
+			msgMock := new(mockTestingT)
+			Len(msgMock, c.v, 1234567)
+			want := fmt.Sprintf(`"%s" should have 1234567 item(s), but has %d`, c.format, c.l)
+			Contains(t, msgMock.errorString(), want)
+		}
```
dolmen, line 1674: I prefer to have the whole expected string in a `want` string and avoid a `want` string built with sprintf. Test code must be boring.

dolmen (changes requested):
Good improvements. We're nearly done.

assert/assertions_test.go
kevinburkesegment: Done

assert/assertions_test.go
```diff
@@ -1644,49 +1644,33 @@ func TestLen(t *testing.T) {
 	ch <- 3
 
 	cases := []struct {
-		v interface{}
-		l int
+		v      interface{}
+		l      int
+		format string
```
dolmen, line 1649: Rename to `expected1234567` (because this isn't a format but the expected message) and add a comment `// message when expecting 1234567 items`.

dolmen (changes requested):

assert/assertions_test.go
kevinburkesegment: Done

k1ng440:
LGTM.

dolmen:
@k1ng440 Use the "Review changes" button to post a review, not a comment. 
image

wwade (approved):
LGTM

k1ng440 (approved):

dolmen (approved):
LGTM.

@kevinburkesegment I'm sorry for the time it took and the multiple exchanges, but we achieved together a short patch with a comprehensive test suite. Thank you!

kevinburkesegment:
No worries! Anything I need to do to get this ready for merge?

kevinburkesegment:
Hi, anything I can help with here?

kevinburkesegment:
Hey, just wondering if there's anything I can do to get this ready for merge.

arjunmahishi:
Hi @kevinburkesegment! 

There seems to be [a few panics](https://github.com/stretchr/testify/actions/runs/6970446567/job/18968782955?pr=1485) the tests. Can you please take a look at that?

kevinburkesegment:
I just rebased against latest master and the tests are passing. I checked the panics and I don't see how they are related to the change that I made.

arjunmahishi:
Cool thanks!
Let me find a maintainer who has access to merge this.

PR #1488, opened 2023-10-16
Deprecate EqualExportedValues

HaraldNordgren:


assert/assertions.go
```diff
@@ -150,6 +150,8 @@ func copyExportedFields(expected interface{}) interface{} {
 // structures.
 //
 // This function does no assertion of any kind.
+//
+// Deprecated: Use EqualExportedValues instead
```
dolmen, line 154: End the sentence.

```suggestion
// Deprecated: Use EqualExportedValues instead.
```

dolmen (changes requested):

assert/assertions.go
HaraldNordgren: Thanks! How does the godoc linking work here?

dolmen (approved):

PR #1491, opened 2023-10-23
assert: Add HTTP builder to combine request x response using builder.

### LINKED ISSUE #598: Assert for HTTP Headers

Matovidlo:
## Summary
Adds new method into assert http_assertions.go. Should be used as generic solution for testing request parameters to response possibilities using 1 method with options.
Previous solutions did not have response header check method. However implementing that it would be necessary to create single test with 3 assert conditions. 
Now it is possible to use single function to test typical HTTP response values(status, body, headers).

## Changes
Add new function `HTTP(...)` that can be customized based on user needs.
Use `go generate ./...` to update generated code.
Add new test `TestHTTPBuilder` without example usage.

## Motivation
Impossible to test HTTP response headers.

`assert.True(mockAssert.HTTP(httpHelloName, "GET", "/", url.Values{"name": []string{"World"}}, WithCode(200), WithResponseHeader(http.Header{"Content-Type": []string{"text/plain; charset=utf-8"}}), WithExpectedBody(*bytes.NewBufferString("Hello, World!"))))`

## Related issues
Closes #598

### DIFF
--- a/assert/http_assertions.go
+++ b/assert/http_assertions.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"reflect"
 	"strings"
 )
 
@@ -162,4 +163,51 @@ func HTTPBodyNotContains(t TestingT, handler http.HandlerFunc, method, url strin
 	}
 
 	return !contains
+}
+
+// HTTP asserts that a specfied handler returns set of expected values given by HttpOptions.
+//
+//	assert.HTTP(t, myHandler, "www.google.com", nil, WithCode(200), WithBody("I'm Feeling Lucky"), WithRequestHeader(http.Header{"a": []string{"b"}}, WithExpectedBody(bytes.NewBuffer("c"))))
+//
+// Returns whether the assertion was successful (true) or not (false).
+func HTTP(t TestingT, handler http.HandlerFunc, method, url string, values url.Values, options ...HttpOption) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+
+	b := &builder{}
+	for _, option := range options {
+		err := option(b)
+		if err != nil {
+			Fail(t, fmt.Sprintf("Failed to build http options, got error: %s", err))
+		}
+	}
+
+	w := httptest.NewRecorder()
+	req, err := http.NewRequest(method, url, b.body)
+	if b.err == nil && err != nil {
+		Fail(t, fmt.Sprintf("Failed to build test request, got error: %s", err))
+	} else if b.err != nil {
+		return err == b.err
+	}
+
+	req.Header = b.requestHeader
+	req.URL.RawQuery = values.Encode()
+	handler(w, req)
+	if w.Code != b.code {
+		Fail(t, fmt.Sprintf("Expected HTTP success status code for %q but received %d", url+"?"+values.Encode(), w.Code))
+		return false
+	}
+
+	if b.responseHeader != nil && !reflect.DeepEqual(w.HeaderMap, b.responseHeader) {
+		Fail(t, fmt.Sprintf("Expected HTTP header to be equal for %q but received %v", url+"?"+values.Encode(), w.HeaderMap))
+		return false
+	}
+
+	contains := strings.Contains(w.Body.String(), b.expectedBody.String())
+	if !contains {
+		Fail(t, fmt.Sprintf("Expected HTTP body to be equal for %q but received %s", url+"?"+values.Encode(), w.Body.String()))
+	}
+
+	return contains
 }
--- a/assert/http_assertions.go
+++ b/assert/http_assertions.go
@@ -175,6 +175,11 @@ func HTTP(t TestingT, handler http.HandlerFunc, method, url string, values url.V
 		h.Helper()
 	}
 
+	if options == nil {
+		Fail(t, fmt.Sprintf("No options selected, no assertions can be executed"))
+		return false
+	}
+
 	b := &builder{}
 	for _, option := range options {
 		err := option(b)
@@ -194,7 +199,7 @@ func HTTP(t TestingT, handler http.HandlerFunc, method, url string, values url.V
 	req.Header = b.requestHeader
 	req.URL.RawQuery = values.Encode()
 	handler(w, req)
-	if w.Code != b.code {
+	if b.code != nil && w.Code != *b.code {
 		Fail(t, fmt.Sprintf("Expected HTTP success status code for %q but received %d", url+"?"+values.Encode(), w.Code))
 		return false
 	}
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -264,6 +264,18 @@ func GreaterOrEqualf(t TestingT, e1 interface{}, e2 interface{}, msg string, arg
 	return GreaterOrEqual(t, e1, e2, append([]interface{}{msg}, args...)...)
 }
 
+// HTTPf asserts that a specfied handler returns set of expected values given by HttpOptions.
+//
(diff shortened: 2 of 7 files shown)


arjunmahishi:
Honestly, I am not a fan of the `assert` package calling the handler function as a part of an assert function/method. IMO, this package should only match values. I would write HTTP assertions like this:

```go
resp, err := makeSomeHTTPCall()
require.NoError(t, err)
defer resp.Body.Close()

assert.HTTPStatus(t, resp, http.StatusOK)
assert.HTTPRespBodyContains(t, resp, "foo")
assert.HTTPRespHeaderContains(t, resp, ...)
```
Simple assertions on a given value.

But it looks like there already are some functions (like [`HTTPBodyContains`](https://github.com/stretchr/testify/blob/v1.8.4/assert/http_assertions.go#L130), [`HTTPStatusCode`](https://github.com/stretchr/testify/blob/v1.8.4/assert/http_assertions.go#L95), etc) that take the handler as an argument and execute the HTTP request as a part of the assertion function. 

---

Would love to hear the thoughts of other maintainers on this.
@dolmen @brackendawson

Matovidlo:
@arjunmahishi 
Hello,
I just added new function to already created module in simillar way. I am not sure how to refactor it so the code base looks  fine. Either it would need first of all refactor of the module and then make this PR as feature to new code base.

Also there would be a major change if we refactor these function to not work with handler. So when projects with already created tests would need to refactor it. Probablly this is for further discussion in some Issue and create for example `v2` branch?

Just to mention the motivation, due to not accepting this PR I need 2 testing frameworks to test the HTTP request, response. The issue I found is that it is not possible to test the response header/headers. However, to simplify the tests I thought it would be better to create 1 assert function. The developer will scope the `HTTP` by what is the developer testing. 

I would like to receive at least some comment whether I have to refactor it to have just `HTTPResponseHeaderContains` Instead of `HTTP`.

Have a nice day

arjunmahishi:
Yes. I understand the intent of this PR. Maybe I did not phrase it correctly. Don't let my comment be a blocker for this PR. The testing strategy i mentioned would require a lot of refactoring. And its probably not worth doing that. I was just putting it out it out there.

assert/http_assertions.go
```diff
@@ -163,3 +164,50 @@ func HTTPBodyNotContains(t TestingT, handler http.HandlerFunc, method, url strin
...
+//
+//	assert.HTTP(t, myHandler, "www.google.com", nil, WithCode(200), WithBody("I'm Feeling Lucky"), WithRequestHeader(http.Header{"a": []string{"b"}}, WithExpectedBody(bytes.NewBuffer("c"))))
+//
+// Returns whether the assertion was successful (true) or not (false).
+func HTTP(t TestingT, handler http.HandlerFunc, method, url string, values url.Values, options ...HttpOption) bool {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+
+	b := &builder{}
```
arjunmahishi, line 178: Since the `options` is optional, should the builder have some defaults?

Consider this:

```go
func handler(w http.ResponseWriter, r *http.Request) {
	w.Write([]byte("hello"))
}

func TestHttp(t *testing.T) {
	assert.HTTP(t, handler, http.MethodGet, "/", nil)
}
```

Fails with:

```
=== RUN   TestHttp
    main_test.go:15: 
        	Error Trace:	/Users/armahishi/dev/personal/forks/testify/example/main_test.go:15
        	Error:      	Expected HTTP success status code for "/?" but received 200
        	Test:       	TestHttp
--- FAIL: TestHttp (0.00s)
FAIL
FAIL	github.com/stretchr/testify/example	0.173s
FAIL
```

assert/http_assertions.go
```diff
@@ -163,3 +164,50 @@ func HTTPBodyNotContains(t TestingT, handler http.HandlerFunc, method, url strin
...
+		return err == b.err
+	}
+
+	req.Header = b.requestHeader
+	req.URL.RawQuery = values.Encode()
+	handler(w, req)
+	if w.Code != b.code {
+		Fail(t, fmt.Sprintf("Expected HTTP success status code for %q but received %d", url+"?"+values.Encode(), w.Code))
+		return false
+	}
```
arjunmahishi, line 200: We can reuse the existing assertions. It already has a lot of diff logic.

```suggestion
	Equal(t, b.code, w.Code)
```

```
=== RUN   TestHttp
    main_test.go:15: 
        	Error Trace:	/Users/armahishi/dev/personal/forks/testify/example/main_test.go:15
        	Error:      	Expected HTTP success status code for "/?" but received 200
        	Test:       	TestHttp
--- FAIL: TestHttp (0.00s)
FAIL
FAIL	github.com/stretchr/testify/example	0.161s
FAIL
```

VS

```
=== RUN   TestHttp
    main_test.go:15: 
        	Error Trace:	/Users/armahishi/dev/personal/forks/testify/example/main_test.go:15
        	Error:      	Expected HTTP success status code for "/?" but received 200
        	Test:       	TestHttp
--- FAIL: TestHttp (0.00s)
FAIL
FAIL	github.com/stretchr/testify/example	0.161s
FAIL
```

assert/http_assertions.go
Matovidlo: Good point I have added the default code for builder.

assert/http_assertions.go
Matovidlo: I see your point, however I wanted to be concise with the code style in the module, and there is not usage of `Equal` even though it is possible. 
As I am more sceptical to use module itself I think it is correct to not use `Equal` function as if it will be changed in future, you have to rework whole `http_assertions.go` module as it is dependend on that.
It is probably bad assumption but I think we can have a discussion about this and possibly make another PR with appropriate refactor of `http_assertions.go` module. :sweat_smile:

assert/http_assertions.go
```diff
@@ -163,3 +164,50 @@ func HTTPBodyNotContains(t TestingT, handler http.HandlerFunc, method, url strin
...
+		h.Helper()
+	}
+
+	b := &builder{}
+	for _, option := range options {
+		err := option(b)
+		if err != nil {
+			Fail(t, fmt.Sprintf("Failed to build http options, got error: %s", err))
+		}
+	}
```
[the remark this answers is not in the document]
Matovidlo: To preserve the variadic options, then the `code` cannot be set as default because there is option that would be set up always and therefore it is not an option, it would be a required parameter of function. 
Therefore I decided to make a `status.code` to be option however to make sure the usage of tests is correct, I throw a fail when no options are set as there is nothing to assert using result of `handler` call.
Basically it is a prevention of missuse of test that it does nothing except the `handler` execution.

assert/http_assertions.go
Matovidlo: I have made a change, based on what you said in next review step.
There are no defaults as the option would override something that is default for the builder. You mentioned that you expected that the `status.code` is not tested as You did not ask for it.
As the options should be optional, I made a change when you select no option, there is nothing to be tested so the test fails. It is possible to to remove this check, however it is incorrectly tested `handler` as there is not `assertion`.

Matovidlo:
@arjunmahishi Is there any update regarding this PR? I would like to push it forward if possible. :slightly_smiling_face:

brackendawson:
> Honestly, I am not a fan of the `assert` package calling the handler function as a part of an assert function/method. IMO, this package should only match values.

This is my thinking too. I note this that proposed function also only works with http.HandlerFunc and not http.Handler, would we add another function to support http.Handler? Better to make assertions against a httptest.ResponseRecorder because this would support both standard library handlers, as well as those from other libraries that differ from stdlib such as gin and restful.

Matovidlo:
@brackendawson Will be there any progress or just close this PR as wont do? I don't see a point what should I do differently or how to send PR with enhancement of this.

brackendawson:
I honestly don't see us adding more assertions that call handlers. Sorry about that.

Issue #1501, opened 2023-11-01
suite: TestSubtestPanic is failing


dolmen:
```console
$ git log -1 --oneline --no-abbrev-commit          
331c520966e81450aef3b50956ea4a5db5e0ccdf (HEAD -> master, origin/master, origin/HEAD) Improve readability
$ go version
go version go1.21.3 darwin/amd64
$ go test -v -race -run TestSubtestPanic -failfast ./suite
=== RUN   TestSubtestPanic
=== RUN   TestSubtestPanic
=== RUN   TestSubtestPanic/TestSubtestPanic
=== RUN   TestSubtestPanic/TestSubtestPanic/subtest
    suite.go:87: test panicked: panic
        goroutine 35 [running]:
        runtime/debug.Stack()
        	/usr/local/Cellar/go/1.21.3/libexec/src/runtime/debug/stack.go:24 +0x67
        github.com/stretchr/testify/suite.failOnPanic(0xc000103040, {0x1646ec0, 0x176d450})
        	/Users/omengue/go/src/github.com/stretchr/testify/suite/suite.go:87 +0x3a
        github.com/stretchr/testify/suite.recoverAndFailOnPanic(0x913f1f8?)
        	/Users/omengue/go/src/github.com/stretchr/testify/suite/suite.go:82 +0x37
        panic({0x1646ec0?, 0x176d450?})
        	/usr/local/Cellar/go/1.21.3/libexec/src/runtime/panic.go:920 +0x270
        github.com/stretchr/testify/suite.(*subtestPanicSuite).TestSubtestPanic.func1()
        	/Users/omengue/go/src/github.com/stretchr/testify/suite/suite_test.go:673 +0x2b
        github.com/stretchr/testify/suite.(*Suite).Run.func1(0xc000103040)
        	/Users/omengue/go/src/github.com/stretchr/testify/suite/suite.go:113 +0x1fa
        testing.tRunner(0xc000103040, 0xc0001355e0)
        	/usr/local/Cellar/go/1.21.3/libexec/src/testing/testing.go:1595 +0x239
        created by testing.(*T).Run in goroutine 34
        	/usr/local/Cellar/go/1.21.3/libexec/src/testing/testing.go:1648 +0x82b
--- FAIL: TestSubtestPanic (0.01s)
    --- FAIL: TestSubtestPanic/TestSubtestPanic (0.00s)
        --- FAIL: TestSubtestPanic/TestSubtestPanic/subtest (0.00s)
[...]

PR #1503, opened 2023-11-01
suite: fix TestSubtestPanic failure (#1501)

dolmen:
## Summary

The [subtest of `TestSubtestPanic`](https://github.com/stretchr/testify/blob/331c520966e81450aef3b50956ea4a5db5e0ccdf/suite/suite_test.go#L673) is expected to fail, but that must not make the testuite of package 'suite' to fail. So call [`Skip`](https://pkg.go.dev/testing#T.Skip) to make `go test` to ignore that expected failure.

## Motivation
`suite` testsuite failure (#1501).

## Related issues
* #1501
* #1471

Cc: @linusbarth @xwjdsh

### DIFF
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -669,9 +669,14 @@ func (s *subtestPanicSuite) TearDownSubTest() {
 }
 
 func (s *subtestPanicSuite) TestSubtestPanic() {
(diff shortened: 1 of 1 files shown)


dolmen:
Replaced by #1504.

PR #1506, opened 2023-11-09
assert: simplify isNil implementation

dolmen:
## Summary
Simpler and faster (also less allocations) in internal function `isNil`.

## Changes
* Simplify `isNil`: use a switch on `reflect.Value.Kind()` instead of browsing a slice of allowed `Kind`s.

## Motivation
Less code, less complexity, less memory usage, more speed.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -632,17 +632,6 @@ func NotNil(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 	return Fail(t, "Expected value not to be nil.", msgAndArgs...)
 }
 
-// containsKind checks if a specified kind in the slice of kinds.
-func containsKind(kinds []reflect.Kind, kind reflect.Kind) bool {
-	for i := 0; i < len(kinds); i++ {
-		if kind == kinds[i] {
-			return true
-		}
-	}
-
-	return false
-}
-
 // isNil checks if a specified object is nil or not, without Failing.
 func isNil(object interface{}) bool {
 	if object == nil {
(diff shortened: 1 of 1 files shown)


assert/assertions.go
```diff
@@ -632,34 +632,20 @@ func NotNil(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
...
-			reflect.Chan, reflect.Func,
-			reflect.Interface, reflect.Map,
-			reflect.Ptr, reflect.Slice, reflect.UnsafePointer},
-		kind)
-
-	if isNilableKind && value.IsNil() {
-		return true
+	switch value.Kind() {
+	case
+		reflect.Chan, reflect.Func,
```
Antonboom, line 644: Also, more dirty option, but protect us from supporting of new possible nillable types:

```go
// isNil checks if a specified object is nil or not, without Failing.
func isNil(object interface{}) (result bool) {
	if object == nil {
		return true
	}
	
	defer func() {
		if err := recover(); err != nil {
			result = false
		}
	}()
	return reflect.ValueOf(object).IsNil()
}
```

Antonboom (approved):

assert/assertions.go
dolmen: Interesting. However I have concerns about performance because of panic/defer. That would require a benchmark.

But if the goal is to be future-proof, at least that alternative implementation would be interesting for validation of `assert.Nil`.

dolmen: As I don't think that that extra safety net is necessary I'm keeping my original version.

dolmen:
@MovieStoreGuy Could you review this?

MovieStoreGuy (approved):

PR #1507, opened 2023-11-14
assert: Add a self-describing Matcher

pdhammond42:
## Summary
Add a Matcher interface and associated assert.Matches function

## Changes
Add interface, assertion functoin and an exmaple matcher implementation.

## Motivation
This allows users to extend the assertions in possibly highly domain specific ways, without having to modify the assert package, and provide a suitably descriptive failure message to help pinpoint the problem.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -2071,7 +2071,11 @@ func buildErrorChainString(err error) string {
 // This allows the assert framework to be extended in specialised ways while maintaining
 // clear error reporting.
 type Matcher interface {
+	// Match returns true if the given actual value meets the criteria embodied
+	// in this Matcher.
 	Match(actual interface{}) bool
+	// Describe is called when the test fails to get a descriptive error message.
+	// It should succinctly describe the condition that is being expected.
 	Describe() string
 }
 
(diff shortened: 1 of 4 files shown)


dolmen (changes requested):
Please run the code generator. Check the Actions result.

dolmen:
* This extends the API surface
* This can be built on top of Testify, in a separate package

At this point I don't want to extend the API surface.
So, closing.

pdhammond42:
Hi @dolmen , I don't see how this can be built on top of Testify. The point is to allow new tests to be built on top of Testify, using Testify's description of the failure. Of course I could do something similar using 
```
stringStartingFoo:=matches.StringStarting("foo")
assert.True(t, stringStartingFoo.Test(x), stringStartingFoo.Describe(x))
```
but then I could also do
```
if c := matches.StringStarting("foo"); !c.Test(x) {
  t.Error(c.Describe(x)
}
```

But it's your code base, your decision. Thanks for looking

dolmen:
"build on top of Testify" means "create your own module that uses the testify/assert API".

PR #1512, opened 2023-11-30
suite.Run: validate Test* methods signature

### LINKED ISSUE #1508: test panicked: reflect: Call with too few input arguments
If a test method on a suite has a parameter there is a panic:

````go
func (suite *SomeTestSuite) TestUpdateSomething(somearg string) {
......
}
````
When running this you get the following panic:
````go
suite.go:87: test panicked: reflect: Call with too few input arguments
....
````
I think the error handling could be more clear here. Something like "method TestUpdateSomething on SomeTestSuite contains arguments"

I can create a pull request if this change is wanted?

### PR DESCRIPTION
**Pull request by Barugoo**
## Summary
Added panic in suite.Run if Test* method has args or returning values

_P.S. This is my first ever opensource contribution, so I'm very open to any code logic/styling criticism!_

## Changes
- Now, before calling Suite method there's an additional check for signature correctness: it should have not more than 1 arg (receiver) and 0 returning values
- If Test* method signature fails the above check test panics with the message: `testify: suite method <MethodName> shouldn't have any arguments or returning values`
- Added test cases for incorrect signature variants

## Motivation
As @wvell mentioned in issue #1508 currently, `suite.Run` returns an ambiguous error when executing with instance of Suite that has incorrect `Test*` methods signatures. 

## Related issues
Closes #1508

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -185,7 +185,7 @@ func Run(t *testing.T, suite TestingSuite) {
 				}()
 
 				if method.Type.NumIn() > 1 || method.Type.NumOut() > 0 {
-					msg := fmt.Sprintf("testify: suite method '%s' shouldn't have any arguments or returning values\n", method.Name)
+					msg := fmt.Sprintf("testify: suite method '%s' shouldn't have any arguments or returning values", method.Name)
 					panic(msg)
 				}
 

### CODE REVIEW

File: suite/suite.go [hunk 1]
```diff
@@ -184,6 +184,11 @@ func Run(t *testing.T, suite TestingSuite) {
 					failOnPanic(t, r)
 				}()
 
+				if method.Type.NumIn() > 1 || method.Type.NumOut() > 0 {
+					msg := fmt.Sprintf("testify: suite method '%s' shouldn't have any arguments or returning values\n", method.Name)
+					panic(msg)
```
Barugoo, line 189: Not sure about using panic here tho, seems like appropriate, but will appreciate if you take a look @dolmen

wvell: Isn't it better to use t.Error(msg)? Then you can directly see the TestMethod violating the contract. I find panic always quite hard.

wvell: Yeah fair point, I just hate reading panics.

Thanks for putting in the work.

dolmen: While I'm a strong advocate of `panic` to catch programmer's mistake, I think a `panic` is not appropriate in this case as the stack trace will not be useful to the user.

Instead we are being called with `*testing.T`, so let's just use `t.Fatal` (or `t.Fatalf`) to report the issue.

However, the main problem with this implementation is that this check must be done much much earlier, line 144, just after the successful call to `methodFilter`.

dolmen (changes requested):

PR #1513, opened 2023-12-02
Add function for supporting JSON equality checks using byte slices

GCrispino:
## Summary
This PR adds a new function into the `assert` package, `JSONEqBytes`, that does the same as `JSONEq`, but by receiving `[]byte` values as parameter instead of `string` ones.

## Changes
`assert/assertions.go`:
The implementation of `JSONEqBytes` is pretty much what was in `JSONEq`, but having it receive `[]byte` values and then not needing to convert these values to `[]byte` when calling `json.Unmarshal` inside it.

Then, the implementation of `JSONEq` was changed such that it just calls `JSONEqBytes` by converting its original arguments to `[]byte` values (which it already did when calling `json.Unmarshal`)

`assert/assertions_test.go`:
Tests for `JSONEqBytes` function were replicated from the tests from `JSONEq`.

## Motivation
It is common in Go to store JSON text in `[]byte` variables instead of `string` values, so having a function that allows to receive these values when doing JSON-equality checks without having to cast them to string can be useful.

Maybe not as important, it would also avoid unnecessary `[]byte`-to-`string` conversions, which could add up when dealing with big JSON payloads.

Example of usage:
```go
package main

import (
    "fmt"
    "testing"
)


func TestIntMinBasic(t *testing.T) {
    actualFile := "/path/to/file.json"
    byteValue, err := ioutil.ReadAll(jsonFile)
    require.NoError(t, err)

    expectedContentsFile := "/path/to/expected.json"
    byteValue, err = ioutil.ReadAll(jsonFile)
    require.NoError(t, err)

    assert.JSONEqBytes(t, expected, byteValue)
}
```

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -2210,6 +2210,57 @@ func TestJSONEq_ArraysOfDifferentOrder(t *testing.T) {
 	False(t, JSONEq(mockT, `["foo", {"hello": "world", "nested": "hash"}]`, `[{ "hello": "world", "nested": "hash"}, "foo"]`))
 }
 
+func TestJSONEqBytes_EqualSONString(t *testing.T) {
+	mockT := new(testing.T)
+	True(t, JSONEqBytes(mockT, []byte(`{"hello": "world", "foo": "bar"}`), []byte(`{"hello": "world", "foo": "bar"}`)))
+}
+
+func TestJSONEqBytes_EquivalentButNotEqual(t *testing.T) {
+	mockT := new(testing.T)
+	True(t, JSONEqBytes(mockT, []byte(`{"hello": "world", "foo": "bar"}`), []byte(`{"foo": "bar", "hello": "world"}`)))
+}
+
+func TestJSONEqBytes_HashOfArraysAndHashes(t *testing.T) {
+	mockT := new(testing.T)
+	True(t, JSONEqBytes(mockT, []byte("{\r\n\t\"numeric\": 1.5,\r\n\t\"array\": [{\"foo\": \"bar\"}, 1, \"string\", [\"nested\", \"array\", 5.5]],\r\n\t\"hash\": {\"nested\": \"hash\", \"nested_slice\": [\"this\", \"is\", \"nested\"]},\r\n\t\"string\": \"foo\"\r\n}"),
+		[]byte("{\r\n\t\"numeric\": 1.5,\r\n\t\"hash\": {\"nested\": \"hash\", \"nested_slice\": [\"this\", \"is\", \"nested\"]},\r\n\t\"string\": \"foo\",\r\n\t\"array\": [{\"foo\": \"bar\"}, 1, \"string\", [\"nested\", \"array\", 5.5]]\r\n}")))
+}
+
+func TestJSONEqBytes_Array(t *testing.T) {
+	mockT := new(testing.T)
+	True(t, JSONEqBytes(mockT, []byte(`["foo", {"hello": "world", "nested": "hash"}]`), []byte(`["foo", {"nested": "hash", "hello": "world"}]`)))
+}
+
+func TestJSONEqBytes_HashAndArrayNotEquivalent(t *testing.T) {
+	mockT := new(testing.T)
+	False(t, JSONEqBytes(mockT, []byte(`["foo", {"hello": "world", "nested": "hash"}]`), []byte(`{"foo": "bar", {"nested": "hash", "hello": "world"}}`)))
+}
+
+func TestJSONEqBytes_HashesNotEquivalent(t *testing.T) {
+	mockT := new(testing.T)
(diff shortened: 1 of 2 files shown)


hendrywiranto:
Hi ! 
this one is great but I think currently maintainers are more focused in fixing bugs rather than expanding the API interface.  
Let's wait their response

assert/assertions_test.go
```diff
@@ -2210,6 +2210,57 @@ func TestJSONEq_ArraysOfDifferentOrder(t *testing.T) {
...
+
+func TestJSONEqBytes_EquivalentButNotEqual(t *testing.T) {
+	mockT := new(testing.T)
+	True(t, JSONEqBytes(mockT, []byte(`{"hello": "world", "foo": "bar"}`), []byte(`{"foo": "bar", "hello": "world"}`)))
+}
+
+func TestJSONEqBytes_HashOfArraysAndHashes(t *testing.T) {
+	mockT := new(testing.T)
+	True(t, JSONEqBytes(mockT, []byte("{\r\n\t\"numeric\": 1.5,\r\n\t\"array\": [{\"foo\": \"bar\"}, 1, \"string\", [\"nested\", \"array\", 5.5]],\r\n\t\"hash\": {\"nested\": \"hash\", \"nested_slice\": [\"this\", \"is\", \"nested\"]},\r\n\t\"string\": \"foo\"\r\n}"),
+		[]byte("{\r\n\t\"numeric\": 1.5,\r\n\t\"hash\": {\"nested\": \"hash\", \"nested_slice\": [\"this\", \"is\", \"nested\"]},\r\n\t\"string\": \"foo\",\r\n\t\"array\": [{\"foo\": \"bar\"}, 1, \"string\", [\"nested\", \"array\", 5.5]]\r\n}")))
```
suqin-haha, line 2226: it would be great if we support `"array\": [{\"bar\": \"foo\"},` , since json is ignore order

brackendawson: Json does not ignore transpositions of keys and values.

brackendawson (approved):
This is a good implementation.

I have questions about whether JSONEq should have ever been added. My preference would always be your API should yield stably formatted output. If JSONEq should have been added then it should have always accepted a []byte rather than a string. I'm not going to merge this without input from other maintainers, any thoughts @dolmen?

It would be useful to see if there is a real world example where the string is prohibitively expensive.

brackendawson:
@GCrispino can you re-base to run the latest version of the tests.

GCrispino:
@brackendawson rebased my branch against master and re ran `go generate` and  `go fmt` 👍🏼

brackendawson (approved):
Still approve, as per my previous comments.

GCrispino:
@brackendawson thanks! Is there something else I should do, or are you only expecting review from another maintainer?

assert/assertions_test.go
suqin-haha: oh I mean different order in array. I mistyped. But I just notice you already did that. So all good. Thanks!

brackendawson:
> @brackendawson thanks! Is there something else I should do, or are you only expecting review from another maintainer?

Nothing else needed from you at this time, just wait for another maintainer to opine.

fredbi:
Note to self: push merge conflict resolve

this is one nice addition in a million. The string stuff should never have existed in the first place but let’s be pragmatic and adopt one api that we test writers may use idiomatically, ie without type conversion

ccoVeille (approved):

GCrispino:
Hi @fredbi !

I have gone ahead and solved the merge conflicts myself. I believe it's all good, as tests are passing.
I've ran `go generate ./...` and `go fmt ./...` again but it appears that no changes happened as a result of this.

fredbi (approved):
Great job

assert/cpu.out
fredbi, line 1: We probably don't need this cpu.out file.

GCrispino: Sorry, I hadn't realized I had added that file. Pushed a commit removing it 👍🏼

fredbi:
@GCrispino  merged in our fork go-openapi/testify

GCrispino:
@dolmen  

> We must have special messages for the `[]byte(nil)` cases.

Currently if `expected == []bytes(nil)`, the `"Expected value ('') is not valid json.\nJSON parsing error: 'unexpected end of JSON input'"` message will be used, and in the case where `actual == []bytes(nil)`, the `"Input ('') needs to be valid json.\nJSON parsing error: 'unexpected end of JSON input'"` one will be used.

I've added a commit that will use the messages of `"Expected value is empty"` and `"Actual value is empty"`, respectively, in case that the provided byte slices are either nil or empty, if you believe we need special messages in these cases, apart from the current ones complaining that the input is not proper JSON.
Otherwise, I can just revert this commit.

Can you please have a look at it?

ccoVeille:
@GCrispino I think you quoted me while you wanted to quote @dolmen

GCrispino:
@ccoVeille oh yeah sorry, that was the case indeed 😅 , thanks for the heads up

Issue #1496, opened 2023-10-29
proposal: analogue of `InDeltaMapValues` with different deltas


Antonboom:
Hi!

Looks like [InDeltaMapValues](https://pkg.go.dev/github.com/stretchr/testify@master/assert#InDeltaMapValues) doesn't make sense if I need different deltas for different keys.

Thanks.

dolmen:
I understand what you don't want. But it is not clear to me what you want instead.

Antonboom:
I made some example 🙏 
https://go.dev/play/p/Rywjnhgua9q

brackendawson:
Genuine question, does this meet [this threshold](https://github.com/stretchr/testify/blob/e2741fa4e9bf2fdfe3ed48d976a7eeebe76c5009/CONTRIBUTING.md?plain=1#L13-L15)? What were you testing?

PR #1500, opened 2023-11-01
assert: Add new assertion InEpsilonMapValues

hendrywiranto:
## Summary
Add new assertion `InEpsilonMapValues` which is the same as InEpsilon, but it compares all values between two maps.

## Changes
- Add new assertion `InEpsilonMapValues`

## Motivation
Quoting https://github.com/stretchr/testify/issues/1497
> To be consistent with [InDeltaMapValues](https://pkg.go.dev/github.com/stretchr/testify@master/assert#InDeltaMapValues) package could provide InEpsilonMapValues

## Example
```
expected := map[string]float64{
	"foo": 2.2,
	"baz": 2.0,
}
actual := map[string]float64{
	"foo": 2.1,
	"baz": 2.1,
}
InEpsilonMapValues(mockT, expected, actual, 0.06, "InEpsilonMapValues example")
```

## Related issues
Closes https://github.com/stretchr/testify/issues/1497

PR #1519, opened 2023-12-17
Implement custom delta comparison for each map value

julian776:
## Summary

Add functionality to compare a custom delta for each map value.

## Changes

Include the function to compare each delta.

## Motivation

These changes now let you compare each map value with its own custom delta, instead of having to use the same  for all of them.

## Example usage

func TestHardcodedValues(t *testing.T) {
	expexted := map[string]float64{
		"foo": 1.0,
		"bar": 7,
		"baz": math.NaN(),
	}

	actual := map[string]float64{
		"foo": 1.01,
		"bar": 7.99,
		"baz": math.NaN(),
	}

	deltas := map[interface{}]float64{
		"foo": 0.1,
		"bar": 1,
		"baz": 0.1,
	}

	assert.InMapDeltasMapValues(
		t,
		expexted,
		actual,
		deltas,
	)
}

## Related issues

Incorporate functionality #1496.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1424,6 +1424,55 @@ func InDeltaMapValues(t TestingT, expected, actual interface{}, delta float64, m
 	return true
 }
 
+// InMapDeltasMapValues is the same as InDelta, but it compares all values between two maps with the matching delta on the delta's map. All maps must have exactly the same keys.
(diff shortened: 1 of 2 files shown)


hendrywiranto:
Hi,
I hate to break this to you but currently testify doesn't accept PR that expands the API as stated by this comment
https://github.com/stretchr/testify/pull/1500#issuecomment-1793674608

but I may be wrong, need to wait maintainer's response

julian776:
I wasn't aware of this, but no problem if that's the case.

PR #1526, opened 2024-01-27
assert: support byte slice in Contains

nickajacks1:
## Summary
Update [`assert.Contains`](https://pkg.go.dev/github.com/stretchr/testify/assert#Contains) and [`assert.NotContains`](https://pkg.go.dev/github.com/stretchr/testify/assert#NotContains) to check if a byte slice contains another byte slice.

## Changes
If both the list and element arguments are byte slices, use [`bytes.Contains`](https://pkg.go.dev/bytes#Contains) to check if element is a subslice. This mirrors the way strings are treated by `assert.Contains`.

## Motivation
The change makes [`Contains`](https://pkg.go.dev/github.com/stretchr/testify/assert#Contains) more consistent and predictable.  One could argue that developers could easily cast byte slices to strings, but the fact that that Contains does not treat strings and byte slices the same way breaks the principle of least astonishment (and probably would warrant a new rule in [`testifylint`](https://github.com/Antonboom/testifylint)).

```go
foo := []byte(`Text`)
b, err := os.ReadFile("file.txt")
require.NoError(t, err)
require.Contains(t, b, foo)
```


assert/assertions.go
```diff
@@ -845,19 +845,26 @@ func containsElement(list interface{}, element interface{}) (ok, found bool) {
...
 			}
 		}
 		return true, false
+	case reflect.Slice:
+		elementValue := reflect.ValueOf(element)
+		if elementValue.Kind() != reflect.Slice ||
+			elementValue.Type().Elem().Kind() != reflect.Uint8 ||
+			listType.Elem().Kind() != reflect.Uint8 {
+			break
+		}
```
arjunmahishi, line 866: This condition looks pretty complex. I think it would be super helpful if a comment explains the intent. Ex:   Why are we discarding `uint8`?

arjunmahishi (commented):
Probably worth updating the examples

Here: https://github.com/stretchr/testify/blob/84e42a9809d9eed42e8ac85a483755f4b0dbafc2/assert/assertions.go#L879-L884

And here: https://github.com/stretchr/testify/blob/84e42a9809d9eed42e8ac85a483755f4b0dbafc2/assert/assertions.go#L902-L907

So, that the godoc reflects that `[]byte` is now supported.

assert/assertions.go
nickajacks1: @arjunmahishi Inverting the `if` made things a lot clearer. Also added a brief comment. Please let me know if it could still be clearer

nickajacks1:
> Would this change be more useful to more people if when `contains` is a slice or array of the same type as `s`, then the assertion checks that the run of `contains` appears at least once in `s`? Rather than this only working with `[]byte`, and not even array of byte.
> 
> eg:
> 
> ```go
> Contains(t, []int{1,2,3,4,5,6,7,8,9}, []int{4,5,6})                        // -> true
> Contains(t, []int{1,2,3,4,5,6,7,8,9}, []int{3,5,7})                        // -> false
> Contains(t, []byte("My hovercraft is full of eels"), []byte("hovercraft")) // -> true
> Contains(t, [...]byte("I am an array"), [...]byte("arr"))                  // -> true
> Contains(t, []any{"hi", []byte{}, false, 9}, []byte{})                     // -> true
> ```

Yeah actually I think I was looking for that functionality before and was disappointed to find that it didn't exist.
This is a good idea, I'll look at reworking this to support "subslice" checking.

dolmen:
@nickajacks1 Let's keep this PR focused on `[]byte` support to allow us to merge this quickly, and submit other changes later.

assert/assertions.go
```diff
@@ -900,6 +907,8 @@ func containsElement(list interface{}, element interface{}) (ok, found bool) {
 //	assert.Contains(t, "Hello World", "World")
 //	assert.Contains(t, ["Hello", "World"], "World")
 //	assert.Contains(t, {"Hello": "World"}, "Hello")
+//	assert.Contains(t, []byte("Hello World"), []byte("World"))
+//	assert.Contains(t, [[]byte("Hello"), []byte("World")], []byte("World"))
```
dolmen, line 911: This line is not valid Go.

Also: any case presented in the doc must be strictly checked in the test suite.

assert/assertions.go
```diff
@@ -923,6 +932,8 @@ func Contains(t TestingT, s, contains interface{}, msgAndArgs ...interface{}) bo
 //	assert.NotContains(t, "Hello World", "Earth")
 //	assert.NotContains(t, ["Hello", "World"], "Earth")
 //	assert.NotContains(t, {"Hello": "World"}, "Earth")
+//	assert.NotContains(t, []byte("Hello World"), []byte("Earth"))
+//	assert.NotContains(t, [[]byte("Hello"), []byte("World")], []byte("Earth"))
```
dolmen, line 936: This line is not valid Go.

Also: any case presented in the doc must be strictly checked in the test suite.

dolmen (changes requested):

assert/assertions.go
```diff
@@ -893,19 +893,25 @@ func containsElement(list interface{}, element interface{}) (ok, found bool) {
...
 		for i := 0; i < len(mapKeys); i++ {
 			if ObjectsAreEqual(mapKeys[i].Interface(), element) {
 				return true, true
 			}
 		}
 		return true, false
+	case reflect.Slice:
+		switch element := element.(type) {
+		case []byte:
+			if list, ok := list.([]byte); ok {
```
Antonboom, line 911: should we have `return false, false` if `!ok`?

nickajacks1: I don't think so, at least with how the code is written - iIf list is a [][]byte, we want to fall through to the loop check on  line 917.
Maybe it's valid to say that if list is neither a []byte or a [][]byte, we should return `false, false`, but I haven't thought very hard about that case yet.

brackendawson (approved):
It's a shame not to go with the generic subslice within slice approach first as it solves this case too. But this does not preclude doing that.

brackendawson:
Could you resolve the conflicts please @nickajacks1?

nickajacks1:
@brackendawson  done!

brackendawson (approved):
This appears fine to me, I've tried using it upside down and backwards.

assert/assertions_test.go
```diff
@@ -990,30 +990,44 @@ func TestContainsNotContains(t *testing.T) {
...
-		{list, "Bar", true},
-		{list, "Salut", false},
+		{"Hello World", "World", true},
+		{"Hello World", "Earth", false},
+		{[]byte("Hello World"), []byte("World"), true},
+		{[]byte("Hello World"), []byte("Earth"), false},
+		{list, "World", true},
+		{list, "Earth", false},
+		{byteSliceList, []byte("World"), true},
+		{byteSliceList, []byte("Earth"), false},
```
brackendawson, line 1023: Ah, while reading through @dolmen's review I noticed that these examples from the doc aren't strictly in the unit tests because simpleMap is a `map[any]any` and not a `map[string]string`.
```go
//	assert.Contains(t, map[string]string{"Hello": "World"}, "Hello")
//	assert.NotContains(t, map[string]string{"Hello": "World"}, "Earth")
```
Could you add another couple of tests for those exact cases please.

brackendawson (changes requested):
Some of the previous review's comments weren't completely implemented. I'll be a lot more comfortable dismissing the other review if we do so.
Thanks.

nickajacks1:
Sorry, I might have made a mistake when rebasing. I'll get to this soon.

assert/assertions.go
```diff
@@ -912,19 +912,25 @@ func containsElement(list interface{}, element interface{}) (ok, found bool) {
...
-
-	if listKind == reflect.Map {
+	case reflect.Map:
 		mapKeys := listValue.MapKeys()
 		for i := 0; i < len(mapKeys); i++ {
 			if ObjectsAreEqual(mapKeys[i].Interface(), element) {
 				return true, true
 			}
 		}
 		return true, false
```
ccoVeille, line 926: So here, I would say the expected behavior is to say that

map[string]string{"foo": "bar"} contains "foo"

And it's OK

But what about 

map[string]string{"foo": "bar"} containing "bar" ?

What is the expected behavior?

Is it tested?

assert/assertions.go
```diff
@@ -912,19 +912,25 @@ func containsElement(list interface{}, element interface{}) (ok, found bool) {
...
+	case reflect.Slice:
+		switch element := element.(type) {
+		case []byte:
+			if list, ok := list.([]byte); ok {
+				return true, bytes.Contains(list, element)
+			}
+		}
```
ccoVeille, line 933: Now about this

[]string{"foo", "bar"} contains "foo"

But what about []string{"foo", "bar"} containing 1, here 1 refer to the index 2 of the list, so "bar"

What is the expected return for Contains?

- an error about object type
- true
- false

Is it tested?

ccoVeille (commented):
I'm raising points of my understanding and knowledge.

Feel free to consider or ignore them

assert/assertions.go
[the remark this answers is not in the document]
ccoVeille: Already reported here, somehow

https://github.com/stretchr/testify/pull/1526#discussion_r1632098374

ccoVeille:
I feel like there is a "problem" with `containsElement`function, especially its returned values.

currently it's  `ok, found bool`, and I would like to suggest it could be refactored to have a `found bool, err error

This way we could report other errors than the simple "couldn't be applied built len()" that the `!ok` catches in Contains and NotContains.

This follows my remark about the expected behavior if contains returning false for "foo" when passing true as element.

It might be considered as a breaking change, and it might need to be addressed in a separate PR.

Maybe it's something that have been reported already

assert/assertions.go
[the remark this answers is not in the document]
brackendawson: I had the exact same concern. The original containsElement implementation should be returning ok=false when the element cannot exist in the container. But to fix this now is a breaking change. I'd accept another PR to document the peculiar behaviour on the functions.

assert/assertions.go
```diff
@@ -912,19 +912,25 @@ func containsElement(list interface{}, element interface{}) (ok, found bool) {
 		}
 	}()
 
-	if listKind == reflect.String {
+	switch listKind {
+	case reflect.String:
 		elementValue := reflect.ValueOf(element)
 		return true, strings.Contains(listValue.String(), elementValue.String())
```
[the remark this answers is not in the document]
brackendawson: I think this should remain false.

assert/assertions.go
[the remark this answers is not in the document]
brackendawson: Yes, the code at the end of the function makes this work, adding a test to cement this behaviour is a good idea. :+1:

brackendawson: I honestly found the Map behaviour testing keys a bit surprising, I'd find the array behaviour testing length even more surprising. Still worth adding the test via a new issue.

assert/assertions.go
nickajacks1: It would surprise me personally as a user for assert.Contains to return true if a value in a map matched.
But on the other hand, I kind of see what you mean since the value of a map is analogous to the value in a slice when you're iterating over a range (`for _, v := range collection {println(v)}`).

assert/assertions.go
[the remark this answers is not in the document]
nickajacks1: This is a good catch, it DOESNT work because `'A'` is a `rune` and not a `byte`.  But `Contains([]byte{'A, ' B'}, byte('A'))` works as expected.  That behavior, though technically correct, is confusing.

What do you guys think? Should that behavior be left alone, fixed in `ObjectsAreEqual`, or fixed only in `Contains`?

ccoVeille:
I think we agreed saying it should not match.

So you can add a test about it being false

brackendawson:
I agree on not including rune in this PR. This will need a re-base because I merged another fix in Contains.

assert/assertions.go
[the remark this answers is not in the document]
dolmen: That is beyond the scope of this PR. Let's focus on the simpler case of where values have the same type (`[]byte`).

assert/assertions_test.go
```diff
@@ -990,30 +990,47 @@ func TestContainsNotContains(t *testing.T) {
...
 
 	complexList := []*A{
 		{"b", "c"},
 		{"d", "e"},
 		{"g", "h"},
 		{"j", "k"},
 	}
-	simpleMap := map[interface{}]interface{}{"Foo": "Bar"}
+
+	interfaceList := []interface{}{"Hello World", 5, 2, []byte("Goodbye")}
```
dolmen, line 1006: Rename to `mixedList`.

dolmen (changes requested):

assert/assertions_test.go
nickajacks1: Done

assert/assertions.go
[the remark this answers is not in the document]
brackendawson: We're definitely not making rune work but I would still like to have a test for `Contains([]byte{"ABC"}, rune('A'))` to stabilise the current behaviour.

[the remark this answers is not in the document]
nickajacks1: Copy that

nickajacks1:
Added tests for []byte and rune.

assert/assertions_test.go
```diff
@@ -1029,30 +1029,50 @@ func TestContainsNotContains(t *testing.T) {
...
-		{"Hello World", "Salut", false},
-		{list, "Bar", true},
-		{list, "Salut", false},
+		{"Hello World", "World", true},
+		{"Hello World", "Earth", false},
+		{[]byte("Hello World"), []byte("World"), true},
+		{[]byte("Hello World"), []byte("Earth"), false},
+		{[]byte("ABC"), 'B', false},
+		{[]byte("ABC"), rune('B'), false},
+		{[]byte("ABC"), byte('B'), true},
```
ccoVeille, line 1077: I feel like these should be added

```suggestion
		{[]byte("ABC"), byte('B'), true},
		{[]byte("ABC"), []byte("B"), true},
		{[]byte("ABC"), []byte{66}, true},
		{[]byte("ABC"), []byte("AB"), true},
		{[]byte("ABC"), []byte{65, 66}, true},
```

https://go.dev/play/p/XTcbu5yNaRh

ccoVeille: In fact, I commented too fast.

This is the unit tests and not the godoc.

So here the equivalent notation such as []byte("AB") and []byte{65,66}  are not needed here

I feel like this would be enough, even if I conceed they are somehow covered by the hello world unit tests already present in the file.

```suggestion
		{[]byte("ABC"), byte('B'), true},
		{[]byte("ABC"), []byte("B"), true},
		{[]byte("ABC"), []byte("AB"), true},
```

ccoVeille:
Because bytes.Contains supports json.RawMessage

https://go.dev/play/p/ES5gYlZxECI

Can we make sure it works as expected with the assert.Contains in addition if the []byte supports ?

But maybe it works and it's just a matter of covering it with the unit tests

nickajacks1:
I probably will not be making more changes to this PR. Maintainers, please feel free to update the PR or supersede.

assert/assertions_test.go
brackendawson: `{[]byte("ABC"), []byte("B"), true},` I will concede, but `{[]byte("ABC"), []byte("AB"), true}` is definitely already there.

brackendawson:
@ccoVeille I've updated the PR to address the comments but as I now have a hand in this code I shouldn't self-approve, I'm re-assigning the review to you, PTAL.

assert/assertions.go
```diff
@@ -928,19 +928,25 @@ func containsElement(list interface{}, element interface{}) (ok, found bool) {
...
 			}
 		}
 		return true, false
+	case reflect.Slice:
+		switch element := element.(type) {
+		case []byte:
+			if list, ok := list.([]byte); ok {
+				return true, bytes.Contains(list, element)
+			}
+		}
```
ccoVeille: It's a bit strange that this new part, is the only one using type assertion while the the rest uses reflection.

I would have kept the same style if possible

Also, if you want/need to keep this code, I wouldn't use a switch type assertion for element, but a `if element, ok := element.(type); ok ` one

brackendawson: The type assertion would be preferable for performance, but your comment raises the point that if named types work for string then they should work for the new byte slice behaviour should too, all these cases should pass:
```go
package kata_test

import (
	"testing"

)

type my string

type alsoMy []byte

func TestIfy(t *testing.T) {
	assert.Contains(t, "Hello", "ell")                                 // true
	assert.Contains(t, my("Hello"), "ell")                             // true
	assert.Contains(t, "Hello", my("ell"))                             // true
	assert.Contains(t, my("Hello"), my("ell"))                         // true
	assert.Contains(t, []byte("Hello"), []byte("ell"))                 // true
	assert.Contains(t, alsoMy([]byte("Hello")), []byte("ell"))         // false
	assert.Contains(t, []byte("Hello"), alsoMy([]byte("ell")))         // false
	assert.Contains(t, alsoMy([]byte("Hello")), alsoMy([]byte("ell"))) // false
}
```

brackendawson: @ccoVeille I've extended the behavior to type definitions (`var myType []byte`) as well. I decided that it should extend to such definitions because the string case does, PTAL.

assert/assertions.go
```diff
@@ -955,8 +973,10 @@ func containsElement(list interface{}, element interface{}) (ok, found bool) {
 // specified substring or element.
 //
 //	assert.Contains(t, "Hello World", "World")
-//	assert.Contains(t, ["Hello", "World"], "World")
-//	assert.Contains(t, {"Hello": "World"}, "Hello")
+//	assert.Contains(t, []string{"Hello", "World"}, "World")
+//	assert.Contains(t, map[string]string{"Hello": "World"}, "Hello")
+//	assert.Contains(t, []byte("Hello World"), []byte("World"))
+//	assert.Contains(t, [][]byte{[]byte("Hello"), []byte("World")}, []byte("World"))
```
ccoVeille, line 979: I feel like this example is a bit too complicated, is there something in the code you add about it?

Also, I feel like the documentation of Contains could be better.

I mean it supports looking for 123 in []int{123, 456}, but also looking for true in []bool{false, true}, but also true in []any{123, true}

The documentation mentions only strings and bytes now, while it supports any slice.

Also, I feel like it's not tested at all

brackendawson: The history of this assertion is that it used to only perform the strings.Contains on string kinds (string and names type `type my string`). It was then expanded to assert that a slice or array kind contains a single element (should be assert.hasElement) and that map kinds contain a key (should be assert.hasKey), but the docs and tests were poorly updated.

We're now looking at adding support for a []byte contains a smaller []byte, because they're often used like strings. This extends the behaviour of slices but only when its kind is []byte or slice of kind of byte. This baloons the complexity.

The fact that it's incredibly difficult to document and test, and that nether the author or another maintainer can get this PR into a shape that the maintainers are happy with is telling. I'm returning to my original assessment of this change which is that we should not do it. assert.Contains is already improperly extended and we should not corrupt it further, adding support for []byte to work the same as an equivalent string is not trivial.

brackendawson:
For the reasons stated [above](https://github.com/stretchr/testify/pull/1526#discussion_r2300047753), I'm going to close this for now.

Proposals for giving []byte the same functionality as string would need to properly consider the interaction between []byte and generic slices, the behaviour of named byte slices (`type MyByteSlice []byte`), the behaviour of slices of named byte types (`type MyByte byte; var b []MyByte`), and how to fully document all the behavior of assert.Contains without placing a burden on the user.

PR #1529, opened 2024-02-02
use hexops/gotextdiff instead

arthurnn:
## Summary

This came up before [here](https://github.com/stretchr/testify/pull/1198). But I am taking a different approach.
Instead of using https://github.com/pmezard/go-difflib I am using https://github.com/hexops/gotextdiff. gotextdiff is a copy of a internal gopls [package](https://github.com/golang/tools/tree/master/internal/diff)

## Motivation
go-difflib is not maintained anymore. And the interface is more complex than gotextdiff

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -18,7 +18,8 @@ import (
 	"unicode/utf8"
 
 	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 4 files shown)


arjunmahishi:
Yikes! Looks like https://github.com/hexops/gotextdiff got archived 4 days ago 👀

PR #1531, opened 2024-02-10
assert: Fix EqualValues to handle overflow/underflow

### LINKED ISSUE #1462: assert: EqualValues does not check for overflows
When asserting two numeric values with EqualValues, it passes if one is the overflowed value of the other.

example: https://go.dev/play/p/KZsFh7FnBVc

```go
func TestEqualValues(t *testing.T) {
	var a int8 = 14
	var b int = 270
	assert.Equal(t, b, a)       // fails because of type mismatch
	assert.EqualValues(t, b, a) // passes
}
```

arjunmahishi:
## Summary

The underlying function `ObjectsAreEqualValues` did not handle overflow/underflow of values while converting one type to another for comparison. For example:

```go
    EqualValues(t, int(270), int8(14))
```

would return `true`, even though the values are not equal. Because, when you convert `int(270)` to `int8`, it overflows and becomes 14 `(270 % 256 = 14)`

## Changes

This commit fixes that by making sure that the conversion always happens from the **smaller type to the larger type**, and then comparing the values. Additionally, this commit also separates out the test cases of `ObjectsAreEqualValues` from `TestObjectsAreEqual`.

## Related issues

Closes #1462

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -165,17 +165,40 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
 		return true
 	}
 
-	actualType := reflect.TypeOf(actual)
-	if actualType == nil {
+	expectedValue := reflect.ValueOf(expected)
+	actualValue := reflect.ValueOf(actual)
+	if !expectedValue.IsValid() || !actualValue.IsValid() {
 		return false
 	}
-	expectedValue := reflect.ValueOf(expected)
-	if expectedValue.IsValid() && expectedValue.Type().ConvertibleTo(actualType) {
-		// Attempt comparison after type conversion
-		return reflect.DeepEqual(expectedValue.Convert(actualType).Interface(), actual)
+
+	expectedType := expectedValue.Type()
+	actualType := actualValue.Type()
+	if !expectedType.ConvertibleTo(actualType) {
+		return false
 	}
 
-	return false
+	if !isNumericType(expectedType) || !isNumericType(actualType) {
+		// Attempt comparison after type conversion
+		return reflect.DeepEqual(
+			expectedValue.Convert(actualType).Interface(), actual,
+		)
+	}
+
+	// If BOTH values are numeric, there are chances of false positives due
+	// to overflow or underflow. So, we need to make sure to always convert
+	// the smaller type to a larger type before comparing.
+	if expectedType.Size() >= actualType.Size() {
+		return actualValue.Convert(expectedType).Interface() == expected
+	}
+
+	return expectedValue.Convert(actualType).Interface() == actual
+}
+
+// isNumericType returns true if the type is one of:
+// int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64,
+// float32, float64, complex64, complex128
+func isNumericType(t reflect.Type) bool {
+	return t.Kind() >= reflect.Int && t.Kind() <= reflect.Complex128
 }
 
 /* CallerInfo is necessary because the assert functions use the testing object
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -135,24 +135,42 @@ func TestObjectsAreEqual(t *testing.T) {
 
 		})
 	}
+}
 
-	// Cases where type differ but values are equal
-	if !ObjectsAreEqualValues(uint32(10), int32(10)) {
-		t.Error("ObjectsAreEqualValues should return true")
-	}
-	if ObjectsAreEqualValues(0, nil) {
-		t.Fail()
-	}
-	if ObjectsAreEqualValues(nil, 0) {
-		t.Fail()
+func TestObjectsAreEqualValues(t *testing.T) {
+	now := time.Now()
+
+	cases := []struct {
+		expected interface{}
+		actual   interface{}
+		result   bool
+	}{
+		{uint32(10), int32(10), true},
+		{0, nil, false},
+		{nil, 0, false},
+		{now, now.In(time.Local), true}, // should be time zone independent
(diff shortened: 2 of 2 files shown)


assert/assertions.go
```diff
@@ -165,19 +165,40 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
...
+			}
+
+			return expectedValue.Convert(actualType).Interface() == actual
+		}
+
 		// Attempt comparison after type conversion
 		return reflect.DeepEqual(expectedValue.Convert(actualType).Interface(), actual)
 	}
 
 	return false
```
MovieStoreGuy, line 192: If it is possible, I would prefer to invert the statements to reduce the indenting?

It makes it easier for me to read (but it doesn't look like it is a practice within the project)

assert/assertions_test.go
```diff
@@ -135,24 +135,34 @@ func TestObjectsAreEqual(t *testing.T) {
...
-	}
+func TestObjectsAreEqualValues(t *testing.T) {
+	now := time.Now()
 
-	tm := time.Now()
-	tz := tm.In(time.Local)
-	if !ObjectsAreEqualValues(tm, tz) {
-		t.Error("ObjectsAreEqualValues should return true for time.Time objects with different time zones")
+	cases := []struct {
+		expected interface{}
```
MovieStoreGuy, line 144: I would ask to replace `interface` with `any`

MovieStoreGuy (commented):
Minor revision from me, and excuse as I become more familiar with helping to maintain the project.

assert/assertions_test.go
arjunmahishi: Sure. I haven't really used generics that much. I keep forgetting that it exists :P

assert/assertions.go
arjunmahishi: Makes sense. will do that and push the changes in a bit.

assert/assertions_test.go
arjunmahishi: We can't use generics yet. This project is still on go `v1.17`

https://github.com/stretchr/testify/blob/c3b0c9b4f50cd8320ccdcdfd3ffd6afc5b109c4a/go.mod#L5

assert/assertions.go
arjunmahishi: Done

MovieStoreGuy (dismissed):

brackendawson (changes requested):
Apparently complex isn't numeric, so this still behaves incorrectly:

```go
	EqualValues(t, complex128(1e+100+1e+100i), complex64(complex(math.Inf(0), math.Inf(0))))
```

I'm open to not fixing this case. Also I wonder why complex isn't numeric.

assert/assertions.go
```diff
@@ -165,17 +165,40 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
...
+	}
+
+	return expectedValue.Convert(actualType).Interface() == actual
+}
+
+// isNumericType returns true if the type is one of:
+// int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64,
+// float32, float64
+func isNumericType(t reflect.Type) bool {
+	return t.Kind() >= reflect.Int && t.Kind() <= reflect.Float64
```
brackendawson, line 201: Aha
```suggestion
	return t.Kind() >= reflect.Int && t.Kind() <= reflect.Complex128
```

arjunmahishi: I don't think `complex` numbers should be included in this bucket. Because We will not be able to convert an `int*` `float*` to `complex*`. So, there is no basis for comparison. 

Also, this condition will make sure that _`int*/float* <> complex*`_ comparisons will always yield `false`
https://github.com/stretchr/testify/blob/2d27e0f92acbd4ef9063399060e58ba3a6867310/assert/assertions.go#L176-L178

---

I tinkered around in the playground to verify this: https://go.dev/play/p/bhIO--yWxMN

brackendawson: But you can convert a complex128 to a complex64. Your test on L176 already fixes your concern by implicitly dividing the bucket into two. These testcases all pass with my proposed change:
```go
		{complex128(1e+100 + 1e+100i), complex64(complex(math.Inf(0), math.Inf(0))), false},
		{complex64(complex(math.Inf(0), math.Inf(0))), complex128(1e+100 + 1e+100i), false},
		{complex128(1e+100 + 1e+100i), 270, false},
		{270, complex128(1e+100 + 1e+100i), false},
		{complex128(1e+100 + 1e+100i), 3.14, false},
		{3.14, complex128(1e+100 + 1e+100i), false},
		{complex128(1e+10 + 1e+10i), complex64(1e+10 + 1e+10i), true},
		{complex64(1e+10 + 1e+10i), complex128(1e+10 + 1e+10i), true},
```

The interesting part is that 10+0i is mathematically equal to 10. But Go's equality operator [will not perform this operation](https://go.dev/play/p/AMYnvdCfYe1) and the types are truly [not convertible](https://go.dev/play/p/6iqAi5A7r-h) without using the helper functions. So I think it's safe and acceptable to include the behaviour in the above testcases.

Can you show me a testcase which behaves incorrectly?

NB: Careful with untyped literals when experimenting here, they [look like they are convertible](https://go.dev/play/p/5-wSmJcEFqG), but that's a static conversion so we will never see it inside ObjectsAreEqualValues.

arjunmahishi: > But you can convert a complex128 to a complex64

Oh! thats right 🤦🏽‍♂️

brackendawson (approved):

PR #1534, opened 2024-02-15
assert: clarify Eventually documentation

### LINKED ISSUE #1443: Clarify documentation of `assert.Eventually`/`assert.EventuallyWithT`
The documentation states:

```go
// Eventually asserts that given condition will be met in waitFor time,
// periodically checking target function each tick.
```

This is a bit ambiguous: does it call the condition function every tick, regardless of whether the previous one finished, or does it wait for the tick _between_ executions of the condition function? (It's actually neither of these)

Conditions cannot overlap as shown in [this test](https://go.dev/play/p/pfPEiQ-6um7). Based on the behavior of the implementation, it would be more precise to say the following:

```go
// Eventually asserts that the given condition will be met in waitFor time. The given condition 
// func is checked periodically. It will be checked every tick unless the execution of the condition 
// takes longer than one tick. If so, it will be called again on the next multiple of tick after it returns.
```

st3penta:
## Summary
The documentation of the `Eventually` function is a bit ambiguous, not specifying what happens when the execution of the condition function takes longer than one tick.

## Changes
Clarify the `Eventually` function description by describing the behavior of the condition function execution in the above scenario.

## Related issues
Closes #1443

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -158,8 +158,9 @@ func ErrorIsf(t TestingT, err error, target error, msg string, args ...interface
 	return ErrorIs(t, err, target, append([]interface{}{msg}, args...)...)
 }
 
-// Eventuallyf asserts that given condition will be met in waitFor time,
-// periodically checking target function each tick.
+// Eventuallyf asserts that given condition will be met in waitFor time. The target function
+// is checked each tick, unless its execution takes longer than one tick.
+// If so, it will be called again on the next multiple of tick after it returns.
 //
 //	assert.Eventuallyf(t, func() bool { return true; }, time.Second, 10*time.Millisecond, "error message %s", "formatted")
 func Eventuallyf(t TestingT, condition func() bool, waitFor time.Duration, tick time.Duration, msg string, args ...interface{}) bool {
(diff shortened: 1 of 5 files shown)


arjunmahishi (commented):
I feel like this comment is answering a very specific question about the implementation. If the user is not concerned with this detail, it might confuse them.

IMO, we can leave the doc comment as it was before and add comments within the function that describe the implementation details.

---

Perhaps here we can describe that "_the ticker is paused while the condition function is being evaluated. If the condition function yields `false` it is resumed again_" 

https://github.com/stretchr/testify/blob/890a441c7aa0cc9fcfb1b2a716f58542a9c440d5/assert/assertions.go#L1880-L1888

st3penta:
Actually i found that there are 3 occurrences of this same ticker logic: [here](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1879), [here](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1955), and [here](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1996). Should we replicate the same comment in all those 3 points?

I liked the idea of moving the explanation closer to the implementation, but actually the code itself is pretty simple, so maybe the comment isn't really necessary, what do you think?

arjunmahishi:
> but actually the code itself is pretty simple

Agreed. We can skip it then.

PR #1535, opened 2024-02-16
suite: fix deadlock in suite.Require()/Assert()

### LINKED ISSUE #1520: `suite.Require` deadlock
If `require` object hasn't been initialized before calling `suite.Require` method, it will be created on the base of  `*testing.T` returned from  `suite.T()` ([code](https://github.com/stretchr/testify/blame/v1.8.4/suite/suite.go#L61)):

```go
// Require returns a require context for suite.
func (suite *Suite) Require() *require.Assertions {
	suite.mu.Lock()
	defer suite.mu.Unlock()
	if suite.require == nil {
		suite.require = require.New(suite.T())
	}
	return suite.require
}
```

This will result in a deadlock because `suite.T()` wants to acquire the lock that has been already acquired:

```go
// T retrieves the current *testing.T context.
func (suite *Suite) T() *testing.T {
	suite.mu.RLock()
	defer suite.mu.RUnlock()
	return suite.t
}
```

arjunmahishi:
As pointed out in issue #1520, if the suite is not initialised properly (by calling the `Run` function), then calling `suite.Require()` or `suite.Assert()` will result in a deadlock.

## Changes

This commit fixes that by panicking if the suite is not initialised properly. This is justified because the suite is intended to be triggered in the right way. If the user does not do that, this panic will nudge them in the right direction.

It has to be a panic because, at this point, we don't have access to any `testing.T` context to gracefully call a `t.Fail()`. Also, these two functions are not expected to return an error.

## Related issues

Fixes #1520

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -58,7 +58,7 @@ func (suite *Suite) Require() *require.Assertions {
 	suite.mu.Lock()
 	defer suite.mu.Unlock()
 	if suite.require == nil {
-		suite.require = require.New(suite.T())
+		panic("'Require' must not be called before 'Run' or 'SetT'")
 	}
 	return suite.require
 }
@@ -72,7 +72,7 @@ func (suite *Suite) Assert() *assert.Assertions {
 	suite.mu.Lock()
 	defer suite.mu.Unlock()
 	if suite.Assertions == nil {
-		suite.Assertions = assert.New(suite.T())
+		panic("'Assert' must not be called before 'Run' or 'SetT'")
 	}
 	return suite.Assertions
 }
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -692,3 +692,27 @@ func TestSubtestPanic(t *testing.T) {
 	assert.True(t, suite.inTearDownTest)
 	assert.True(t, suite.inTearDownSuite)
 }
+
+type unInitialisedSuite struct {
+	Suite
+}
+
+// TestUnInitialisedSuites asserts the behaviour of the suite methods when the
+// suite is not initialised
+func TestUnInitialisedSuites(t *testing.T) {
+	t.Run("should panic on Require", func(t *testing.T) {
+		suite := new(unInitialisedSuite)
+
+		assert.Panics(t, func() {
+			suite.Require().True(true)
+		})
+	})
+
+	t.Run("should panic on Assert", func(t *testing.T) {
+		suite := new(unInitialisedSuite)
+
+		assert.Panics(t, func() {
+			suite.Assert().True(true)
+		})
+	})
+}


brackendawson (changes requested):
This is a breaking change.

Instead, you can fix the deadlock by making two versions of the `T` method:
```go
// T retrieves the current *testing.T context.
func (suite *Suite) T() *testing.T {
	suite.mu.RLock()
	defer suite.mu.RUnlock()
	return suite._t()
}

func (suite *Suite) _t() *testing.T {
	return suite.t
}
```

And then calling the internal version from `Require` and `Assert`, since they are alreadty protected by the lock.

arjunmahishi:
> This is a breaking change.

How? This would have never worked in the previous version right?

> You can fix the deadlock by making two versions of the T method

But just like `suite.require`, `suite.t` would also not be initialised without calling `Run`. So, the user would run into a _nil pointer deref_ panic.

```
                                          ┌─────────────────────────────────────────┐ 
                                          │func (suite *Suite) SetT(t *testing.T) { │ 
                                          │  suite.mu.Lock()                        │
┌─────┐        ┌─────────────┐            │  defer suite.mu.Unlock()                │
│Run()├───────►│suite.SetT(t)├───────────►│  suite.t = t                            │
└─────┘        └─────────────┘            │  suite.Assertions = assert.New(t)       │
                                          │  suite.require = require.New(t)         │
                                          │}                                        │ 
                                          └─────────────────────────────────────────┘ 

```

brackendawson:
At v1.0 (which is difficult to require from a modern module) `SetT` exists and does not initialise `suite.Suite.Require`, and the code path we are modifying was useful, it sucessfully returned a functioning `Require` to this contrived example:
```go
package kata_test

import (
	"os"
	"testing"

)

type mySuite struct {
	suite.Suite
}

func (suite *mySuite) TestSomething() {
	require := suite.Require()
	require.True(true)
}

func TestSuite(t *testing.T) {
	m := &mySuite{}
	m.SetT(t)
	f, err := os.Create("testfile")
	m.Require().NoError(err)
	defer f.Close()
	suite.Run(t, m)
}
```

This code works today because `SetT` also initialises both `suite.Suite.Assert` and `suite.Suite.Require`, meaning we don't traverse the code we're modifying, which has at various times been a nil pointer dereference, a deadlock, or now a deliberate panic.

So I think you are right, I can't contrive an example that is broken by this change, it's not a breaking change.

Perhaps though the panic message should encompass something like: *"Require must not be called before Run or SetT"*.

arjunmahishi:
Ohh got it. Thanks for the context!
Are you on board with panicking though? (I've changed the message)

brackendawson (approved):

Issue #843, opened 2019-11-12
Ignore fields using struct tags


hiimoliverwang:
Hello, 

Is there a way to ignore certain fields when asserting equal or as the argument of .On?

e.g. 
```
type A struct {
  Name string
  UpdatedAt time.Time `mock: "ignore"`
}

expected := A{
    Name: "Expected Name"
}
mocks.MockController.On("Update", expected).Return(whatever)
...
```

We don't really care about the UpdateAt time, and this would also apply to random UUID's etc. 
Otherwise we would have to assert each sub field in order for the test to pass.

boyan-soubachov:
Not that I am aware of, would it make sense in your use case to explicitly assert the fields (if the include list is small) or conversely shallow-copy the struct and remove the ones you don't want to compare (if the exclude list is small)?

Khyme:
Hello. Is there any news on this? I have the issue described, and I cannot use the suggested ways of @boyan-soubachov because it's intricated controllers so I don't get to manipulate the intermediate result used by the mock.

dolmen:
@brackendawson I like your proposal, but the original issue is about `mock`. Could you put your proposal as a separate issue?

PR #1537, opened 2024-02-18
Revert "assert: ObjectsAreEqual: use time.Equal for time.Time type"

### LINKED ISSUE #1536: assert: Equal does not consider Zone information in time.Time
This test should not pass:
```go
package kata_test

import (
	"testing"
	"time"

)

func ToSydney(t time.Time) time.Time {
	// zoneSyd, err := time.LoadLocation("Australia/Sydney")
	zoneSyd, err := time.LoadLocation("Europe/London")
	if err != nil {
		panic("Australia is gone")
	}
	return t.In(zoneSyd)
}

func TestToSydney(t *testing.T) {
	zoneSyd, err := time.LoadLocation("Australia/Sydney")
	require.NoError(t, err)
	now := time.Now()
	require.Equal(t, now.In(zoneSyd), ToSydney(now))
}
```

And on v1.8.4 it correctly fails, but on master it incorrectly passes.

This is a regression introduced by #1464. time.Time.Equal does not mean the two instances are equal, it means they represent the same instant. There is more information contained in a time.Time instance than the instant, this regression makes it impossible to test for state which is a best practice. If the user only cares about the instant represented then `WithinDuration(t, expected, actual, 0)` is available.

brackendawson:
## Summary
This reverts commit 34763e0df3d560ca247996a0e9291fa919a3abc6.

time.Time.Equal only tests that the two instances refer to the same instant, but time.Time also carries zone information, so this caused two non-equal instances to be considered equal.

## Changes
Revert the change which cases assert.Equal to consider time.Time objects with different zones equal.

## Motivation
* We can't assume the user doesn't care avout the zone information, we should support testing for state first.
* This would be a breaking change.
* WithinDuration already covered the desired case.

## Related issues
Fixes #1536

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -148,8 +148,8 @@ func TestObjectsAreEqualValues(t *testing.T) {
 		{uint32(10), int32(10), true},
 		{0, nil, false},
 		{nil, 0, false},
-		{now, now.In(time.Local), true}, // should be time zone independent
-		{int(270), int8(14), false},     // should handle overflow/underflow
+		{now, now.In(time.Local), false}, // should not be time zone independent
+		{int(270), int8(14), false},      // should handle overflow/underflow
 		{int8(14), int(270), false},
 		{[]int{270, 270}, []int8{14, 14}, false},
 		{complex128(1e+100 + 1e+100i), complex64(complex(math.Inf(0), math.Inf(0))), false},
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -59,25 +59,20 @@ func ObjectsAreEqual(expected, actual interface{}) bool {
 	if expected == nil || actual == nil {
 		return expected == actual
 	}
-	switch exp := expected.(type) {
-	case []byte:
-		act, ok := actual.([]byte)
-		if !ok {
-			return false
-		}
-		if exp == nil || act == nil {
-			return exp == nil && act == nil
-		}
-		return bytes.Equal(exp, act)
-	case time.Time:
-		act, ok := actual.(time.Time)
-		if !ok {
-			return false
-		}
-		return exp.Equal(act)
-	default:
+
+	exp, ok := expected.([]byte)
+	if !ok {
 		return reflect.DeepEqual(expected, actual)
 	}
+
+	act, ok := actual.([]byte)
+	if !ok {
+		return false
+	}
+	if exp == nil || act == nil {
+		return exp == nil && act == nil
+	}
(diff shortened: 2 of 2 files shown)


assert/assertions_test.go
```diff
@@ -148,8 +148,8 @@ func TestObjectsAreEqualValues(t *testing.T) {
 		{uint32(10), int32(10), true},
 		{0, nil, false},
 		{nil, 0, false},
-		{now, now.In(time.Local), true}, // should be time zone independent
-		{int(270), int8(14), false},     // should handle overflow/underflow
+		{now, now.In(time.Local), false}, // should not be time zone independent
```
arjunmahishi, line 151: Did you mean to add the test here or in `TestObjectsAreEqual`. Although it fit in here also.

arjunmahishi (commented):
Code changes LGTM

---

While you are at it, you might as well fix this:
https://github.com/stretchr/testify/blob/7b3de0842561fd516ab69ed866bc50c95a48e664/assert/assertions_test.go#L123

I think this was meant to be 
```go
	{time.Now(), time.Now(), false},
```

---

> we should support testing for state first

Can you explain this please?

arjunmahishi (approved):

Antonboom:
offtop: please introduce more obvious API or improve documentation – when people should use `assert.Equal`, when `assert.True(t, t1.Equal(t2)`, when `assert.EqualUTCTime(...)` (or similar)

brackendawson:
> While you are at it, you might as well fix this:
> 
> https://github.com/stretchr/testify/blob/7b3de0842561fd516ab69ed866bc50c95a48e664/assert/assertions_test.go#L123
> 
> I think this was meant to be
> 
> ```go
> 	{time.Now(), time.Now(), false},
> ```
> 
> > we should support testing for state first
> 
> Can you explain this please?

Coincidence, I think. That's saying that functions are not comparable, it just happens that the author chose time.Now. If we did change it to your suggestion then the test would flake if the two calls were made in the same nanosecond [which seems likely](https://go.dev/play/p/AVWOlZZy76e).

brackendawson:
> offtop: please introduce more obvious API or improve documentation – when people should use `assert.Equal`, when `assert.True(t, t1.Equal(t2)`, when `assert.EqualUTCTime(...)` (or similar)

Thanks, I think #1078 might be the best approach here? Make the failure easier to understand, rather than clutter the Equal doc with information about something you may not be trying to do.

Also for nested time.Time objects something along the lines of #843 might be a good approach?

PR #1546, opened 2024-02-22
assert: switch the diff library to go-diff

mitioshi:
## Summary
This PR changes the diff library used in the assert package to go-diff.

## Changes
the library used for diffing was replaced to go-diff/diffmatchpatch which is actively maintained and enables us to access structured diffs.

## Motivation
See the [parent PR](https://github.com/stretchr/testify/pull/1480) for more context

`difflib` is no longer maintained and cannot be extended to modify the diff structure. We end up outputting whatever the library spews without being able to modify it easily. Having the access to the diff structure gives us a more fine-grained control over what we show to users. 
The initial motivation for this PR was to add colorized diffs to `assert.Equal`. Accessing the diff structure would make this task trivial

## Related issues
Split from #1480 as per @dolmen's suggestion


assert/assertions.go
```diff
@@ -12,13 +12,15 @@ import (
...
 	"runtime"
 	"runtime/debug"
+	"strconv"
 	"strings"
 	"time"
 	"unicode"
 	"unicode/utf8"
 
+	"github.com/sergi/go-diff/diffmatchpatch"
+
```
dolmen, line 22: Remove this empty line.
```suggestion
```

assert/assertions.go
```diff
@@ -145,8 +147,6 @@ func copyExportedFields(expected interface{}) interface{} {
 // structures.
 //
 // This function does no assertion of any kind.
-//
-// Deprecated: Use [EqualExportedValues] instead.
```
dolmen, line 149: Why are you removing this?

Devourian:
Why its not merged yet?

sleeyax (approved):

stefanpenner:
late to the party but @mitioshi thanks for going down this path, complex diffs in tests in golang assertion libs could use some of TLC.

Devourian:
Merge it at all costs

dolmen:
From https://github.com/stretchr/testify/pull/1546#issuecomment-2647785759:

> Merge it at all costs

@Devourian I consider this to be abusive because not respectful of the benevolent work of maintainers who will have to endure the consequences of a such change.

dolmen (commented):
This patch is incomplete: difflib is not replaced in package `mock`. The result is that it adds a direct dependency (and many indirect dependencies, including circular dependency on older Testify), but doesn't remove the original dependency.

dolmen:
Our preference goes to #1708 which imports the frozen `difflib` into this module: there is a lower risk of regressions.

However, an alternative way would be to provide a mean for the user to plug his chosen alternate `difflib`, without having that dependency linked (via go.mod) in Testify itself. See how this is done for YAML in #1579.

PR #1556, opened 2024-03-01
Create errorAssertionFunction creators

JERHAV:
To clean up table driven test with often repeated functions we can add a simple function creator that generates a errorAssertionFunction

## Changes

Add creators for errorAssertionFunctions for ErrorIs and ErrorAs to be able to use these functions simmilar to NoError and Error in test tables

## Motivation

```errorAssertionFunc: require.CreateErrorIsAssertionFunc(testedpkg.ExpectedError)```

### CODE REVIEW

**Comment by brackendawson** [comment]
This is quite a terse issue, but I actually think I know what you mean because I've done it. I'd also like some high-order functions to create assertion functions for ErrorIs and ErrorAs which match the signature of Error and NoError.

Example:
```go
package kata_test

import (
	"testing"

	"github.com/brackendawson/kata"
)

func TestIfy(t *testing.T) {
	for name, test := range map[string]struct {
		expectErr func(assert.TestingT, error, ...interface{}) bool
	}{
		"should work": {
			expectErr: assert.NoError,
		},
		"should return any error": {
			expectErr: assert.Error,
		},
		"should return a specific error": {
			expectErr: errorIs(kata.SmellyError),
		},
	} {
			err := kata.Funk()
			test.expectErr(t, err)
		})
	}
}

// errorIs could be part of testify, same for errorAs
func errorIs(target error) func(assert.TestingT, error, ...interface{}) bool {
	return func(t assert.TestingT, err error, msgAndArgs ...interface{}) bool {
		return assert.ErrorIs(t, err, target, msgAndArgs...)
	}
}

```

brackendawson (changes requested):
I think this needs an example, somewhere, perhaps:
```go
func TestIfy(t *testing.T) {
	for name, test := range map[string]struct {
		expectErr assert.ErrorAssertionFunc
	}{
		"returns no error": {
			expectErr: assert.NoError,
		},
		"returns error wrapping io.EOF": {
			expectErr: assert.ErrorIsFor(io.EOF),
		},
		"retuens error wrapping fs.PathError": {
			expectErr: assert.ErrorAsFor(&fs.PathError{}),
		},
	} {
			err := Do()
			test.expectErr(t, err)
		})
	}
}
```

In practice, for the ErrorAs case I've also wanted to make assertions against the error's fields, can `ErrorAsFor` have a function argument for this?

```
func TestIfy(t *testing.T) {
	for name, test := range map[string]struct {
		expectErr assert.ErrorAssertionFunc
	}{
		"retuens error wrapping fs.PathError": {
			expectErr: assert.ErrorAsFor(func(t assert.TestingT, err fs.PathError) { // testify can infer the type from the function signature 
				assert.Equal(t, "filename.txt", err.Path)
			}),
		},
	} {
			err := Do()
			test.expectErr(t, err)
		})
	}
}
```

assert/assertions.go
```diff
@@ -48,6 +48,20 @@ type ErrorAssertionFunc func(TestingT, error, ...interface{}) bool
 // Comparison is a custom function that returns true on success and false on failure
 type Comparison func() (success bool)
 
+// CreateErrorIsAssertionFunc returns an [ErrorAssertionFunc] which tests if the error is a specific value.
+func CreateErrorIsAssertionFunc(expectedError error) ErrorAssertionFunc {
+	return func(t TestingT, err error, i ...interface{}) bool {
```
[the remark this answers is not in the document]
JERHAV: It can't call t.Helper, there is no helper in this interface.

require/requirements.go
```diff
@@ -26,4 +26,18 @@ type BoolAssertionFunc func(TestingT, bool, ...interface{})
 // for table driven tests.
 type ErrorAssertionFunc func(TestingT, error, ...interface{})
 
+// CreateErrorIsAssertionFunc returns an [ErrorAssertionFunc] which tests if the error is a specific value.
+func CreateErrorIsAssertionFunc(expectedError error) ErrorAssertionFunc {
+	return func(t TestingT, err error, i ...interface{}) {
```
dolmen, line 31: Call `t.Helper()`.

require/requirements.go
```diff
@@ -26,4 +26,18 @@ type BoolAssertionFunc func(TestingT, bool, ...interface{})
...
+		ErrorIs(t, err, expectedError, i...)
+	}
+}
+
+// CreateErrorIsAssertionFunc returns an [ErrorAssertionFunc] which tests if the error is a specific value.
+func CreateErrorAsAssertionFunc(expectedInterface interface{}) ErrorAssertionFunc {
+	return func(t TestingT, err error, i ...interface{}) {
```
dolmen, line 38: Call `t.Helper()`.

dolmen (changes requested):

require/requirements.go
```diff
@@ -26,4 +26,27 @@ type BoolAssertionFunc func(TestingT, bool, ...interface{})
 // for table driven tests.
 type ErrorAssertionFunc func(TestingT, error, ...interface{})
 
+// ErrorIsFunc returns an [ErrorAssertionFunc] which tests if the error wraps target.
+func ErrorIsFor(expectedError error) ErrorAssertionFunc {
+	return func(t TestingT, err error, i ...interface{}) {
```
dolmen, line 31: Rename `i` to `msgAndArgs`.

dolmen (changes requested):

dolmen:
I'm not convinced by the need for `ErrorAsFor`. I need to see a real use case to be convinced.

JERHAV:
I agree with the ErrorAsFor being less useful, but sometimes you have a function that returns something like:

`type ErrNotOKStatusCode struct{StatusCode int}`,
 in which case if I then want to use this error in a higher up function I'd test if it is the correct interface.

JERHAV:
I do not exactly know what you mean, I have added a simple type assertion function in the fixup commit. But I'm not sure that is what you mean.

assert/assertions.go
```diff
@@ -48,6 +48,40 @@ type ErrorAssertionFunc func(TestingT, error, ...interface{}) bool
...
+		if h, ok := t.(tHelper); ok {
+			h.Helper()
+		}
+
+		return ErrorAs(t, err, target, msgAndArgs...)
+	}
+}
+
+// ErrorTypeFor returns an [ErrorAssertionFunc] which tests whether the type of the error matches the targets type.
+func ErrorTypeFor(target interface{}) ErrorAssertionFunc {
```
dolmen, line 75: `target` should be of type `error`, not any.

```suggestion
func ErrorTypeFor(target error) ErrorAssertionFunc {
```

require/requirements.go
```diff
@@ -26,4 +26,38 @@ type BoolAssertionFunc func(TestingT, bool, ...interface{})
...
+		if h, ok := t.(tHelper); ok {
+			h.Helper()
+		}
+
+		ErrorAs(t, err, expectedInterface, msgsAndArgs...)
+	}
+}
+
+// ErrorTypeFor returns an [ErrorAssertionFunc] which tests whether the type of the error matches the targets type.
+func ErrorTypeFor(target interface{}) ErrorAssertionFunc {
```
dolmen, line 53:
```suggestion
func ErrorTypeFor(target error) ErrorAssertionFunc {
```

require/requirements_test.go
```diff
@@ -665,14 +665,26 @@ func ExampleErrorAssertionFunc() {
...
 func TestErrorAssertionFunc(t *testing.T) {
+	var testError = errors.New("test error")
 	tests := []struct {
 		name      string
 		err       error
 		assertion ErrorAssertionFunc
 	}{
 		{"noError", nil, NoError},
 		{"error", errors.New("whoops"), Error},
+		{"errorIs", testError, ErrorIsFor(testError)},
```
dolmen, line 685: Add a test case with `testError` wrapped: `fmt.Errorf("wrapped %w", testError)`

require/requirements_test.go
```diff
@@ -665,14 +665,26 @@ func ExampleErrorAssertionFunc() {
...
+		{"errorAs", testingError{extraInfo: "something"}, ErrorAsFor(&testingError{})},
```
dolmen, line 686: Add a test case with a wrapped `testingError`.

dolmen (changes requested):

dolmen:
@brackendawson Here is a rewrite of [your example of ErrorAsFor](https://github.com/stretchr/testify/pull/1556#issuecomment-1977259141) that I find much more readable (it doesn't require to know the special function `ErrorAsFor` that this PR adds):

```go
func TestIfy(t *testing.T) {
	for name, test := range map[string]struct {
		expectErr assert.ErrorAssertionFunc
	}{
		"returns error wrapping fs.PathError": {
			expectErr: func(t TestingT, err error, _ ...interface{}) bool {
				var errPath fs.PathError
				assert.ErrorAs(t, err, &errPath)
				assert.Equal(t, "filename.txt", errPath.Path)
			}),
		},
	} {
			err := Do()
			test.expectErr(t, err)
		})
	}
}
```

brackendawson:
Yes, that's a better approach. Let's use this PR to add ErrorIsFor, because this function increases readability of people's tests.

We can do ErrorAsFor later if there is a good API design. The implementation here does not encourage good testing practice because it can only be used to check that an `error` in `err`'s chain is assignable to `target`. In good practice a developer testing for state should assert that the fields or behaviour of that error are also as intended.

brackendawson (approved):
LGTM

segogoreng:
Hi @JERHAV @brackendawson @dolmen 
Do we still want these changes?

PR #1558, opened 2024-03-03
Added support for getting the arguments of mock function call for a specific call count

mahigadamsetty:
## Summary
The purpose of this change is to enable asserting the arguments of a mock function for a specific call count.

## Changes
Updated the Mock struct to hold the map of methodName to a slice of calls instead of just a slice of Calls, this will be used to extract the calls for a given methodName easily 

## Motivation
For instance, when a method is invoked multiple times within the code, and there's a need to verify the arguments for a particular call count, this enhancement proves beneficial.

```
actualObject.ProcessData("data1")
actualObject.ProcessData("data2")
```

with the new API, can assert the argument for a second call like this .

```
returnArgs := mockedObject.ArgsForCallCount(t, "ProcessData", 1)
if assert.Equal(t, 1, len(returnArgs)) {
     assert.Equal(t, "data2", returnArgs[0])			
}
```

## Related issues
N/a

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -282,7 +282,7 @@ type Mock struct {
 	ExpectedCalls []*Call
 
 	// Holds the calls that were made to this mocked object.
-	Calls []Call
+	Calls map[string][]Call
 
 	// test is An optional variable that holds the test struct, to be used when an
 	// invalid mock call was made.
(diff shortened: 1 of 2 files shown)


mock/mock.go
```diff
@@ -282,7 +282,7 @@ type Mock struct {
 	ExpectedCalls []*Call
 
 	// Holds the calls that were made to this mocked object.
-	Calls []Call
+	Calls map[string][]Call
```
brackendawson, line 285: This is a breaking change.

mock/mock.go
```diff
@@ -712,6 +727,24 @@ func (m *Mock) IsMethodCallable(t TestingT, methodName string, arguments ...inte
 	return false
 }
 
+// ArgsForCallCount returns the arguments of a function for a specific call count(0 based index)
+func (m *Mock) ArgsForCallCount(t TestingT, methodName string, count int) Arguments {
+	fmt.Println("ArgsForCallCount")
+	fmt.Println(m.Calls)
```
brackendawson, line 733: Do not print.

brackendawson (changes requested):
In general, I do not think this API should be added because it does not encourage best testing practices. If testing for state then a mock can be prepared with the expected calls and AssertExpectations used.

Can you demonstrate a specific test case which can't be written using the existing API?

It's also quite unclear what a call "count" is in this context.

PR #1559, opened 2024-03-03
assert: handle TokenTooLong error scenario

### LINKED ISSUE #1525: assert/require.Len doesn't print anything if the slice is too long
Hey there, I encountered this when attempting to use `require.Len` as shown below.

```go
package tester_test

import (
	"testing"

)

func TestVeryLongSlice(t *testing.T) {
	var bigSlice []string
	for i := 0; i < 20000; i++ {
		bigSlice = append(bigSlice, "hello")
	}

	require.Len(t, bigSlice, 10)
}
```

Output of test:

```
=== RUN   TestVeryLongSlice
    tester_test.go:15:
                Error Trace:    C:/Users/Fots/source/tester/tester/tester_test.go:15
                Error:
                Test:           TestVeryLongSlice
--- FAIL: TestVeryLongSlice (0.00s)
FAIL
FAIL    github.com/fgimian/tester/tester        0.150s
FAIL
```

I haven't quite figured out why this happens, although I can imagine the output would be impractically long anyway. Worst of all, the main assertion message is not displayed which is quite confusing (especially to someone new using the library, like me 😄).

Personally, I think that it would be best not to display the slice at all for the `Len` function, and instead just display:

```
Should have X item(s), but has Y
```

However, I suspect my view on that may not be what some people want. The reason I think it is unwise to print the slice in any form, is that typically when many people are testing for the length of a slice, they may be comparing to a relatively large slice of items (e.g. the output of a mocked API call or similar).

An additional point that I believe has been discussed before is the inconsistency with argument ordering for the `Len` function which takes the `actual` first and `expected` second.

Cheers
Fotis

arjunmahishi:
## Summary

As pointed out in #1525, when the assertion message is too long, it gets completely truncated in the final output. This is because `bufio.Scanner.Scan()` has a default `MaxScanTokenSize` set to `65536` characters (64 * 1024). The `Scan()` function returns false whenever the line being scanned exceeds that max limit. This leads to the final assertion message being truncated.

## Changes

This commit fixes that by manually setting the internal scan buffer size to `len(message) + 1` to make sure that the above scenario never occurs.

## Related issues

Fixes #1525


assert/assertions_test.go
```diff
@@ -3192,3 +3192,45 @@ func TestErrorAs(t *testing.T) {
...
+			expected:        "Hello\n\t            \tWorld",
+		},
+		{
+			name:            "single line - extreamly long",
+			msg:             strings.Repeat("hello ", 20000),
+			longestLabelLen: 11,
+			expected:        strings.Repeat("hello ", 20000),
+		},
+		{
+			name:            "multi line - extreamly long",
```
hendrywiranto, line 3222: just a nitpick
```suggestion
			name:            "single line - extremely long",
			msg:             strings.Repeat("hello ", 20000),
			longestLabelLen: 11,
			expected:        strings.Repeat("hello ", 20000),
		},
		{
			name:            "multi line - extremely long",
```

arjunmahishi: Done. Thanks

assert/assertions.go
```diff
@@ -301,13 +301,17 @@ func messageFromMsgAndArgs(msgAndArgs ...interface{}) string {
 func indentMessageLines(message string, longestLabelLen int) string {
 	outBuf := new(bytes.Buffer)
 
-	for i, scanner := 0, bufio.NewScanner(strings.NewReader(message)); scanner.Scan(); i++ {
-		// no need to align first line because it starts at the correct location (after the label)
-		if i != 0 {
-			// append alignLen+1 spaces to align with "{{longestLabel}}:" before adding tab
+	msgScanner := bufio.NewScanner(strings.NewReader(message))
+	msgScanner.Buffer([]byte{}, len(message)+1)
```
dolmen, line 311: Set the buffer to `nil` instead of a useless zero-length buffer (`[]byte{}`) that will have to be reallocated in all cases.

```go
msgScanner.Buffer(nil, len(message)+1)
```

assert/assertions_test.go
```diff
@@ -3192,3 +3192,45 @@ func TestErrorAs(t *testing.T) {
 		})
 	}
 }
+
+func Test_indentMessageLines(t *testing.T) {
```
dolmen, line 3196: Move this new test to a separate file `assertions_priv_test.go`: we aim to move `assertions_test.go` to package `assert_test` and this new test will get in the way for that project.

dolmen (changes requested):

assert/assertions_test.go
arjunmahishi: > we aim to move assertions_test.go to package assert_test

Is this documented/discussed anywhere? Would love to get more context.

arjunmahishi: Made that change

dolmen: @arjunmahishi No, I've not discussed this anywhere I remember. However, to ease maintenance I think it would be better to have clear separation of unit tests written to check implementation details from higher level unit tests that test the Testify external API.

A related effort is #1431 by @brackendawson. While I approve the principle of that effort, I think we must first clear most of the PR queue to avoid missing a valuable pending bug fix that would be broken by such a major refactor. I have started co-maintenance in July 2023, but so far I haven't been able to review _all_ issues and PRs (those without labels haven't been reviewed).

dolmen (changes requested):

assert/assertions_priv_test.go
```diff
@@ -0,0 +1,48 @@
...
+		},
+		{
+			name:            "single line - extremely long",
+			msg:             strings.Repeat("hello ", 20000),
+			longestLabelLen: 11,
+			expected:        strings.Repeat("hello ", 20000),
+		},
+		{
+			name:            "multi line - extremely long",
+			msg:             strings.Repeat("hello\n", 20000),
```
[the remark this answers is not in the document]
arjunmahishi: Can you explain the 2nd point? do we want to check that 20_000 is greater than `MaxScanTokenSize` in case that is change in the package?

dolmen (changes requested):

assert/assertions_priv_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"bufio"
+	"strings"
+	"testing"
+)
+
+const (
+	// set maxScanTokenSize to 1 more than the default set by bufio. This will
+	// cover the case where a single line is longer than the default
+	// maxScanTokenSize
+	maxScanTokenSize = bufio.MaxScanTokenSize + 1
```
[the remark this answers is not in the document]
arjunmahishi: Done

assert/assertions.go
```diff
@@ -301,13 +301,24 @@ func messageFromMsgAndArgs(msgAndArgs ...interface{}) string {
...
+	msgScanner.Buffer(nil, len(message)+1)
+
+	first := true
+	for msgScanner.Scan() {
+		if !first {
 			outBuf.WriteString("\n\t" + strings.Repeat(" ", longestLabelLen+1) + "\t")
 		}
-		outBuf.WriteString(scanner.Text())
+
+		outBuf.WriteString(msgScanner.Text())
```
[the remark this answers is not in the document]
arjunmahishi: Good catch. Thanks.
Done

arjunmahishi:
@dolmen Are we good to merge?

dolmen:
I'm not yet convinced that the tests provide enough coverage. I would like to see more clearly a check related to bufio boundaries.

I feel that we are testing with strings much longer than the boundary, and not just below and at the boundary.
I have to dig into the code more carefully.

assert/assertions.go
```diff
@@ -301,13 +301,24 @@ func messageFromMsgAndArgs(msgAndArgs ...interface{}) string {
...
+
+	first := true
+	for msgScanner.Scan() {
+		if !first {
 			outBuf.WriteString("\n\t" + strings.Repeat(" ", longestLabelLen+1) + "\t")
 		}
-		outBuf.WriteString(scanner.Text())
+
+		outBuf.Write(msgScanner.Bytes())
+		first = false
```
dolmen, line 321: Move this line in the `if !first` block as it will have no effect on the subsequent iterations.

assert/assertions_priv_test.go
```diff
@@ -0,0 +1,77 @@
...
+			name:            "multi line - over the bufio default limit",
+			msg:             strings.Repeat("hello\n", bufio.MaxScanTokenSize+10),
+			longestLabelLen: 3,
+			expected: strings.TrimSpace(
+				strings.TrimPrefix(strings.Repeat("\thello\n\t    ", bufio.MaxScanTokenSize+10), "\t"),
+			),
+		},
+		{
+			name:            "single line - just under the bufio default limit",
+			msg:             strings.Repeat("hello ", bufio.MaxScanTokenSize-10),
```
dolmen, line 44: The boundary is not `"hello"` repeated `bufio.MaxScanTokenSize` times, but a string which has a line whose length is `bufio.MaxScanTokenSize` bytes.

We probably need a utility function to build such an input string. At least `strings.Repeat` alone doesn't help.

dolmen (changes requested):

assert/assertions.go
```diff
@@ -301,13 +301,24 @@ func messageFromMsgAndArgs(msgAndArgs ...interface{}) string {
...
+	// any additional allocations. We set the maximum buffer length to 1 more
+	// than the length of the message to avoid exceeding the default
+	// MaxScanTokenSize while scanning lines. This can happen when there is a
+	// single very long line. Refer to issue #1525
+	msgScanner.Buffer(nil, len(message)+1)
+
+	first := true
+	for msgScanner.Scan() {
+		if !first {
 			outBuf.WriteString("\n\t" + strings.Repeat(" ", longestLabelLen+1) + "\t")
```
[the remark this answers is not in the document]
arjunmahishi: Good point. Done

assert/assertions_priv_test.go
arjunmahishi: Ohh! that's right. 

Changed the approach a little bit. The new commit generates the input as you suggested. And instead of matching against an exact output, it asserts the pattern of the output. Like the leading white spaces etc. Because for generating the expected output, we will end up rewriting the actual function logic in the test.

assert/assertions.go
```diff
@@ -300,14 +300,25 @@ func messageFromMsgAndArgs(msgAndArgs ...interface{}) string {
...
-			outBuf.WriteString("\n\t" + strings.Repeat(" ", longestLabelLen+1) + "\t")
+	msgScanner := bufio.NewScanner(strings.NewReader(message))
+
+	// a buffer is set manually to store the tokenization of the message string
+	// so that we can re-use the same buffer for each line of the message without
+	// any additional allocations. We set the maximum buffer length to 1 more
+	// than the length of the message to avoid exceeding the default
+	// MaxScanTokenSize while scanning lines. This can happen when there is a
+	// single very long line. Refer to issue #1525
+	msgScanner.Buffer(nil, len(message)+1)
```
brackendawson, line 311: Is this safe? As we know from the assert.Len case this can contain a very large formatted slice. Having more than doubled the memory consumed by the large slice by creating a string containing the formatted version of it. Do we want to double the formatted version again? Plus all the intermediate buffers that were allocated on the heap by msgScanner before the next GC cycle?

We could at least pre-allocate the buffer for msgScanner rather than passing in `nil`. But I'd prefer that we either split or truncate lines that are unreasonably long, or we could just truncate the entire message to less than `bufio.MaxScanTokenSize`?

arjunmahishi: > We could at least pre-allocate the buffer for msgScanner rather than passing in nil.

Agreed. I've made this change.

> But I'd prefer that we either split or truncate lines that are unreasonably long, or we could just truncate the entire message to less than `bufio.MaxScanTokenSize`?

If we decide to truncate, we could probably truncate based on the dimensions of the terminal? This makes more sense from a UX point of view.

brackendawson: Hey, @arjunmahishi. I really think we should truncate rather than allocate unknown amounts of memory. It turns out `Equal` actually already uses a function to truncate its values. Sorry to steal the issue but I've opened #1646 with how I think it could be done. What do you think?

arjunmahishi: Cool. Better to reuse the existing functionality and keep it consistent. Closing this now.

PR #1561, opened 2024-03-04
deps: downgrade objx to v0.5.1

dolmen:
## Summary
objx `v0.5.2` unnecessarly upgrades the minimum Go version from 1.13 to 1.20. `v0.5.1` is just enough.

## Changes
* This reverts commit 814075f391adffd2bf2b5110a74c51827ba132c4 (#1552).

## Motivation

objx `v0.5.1` is just enough. But stretchr/objx#149 changed the minimum Go version from 1.13 to 1.20.

We don't want to force users to upgrade Go to use the latest testify version while we don't use specific features of those later versions.

## Related issues
Revert #1552.
See stretchr/objx#149.


brackendawson:
v1.17 builds all passed, what is the actual minimum version this imposes on users?

brackendawson (changes requested):
I believe this is not necessary.

Because a go directive before 1.21 is advisory and objx set it to 1.20 (for no reason), it is not possible for the configuration in v1.9.0 to generate an error. Testify's advisory 1.17 takes precedence, for what that is worth.

If objx set it to 1.22, then 1.21 toolchains would refuse to import testify v1.9.0 and our `oldstable` and `1.21` builds would have failed. This would be a good reason to keep every version between `1.17` and `stable` in our tests.

PR #1562, opened 2024-03-04
assert: make tHelper a type alias

dolmen:
## Summary
The tHelper interface is defined only for internal usage. We don't need a "hard" type for it as we don't define methods on it. Let's make is just a alias to the anonymous interface it represents.

Note: we are already using type aliases elswhere, and type aliases were introduced a long time ago in Go 1.9.

## Changes
* assert: `tHelper` becomes a type alias
* require: `tHelper` becomes a type alias

## Motivation
The `tHelper` symbol is not needed at runtime. This should reduce resource usage (yes, negligible).

## Related issues
* #1422 is also related to the `interface { Helper() }` type, but not dependent

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1872,7 +1872,7 @@ var spewConfigStringerEnabled = spew.ConfigState{
 	MaxDepth:                10,
 }
 
-type tHelper interface {
+type tHelper = interface {
 	Helper()
 }
 
--- a/require/requirements.go
+++ b/require/requirements.go
@@ -6,7 +6,7 @@ type TestingT interface {
 	FailNow()
 }
 
-type tHelper interface {
+type tHelper = interface {
 	Helper()
 }
 


brackendawson (approved):
You would normally do this to facilitate moving/combining previously declared types, which we aren't. If there is a performance difference, it is too small to measure. But I have no other reason to object to this.

<details>
<summary>Benchmarks</summary>

```go
type impl struct{}

func (impl) Helper() {
	_ = math.Abs(-1)
}

func BenchmarkTypeAssertionWithDeclaredType(b *testing.B) {
	type helper interface {
		Helper()
	}

	var t any = &impl{}

	for i := 0; i < b.N; i++ {
		if t, ok := t.(helper); ok {
			t.Helper()
		}
	}
}

func BenchmarkTypeAssertionWithTypeAlias(b *testing.B) {
	type helper = interface {
		Helper()
	}

	var t any = &impl{}

	for i := 0; i < b.N; i++ {
		if t, ok := t.(helper); ok {
			t.Helper()
		}
	}
}
```

```
goos: darwin
goarch: arm64
pkg: github.com/brackendawson/kata
BenchmarkTypeAssertionWithDeclaredType-10       525305594                2.220 ns/op           0 B/op          0 allocs/op
BenchmarkTypeAssertionWithTypeAlias-10          544804904                2.235 ns/op           0 B/op          0 allocs/op
PASS
ok      github.com/brackendawson/kata   2.963s
```

</details>

dolmen:
@brackendawson There is no performance improvement expected. The resource usage I claim is just the absence of the type definition in the final binary, so a smaller binary.

Compare the output of this command on this branch vs `master`:
```console
$ (cd assert ; go test -c && strings assert.test | grep BoolAssertionFunc)
```

PR #1337, opened 2023-02-01
Add PanicAssertionFunc

fahimbagar:
## Summary
Add a `PanicAssertionFunc` to ease writing table-driven tests for panic assertion.

## Changes
- Add `PanicAssertionFunc` for table driven.

## Motivation
- My team uses panic assertion a lot :smile: 
- Useful for building panic assertion
- There is an issue reported for this one as well: #730

```go
func ExamplePanicAssertionFunc() {
	t := &testing.T{} // provided by test

	tests := []struct {
		name      string
		panicFn   func()
		assertion PanicAssertionFunc
	}{
		{"with panic", func() { panic(nil) }, Panics},
		{"without panic", func() {}, NotPanics},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			tt.assertion(t, tt.panicFn)
		})
	}
}
```
https://github.com/fahimbagar/testify/blob/cb954e70769531b0462ae2f713c12c74e64ad69c/assert/assertions_test.go#L2410-L2427

## Related issues
Closes #730

PR #1563, opened 2024-03-04
assert: make *AssertionFunc types just aliases

dolmen:
## Summary
We don't need to define "hard" types for `*AssertionFunc` types as we don't attach methods to them. Also, making them just type aliases will give more flexibility to users of our API.
So [ComparisonAssertionFunc](https://pkg.go.dev/github.com/stretchr/testify/assert#ComparisonAssertionFunc) and other `*AssertionFunc` types are now just type aliases.

## Changes
Make [`ComparisonAssertionFunc`](https://pkg.go.dev/github.com/stretchr/testify/assert#ComparisonAssertionFunc), [`ValueAssertionFunc`](https://pkg.go.dev/github.com/stretchr/testify/assert#ValueAssertionFunc), [`BoolAssertionFunc`](https://pkg.go.dev/github.com/stretchr/testify/assert#BoolAssertionFunc) and [`ErrorAssertionFunc`](https://pkg.go.dev/github.com/stretchr/testify/assert#ErrorAssertionFunc) just type aliases instead of "hard" types.

## Motivation
* More flexibility for API users
* Less resource usage at runtime (type aliases are used only for compile step) (yes, negligible difference)

## Related issues
* #1337
* #1562

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -32,19 +32,19 @@ type TestingT interface {
 
 // ComparisonAssertionFunc is a common function prototype when comparing two values.  Can be useful
 // for table driven tests.
-type ComparisonAssertionFunc func(TestingT, interface{}, interface{}, ...interface{}) bool
+type ComparisonAssertionFunc = func(TestingT, interface{}, interface{}, ...interface{}) bool
 
 // ValueAssertionFunc is a common function prototype when validating a single value.  Can be useful
 // for table driven tests.
-type ValueAssertionFunc func(TestingT, interface{}, ...interface{}) bool
+type ValueAssertionFunc = func(TestingT, interface{}, ...interface{}) bool
 
 // BoolAssertionFunc is a common function prototype when validating a bool value.  Can be useful
 // for table driven tests.
-type BoolAssertionFunc func(TestingT, bool, ...interface{}) bool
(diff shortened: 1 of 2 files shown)


arjunmahishi (approved):
LGTM. The only concern is, is this a backward compatible change? I believe type assertions work differently with type aliases and type definitions?

brackendawson:
> LGTM. The only concern is, is this a backward compatible change? I believe type assertions work differently with type aliases and type definitions?

I believe the only observable change is the type name. I can fabricate code that would be impacted by this change:
```go
func TestIfy(t *testing.T) {
	var f assert.ErrorAssertionFunc
	assert.Equal(t, reflect.TypeOf(f).String(), "assert.ErrorAssertionFunc")
}
```

But it's extraordinarily unlikely anyone relies on that. Whether this is a breaking change is up to our opinion and I have seen Go make more noticeable changes in the standard library (the new optional `Unwrap() []error` method on errors for example).

I would make this change if it were necessary, since it's not strictly necessary for us I would personally not make this change in v1. I'm not going to block the PR though, I will leave the ultimate decision up to consensus of the maintainers.

MovieStoreGuy (approved):

PR #1569, opened 2024-03-07
mock: document more alternatives to deprecated AnythingOfTypeArgument

dolmen:
## Summary
Follow-up to #1441: more doc for deprecated [`mock.AnythingOfTypeArgument`](https://pkg.go.dev/github.com/stretchr/testify/mock#AnythingOfTypeArgument).

## Changes
Improve go doc.

## Motivation
I had a look to usages of [`mock.AnythingOfTypeArgument`](https://pkg.go.dev/github.com/stretchr/testify/mock#AnythingOfTypeArgument) in the wild on GitHub:
https://github.com/search?q=mock.AnythingOfTypeArgument+language%3Ago+NOT+repo%3Astretchr%2Ftestify&type=code

This change should give more guidance for fixing the issue in downstream projects.

## Related issues
* #1441, #1434
* mattermost/mattermost-plugin-confluence#108, mattermost/mattermost-plugin-cloud#154, mattermost/mattermost-plugin-jira#1030

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -766,7 +766,25 @@ const (
 // AnythingOfTypeArgument contains the type of an argument
 // for use when type checking.  Used in Diff and Assert.
 //
-// Deprecated: this is an implementation detail that must not be used. Use [AnythingOfType] instead.
+// Deprecated: this is an implementation detail that must not be used. Use the [AnythingOfType] constructor instead.
+//
+// If AnythingOfTypeArgument is used as a function return value type you can usually replace
+// the function with a variable. Example:
+//
+//	func anyString() mock.AnythingOfTypeArgument {
+//	    return mock.AnythingOfType("string")
+//	}
+//
+// Can be replaced by:
+//
+//	func anyString() interface{} {
+//	    return mock.IsType("")
+//	}
+//
+// It could even be replaced with a variable:
+//
+//	var anyString = mock.AnythingOfType("string")
+//	var anyString = mock.IsType("") // alternative
 type AnythingOfTypeArgument = anythingOfTypeArgument
 
 // anythingOfTypeArgument is a string that contains the type of an argument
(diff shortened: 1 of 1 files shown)


mock/mock.go
```diff
@@ -766,7 +766,25 @@ const (
...
+// Can be replaced by:
+//
+//	func anyString() interface{} {
+//	    return mock.IsType("")
+//	}
+//
+// It could even be replaced with a variable:
+//
+//	var anyString = mock.AnythingOfType("string")
+//	var anyString = mock.IsType("") // alternative
```
brackendawson, line 787: This suggests to replace with a variable, then shows an example which does not do that. Then it suggests again that it can be replaced with a variable and shows two examples.

There are three uses of the type previously, right?

```go
// conversion is replaced by constructor

```


```suggestion
// If AnythingOfTypeArgument is used as a function return value type you can usually replace
// the function with a variable. Example:
//
//	func anyString() mock.AnythingOfTypeArgument {
//	    return mock.AnythingOfType("string")
//	}
//
// Can be replaced by:
//
//	var anyString = mock.AnythingOfType("string")
```

brackendawson, line 787: This suggests to replace with a variable, then shows an example which does not do that. Then it suggests again that it can be replaced with a variable and shows two examples.

This also suggests IsType as an alternative, which I agree is a better matcher, but it's not what the deprecation line says to use. It might also not be as easy to switch to for people with tabled tests that already define the type by its string name.

I rendered this with pkgsite and with the amount of code blocks it becomes hard to read, especially with the "bad" code examples present, users might take them as a "good" example.

How about something more simple?

```suggestion
// Deprecated: this is an implementation detail that must not be used. Use the [AnythingOfType] constructor instead, example:
//
//	m.On("Do", mock.AnythingOfType("string"))
//
// All explicit type declarations can be replaced with interface{} as is expected by [Mock.On], example:
//
//	func anyString interface{} {
//		return mock.AnythingOfType("string")
//	}
```

Which will look like:
image

brackendawson (changes requested):

brackendawson (approved):

Issue #1532, opened 2024-02-13
Fix CVE-2022-28948 - Remove `gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c`


vmatyusGitHub:
Github Advisor reported a vulnerable package: `gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c`
Here is the [CVE report](https://github.com/advisories/GHSA-hp87-p4gw-j4gq).

One of my application uses`testify` package as dependency, in the current setup my application is vulnerable, this is why I am asking from you to correct this vulnerability.

I checked the dependency usage in the following way:

```
go mod graph | grep "gopkg.in/yaml.v3@v3.0.0-20200313102051-9f266ea9e77c"
github.com/stretchr/testify@v1.7.1 gopkg.in/yaml.v3@v3.0.0-20200313102051-9f266ea9e77c

go mod graph | grep "github.com/stretchr/testify@v1.7.1"                 
github.com/stretchr/objx@v0.4.0 github.com/stretchr/testify@v1.7.1

go mod graph | grep "github.com/stretchr/objx@v0.4.0"   
github.com/stretchr/testify@v1.8.0 github.com/stretchr/objx@v0.4.0

go mod graph | grep "github.com/stretchr/testify@v1.8.0"
github.com/stretchr/objx@v0.5.0 github.com/stretchr/testify@v1.8.0

go mod graph | grep "github.com/stretchr/objx@v0.5.0"   
github.com/stretchr/testify@v1.8.4 github.com/stretchr/objx@v0.5.0

go mod graph | grep "github.com/stretchr/testify@v1.8.4"
github.ibm.com/cloudant/rc-sync github.com/stretchr/testify@v1.8.4
github.com/stretchr/testify@v1.8.4 github.com/davecgh/go-spew@v1.1.1
github.com/stretchr/testify@v1.8.4 github.com/pmezard/go-difflib@v1.0.0
github.com/stretchr/testify@v1.8.4 github.com/stretchr/objx@v0.5.0
github.com/stretchr/testify@v1.8.4 gopkg.in/yaml.v3@v3.0.1
github.ibm.com/IAM/context-token@v0.2.3 github.com/stretchr/testify@v1.8.4
github.ibm.com/IAM/pep/v4@v4.2.1-release github.com/stretchr/testify@v1.8.4
github.ibm.com/IAM/token/v5@v5.2.5 github.com/stretchr/testify@v1.8.4
```

[...]

Issue #1292, opened 2022-10-31
Yaml3.0.0 vulnerability via objx v0.5.0


rohanthewiz:
testify@v1.8.1 includes objx v0.5.0 which includes testify@**v1.8.0** which eventually includes yaml.v3@v3.0.0 which has a Denial of Service [CWE 400](https://cwe.mitre.org/data/definitions/400.html) and a Null Pointer deference [CWE 476](https://cwe.mitre.org/data/definitions/476.html) vulnerability.

`go mod graph` (excerpts):

github.com/stretchr/testify@v1.8.1 github.com/stretchr/objx@v0.5.0

github.com/stretchr/objx@v0.5.0 github.com/stretchr/testify@v1.8.0

github.com/stretchr/testify@v1.8.0 github.com/stretchr/objx@v0.4.0

github.com/stretchr/objx@v0.4.0 github.com/stretchr/testify@v1.7.1

github.com/stretchr/testify@v1.7.1 gopkg.in/yaml.v3@v3.0.0-20200313102051-9f266ea9e77c

Perhaps a fix would be to tag the latest objx to say 0.5.1, update testify to point to that version and cut a new tag for testify that everyone can include.

Thanks.

joaocbarbosa:
I'm wondering if we really can fix this problem without changing the objx library or breaking the circular dependency somehow.
Even if we increase the testify version, pointing to a new objx tag (0.5.1 as example), the dependency tree will be like this:

github.com/stretchr/objx@v0.5.0 github.com/stretchr/testify@v1.8.2

github.com/stretchr/testify@v1.8.1 github.com/stretchr/objx@v0.5.1

github.com/stretchr/objx@v0.5.0 github.com/stretchr/testify@v1.8.1

github.com/stretchr/testify@v1.8.1 github.com/stretchr/objx@v0.5.0

github.com/stretchr/objx@v0.5.0 github.com/stretchr/testify@v1.8.0

github.com/stretchr/testify@v1.8.0 github.com/stretchr/objx@v0.4.0

github.com/stretchr/objx@v0.4.0 github.com/stretchr/testify@v1.7.1


It will be inside this loop until the vulnerable version.

rohanthewiz:
I am seeing the same issue. Because of the circular dependency, updating one package will still leave a reference to respective dependency, and so we go down the stairs back to the older versions with the vulnerabilities.
We would somehow have to update both at once...

brackendawson:
[...]

PR #1280, opened 2022-10-15
upgrade objx to fix vulnerability

joaocbarbosa:
## Summary
This PR is to upgrade the objx dependency from 0.4.0 to 0.5.0

## Changes
Changed `go.mod` and `go.sum` files

## Motivation
Since the objx package 0.5.0 was released with a vulnerability fix made by dependabot, we could use the newest version of it.
This also will fix a vulnerability found in objx 0.4.0, by using the yamlv3 3.0.0 indirectly throught testify 1.7.1.

**current dependency tree**
```
- objx v0.4.0
   - testify v1.7.1
      - yaml.v3 v3.0.0
```

**Vulnerability details:**
https://security.snyk.io/package/golang/github.com%2Fgo-yaml%2Fyaml
https://security.snyk.io/vuln/SNYK-GOLANG-GOPKGINYAMLV3-2952714

**What was added between 0.4.0 and 0.5.0**
https://github.com/stretchr/objx/compare/v0.4.0...v0.5.0

## Related issues
https://github.com/stretchr/objx/issues/124

Issue #1241, opened 2022-08-02
Please migrate to new Yaml Version V3


reachlakstar:
FORTIFY is throwing vulnerability with Yaml V2, can you please migrate to V3

brackendawson:
v1.8.0 does not use Yaml V2:
```
% go mod why -m "gopkg.in/yaml.v2"
# gopkg.in/yaml.v2
(main module does not need module gopkg.in/yaml.v2)
```
Any nancy is happy:
```
% go list -json -deps | nancy sleuth
Checking for updates...
Already up-to-date.
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Summary                     ┃
┣━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━┫
┃ Audited Dependencies    ┃ 4 ┃
┣━━━━━━━━━━━━━━━━━━━━━━━━━╋━━━┫
┃ Vulnerable Dependencies ┃ 0 ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━┻━━━┛
```

What version of testify are you using?

PR #1579, opened 2024-03-19
assert: make YAML dependency pluggable via build tags

dolmen:
## Summary

Make the YAML module dependency required for [`assert.YAMLEq`](https://pkg.go.dev/github.com/stretchr/testify/assert.YAMLEq) (and family) pluggable.

## Changes
* Add package `github.com/stretchr/testify/assert/yaml` that wraps the dependency. It exposes just `func Unmarshal([]byte, interface{}) error` (the only function imported from `gopkg.in/yaml.v3` for `assert`).
* Add 3 implementations of that package which are activated by build tags. The default implementation is the existing behavior and just wraps [`gopkg.in/yaml.v3.Unmarshal`](https://pkg.go.dev/gopkg.in/yaml.v3#Unmarshal).

## Motivation

The dependency on `gopkg.in/yaml.v3` for YAML deserialization (used by `assert.YAMLEq`) is causing painful maintenance work. Both for this project, and for downstream projects (end users). Unfortunately we can't just drop `assert.YAMLEq` until `v2`.

However this change allows to mitigate the issue by giving freedom to users to avoid the link constraints. This allows:
* avoid linking constraints of the license of [`gopkg.in/yaml.v3`](https://pkg.go.dev/gopkg.in/yaml.v3): see #1120
* should avoid future shower of irrelevant reports of vulnerabilities related to bugs in `gopkg.in/yaml.v3`: like previous #1192, #1193, #1194, #1280, #1241, #1292, #1532 (I wrote *irrelevant* because [selecting a particular version of a module](https://research.swtch.com/vgo-principles) is the final responsibility of users in downstream projects who assemble a program: downstream projects have all the infrastructure in the Go toolchain to handle such issues without harassing upstream projects' maintainers)

The `github.com/stretchr/testify/assert/yaml` implementation can be selected using build tags:
* `testify_yaml_default` (default): [`gopkg.in/yaml.v3`](https://pkg.go.dev/gopkg.in/yaml.v3) is used, like before
* `testify_yaml_fail`: YAML deserialization is not implemented and always fails. So `assert.YAMLEq` always fails. This is useful if the test suite package doesn't use `assert.YAMLEq` (very common case).
* `testify_yaml_custom`: the (new) `github.com/stretchr/testify/assert/yaml` package exposes an `Unmarshal` variable of type `func([]byte, any) error` (same as [`gopkg.in/yaml.v3.Unmarshal`](https://pkg.go.dev/gopkg.in/yaml.v3#Unmarshal)) that allows to plug any alternate implementation. For example [`github.com/goccy/go-yaml.Unmarshal`](https://pkg.go.dev/github.com/goccy/go-yaml#Unmarshal).

Usage:
```
go test -tags testify_yaml_fail
go test -tags testify_yaml_custom
```

To install an alternate implementation with `testify_yaml_custom` (as requested in #1120):

```go
//go:build testify_yaml_custom

package my_pkg_test

import (
	goyaml "github.com/goccy/go-yaml"
	"github.com/stretchr/testify/assert/yaml"
)

func init() {
	yaml.Unmarshal = goyaml.Unmarshal
}
```

## Related issues
* #1120

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -19,7 +19,9 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/pmezard/go-difflib/difflib"
-	"gopkg.in/yaml.v3"
+
+	// Wrapper around gopkg.in/yaml.v3
+	"github.com/stretchr/testify/assert/yaml"
 )
 
 //go:generate sh -c "cd ../_codegen && go build && cd - && ../_codegen/_codegen -output-package=assert -template=assertion_format.go.tmpl"
--- a/assert/yaml/yaml_custom.go
+++ b/assert/yaml/yaml_custom.go
@@ -1,0 +1,25 @@
+//go:build testify_yaml_custom && !testify_yaml_fail && !testify_yaml_default
+// +build testify_yaml_custom,!testify_yaml_fail,!testify_yaml_default
+
+// Package yaml is an implementation of YAML functions that calls a pluggable implementation.
+//
+// This implementation is selected with the testify_yaml_custom build tag.
+//
+//	go test -tags testify_yaml_custom
+//
+// This implementation can be used at build time to replace the default implementation
+// to avoid linking with [gopkg.in/yaml.v3].
+//
+// In your test package:
+//
+//		import assertYaml "github.com/stretchr/testify/assert/yaml"
+//
+//		func init() {
+//			assertYaml.Unmarshall = func (in []byte, out interface{}) error {
+//				// ...
+//	     			return nil
+//			}
+//		}
+package yaml
+
+var Unmarshal func(in []byte, out interface{}) error
--- a/assert/yaml/yaml_default.go
+++ b/assert/yaml/yaml_default.go
@@ -1,0 +1,37 @@
+//go:build !testify_yaml_fail && !testify_yaml_custom
+// +build !testify_yaml_fail,!testify_yaml_custom
+
+// Package yaml is just an indirection to handle YAML deserialization.
+//
+// This package is just an indirection that allows the builder to override the
+// indirection with an alternative implementation of this package that uses
+// another implemantation of YAML deserialization. This allows to not either not
+// use YAML deserialization at all, or to use another implementation than
+// [gopkg.in/yaml.v3] (for example for license compatibility reasons, see [PR #1120]).
+//
+// Alternative implementations are selected using build tags:
+//
+//   - testify_yaml_fail: [Unmarshal] always fails with an error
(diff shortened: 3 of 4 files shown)


dolmen:
Cc: @Al2Klimov as you proposed #1120. I think this should fit you use case.

Cirruslucent:
https://github.com/verb/kubernetes-enhancements/pull/17

brackendawson:
I am hesitant to use this approach. Are there any unforseen consequences? This will not remove yaml from the go.mod or that of any of module importing assert.

Yes, you can show that you don't actually use yaml in your build, but you can already show that you don't compile it in as you only use it in your test.

If the original reporter does actually have utility for this then I might be more swayed.

dolmen:
> Are there any unforseen consequences?

None.

> This will not remove yaml from the go.mod or that of any of module importing assert.

No.

> Yes, you can show that you don't actually use yaml in your build, but you can already show that you don't compile it in as you only use it in your test.

In #1120 I expect the requester didn't want even to have `gopkg.in/yaml.v3` linked when running tests as that would raise a licencing issue. The `-tags testify_yaml_fail` would enforce that. The user could use a `//go:build testify_yaml_fail` for all of its testsuite sources to ensure that his testsuite never runs with `gopkg.in/yaml.v3` linked.

dolmen:
Note that I have also serious concerns about the state of the maintenance of `gopkg.in/yaml.v3`. Here are 2 PR that I filled myself and that haven't yet got reviews:
* go-yaml/yaml#1027
* go-yaml/yaml#1028

dolmen:
Cc: @jasdel In project github.com/jmespath/go-jmespath [you vendored](https://github.com/jmespath/go-jmespath/commit/ff168ca20786696708c0f3eb2cc29d7912467599) Testify at v1.5.1 because of the upgrade of go-yaml from v2 to v3.
As the go-jmespath testsuite doesn't use YAML function at all, I think that this change could help you to drop that fork and just use the maintained version of Testify.
What do you think about it? Could you review the changes?

jasdel:
@dolmen thanks for the ping. I'm not a maintainer of that project anymore. But with that said, splitting the yaml as a non-required dependency would address the issue I ran into when originally filing the issue.

MovieStoreGuy (approved):

andig:
> Note that I have also serious concerns about the state of the maintenance of gopkg.in/yaml.v3. Here are 2 PR that I filled myself and that haven't yet got reviews:

@dolmen OT, but I've opened https://github.com/go-yaml/yaml/issues/1034

PR #1582, opened 2024-03-29
Fix time.Time compare

### LINKED ISSUE #1581: Integer overflow in time.Time comparison
## Description
time.Time comparison rely on https://pkg.go.dev/time#Time.UnixNano, but documentation explicitly said, that it has undefined behavior is data is bigger than year 2262 (since it not fit in int64)

## Step To Reproduce
```go
func TestTime(t *testing.T) {
	require.Less(t, time.Now(), time.Date(2263, 0, 0, 0, 0, 0, 0, time.Local))
}
```
Error:      	"2024-03-29 08:20:14.8719425 +0300 MSK m=+0.001135586" is not less than "2262-11-30 00:00:00 +0300 MSK"

myxo:
fix #1581

I was eager to use [Time.Compare](https://pkg.go.dev/time#Time.Compare), but it was added only in go 1.20

### DIFF
--- a/assert/assertion_compare.go
+++ b/assert/assertion_compare.go
@@ -328,7 +328,13 @@ func compare(obj1, obj2 interface{}, kind reflect.Kind) (compareResult, bool) {
 				timeObj2 = obj2Value.Convert(timeType).Interface().(time.Time)
 			}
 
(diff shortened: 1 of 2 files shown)


brackendawson (approved):

brackendawson:
LGTM

Issue #1584, opened 2024-04-11
EqualExportedValues can compare two slices


redachl:
## Description

`EqualExportedValues` does not handle slices for now, although `Equal` and `EqualValues` compare each element of 2 slices.

## Proposed solution

Removing the check that the provided variables are structs or pointers to structs here allows comparing two slices.
https://github.com/stretchr/testify/blob/352d2438b9da776308844714c38c0c9a16f3b5b6/assert/assertions.go#L629

## Use case

The implementation of EqualExportedValues can actually compare slices, if they are inside a struct.

Here is an example.

```go
func TestSomething(t *testing.T) {
	type sliceWrapper struct {
		ExportedSlice []int
	}
	a := sliceWrapper{ExportedSlice: []int{1, 2, 3}}
	b := sliceWrapper{ExportedSlice: []int{1, 2, 3}}

	assert.Equal(t, a.ExportedSlice, b.ExportedSlice)               // OK
	assert.EqualValues(t, a.ExportedSlice, b.ExportedSlice)         // OK
	assert.EqualExportedValues(t, a.ExportedSlice, b.ExportedSlice) // Types expected to both be struct or pointer to struct

	// Possible workaround: use a struct wrapper.
	assert.Equal(t, a, b)               // OK
	assert.EqualValues(t, a, b)         // OK
	assert.EqualExportedValues(t, a, b) // OK
}	
```

Antonboom:
The word `Exported` binds this method to structs, not to specific struct's field.

How does this relate to slices and why is this function expected to behave the same as `Equal` and `EqualValues`?

Sounds like nonsense.

redachl:
That's one way to look at it.

But I think it's also a pity that you _have_ to use a slice wrapper when testing a function that returns a slice.
Especially when it's a slice of proto messages, which with the latest code generation have unexported fields.

Another option is to add a new function, but I'm not sure it's the best thing to do.

Or we just accept using a slice wrapper is enough.

brackendawson:
For me it's key that the doc says *"objects"*, which is not defined in the go spec, and not *"structs"*. I think this leaves scope for nested structs (already supported) and structs nested in other container types.

PR #1586, opened 2024-04-11
assert.EqualExportedValues: accepts everything

redachl:
Remove type validation in EqualExportedValues

See #1584.


HaraldNordgren (approved):
Looks good! I wrote a big part of that function. For me, it was never intended to _not_ work for this use case.

Let's merge this @dolmen?

brackendawson (approved):
LGTM

I think the existing doc string for EqualExportedValues is approprioate. In fact I think this code is a more faithful implementation of the documented promise.

It works, I've tried to break it. It is well tested.

Going to allow a little time for other maintainers to review.

Issue #1585, opened 2024-04-11
Regexp accepts interface{}


kevinburkesegment:
The signature for [`assert.Regexp`](https://pkg.go.dev/github.com/stretchr/testify/assert#Regexp) (and associated methods) is:

```go
func Regexp(t TestingT, rx interface{}, str interface{}, msgAndArgs ...interface{}) bool {
```

However, these methods immediately call `matchRegexp`, which is this:

```go
func matchRegexp(rx interface{}, str interface{}) bool {

	var r *regexp.Regexp
	if rr, ok := rx.(*regexp.Regexp); ok {
		r = rr
	} else {
		r = regexp.MustCompile(fmt.Sprint(rx))
	}

	return (r.FindStringIndex(fmt.Sprint(str)) != nil)

}
```

Given this - wouldn't it make more sense for the signature to just be `*regexp.Regexp` and `string` ? The latter cast would likely help prevent unexpected behavior around the `fmt.Sprint` call.

dolmen:
Changing the signature of `assert.Regexp`, `assert.NotRegexp` (...) with stricter types would break compatibility. At this point detecting usage errors can only be made via a linter like testifylint.

ccoVeille:
> Changing the signature of `assert.Regexp`, `assert.NotRegexp` (...) with stricter types would break compatibility. At this point detecting usage errors can only be made via a linter like testifylint.

I agree with you about the compatibility issue if it was chanted, but what about changing the signature for testify v2?

Would it be acceptable?

ccoVeille:
For record, I noticed by myself the way to solve this in testify is way more complex than I thought

I'm talking about it here

https://github.com/Antonboom/testifylint/issues/81#issuecomment-2060642389

PR #1587, opened 2024-04-11
assert.Regexp: handle []byte array properly

kevinburkesegment:
## Summary

The regexp package works more efficiently on bytes; if you have bytes, it is easier to pass these directly to assert.Regexp than to convert them to a string first.

In addition, FindIndex/FindStringIndex are unnecessary here because we immediately throw away the result - let's just call Match() instead.

## Motivation

More efficient test calls, fewer casts in test code. 

## Related issues

Updates #1585.


dolmen (approved):
LGTM

MovieStoreGuy (approved):

PR #1588, opened 2024-04-12
Correct the EventuallyWithT and EventuallyWithTf example

JonCrowther:
## Summary
The example used for `EventuallyWithT` documentation doesn't actually pass.

## Changes
Updated the example to be correct by swapping the times used for `waitFor` and `tick`

## Motivation
I was looking at the docs and, based on the example, mistakenly set `waitFor` to be a shorter time than `tick`, which means `condition` never gets called. Hopefully this helps other people avoid the same mistake.


dolmen (approved):
LGTM

PR #1595, opened 2024-05-03
Assertion for minimum number of method/function invocations

shalinlk:
## Summary
Method for asserting minimum number of invocations made to a method.

## Changes
Added assertion method named `AssertMinNumberOfCalls` which takes expected minimum number of calls along with the method name for assertion. Assertion fails if specified minimum number of invocations are not made to the method.

## Motivation
There are cases where exact number of invocations of a method cannot be predetermined accurately. Eg, number of network calls made by a retry logic which is governed by time out and time delay. Number of calls made in such cases will depend on time of previous call and time out of governing context. Hence ability to ensure a minimum threshold is required

## Example usage
```
mockClient.AssertMinNumberOfCalls(t, "ModifyRule", 6)
```

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -648,6 +648,22 @@ func (m *Mock) AssertNumberOfCalls(t TestingT, methodName string, expectedCalls 
 	return assert.Equal(t, expectedCalls, actualCalls, fmt.Sprintf("Expected number of calls (%d) does not match the actual number of calls (%d).", expectedCalls, actualCalls))
 }
 
+// AssertMinNumberOfCalls asserts that the method was called at-least expectedMinCalls times.
+func (m *Mock) AssertMinNumberOfCalls(t TestingT, methodName string, expectedMinCalls int) bool {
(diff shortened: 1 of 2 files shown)


riyasyash (approved):

riyasyash (approved):

brackendawson:
How does this behave if the method has multiple calls? The doc on the new method doesn't say.

I would prefer not to expand the API with more methids like this. I don't think `AssertNumberOfCalls` should have been added because it has the same problem. It's better to set the expectations up front and use `AssertExpectations`. This is how you can cleanly assert a minimum number of calls up front:
```go
func TestMyClassDoThings(t *testing.T) {
	mockThing := &myMockThing{}
	mockThing.Test(t)
	defer mockThing.AssertExpectations(t)
	mockThing.On("ModifyRule").Return().Times(6)
	mockThing.On("ModifyRule").Return().Maybe()
	
	c := MyClass{
		Thing: mockThing,
	}
	c.DoThings()
}
```

Issue #1597, opened 2024-05-14
Mock code is prone to data races


petergardfjall:
## Description

Testify mocks are currently prone to data races when supplied a pointer argument (that may be concurrently updated elsewhere).

In several instances of our code, we have started seeing data races when running `go test -race` that originate from testify code. More specifically these data races occur whenever a mock pointer argument is concurrently modified. The reason being that `Arguments.Diff` uses the `%v` format specifier to get a presentable string for the argument. This also traverses the pointed-to data structure, which would lead to the data race.

We did not see these issues earlier, but they started appearing lately after we upgraded to go 1.22. Not sure if that is a coincidence, but maybe performance improved in such a way that these data races suddenly started surfacing.

We need to test our code for race conditions with `go test -race`, and we cannot have testify be the offender of data races.

A PR to resolve the issue is suggested in TODO
(It should be noted that it is heavily inspired by https://github.com/stretchr/testify/pull/1229, but since that PR hasn't seen any updates in over a year I decided to start fresh).

## Step To Reproduce
See test case in suggested fix PR: TODO

lbcjbb:
You can also trigger this issue with this simple code below using `io.Pipe`, with Go `1.22.5`. No problem with Go `1.21.12`.

```go
package foo

import (
	"io"
	"testing"

	"github.com/stretchr/testify/mock"
)

type Mock struct {
	mock.Mock
}

func (m *Mock) Read(r io.Reader) {
	m.Called(r)
}

func TestDataRace(t *testing.T) {
	var m Mock
	m.On("Read", mock.Anything)

	r, w := io.Pipe()

	go func() {
		_ = w.Close()
	}()

	m.Read(r) // trigger m.Called(r) in Mock::Read() --> DATA RACE

	_ = r.Close()
}
```

The fix proposed in #1598 fixes also this issue in this usecase.

RouquinBlanc:
This is also an issue we have each time we want to test things passing references to complex structs with private data protected by locks.

Consider the following example:

```golang
type Connection interface {
    Write([]byte) error
    Close() error
}

type EventHandler interface {
    HandleEvent(connection Connection, data []byte)
}
```

This issue triggers race checks frequently; the actual struct implementing `Connection` does change during the processing of code, but it is something normal, and the changes are protected by a mutex or other synchronization.

When doing something like this (using mockery):

```golang
myConnection := NewRealConnection() // concrete struct pointer implementing Connection

eventHandler = NewMockEventHandler(t)

[...]

PR #1229, opened 2022-07-14
Workaround race condition in printing of diff values.

dokterbob:
## Summary
When a pointer argument is used in `Run()` and this object is modified in an asynchronous fashion, the evaluation of `%v` in `Arguments.Diff` causes a race condition.

### Example output
```
==================
WARNING: DATA RACE
Write at 0x00c000161a80 by goroutine 9:
  github.com/ipfs-search/ipfs-search/components/crawler.(*CrawlerTestSuite).TestCrawlAddReference.func1()
      /Users/drbob/Development/ipfs-search/components/crawler/crawler_test.go:1040 +0x108
  github.com/stretchr/testify/mock.(*Mock).MethodCalled()
      /Users/drbob/Development/testify/mock/mock.go:538 +0x118e
  github.com/stretchr/testify/mock.(*Mock).MethodCalled()
      /Users/drbob/Development/testify/mock/mock.go:460 +0xb9
  github.com/stretchr/testify/mock.(*Mock).Called()
      /Users/drbob/Development/testify/mock/mock.go:450 +0x1a4
  github.com/ipfs-search/ipfs-search/components/index.(*Mock).Get()
      /Users/drbob/Development/ipfs-search/components/index/mock.go:27 +0x1f3
  github.com/ipfs-search/ipfs-search/components/index.MultiGet.func1()
      /Users/drbob/Development/ipfs-search/components/index/multiget.go:25 +0x121
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /Users/drbob/go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57 +0x91

Previous read at 0x00c000161a80 by goroutine 11:
  reflect.packEface()
      /usr/local/Cellar/go/1.18.2/libexec/src/reflect/value.go:130 +0xfd
  reflect.valueInterface()
      /usr/local/Cellar/go/1.18.2/libexec/src/reflect/value.go:1460 +0x188
  reflect.Value.Interface()
      /usr/local/Cellar/go/1.18.2/libexec/src/reflect/value.go:1430 +0xc9
  fmt.(*pp).printValue()
      /usr/local/Cellar/go/1.18.2/libexec/src/fmt/print.go:722 +0xca
  fmt.(*pp).printValue()
      /usr/local/Cellar/go/1.18.2/libexec/src/fmt/print.go:806 +0x1c92
  fmt.(*pp).printValue()
[...]

PR #1598, opened 2024-05-14
mock: avoid data races in Arguments.Diff

petergardfjall:
Fixes a concurrency issue that would lead to testify mocks producing data races detected by go test -race. These data races would occur whenever a mock pointer argument was concurrently modified. The reason being that `Arguments.Diff` uses the `%v` format specifier to get a presentable string for the argument. This also traverses the pointed-to data structure, which would lead to the data race.

It should be noted that it is heavily inspired by https://github.com/stretchr/testify/pull/1229, but since that PR hasn't seen any updates in over a year I decided to start fresh.

## Summary
Avoids testify mocks causing data races for (concurrently accessed) pointer arguments when running `go test -race`.

## Changes
`Arguments.Diff` now uses `%p` to produce a "presentable string" for pointers rather than `%v`, since the latter traverses the pointed to data structure.

## Motivation
We need to test our code for race conditions with go test -race, and we cannot have testify be the offender of data races.

## Related issues
https://github.com/stretchr/testify/issues/1597

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -947,7 +947,15 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			actualFmt = "(Missing)"
 		} else {
 			actual = objects[i]
-			actualFmt = fmt.Sprintf("(%[1]T=%[1]v)", actual)
+			// Note: avoid %v format specifier for pointer arguments. The %v format
+			// specifier traverses the data structure, and for situations where the
+			// argument is a pointer (that may be updated concurrently) this can result
+			// in the mock code causing a data race when running go test -race.
+			if isPtr(actual) {
+				actualFmt = fmt.Sprintf("(%[1]T=%[1]p)", actual)
+			} else {
+				actualFmt = fmt.Sprintf("(%[1]T=%[1]v)", actual)
+			}
 		}
 
 		if len(args) <= i {
@@ -955,7 +963,15 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			expectedFmt = "(Missing)"
 		} else {
 			expected = args[i]
-			expectedFmt = fmt.Sprintf("(%[1]T=%[1]v)", expected)
+			// Note: avoid %v format specifier for pointer arguments. The %v format
+			// specifier traverses the data structure, and for situations where the
+			// argument is a pointer (that may be updated concurrently) this can result
+			// in the mock code causing a data race when running go test -race.
+			if isPtr(expected) {
+				expectedFmt = fmt.Sprintf("(%[1]T=%[1]p)", expected)
+			} else {
+				expectedFmt = fmt.Sprintf("(%[1]T=%[1]v)", expected)
+			}
 		}
 
 		if matcher, ok := expected.(argumentMatcher); ok {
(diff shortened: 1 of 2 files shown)


dolmen (approved):

petergardfjall:
Friendly reminder. :slightly_smiling_face:

petergardfjall:
Any chance of this being accepted?

ndk:
It would be great to see the fix in `master` branch

brackendawson (changes requested):
This solution only works for reflect.Pointer kinds, but there are other kinds which are inherent pointers, such as maps.

Consider this test:
```go
type mapArgMock struct{ Mock }

func (m *mapArgMock) Question(arg map[string]string) int {
	args := m.Called(arg)
	return args.Int(0)
}

func Test_CallMockWithConcurrentlyModifiedMapArg(t *testing.T) {
	m := &mapArgMock{}
	m.On("Question", Anything).Return(42)

	mapArg := map[string]string{"Question": "What's the meaning of life?"}

	// Emulates a situation where the pointer value gets concurrently updated by another thread.
	wg := sync.WaitGroup{}
	wg.Add(1)
	go func() {
		defer wg.Done()
		mapArg["Question"] = "What is 7 * 6?"
	}()

	// This is where we would get a data race since Arguments.Diff would traverse the pointed to
	// struct while being updated. Something go test -race would identify as a data race.
	value := m.Question(mapArg)
	assert.Equal(t, 42, value)
	wg.Wait()

	m.AssertExpectations(t)
}
```

`IsPtr` needs to be replaced with a more comprehensive test for pointyness.

I also fear the wrath of the users passing complex struct pointers with small errors into mocks, they won't see the diff any more. But I probably care more about fixing the data race.

Is this the only way to fix the issue? Given that we consider equivelant instances of structs to be equal, could `On` be made to create a deep copy of the argument?

PS. Sorry for my tardiness on this review.

petergardfjall:
> Thinking more on the above, does `IsPtr` need to recurse deep objects up to the first pointer shaped object to catch pointer fields in non-pointer structs? It's turtles all the way down.

Yes, I think you are right. The `IsPtr` only works for pointer arguments, but not for other "mutable" types such as `Map`, `Slice`, `Array`. So a more comprehensive check would be something like:

```go
// isMutable indicates if the supplied value is of a mutable type.
func isMutable(v interface{}) bool {
	switch reflect.ValueOf(v).Kind() {
	case reflect.Array, reflect.Chan, reflect.Map, reflect.Ptr, reflect.Slice:
		return true
	}
	return false
}
```

And then it still won't catch the case where a struct _value_ (not _pointer_) gets passed which contains a nested pointer field that gets updated.

Not sure how to solve this as recursing down `v` in search of a mutable/"pointerish" field traverses the value which is what we were trying to avoid to begin with (then we might as well just use `%v`). :shrug:

brackendawson:
It is possible to write a test to show if any leaf of the structure is "pointerish", but I like it about as much as you do.

I considered if we could make a deep copy of the object rather than relying on the pointers but this isn't possible. Even with reflection you cannot copy unexported struct fields, but unexported struct fields' values are used when comparing structs with `reflect.DeepEqual` (and with `==`).

johanneswuerbach:
Would it be an option to stringify the expect and actual args only in case of an error or and when the output is actually being used?

E.g. when using a "count only" diff in places like https://github.com/stretchr/testify/blob/master/mock/mock.go#L378-L384 the data race seems also to be resolved.

Issue #1594, opened 2024-04-29
Proposal: assert.NotElementsMatch()


jdycaico:
## Description
As per https://github.com/stretchr/testify/discussions/1592#discussioncomment-9261557:
A method that does the opposite of `assert.ElementsMatch()`

## Use case
It would appear that there are no stable workarounds, and I have been suggested to make a FR

hendrywiranto:
hello @jdycaico do you plan to work on this and make a PR eventually? I can try to open a PR if you won't work on this, thanks

hendrywiranto:
fyi @jdycaico this has been merged https://github.com/stretchr/testify/pull/1600

hendrywiranto:
btw this can be closed right? @brackendawson

brackendawson:
Yes :-)

PR #1600, opened 2024-05-21
assert: new assertion NotElementsMatch

hendrywiranto:
## Summary
Add new assertion `NotElementsMatch` which asserts whether 2 lists are NOT equal ignoring the order of the elements. If there are duplicate elements, the number of appearances of each of them in both lists should not match. This is an inverse of `ElementsMatch`.

## Changes
- Add new assertion `NotElementsMatch` 

## Motivation
- No assertion for comparing 2 lists do NOT contain the exact same elements, irrespective of order. (inverse of `ElementsMatch`)

## Related issues
https://github.com/stretchr/testify/issues/1594
https://github.com/stretchr/testify/discussions/1592

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1176,6 +1176,39 @@ func formatListDiff(listA, listB interface{}, extraA, extraB []interface{}) stri
 	return msg.String()
 }
 
+// NotElementsMatch asserts that the specified listA(array, slice...) is NOT equal to specified
+// listB(array, slice...) ignoring the order of the elements. If there are duplicate elements,
+// the number of appearances of each of them in both lists should not match.
+// This is an inverse of ElementsMatch.
+//
+// assert.NotElementsMatch(t, [1, 1, 2, 3], [1, 1, 2, 3]) -> false
+//
+// assert.NotElementsMatch(t, [1, 1, 2, 3], [1, 2, 3]) -> true
+//
+// assert.NotElementsMatch(t, [1, 2, 3], [1, 2, 4]) -> true
+func NotElementsMatch(t TestingT, listA, listB interface{}, msgAndArgs ...interface{}) (ok bool) {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	if isEmpty(listA) && isEmpty(listB) {
+		return Fail(t, "A and B are the same", msgAndArgs)
+	}
+
+	if !isList(t, listA, msgAndArgs...) {
+		return Fail(t, "A is not a list", msgAndArgs...)
+	}
+	if !isList(t, listB, msgAndArgs...) {
+		return Fail(t, "B is not a list", msgAndArgs...)
+	}
+
+	extraA, extraB := diffLists(listA, listB)
+	if len(extraA) == 0 && len(extraB) == 0 {
+		return Fail(t, "A and B are the same", msgAndArgs)
+	}
+
+	return true
+}
+
 // Condition uses a Comparison to assert a complex condition.
 func Condition(t TestingT, comp Comparison, msgAndArgs ...interface{}) bool {
 	if h, ok := t.(tHelper); ok {
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1191,19 +1191,19 @@ func NotElementsMatch(t TestingT, listA, listB interface{}, msgAndArgs ...interf
 		h.Helper()
 	}
 	if isEmpty(listA) && isEmpty(listB) {
-		return Fail(t, "A and B are the same", msgAndArgs)
+		return Fail(t, "listA and listB contain the same elements", msgAndArgs)
 	}
 
 	if !isList(t, listA, msgAndArgs...) {
-		return Fail(t, "A is not a list", msgAndArgs...)
+		return Fail(t, "listA is not a list type", msgAndArgs...)
 	}
 	if !isList(t, listB, msgAndArgs...) {
-		return Fail(t, "B is not a list", msgAndArgs...)
+		return Fail(t, "listB is not a list type", msgAndArgs...)
 	}
 
 	extraA, extraB := diffLists(listA, listB)
 	if len(extraA) == 0 && len(extraB) == 0 {
-		return Fail(t, "A and B are the same", msgAndArgs)
+		return Fail(t, "listA and listB contain the same elements", msgAndArgs)
 	}
 
 	return true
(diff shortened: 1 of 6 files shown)


assert/assertions.go
```diff
@@ -1174,6 +1174,46 @@ func formatListDiff(listA, listB interface{}, extraA, extraB []interface{}) stri
...
+func NoElementsMatch(t TestingT, listA, listB interface{}, msgAndArgs ...interface{}) (ok bool) {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	if isEmpty(listA) || isEmpty(listB) {
+		return true
+	}
+
+	if !isList(t, listA, msgAndArgs...) || !isList(t, listB, msgAndArgs...) {
+		return false
```
brackendawson, line 1192: We mustn't return false, always return a call to Fail.

assert/assertions.go
```diff
@@ -1174,6 +1174,46 @@ func formatListDiff(listA, listB interface{}, extraA, extraB []interface{}) stri
...
+		element := aValue.Index(i).Interface()
+		memo[element] = true
+	}
+
+	for j := 0; j < bLen; j++ {
+		element := bValue.Index(j).Interface()
+		if _, found := memo[element]; found {
+			return Fail(t, fmt.Sprintf("Found common element %#v in both lists", element), msgAndArgs...)
+		}
+	}
```
brackendawson, line 1212: I'd suggest either make this more efficient with:
```suggestion
	memo := make(map[interface{}]struct{}, 0)
	for i := 0; i < aLen; i++ {
		element := aValue.Index(i).Interface()
		memo[element] = struct{}{}
	}

	for j := 0; j < bLen; j++ {
		element := bValue.Index(j).Interface()
		if _, found := memo[element]; found {
			return Fail(t, fmt.Sprintf("Found common element %#v in both lists", element), msgAndArgs...)
		}
	}
```

Or utilise the boolean with:
```suggestion
	memo := make(map[interface{}]bool, 0)
	for i := 0; i < aLen; i++ {
		element := aValue.Index(i).Interface()
		memo[element] = true
	}

	for j := 0; j < bLen; j++ {
		element := bValue.Index(j).Interface()
		if memo[element] {
			return Fail(t, fmt.Sprintf("Found common element %#v in both lists", element), msgAndArgs...)
		}
	}
```

brackendawson (changes requested):

brackendawson:
I had read the user's request as wanting NotElementsMatch (inverse of ElementsMatch) rather than NoElementsMatch, which a new and different assertion.

NotElementsMatch is an easy sell because by not having a negative it's inconsistent with the rest of the API and because a user requested it.

NoElementsMatch makes sense but is harder to justify because it's a distinct assertion and because nobody asked for it. Does it have an opposite? NotNoElementsMatch? OneElementMatches?

assert/assertions.go
hendrywiranto: I was referring to its sibling function `ElementsMatch` that also return `true` or `false`   
https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1097

hendrywiranto:
> I had read the user's request as wanting NotElementsMatch (inverse of ElementsMatch) rather than NoElementsMatch, which a new and different assertion.
> 
> NotElementsMatch is an easy sell because by not having a negative it's inconsistent with the rest of the API and because a user requested it.
> 
> NoElementsMatch makes sense but is harder to justify because it's a distinct assertion and because nobody asked for it. Does it have an opposite? NotNoElementsMatch? OneElementMatches?

Hello, thanks for the review
I agree with you, should've just made the `NotElementsMatch`
I never intend to make a new assertion, it's a misunderstanding between the two assertion, I was thinking those two would be the same
I'll update the PR

assert/assertions.go
hendrywiranto: adjusted to return a failure call to `Fail`

hendrywiranto:
I adjusted the PR to `NotElementsMatch`, please kindly review again @brackendawson thank you 😄

assert/assertions.go
```diff
@@ -1174,6 +1174,39 @@ func formatListDiff(listA, listB interface{}, extraA, extraB []interface{}) stri
...
+//
+// assert.NotElementsMatch(t, [1, 1, 2, 3], [1, 2, 3]) -> true
+//
+// assert.NotElementsMatch(t, [1, 2, 3], [1, 2, 4]) -> true
+func NotElementsMatch(t TestingT, listA, listB interface{}, msgAndArgs ...interface{}) (ok bool) {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+	if isEmpty(listA) && isEmpty(listB) {
+		return Fail(t, "A and B are the same", msgAndArgs)
```
brackendawson, line 1192:
```suggestion
		return Fail(t, "listA and listB contain the same elements", msgAndArgs)
```

assert/assertions.go
```diff
@@ -1174,6 +1174,39 @@ func formatListDiff(listA, listB interface{}, extraA, extraB []interface{}) stri
...
+	}
+
+	if !isList(t, listA, msgAndArgs...) {
+		return Fail(t, "A is not a list", msgAndArgs...)
```
brackendawson, line 1196:
```suggestion
		return Fail(t, "listA is not a list type", msgAndArgs...)
```

assert/assertions.go
```diff
@@ -1174,6 +1174,39 @@ func formatListDiff(listA, listB interface{}, extraA, extraB []interface{}) stri
...
+	}
+	if !isList(t, listB, msgAndArgs...) {
+		return Fail(t, "B is not a list", msgAndArgs...)
```
brackendawson, line 1199:
```suggestion
		return Fail(t, "listB is not a list type", msgAndArgs...)
```

assert/assertions.go
```diff
@@ -1174,6 +1174,39 @@ func formatListDiff(listA, listB interface{}, extraA, extraB []interface{}) stri
...
+	}
+
+	extraA, extraB := diffLists(listA, listB)
+	if len(extraA) == 0 && len(extraB) == 0 {
+		return Fail(t, "A and B are the same", msgAndArgs)
```
brackendawson, line 1204:
```suggestion
		return Fail(t, "listA and listB contain the same elements", msgAndArgs)
```

brackendawson (changes requested):
Looks great. I considered if ElementsMatch should be moved to a private function and wrapped by ElementsMatch and NotElementsMatch, but this is already done with diffLists. :+1:

Would like to see the references to the lists in the failure messages match the parameter names, though.

brackendawson (approved):
LGTM

PR #1602, opened 2024-05-27
Update minimum required go version in README.md

lord63:
## Summary

Update the minimum required go version in readme. Since the go.mod and the ci says the minimum version should be 1.17.

## Changes

No other changes, just update the readme.

## Motivation

Update and clarify the minimum required go version, so 1.17 and 1.18 go version users will have a try.

## Related issues

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,7 +348,7 @@ To update Testify to the latest version, use `go get -u github.com/stretchr/test
 Supported go versions
 ==================
 
-We currently support the most recent major Go versions from 1.19 onward.
+We currently support the most recent major Go versions from 1.17 onward.
 
 ------
 


brackendawson:
The supported version is the standard we hold ourselves to providing support for. We deliberately leave the go directive set to a correct value though, which lags behind, in order for it to be correct (the source code itself really does "target" that release), and going forward past 1.20 to not flatly break the people we don't support.

dolmen:
While we still aim to support Go from 1.17, this is a best effort and we don't want our users to rely to much on this as time passes.

So I'm fine with the README mentioning a more recent version than what is in go.mod. Go 1.19 is already quite old.

PR #1605, opened 2024-05-30
mock: Conditionally toggle mock Call

piotrpersona:
## Summary

Toggle on/off Unset method with an option.

## Changes

- New type `UnsetOption` was defined
- New function that return `UnsetOption` were defined: `WithUnsetToggle`, `WithUnsetEnabled`, `WithUnsetDisabled`
- `Unset` method will accept variadic list of `UnsetOptions` to be called, which makes the new change entirely backward-compatible
- By default `Unset` will be enabled, which is also backward-compatible

## Motivation

Consider the following example test scenario:

The goal is to test http server API. The test may set up a local http server, database and mocks to 3rd party services. Since the tests are calling a mocked server, it may be useful to use mock `EXPECT()` calls. Test structure:
1. Setup local server, database and mocks
2. Mock 3rd party service calls
3. Send request to local server
4. Assert response
5. Verify side effects - execute database query

As an enhancement, such tests can easily be called against deployed application, by simply replacing server and database URLs. The only catch are mocked calls, which won't be executed, since the request is sent to a remote server.

In such case, it may be possible to call `Unset()` method with `WithUnsetToggle` option, which value is set based on the current environment. For example:
```go
func Env() mock.UnsetOption {
    if os.Getenv("local") {
        return mock.WithUnsetToggle(false)
    }
    return mock.WithUnsetToggle(true)
}

// In test ...
func TestAPI(t *testing.T) {
    // setup server, database and mocks
    mock.Call3rdParty(ctx, argument).Return(nil).Unset(Env())

    response := client.Post(request)
    // assertions and database query
}
```

In the above example, if environment is "local" Unset will be disabled, meaning that all mocks will execute.
If environment is other than "local" the Unset will be enabled, meaning that mocks will not be called.

This allows to run the same suite of tests regardless of environment and setup.

### Alternative approach

> Edit: I also implemened `Toggle` method, that is way simpler and may be used in a similar way. I am looking forward to your feedback.

```go
func Env() bool {
    return os.Getenv("ENV") == "local"
}
// ...
mock.Call3rdParty(ctx, argument).Return(nil).Toggle(Env())
```

P.S. If there is an easier way to achieve similar behaviour than what I proposed, please let me know. I wasn't able to find any workaround, other than calling `Unset` in `if` block. I am also open to any suggestions and would definitely like to implement this logic somewhere in testify package, only if you think this is the right approach.

## Related issues

None

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -208,10 +208,49 @@ func (c *Call) On(methodName string, arguments ...interface{}) *Call {
 	return c.Parent.On(methodName, arguments...)
 }
 
+// unsetConfiguration stores Unset method call configuration.
+type unsetConfiguration struct {
+	enabled bool
+}
+
+// UnsetOption will allow to configure Unset method invocation.
+type UnsetOption func(*unsetConfiguration)
+
+// WithUnsetToggle will configure unset value. If enabled == false the Unset
+// function will not be called. Default: true - the Unset will be called.
+// This option may be used to conditionally enable or disable mock calls.
+func WithUnsetToggle(enabled bool) UnsetOption {
+	return func(uc *unsetConfiguration) {
+		uc.enabled = enabled
+	}
+}
+
+// WithUnsetEnabled is shorthand for WithUnsetToggle(true).
+func WithUnsetEnabled() UnsetOption {
+	return func(uc *unsetConfiguration) {
+		uc.enabled = true
+	}
+}
+
+// WithUnsetEnabled is shorthand for WithUnsetToggle(false).
+func WithUnsetDisabled() UnsetOption {
+	return func(uc *unsetConfiguration) {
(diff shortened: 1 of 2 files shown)


brackendawson:
> calling `Unset` in `if` block

Please, just do that, it is far more readable. Better yet, inverse the logic of the if block and never set the expectation in the first place:

```go
// less good
m := mock.Call3rdParty(ctx, argument).Return(nil)
if os.Getenv("local") {
  m.Unset()
}

// better
if !os.Getenv("local") {
  mock.Call3rdParty(ctx, argument).Return(nil).Once()
}
```

It is my opinion that Unset should not have been added to begin with, its use makes it difficult to grok what expectations a test actually has. I would urge us to not complicate it further, sorry.

Issue #1087, opened 2021-06-16
Feature: assert.Consistently


shaneutt:
I've been using testify for a long time now, I appreciate you building and maintaining this for us all :bow: 

I'm looking for functionality that is inspired by other testing frameworks that would be similar to `assert.Eventually` and `require.Eventually`, but instead would be:

```
stagger := time.Millisecond * 200
tries := 15
assert.Consistently(t, func() bool {
   fmt.Println("my logic here")
   return result()
}, time.Millisecond * 200, )
```

The point of this would be to ensure that a result is _consistently_ occurring over a period of time. In the above example, that the fake `result()` function is returning true consistently every 200ms for 3 seconds total.

Does testify already have something like this and I've just missed it?

If not, are you cool with me contributing something like this?

Thanks!

brackendawson:
Would `Consistently()` be more readable over say this code:
```go
for i := 1 ; i <= 15 ; i++ {
    t.Logf("attempt %d", i)
    time.Sleep(200*time.Millisecond)
    got := myCode()
    assert.Equal(t, expectation, got)
}
```

Another concern I have is what if someone calls `t.Error()` inside the function? The testify framework will have no way to know they did that, so the `Consistently()` assertion will pass but the test will fail and you wont be told on which attempt it failed.

shaneutt:
Readability is subjective: I'm more interested in the prescriptiveness it would bring: setting a clear standard in tests that use testify for how to achieve this and leaving less room for interpretation for contributors. As far as your concerns about t.Error does that problem also exist for `Eventually` or no?

brackendawson:
`Eventually` probably does have the same quirk with use of t inside it's function. And it does set a precedent for your suggested function signature.

I think I wouldn't object to adding this.

brackendawson:
Maybe it should take two durations, a `tick` and a `testFor`, a bit like `Eventually` does?

If the function takes longer than the tick do you run them concurrently or fail the assertion?

shaneutt:
> Maybe it should take two durations, a `tick` and a `testFor`, a bit like `Eventually` does?

Right, in the original post I had mocked it up that way but apparently messed up the var names so it's a little confusing, here's a redux:

```go
assert.Consistently(t, func() bool {
   fmt.Println("my logic here")
   return result()
}, testFor, tick)
```

> If the function takes longer than the tick do you run them concurrently or fail the assertion?

[...]

PR #1606, opened 2024-06-04
feat: add assert.Consistently

jfmyers9:
## Summary
This changeset adds the `assert.Consistently` and it's associated functions to assert that a condition is true over the entire period of `waitFor`. This is useful when testing the behavior of asynchronous functions.

## Changes

- `assert.Consistently` checks that a provided function is true for the entirety of `waitFor`
- `assert.Consistentlyf` checks that a provided function is true for the entirety of `waitFor` and allows for formatting the error message
- `assert.ConsistentlyWithT` checks that a provided function that accepts a `testing.T` is true for the entirety of `waitFor`
- `assert.ConsistentlyWithTf` checks that a provided function that accepts a `testing.T` is true for the entirety of `waitFor` and allows for formatting the error message

## Motivation

See: https://github.com/stretchr/testify/issues/1087#issuecomment-2139222285

## Related issues
Closes https://github.com/stretchr/testify/issues/1087


dolmen:
I don't want more uses of [`CollectT`](https://pkg.go.dev/github.com/stretchr/testify/assert#CollectT) (which should never have been created as a concrete type in the first place). So let's remove all the `WithT` functions for now.

dolmen:
We already have [`Never`](https://pkg.go.dev/github.com/stretchr/testify/assert#Never) and `Neverf` which match the use case: it just need a reversed condition.

z0marlin:
> I don't want more uses of [`CollectT`](https://pkg.go.dev/github.com/stretchr/testify/assert#CollectT) (which should never have been exposed as a concrete type in the first place). So let's remove all the `WithT` functions for now.

@dolmen, curious, what are the issues with `CollectT`? Also, I cannot say much about the `CollectT` type but I have found `WithT` quite useful in cases when I already have a certain set of assertions and want to assert that they succeed eventually (testing k8s controllers).

dolmen:
`CollectT` should have been an interface to keep its implementation opaque.

ccoVeille:
> We already have [`Never`](https://pkg.go.dev/github.com/stretchr/testify/assert#Never) and `Neverf` which match the use case: it just need a reversed condition.

Make sense, what would be the reversed condition asserter ? NotNever ? The name would be strange. But using another name that is not the opposite of Never would be strange no?

jfmyers9:
In my opinion the opposite of `Never` would be `Always`.

`Consistently` works better as a name when paired with `Eventually`.

I don't have a strong preference on naming here.

assert/assertions.go
```diff
@@ -2038,6 +2038,93 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
...
+	for tick := ticker.C; ; {
+		select {
+		case <-timer.C:
+			return true
+		case <-tick:
+			tick = nil
+			go func() { ch <- condition() }()
+		case v := <-ch:
+			if !v {
+				return Fail(t, "Condition never satisfied", msgAndArgs...)
```
z0marlin, line 2067: nit: not sure if this is the right error message. To me, "Condition never satisfied"  makes sense in the case of `Eventually`. The term never is not applicable in the context of `Consistently`.

assert/assertions.go
```diff
@@ -2038,6 +2038,93 @@ func EventuallyWithT(t TestingT, condition func(collect *CollectT), waitFor time
...
+			go func() {
+				collect := new(CollectT)
+				defer func() {
+					ch <- collect.errors
+				}()
+				condition(collect)
+			}()
+		case errs := <-ch:
+			if len(errs) > 0 {
+				return Fail(t, "Condition never satisfied", msgAndArgs...)
```
z0marlin, line 2120: add the errors to `t` first before calling `Fail`. Reference: https://github.com/stretchr/testify/blob/master/assert/assertions.go#L2017

z0marlin (changes requested):

jfmyers9:
If we are happy with `Always`, then this just becomes:
```
func Always(t TestingT, condition func() bool, waitFor time.Duration, tick time.Duration, msgAndArgs ...interface{}) bool {
	return Never(t, func() bool { return !condition() }, waitFor, tick, msgAndArgs)
}
```

I missed the fact that `Never` exists when implementing this. Thoughts on moving to that approach?

z0marlin:
> I missed the fact that `Never` exists when implementing this. Thoughts on moving to that approach?

My only concern here is that using `Never` will not allow having something like `ConsistentlyWithT` or `AlwaysWithT` (not really concerned about the naming). Not sure about how big of a use-case it is though.

dolmen:
`Never` isn't a great name as it isn't what users expect. `Always` is in the same bucket, or even worse. I think that `Consistently` is a better name as it is consistent (pun intended) with `Eventually`, even if inconsistent with `Never`.

I was reluctant to add `Consistently` (I want to limit the increase the API surface), but as `Never` already exists that would be consistent (pun intended). I understand also that `ConsistentlyWithT` would be convenient as it would allow to use `require`.

About the implementation, I think it's a bit early because we are not yet done with `Eventually` bugs. I would like every eyes to focus on #1611.

Issue #1609, opened 2024-06-08
Incorrect require.Error() docs 


andig:
## Description

The error docs read:

```go
// Error asserts that a function returned an error (i.e. not `nil`).
//
//	  actualObj, err := SomeFunction()
//	  if assert.Error(t, err) {
//		   assert.Equal(t, expectedError, err)
//	  }
```

This indicates that `Error()` has a `bool` return value which it hasn't. There is no way to check `expectedError` like this.

## Expected behavior
Error docs should show actual usage. It would be nice if `Error()` actually allowed for checking the specific error value.

hendrywiranto:
hello @andig 

is this the `Error()` that you meant? https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1596
Error has a `bool` return value, please do elaborate more if I misunderstood your meanings.

hendrywiranto:
yea I think so, but I checked the docs for `require`, they all used `assert` as example in the comments, I don't know if it's intended or not
I think if we want to change we need to adjust all the comments on the package.

brackendawson:
The doc-string for Error needs to be made into one which works for both the assert and require packages. Especially after #1610 is merged.

I'd suggest simply removing the if statement and the Equal call from the example. Especially since EqualError exists.

architagr:
Hi @brackendawson,

I can fix this doc-string for assert and require. To reconfirm we want to update the doc not to have an example that uses if  block and then `Equal` function

```go

``` // Error asserts that a function returned an error (i.e. not `nil`).
//
//	  actualObj, err := SomeFunction()
//	  assert.Error(t, err) 
func Error(t TestingT, err error, msgAndArgs ...interface{}) bool {
	if err == nil {
		if h, ok := t.(tHelper); ok {
			h.Helper()
		}
		return Fail(t, "An error is expected but got nil.", msgAndArgs...)
	}

	return true
}

and similar change in https://github.com/stretchr/testify/blob/master/require/require.go#L280

brackendawson:
Yes. The change in require is automatic, just run `go generate ./...`.

architagr:
> Yes. The change in require is automatic, just run `go generate ./...`.

I have created pull request #1675

PR #1610, opened 2024-06-13
Generate better comments for require package

Neokil:
## Summary

The comments for the require package were just copied over from the assert package when generating the functions. This could lead to confusion because
1. The code-examples were showing examples using the assert package instead of the require package
2. The function-documentation was not mentioning that the functions were calling `t.FailNow()` which is some critical information when using this package.

## Changes
1. Added a "replace" function to the template
2. require.go.tmpl now replaces the package name accordingly
3. require.go.tmpl now adds another line to the function-docs explaining the difference to the assert-function

## Motivation
When new developers are coming in they most likely did not go through the github-documentation for every package used in the application. But what they do is read the function-docs when they go through code to understand what it is doing.
In case of test-cases using both `require` and `assert` it gets confusing because the function docs for the `require` package are referencing the `assert` package and don't really document the actual function.

### DIFF
--- a/require/require.go
+++ b/require/require.go
@@ -10,7 +10,7 @@ import (
 )
 
 // Condition uses a Comparison to assert a complex condition.
-// Instead of returning a boolean result this function calls `t.FailNow()` on failure.
+// Failure of this check is fatal ([testing.T.FailNow] is called on failure).
 func Condition(t TestingT, comp assert.Comparison, msgAndArgs ...interface{}) {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
@@ -22,7 +22,7 @@ func Condition(t TestingT, comp assert.Comparison, msgAndArgs ...interface{}) {
 }
 
 // Conditionf uses a Comparison to assert a complex condition.
-// Instead of returning a boolean result this function calls `t.FailNow()` on failure.
+// Failure of this check is fatal ([testing.T.FailNow] is called on failure).
 func Conditionf(t TestingT, comp assert.Comparison, msg string, args ...interface{}) {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
@@ -40,7 +40,7 @@ func Conditionf(t TestingT, comp assert.Comparison, msg string, args ...interfac
 //	require.Contains(t, ["Hello", "World"], "World")
 //	require.Contains(t, {"Hello": "World"}, "Hello")
 //
-// Instead of returning a boolean result this function calls `t.FailNow()` on failure.
+// Failure of this check is fatal ([testing.T.FailNow] is called on failure).
 func Contains(t TestingT, s interface{}, contains interface{}, msgAndArgs ...interface{}) {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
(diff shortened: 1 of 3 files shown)


require/require.go.tmpl
```diff
@@ -1,4 +1,5 @@
-{{.Comment}}
+{{ replace .Comment "assert." "require."}}
+// Instead of returning a boolean result this function calls `t.FailNow()` on failure.
```
ccoVeille, line 2: The "FailNow on failure" sounds a bit strange to me

Maybe:
- calls FailNow when a test fails
- calls FailNow on first invalid assertion

you could also add it has the consequence to stop all tests execution

These are wording idea, more a functional point than a real wording

hendrywiranto:
this will close https://github.com/stretchr/testify/issues/1609 too

Neokil:
somehow I cannot add reviewers the to the PR, but maybe @boyan-soubachov, @dolmen, @MovieStoreGuy, @arjunmahishi or @brackendawson can one of you have a look?

Neokil:
any change to get someone to review/merge this?

brackendawson (changes requested):
The change to swap `assert.` with `require.` is good. 👍 It should be safe as if any assert examples ever use `msgAndArgs` the syntax will be compatible with require. I considered if inline examples should use the package name at all, in the standard library they do, despite links to functions not doing.

Adding a note about `t.FailNow` to every requirement is too much. The doc for require [already says they stop test execution](https://github.com/stretchr/testify/blob/c4b8421a1fb1a1eed787101978a3bcf7ae8a5d0f/require/doc.go#L2), we should just change this sentence to mention that it uses `t.FailNow`. The doc for tetify [also mentions t.FailNow](https://github.com/stretchr/testify/blob/c4b8421a1fb1a1eed787101978a3bcf7ae8a5d0f/README.md?plain=1#L108), but tese two docs should probably match one-another.

brackendawson:
> this will close #1609 too

I'm afraid it wont, requre.Error has no return value. That docstring needs to be made into something that will work for both pacakges.

Neokil:
@brackendawson just removed the additional comment as you requested. 

I have to say that IMO a hint like this would be really helpful after all to all people that are new to the package. When you encounter it in an existing project you will only see the function-comments in your IDE with no hint to the `t.FailNow()`. It is only when you are actually looking into the Github of the library that you will find that information. If you land on the [pkg.go.dev](https://pkg.go.dev/github.com/stretchr/testify/require#DirExists) page because you want to see the documentation for the function to see why it terminated your test you need to know that you have to scroll up to the top or else you will not find that info.

brackendawson (approved):

brackendawson:
> I have to say that IMO a hint like this would be really helpful after all to all people that are new to the package. When you encounter it in an existing project you will only see the function-comments in your IDE with no hint to the `t.FailNow()`. It is only when you are actually looking into the Github of the library that you will find that information. If you land on the [pkg.go.dev](https://pkg.go.dev/github.com/stretchr/testify/require#DirExists) page because you want to see the documentation for the function to see why it terminated your test you need to know that you have to scroll up to the top or else you will not find that info.

I hear you. And it's not that I disagree with any of that. But there are a lot of other notes that we could add to each function's docstring and there is a point where we're better not making them larger. For me the concept of t.Error vs t.Fatal doesn't feel so surprising, even if you might not be familiar with our nomenclature of assert and require. Just my opinion.

Going to allow some time for other maintainers to review before I merge.

require/require.go
```diff
@@ -415,9 +415,9 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 //		time.Sleep(8*time.Second)
 //		externalValue = true
 //	}()
-//	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+//	require.EventuallyWithT(t, func(c *require.CollectT) {
```
machine424, line 418: I'm afraid `require.CollectT` is not a thing

PR #1614, opened 2024-06-24
Call stack perf change for CallerInfo

mikeauclair:
## Summary

- Switches `CallerInfo` to use `runtime.Callers()` to pull the stack instead of repeatedly calling `runtime.Caller()`

## Changes
- Buffer 10 stack frames at a time by switching to a series of calls to `Callers()` instead of repeated `Caller` invocations, because there is a cost to querying the stack
- Iterate over the frames with the `frames.Next()` iterator

## Motivation

- We heavily use mocks at DevotedHealth and the repeated calls to pull the stack were flagging as a contributor when I profiled some of our slow tests - see example of `Caller` contributing a CPU cost of 7660ms out of a total cost of 13740ms
Screenshot 2024-06-24 at 1 10 26 PM


assert/assertions.go
```diff
@@ -212,19 +214,26 @@ the problem actually occurred in calling code.*/
...
-		pc, file, line, ok = runtime.Caller(i)
-		if !ok {
-			// The breaks below failed to terminate the loop, and we ran off the
-			// end of the call stack.
-			break
-		}
+	pcs := make([]uintptr, stackFrameBufferSize)
+	offset := 1
+	n := runtime.Callers(offset, pcs)
+	maybeMore := n == stackFrameBufferSize
```
ccoVeille, line 225: I would move this line, just before the `for`

assert/assertions.go
```diff
@@ -263,6 +272,23 @@ func CallerInfo() []string {
 			isTest(name, "Example") {
 			break
 		}
+
+		if !more {
```
ccoVeille, line 276:
```suggestion
		if more {
         continue
    }
```

This would be simpler and more readable, no?

assert/assertions.go
ccoVeille: same for offset and pcs

They are not used when return via the `if`

mikeauclair: `pcs` is filled by `runtime.Callers` before the `if` so I'll keep that there; I'll move the other two lines, though

ccoVeille: Indeed, thanks for applying my first suggestion

ccoVeille (approved):

assert/assertions.go
```diff
@@ -263,6 +273,24 @@ func CallerInfo() []string {
...
+		if !maybeMore {
+			break
+		}
+		offset += stackFrameBufferSize
+		n = runtime.Callers(offset, pcs)
+		if n == 0 {
+			break
+		}
+
+		maybeMore = n == stackFrameBufferSize
```
brackendawson, line 290: Copying this comment from #1615 which also contains this change.

I'm not a fan of `more` and `maybeMore`, this code is actually two nested loops and I think it would be more readable if you coded it that way. I think that pure behaviour is usually better than data (varibles) used only to alter behaviour.

brackendawson (changes requested):

brackendawson:
The performance enhancement in this code is a change I'd be happy to include, with some slight alteration.

assert/assertions.go
mikeauclair: I just pushed a commit that moves this to two loops (though still using `more` to `break` out of the inner loop). LMK if that's along the lines of what you were thinking.

brackendawson (approved):
I'm happy with the `more` variable because it's in the idiomatic usage example too: https://pkg.go.dev/runtime#Frames

Issue #643, opened 2018-08-01
mock: Infinite recursion deadlock if mocking String()


natemara:
This is a fairly contrived example, I admit. But this was a real head-scratcher in the wild.

The following code will never complete.

```
package main

import (
	"fmt"

	"github.com/stretchr/testify/mock"
)

type K struct {
	mock.Mock
}

func (f *K) String() string {
	return f.Called().Get(0).(string)
}

func (f *K) Foo(b B) string {
	return f.Called(b).Get(0).(string)
}

type B struct {
	Kay *K
}

func main() {
	k := &K{}
	b := B{
		Kay: k,
	}

	k.On("String").Return("string-return")
	k.On("Foo", b).Return("foo-return")

	fmt.Printf(k.Foo(b))
}
```

This fails because `Mock.Called` calls `Mock.MethodCalled`, which locks the Mutex, and eventually calls `Arguments.Diff`. When `Arguments.Diff` calls `fmt.Sprintf`, `fmt.Sprintf` sees the `Kay` field of `B`. Since `K` implements `fmt.Stringer`, `fmt.Sprintf` calls `K.String`. Since this is a mocked function, `K.String` jumps into `Mock.Called`, which calls `Mock.MethodCalled` which waits for the still-locked mutex to unlock, which never happens, because the lock is further up the call chain.

I don't know if this should be considered a bug or not, but it certainly was unexpected, and took a lot of digging to figure out.

natemara:
Here's a stack trace from a debug session just before the mutex lock deadlock

```
github.com/stretchr/testify/mock.(*Mock).MethodCalled (github.com/stretchr/testify/mock/mock.go:318)
github.com/stretchr/testify/mock.(*Mock).Called (github.com/stretchr/testify/mock/mock.go:310)
main.(*K).String (main.go)
fmt.(*pp).handleMethods (/usr/local/Cellar/go/1.10/libexec/src/fmt/print.go:603)
fmt.(*pp).printValue (/usr/local/Cellar/go/1.10/libexec/src/fmt/print.go:700)
fmt.(*pp).printValue (/usr/local/Cellar/go/1.10/libexec/src/fmt/print.go:783)
fmt.(*pp).printValue (/usr/local/Cellar/go/1.10/libexec/src/fmt/print.go:853)
fmt.(*pp).printArg (/usr/local/Cellar/go/1.10/libexec/src/fmt/print.go:689)
[...]

PR #1615, opened 2024-06-25
Lazily render mock diff output on successful match

mikeauclair:
## Summary

- For mocks on functions that take complex types as arg, `Sprintf`-ing the `actual` value can be costly. Switch successful matches to use thunks to lazily render their output, and fully avoid rendering their output when there are no differences since we return a constant string in that case anyways

## Changes
- Switch to a string builder for accumulating output for better performance even when we *do* render

## Motivation
- We heavily use mocks at DevotedHealth, and use them frequently in gRPC contexts where we're mocking RPCs that take `<Whatever>Request` structs that are costly to `%v` via `Sprintf`. `Sprintf` showed up as a tangible contributor to runtime performance when profiling some of our heavier tests. The below ~5s cpu time spent on `Sprintf` goes away completely here in one of our exemplar tests

Screenshot 2024-06-25 at 12 16 35 PM

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -942,8 +942,8 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 
 	outputRenderers := []outputRenderer{}
 
-	for j := 0; j < maxArgCount; j++ {
-		i := j
+	for i := 0; i < maxArgCount; i++ {
+		i := i
 		var actual, expected interface{}
 		var actualFmt, expectedFmt func() string
 


brackendawson:
This includes the CallerInfo change from #1614

brackendawson (changes requested):
I've considered and tried a similar approach before when trying to address #1597. It only removes the race from passing tests, though. If it comes with a sizable performance benefit I'm more than happy to lazily render these.

I'd like it to be separated from #1614, they're not dependant on one another, right?

Then there's some changes I'd like around the use of `outputRenderers`. I'm fairly sure the slice isn't even necessary, you can change `actualFmt`'s type from `string` to `func() string` and change all its references to `actualFmt()`.

mikeauclair:
> I'd like it to be separated from #1614, they're not dependant on one another, right?

I honestly don't remember if they are or are not given how long it's been since I wrote this. I can try to take a pass on cleaning this and the other PR up after next week (short week for thanksgiving in the states)

mock/mock.go
```diff
@@ -923,56 +923,73 @@ func (args Arguments) Is(objects ...interface{}) bool {
...
+	var outputBuilder strings.Builder
 	var differences int
 
 	maxArgCount := len(args)
 	if len(objects) > maxArgCount {
 		maxArgCount = len(objects)
 	}
 
-	for i := 0; i < maxArgCount; i++ {
+	outputRenderers := []outputRenderer{}
```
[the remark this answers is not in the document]
mikeauclair: Yeah I waffled back and forth on that for a bit. Can definitely simplify every down to `func() string`

mock/mock.go
```diff
@@ -923,56 +923,75 @@ func (args Arguments) Is(objects ...interface{}) bool {
...
+
+	for j := 0; j < maxArgCount; j++ {
+		i := j
```
brackendawson, line 946: What's the difference between i and j?

brackendawson (changes requested):
I think if #1693 were merged then this would still offer a performance benefit in finding the closest match when there are no exact matches.

Though, a better approach might be to fully separate finding the first exact match, finding the closest match, then rendering the diff of the closest match. Diff is getting awfully complicated and has a few too many responsibilities.

mock/mock.go
mikeauclair: We close over the loop variable in the renderers, so this is to avoid the pre-1.22 loopvar behavior issues. Can definitely drop depending on support requirements

brackendawson: Ah of course we do. I mistook it for an artefact from the removal of the outer loop. Could you make it into `i := i`? It is legal and I think more clearly communicates to most gophers that you're deliberately re-scoping the variable.

brackendawson (approved):
I'm happy with the code and I'm happy enough with the approach being a step towards argument matching in Mock.Called with fewer race conditions.

This should make no change to the text output of the library. But due to the frailty of the area of code it touches I'm going to allow a couple of weeks for other maintainers to comment.

brackendawson:
Thank you for your contribution.

brackendawson:
This was reverted by #1786 because it did not properly implement lazy rendering when there was more than one expected call. I anticipate that the fix for #643 will properly implement lazy rendering.

Issue #1621, opened 2024-07-10
`Unset` removes all expected calls that match arguments in call being unset


techfg:
## Description
When calling `Unset`, any expected call that matches the arguments in the call being unset is removed from `ExpectedCalls`.

For example, in the following scenarios, both the `Anything, Anything` and `3,3` calls are removed:

```go
// Scenario 1- Explicit Unset on the call
mockedService.
    On("TheExampleMethod1", Anything, Anything).Return(0).
    On("TheExampleMethod1", 2, 2).Return(0)
callToUnset := mockedService.On("TheExampleMethod1", 3, 3).Return(0)
callToUnset.Unset()

// Scenario 2 - Adding/Unsetting a call
mockedService.
    On("TheExampleMethod1", 2, 2).Return(0).
    On("TheExampleMethod1", 3, 3).Return(0).
    On("TheExampleMethod1", Anything, Anything).Return(0)
mockedService.On("TheExampleMethod1", 3, 3).Unset()
```

## Reproduction

[Go Playground](https://go.dev/play) is timing out on build when `github.com/stretchr/testify/mock` is used (likely due to dependencies downloading), so submitted [PR]() with tests to reproduce.

## Step To Reproduce

Run any/all of the following tests:

1. [Test_Mock_UnsetOnlyUnsetsCurrentCall]()
2. [Test_Mock_UnsetOnlyUnsetsExactArgsMatchCall]()
3. [Test_Mock_UnsetOnlyUnsetsExactArgsMatchCallWhenPartialMatch]()
4. [Test_Mock_UnsetOnlyUnsetsExactArgsMatchCallWhenAnythingSomeArg]()
5. [Test_Mock_UnsetOnlyUnsetsExactArgsMatchCallWhenAnythingEveryArg]()

## Expected behavior
Only the call(s) with the _exact_ matched arguments should be unset/removed. 

Ideally only the call that `Unset` is invoked against is removed but the current API is written in a manner that requires a `Call` to unset so unless the call is tracked (e.g., `c := mockService.On(....)`), `On` must be called first then `Unset` which adds the expectation only to immediately remove it along with any other "matched" expectations.  #982 which added `Unset` actually started with `Unset` on `Mock` but then [moved it](https://github.com/stretchr/testify/pull/982/commits/3ca34d998b448a4cfeef174c27f2f039dbfd4386#diff-f1f50cce7e03599a6f1dff51932f13530c2b3393e70aa456fb18ba2513fac8cdL292
[...]

PR #1622, opened 2024-07-10
test: add tests for unset call matching #1621

techfg:
**NOTE: The build for this PR is expected to fail as the PR currently only adds tests to reproduce the issues in #1621**

## Summary
Add tests to reproduce issues in #1621

## Changes
Adds 5 tests to reproduce issue

## Motivation
See #1621

## Related issues
#1621


brackendawson:
Looking into #1621 it appears the current behaviour is working as designed. I dislike the design, but specifically making Unset only unset its own Call instance will break the intended use case which is encoded in tests added with the feature:
```go
myMock.On(1).Return()
myMock.On(1).Return().Unset()
// There are now no expectations to be called with `1`.
```

PR #1381, opened 2023-05-11
Remove functional option caller name match expectation

dillonstreator:
## Summary
Fixes #1380 by removing expectations around optional functions calling function names. These assertions are not necessary to satisfy the testing use case for functional options and are in fact just outright incorrect due to the curried nature of the optional function. The time at which the return function from the optional function is evaluated, the optional function name is no longer accessible/in-scope.

## Changes
* Removes expectation around functional options caller name matching as it is not guaranteed and not truly asserting the functional options function name.
* Adds test to cover additional use case.
* Introduces a strict length check on the function options.

## Motivation
Fix bug and improve test to cover additional use cases.

## Related issues
Closes #1380

PR #1626, opened 2024-07-23
fix: compare functional option names for indirect calls

### LINKED ISSUE #1380: Functional Options testing broken for indirect calls
The changes introduced to support testing of functional options (https://github.com/stretchr/testify/pull/1023) are broken with indirect calls.

When the service which exposes the method that supports functional options is injected into another function and invoked, the mock assertions are broken.

The following test and mocks should be supported.

```diff
diff --git a/mock/mock_test.go b/mock/mock_test.go
index 60cf5d0..1c7ca47 100644
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -20,6 +20,7 @@ import (
 // ExampleInterface represents an example interface.
 type ExampleInterface interface {
 	TheExampleMethod(a, b, c int) (int, error)
+	TheExampleMethodFunctionalOptions(x string, opts ...OptionFn) error
 }
 
 // TestExampleImplementation is a test implementation of ExampleInterface
@@ -1455,6 +1456,27 @@ func Test_Mock_AssertExpectationsFunctionalOptionsType(t *testing.T) {
 
 }
 
+func Test_Mock_AssertExpectationsFunctionalOptionsType_indirect(t *testing.T) {
+
+	indirectFunctionalOptions := func(ei ExampleInterface) {
+		ei.TheExampleMethodFunctionalOptions("testing", OpStr("test"), OpNum(1))
+	}
+
+	var mockedService = new(TestExampleImplementation)
+
+	mockedService.On("TheExampleMethodFunctionalOptions", "testing", FunctionalOptions(OpStr("test"), OpNum(1))).Return(nil).Once()
+
+	tt := new(testing.T)
+	assert.False(t, mockedService.AssertExpectations(tt))
+
+	// make the call now
+	indirectFunctionalOptions(mockedService)
+
+	// now assert expectations
+	assert.True(t, mockedService.AssertExpectations(tt))
+
+}
+
 func Test_Mock_AssertExpectationsFunctionalOptionsType_Empty(t *testing.T) {
 
 	var mockedService = new(TestExampleImplementation)
```

but instead results in the following error:

```sh
mock: Unexpected Method Call
-----------------------------

TheExampleMethodFunctionalOptions(string,[]mock.OptionFn)
		0: "testing"
		1: []mock.OptionFn{(mock.OptionFn)(0x11a8020), (mock.OptionFn)(0x11a8000)}

The closest call I have is: 

TheExampleMethodFunctionalOptions(string,*mock.FunctionalOptionsArgument)
		0: "testing"
		1: []mock.OptionFn{(mock.OptionFn)(0x11a0980), (mock.OptionFn)(0x11a0960)}


Diff: 0: PASS:  (string=testing) == (string=testing)
	1: FAIL:  [mock.Test_Mock_AssertExpectationsFunctionalOptionsType_indirect.func1 mock.Test_Mock_AssertExpectationsFunctionalOptionsType_indirect.func1] != [mock.Test_Mock_AssertExpectationsFunctionalOptionsType_indirect mock.Test_Mock_AssertExpectationsFunctionalOptionsType_indirect]
```

The culprit seems to be the [`funcName`](https://github.com/stretchr/testify/blob/4b2f4d2bcff3848b6a4e63d462da6ce3c21e9c78/mock/mock.go#L1220C1-L1223) call which is used in [`assertOpts`](https://github.com/stretchr/testify/blob/master/mock/mock.go#L1167C1-L1173) returning unexpected values.

arjun-1:
## Summary
Closes Functional Options testing broken for indirect calls #1380. The fix is consists of not comparing the the function names, as it is already asserted when comparing `expectedValues` and `actualValues`

## Changes
* Adds test `Test_Mock_AssertExpectationsFunctionalOptionsTypeIndirectly` which would fail without the fix
* Adds test `Test_Mock_AssertExpectationsFunctionalOptionsType_Diff` to examine behavior of slightly different functional options in expected vs actual.
* Changes the `funcName` implementation to only compare the relevant last part, if applicable. This aids in reading the diff that is produced when the function names don't. match (for go1.20 or higher)

## Motivation
The choice was made to omit the comparison of function names completely, instead of comparing only the relevant last part. The reason is that go1.19 and lower don't give enough any meaningful comparable part in the function name. Nevertheless the most relevant part of the function name is included in the diff result. Without it, the diff yielded in `Test_Mock_AssertExpectationsFunctionalOptionsType_Diff` would have resulted in (go1.20):
```
mock.Test_Mock_AssertExpectationsFunctionalOptionsType_Diff.OpNum &{num:0 str:1} != mock.Test_Mock_AssertExpectationsFunctionalOptionsType_Diff.OpNum &{num:1 str:}
```
The current implementation instead yields a diff:
```
OpStr &{num:0 str:1} != OpNum &{num:1 str:}
```

Though there could be added benefit in displaying the full location of the function, I can revert that part if desired.

## Related issues
Closes #1380

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -1204,31 +1204,16 @@ type tHelper interface {
 func assertOpts(expected, actual interface{}) (expectedFmt, actualFmt string) {
 	expectedOpts := reflect.ValueOf(expected)
 	actualOpts := reflect.ValueOf(actual)
-	var expectedNames []string
-	for i := 0; i < expectedOpts.Len(); i++ {
-		expectedNames = append(expectedNames, funcName(expectedOpts.Index(i).Interface()))
-	}
-	var actualNames []string
-	for i := 0; i < actualOpts.Len(); i++ {
-		actualNames = append(actualNames, funcName(actualOpts.Index(i).Interface()))
-	}
-	if !assert.ObjectsAreEqual(expectedNames, actualNames) {
-		expectedFmt = fmt.Sprintf("%v", expectedNames)
-		actualFmt = fmt.Sprintf("%v", actualNames)
+
+	if expectedOpts.Len() != actualOpts.Len() {
+		expectedFmt = fmt.Sprintf("%v", expectedOpts)
+		actualFmt = fmt.Sprintf("%v", actualOpts)
 		return
 	}
 
 	for i := 0; i < expectedOpts.Len(); i++ {
 		expectedOpt := expectedOpts.Index(i).Interface()
 		actualOpt := actualOpts.Index(i).Interface()
-
-		expectedFunc := expectedNames[i]
-		actualFunc := actualNames[i]
-		if expectedFunc != actualFunc {
-			expectedFmt = expectedFunc
-			actualFmt = actualFunc
-			return
-		}
 
 		ot := reflect.TypeOf(expectedOpt)
 		var expectedValues []reflect.Value
@@ -1248,8 +1233,8 @@ func assertOpts(expected, actual interface{}) (expectedFmt, actualFmt string) {
 
 		for i := 0; i < ot.NumIn(); i++ {
 			if !assert.ObjectsAreEqual(expectedValues[i].Interface(), actualValues[i].Interface()) {
-				expectedFmt = fmt.Sprintf("%s %+v", expectedNames[i], expectedValues[i].Interface())
-				actualFmt = fmt.Sprintf("%s %+v", expectedNames[i], actualValues[i].Interface())
+				expectedFmt = fmt.Sprintf("%s %+v", funcName(expectedOpts.Index(i).Interface()), expectedValues[i].Interface())
+				actualFmt = fmt.Sprintf("%s %+v", funcName(actualOpts.Index(i).Interface()), actualValues[i].Interface())
 				return
 			}
 		}
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -1205,21 +1205,31 @@ func assertOpts(expected, actual interface{}) (expectedFmt, actualFmt string) {
 	expectedOpts := reflect.ValueOf(expected)
 	actualOpts := reflect.ValueOf(actual)
 
+	var expectedFuncs []*runtime.Func
+	var expectedNames []string
+	for i := 0; i < expectedOpts.Len(); i++ {
+		f := runtimeFunc(expectedOpts.Index(i).Interface())
+		expectedFuncs = append(expectedFuncs, f)
+		expectedNames = append(expectedNames, funcName(f))
+	}
+	var actualFuncs []*runtime.Func
+	var actualNames []string
+	for i := 0; i < actualOpts.Len(); i++ {
+		f := runtimeFunc(actualOpts.Index(i).Interface())
+		actualFuncs = append(actualFuncs, f)
+		actualNames = append(actualNames, funcName(f))
+	}
+
 	if expectedOpts.Len() != actualOpts.Len() {
-		expectedFmt = fmt.Sprintf("%v", expected)
-		actualFmt = fmt.Sprintf("%v", actual)
+		expectedFmt = fmt.Sprintf("%v", expectedNames)
+		actualFmt = fmt.Sprintf("%v", actualNames)
 		return
 	}
 
-	var funcNames []string
-
 	for i := 0; i < expectedOpts.Len(); i++ {
-		expectedFunc := getRuntimeFunc(expectedOpts.Index(i).Interface())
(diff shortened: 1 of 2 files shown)


arjun-1:
Looks like the approach to compare a relevant part only fails for older Go versions, I'll rethink that

arjun-1:
Hey @dolmen I believe you were involved in reviewing related change https://github.com/stretchr/testify/pull/1381. Anything specific required for this review?

pmokeev:
@dolmen can you have a look at this issue? I've also encountered the same problem

brackendawson (changes requested):
This PR would regresses this highly contrived test:
```go
package kata_test

import (
	"testing"

)

type MyStruct struct {
	A []byte
}

type Opt func(*MyStruct)

func WithString(s string) Opt {
	return func(m *MyStruct) {
		m.A = []byte(s)
	}
}

func WithBytes(b []byte) Opt {
	return func(m *MyStruct) {
		m.A = b
	}
}

type myMockFactory struct {
	mock.Mock
}

func (m *myMockFactory) New(opts ...Opt) *MyStruct {
	return m.Called(opts).Get(0).(*MyStruct)
}

func TestIfy(t *testing.T) {
	mockFactory := &myMockFactory{}
	mockFactory.Test(t)
	mockFactory.On("New", mock.FunctionalOptions(WithBytes([]byte("this")))).Return(&MyStruct{})
	mockFactory.New(WithString("this"))
}
```

The existing function name checking doesn't sit well with me either because it stripped the package name off. Can we use [reflect.Value.Pointer](https://pkg.go.dev/reflect#Value.Pointer) to check that it's the same function? The docs suggest that multiple functions can share a code pointer but I don't know in what cases that happens. If it's not practical lets drop to testing only the behaviour.

On the diff output, we should show what we compare, which in the current state of this PR is just the behaviour, so I'd like the diff to look somewhat like:
```
Diff: 0: FAIL:  &MyStruct{A:1 B:0} != &MyStruct{A:0 B:1} 
```

brackendawson:
Considering the diff some more, the function name could make a mistake more obvious to the user. But I'd want some clearer separation between the function name and the bahaviour because it's quite hard to grok, how about:
```
Diff: 0: FAIL:  WithB(*MyMock) -> &MyMock{A:0 B:1} != WithA(*MyMock) -> &MyMock{A:1 B:0}
```

brackendawson:
The docstring for `FunctionalOptions` should also be changes to state that only the behaviour of the functional option argument is compared, and that functional options with identical outcomes compare as equal. Currently it says:

> ```
> FunctionalOptions returns an [FunctionalOptionsArgument] object containing
> the expected functional-options to check for.
>
> For example:
>
>	Assert(t, FunctionalOptions(foo.Opt1("strValue"), foo.Opt2(613)))
> ```

arjun-1:
@brackendawson  I agree with you that the regression you mention is not ideal. Originally I removed the function name comparison (only focussing on behavior) to make the tests (fixing the bug of this issue) pass for older Go versions older than 1.20. The function name retrieved via `FuncForPC` yields false unexpected calls for these versions. For example

`github.com/stretchr/testify/mock.Test_Mock_AssertExpectationsFunctionalOptionsType_Indirectly.func1`
`github.com/stretchr/testify/mock.TheExampleMethodFunctionalOptionsIndirect.func1`

I investigated the solution of using `Pointer()` directly on the reflected value. The pointer value is again falsely different, as I believe it points to the inner closed anonymous function (`func1`).

I explored a different comparison using `FileLine`. It's imperfect as two different function values would not be detected as equal. But it is the only solution that prevents the regression you mention (see `Test_Mock_AssertExpectationsFunctionalOptionsType_Diff_Func`) and fixes the original issue (`Test_Mock_AssertExpectationsFunctionalOptionsType_Indirectly`) for all Go versions.

The result is now a comparison based on function file line, and if that is the same a comparison based on function behavior. Let me know if the docstring for `FunctionalOptions` should still be updated.

brackendawson (approved):
Thanks for pulling on all those threads. I'm happy to take a function matching strategy that falsely matches different functions with the same behaviour defined on the same line because the original match also wasn't complete, it considered the same function defined in different packages to be equal.

* Your proposed change fixes the described bug, anonymous functions now match. They are actually quite a common code pattern to use.
* Different functions with the same behaviour still fail (unless they are defined on the same line).
* Different functions defined in different packages with the same name and the same behaviour now fail. This is faithful to our function's promise and is such a contrived example that I'm not concerned by the change to behaviour.

I'm confident that this is a safe enough way to resolve the issue. Thanks for all your work here. I'm also going to give other maintainers some time to review.

PR #1636, opened 2024-08-29
Record Return Arguments of a Call

jayd3e:
## Summary
I had a situation where I did not have full visibility into what was returned by one of my mocked functions.  In this light, it is helpful to have the recorded arguments in the `mock.Calls` field.  I was surprised to find that for the recorded `Call`s `ReturnArguments` is left empty and is only used for `ExpectedCalls`.  This PR now properly records the ReturnArguments for a `Call`.

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -80,12 +80,12 @@ type Call struct {
 	requires []*Call
 }
 
-func newCall(parent *Mock, methodName string, callerInfo []string, methodArguments ...interface{}) *Call {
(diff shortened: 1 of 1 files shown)


brackendawson (approved):
`call.ReturnArguments` should probably have never been exported. It's also not guaranteed to be the arguments that were returned by the class embedding `mock.Mock` because there could be additional behaviour in its methods.

`mock.Calls` and `call.Arguments` are fine to be exported because they allow the "spy" pattern of mock to be implemented.

I think I'm fine with this change given we did export `ReturnArguments`. This can also be compatible with and even compliment #742 were it ever to be included.

Before I merge, what do you think @dolmen ?

PR #1637, opened 2024-09-06
mock: in order mock calls

### LINKED ISSUE #1639: (syntax sugar) Implement an `InOrder` func for mocks
## Description

Since [gomock](https://github.com/golang/mock) has been archived, I am switching to testify's mock pkg.

One extremely neat feature that gomock had is the `InOrder` func:
https://pkg.go.dev/github.com/golang/mock/gomock#InOrder

Which allowed to very easily enforce expectation order by going from:
```go
someMock.EXPECT().ValidateArgs(gomock.Any(), 1).Return(nil),
someMock.EXPECT().ValidateArgs(gomock.Any(), 2).Return(nil)
```
to
```go
gomock.InOrder(
    someMock.EXPECT().ValidateArgs(gomock.Any(), 1).Return(nil),
    someMock.EXPECT().ValidateArgs(gomock.Any(), 2).Return(nil)
)
```

As opposed to the current API implemented in testify which is more cumbersome and less intuitive with the need to declare vars and pass them inside the last chained method.
https://github.com/stretchr/testify/issues/741
https://github.com/stretchr/testify/pull/1106

```go
call1 := mockThing.On("Init").Return(nil)
call2 := mockThing.On("Do").Return(nil).NotBefore(call1)
mockThing.On("Close").Return(nil).NotBefore(call1, call2)
```

Or I guess do some kind of nesting which reverse the chronological order of events and creates nesting.
```go
mockThing.On("Close").Return(nil).NotBefore(
    mockThing.On("Do").Return(nil).NotBefore(
        mockThing.On("Init").Return(nil).NotBefore(
            ... nesting
)
```

Ideally could be turned into:

```go
mock.InOrder(
    mockThing.On("Init").Return(nil),
    mockThing.On("Do").Return(nil),
    mockThing.On("Close").Return(nil)
)
```

## Proposed solution

I didn't looked into the codebase yet to know if that's easily feasible. 
But my gut feeling says it could be trivial to chain the `NotBefore` inside the `mock.InOrder` variadic function.
This issue is mostly to gather feedback.

## Use case

More intuitive API.

ReyOrtiz:
## Summary

Add helper method to declare the order of the mock calls.

## Changes

* add `InOrder(calls ...*Call)` helper method to declare the order of mock calls by calling `NotBefore()`

## Motivation

This is probably the only feature I miss from `gomock.InOrder()`, it is more intuitive and cleaner to declare the order of the mock calls. Probably most cases can be satisfied with this method.

```
InOrder(
	mockedService1.On("TheExampleMethod", 1, 2, 3).Return(4, nil),
	mockedService2.On("TheExampleMethod2", true).Return(),
	mockedService3.On("TheExampleMethod3", "abc", 3).Return(false, nil),
)

```
## Related issues
Closes #1639

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -274,6 +274,13 @@ func (c *Call) NotBefore(calls ...*Call) *Call {
 }
 
 // InOrder defines the order in which the calls should be made
+//
+//	For example:
+//
+//	InOrder(
+//		Mock.On("init").Return(nil),
+//		Mock.On("Do").Return(nil),
+//	)
 func InOrder(calls ...*Call) {
 	for i := 1; i < len(calls); i++ {
 		calls[i].NotBefore(calls[i-1])
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -273,6 +273,13 @@ func (c *Call) NotBefore(calls ...*Call) *Call {
 	return c
 }
 
+// InOrder defines the order in which the calls should be made
+func InOrder(calls ...*Call) {
+	for i := 1; i < len(calls); i++ {
+		calls[i].NotBefore(calls[i-1])
+	}
+}
+
 // Mock is the workhorse used to track activity on another object.
 // For an example of its usage, refer to the "Example Usage" section at the top
 // of this document.
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -940,19 +940,15 @@ func Test_Mock_Return_NotBefore_In_Order(t *testing.T) {
 func Test_Mock_Return_InOrder_Uses_NotBefore(t *testing.T) {
 	var mockedService = new(TestExampleImplementation)
 
-	b := mockedService.
-		On("TheExampleMethod", 1, 2, 3).
-		Return(4, nil)
-	c := mockedService.
-		On("TheExampleMethod2", true).
-		Return()
-
 	InOrder(
-		b,
-		c,
+		mockedService.
+			On("TheExampleMethod", 1, 2, 3).
+			Return(4, nil),
+		mockedService.
+			On("TheExampleMethod2", true).
+			Return(),
 	)
 
-	require.Equal(t, []*Call{b, c}, mockedService.ExpectedCalls)
 	require.NotPanics(t, func() {
 		mockedService.TheExampleMethod(1, 2, 3)
 	})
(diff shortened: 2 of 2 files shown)


brackendawson:
Ah, an earlier issue, I suspect we'll see this a lot from gomock expats over the coming months.

I'm going to close this is a duplicate of #1639. I know this one was first but the other is more in depth. Please do go add your 👍 to it.

brackendawson (changes requested):
Can we get a unit test, please.

ReyOrtiz:
unit test added

mock/mock_test.go
```diff
@@ -937,6 +937,30 @@ func Test_Mock_Return_NotBefore_In_Order(t *testing.T) {
...
+	c := mockedService.
+		On("TheExampleMethod2", true).
+		Return()
+
+	InOrder(
+		b,
+		c,
+	)
+
+	require.Equal(t, []*Call{b, c}, mockedService.ExpectedCalls)
```
brackendawson, line 955: This line is testing the implementation. A cost effective test would only test the behaviour, as you are doing just fine after this line.

If you remove this assertion then we are free to change the implementation of the intended behaviour without breaking the test. It also removes the need to declare `b` and `c`, so the test can be written as `InOrder` is intended to be used.

mock/mock.go
```diff
@@ -273,6 +273,13 @@ func (c *Call) NotBefore(calls ...*Call) *Call {
 	return c
 }
 
+// InOrder defines the order in which the calls should be made
```
brackendawson, line 276: This could do with a small usage example for the docs.

brackendawson (changes requested):
Positive and negative tests, nice one. 👍  Just a couple of small comments.

mock/mock_test.go
brackendawson: Also, it's just comparing pointer values. If InOrder did literally nothing then that assertion would still pass.

mock/mock.go
brackendawson: It's also the order they need to be "satisfied", not the order they need to be called. In the case where they can be called multiple times.

ReyOrtiz:
Good comments, thanks! Requested changes applied.

brackendawson (approved):
Looks good to me, should be helpful to all the developers moving from gomock. I'll give some time for other maintainers to review before I merge.

jybp:
Ah you beat me to it, thanks!

ReyOrtiz:
Hello, any ETA on merging this PR?

brackendawson:
Thanks @ReyOrtiz. This can be consumed now by gomock expats with: `go get github.com/stretchr/testify@master`. Once testify v1.10 is released it can be consumed as per the README. I'll have a look at drafting a release soon.

ReyOrtiz:
Thanks @brackendawson for all the help! 🙏

PR #1644, opened 2024-09-30
mock: caller information for unexpected method call

spirin:
## Summary
When using mocks in large integration tests, it is very difficult to find places of unexpected function calls.

## Changes
I suggest adding information about the location of the call.

## Motivation
371770366-d54f21fd-e916-4ce9-805a-856a80bd3a20

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -491,11 +491,12 @@ func (m *Mock) MethodCalled(methodName string, arguments ...interface{}) Argumen
 		m.mutex.Unlock()
 
 		if closestCall != nil {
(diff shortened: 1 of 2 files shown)


brackendawson (changes requested):
This makes sense as an addition, especially given the else case already does it.

Can you update a test too please? Possibly `TestArgumentMatcherToPrintMismatch` but maybe something else. Confirming that the test's own file base name is in the stack should be sufficient.

spirin:
> Can you update a test too please? Possibly `TestArgumentMatcherToPrintMismatch` but maybe something else. Confirming that the test's own file base name is in the stack should be sufficient.

Done

brackendawson (approved):
Thank you.

Issue #934, opened 2020-04-27
TearDownTest is being executed while parallel sub tests are still running


swithek:
Consider the following example:
```go
type Suite struct {
	suite.Suite
}

func (s *Suite) TearDownSuite() {
	s.T().Log(">>>>>> suite tear down")
}

func (s *Suite) TearDownTest() {
	s.T().Log(">>>>>> single test tear down")
}

func (s *Suite) TestOne() {
	for _, v := range []string{"sub1", "sub2", "sub3"} {
		s.Run(v, func() {
			s.T().Parallel()
		})
	}
}

func (s *Suite) TestTwo() {
	for _, v := range []string{"sub1", "sub2", "sub3"} {
		s.Run(v, func() {
			s.T().Parallel()
		})
	}
}

func TestLogic(t *testing.T) {
	suite.Run(t, &Suite{})
}
```

After executing that you would get a similar output to this:
```
=== RUN   TestLogic
=== RUN   TestLogic/TestOne
=== RUN   TestLogic/TestOne/sub1
=== PAUSE TestLogic/TestOne/sub1
=== RUN   TestLogic/TestOne/sub2
=== PAUSE TestLogic/TestOne/sub2
=== RUN   TestLogic/TestOne/sub3
=== PAUSE TestLogic/TestOne/sub3
    TestLogic/TestOne: prog.go:18: >>>>>> single test tear down
=== CONT  TestLogic/TestOne/sub1
=== CONT  TestLogic/TestOne/sub3
=== CONT  TestLogic/TestOne/sub2 >>> test finished here
=== RUN   TestLogic/TestTwo
=== RUN   TestLogic/TestTwo/sub1
=== PAUSE TestLogic/TestTwo/sub1
=== RUN   TestLogic/TestTwo/sub2
=== PAUSE TestLogic/TestTwo/sub2
=== RUN   TestLogic/TestTwo/sub3
=== PAUSE TestLogic/TestTwo/sub3
    TestLogic/TestTwo: prog.go:18: >>>>>> single test tear down
=== CONT  TestLogic/TestTwo/sub1
=== CONT  TestLogic/TestTwo/sub3
=== CONT  TestLogic/TestTwo/sub2 >>> test finished here
    TestLogic: prog.go:14: >>>>>> suite tear down
```

Live example: https://play.golang.com/p/fPY3DZFNNok

[...]

PR #1645, opened 2024-10-01
Document suite's lack of support for t.Parallel

brackendawson:
## Summary
Document that at this time suite cannot support Parallel tests

## Changes
Document this in README.md and suite's godoc.

## Motivation
Suite will always be broken when using parallel tests in testify v1. The only clean fix is a breaking change described in #1109 

## Related issues
Related to
* #934 
* #187


Antonboom:
Covered by [testifylint#suite-broken-parallel](https://github.com/Antonboom/testifylint?tab=readme-ov-file#suite-broken-parallel)

dolmen (approved):

PR #1646, opened 2024-10-03
assert: truncate very long objects in test failure messages

### LINKED ISSUE #1525: assert/require.Len doesn't print anything if the slice is too long

brackendawson:
## Summary
Truncate object representations in test failure output messages that would cause the message to be too long to scan.

This is a different approach than that in #1559. I feel this is more appropriate because is it safer. The other approach has the potential to allocate excessive memory and to print excessive output to the console.

## Changes
* Make the message line indenter show an appropriate error if it fails to scan a line from the message.
* Use the existing `truncatingFormat` function to truncate potentially long objects in assertions which format expected/actual values in full.
* Parameterize the format string used by `truncatingFormat`.
* Reduce the limit used by `truncatingFormat` by almost half to support failure messages showing both long expected and long actual values on the same line.
* Change the format of `NotSubset` from `%q` to `%#v` to match that of `Subset`.

## Motivation
So that people don't see no output in the failure message when making assertions (such as `Len`) against very long objects.

## Related issues
Closes #1525

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -622,9 +622,9 @@ func formatUnequalValues(expected, actual interface{}) (e string, a string) {
 func truncatingFormat(format string, data interface{}) string {
 	value := fmt.Sprintf(format, data)
 	// Give us space for two truncated objects and the surrounding sentence.
-	max := bufio.MaxScanTokenSize/2 - 100
-	if len(value) > max {
-		value = value[0:max] + "<... truncated>"
+	maxMessageSize := bufio.MaxScanTokenSize/2 - 100
+	if len(value) > maxMessageSize {
+		value = value[0:maxMessageSize] + "<... truncated>"
 	}
 	return value
 }
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -325,13 +325,15 @@ func messageFromMsgAndArgs(msgAndArgs ...interface{}) string {
 func indentMessageLines(message string, longestLabelLen int) string {
 	outBuf := new(bytes.Buffer)
 
-	for i, scanner := 0, bufio.NewScanner(strings.NewReader(message)); scanner.Scan(); i++ {
-		// no need to align first line because it starts at the correct location (after the label)
-		if i != 0 {
-			// append alignLen+1 spaces to align with "{{longestLabel}}:" before adding tab
-			outBuf.WriteString("\n\t" + strings.Repeat(" ", longestLabelLen+1) + "\t")
+	scanner := bufio.NewScanner(strings.NewReader(message))
+	for firstLine := true; scanner.Scan(); firstLine = false {
+		if !firstLine {
+			fmt.Fprint(outBuf, "\n\t"+strings.Repeat(" ", longestLabelLen+1)+"\t")
 		}
-		outBuf.WriteString(scanner.Text())
+		fmt.Fprint(outBuf, scanner.Text())
+	}
+	if err := scanner.Err(); err != nil {
+		return fmt.Sprintf("cannot display message: %s", err)
 	}
 
 	return outBuf.String()
@@ -544,9 +546,8 @@ func Same(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) b
 	if !same {
 		// both are pointers but not the same type & pointing to the same address
 		return Fail(t, fmt.Sprintf("Not same: \n"+
-			"expected: %#[1]v (%[1]T)(%[1]p)\n"+
-			"actual  : %#[2]v (%[2]T)(%[2]p)",
-			expected, actual), msgAndArgs...)
+			"expected: %[2]s (%[1]T)(%[1]p)\n"+
+			"actual  : %[4]s (%[3]T)(%[3]p)", expected, truncatingFormat("%#v", expected), actual, truncatingFormat("%#v", actual)), msgAndArgs...)
 	}
 
 	return true
@@ -571,8 +572,8 @@ func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
 
 	if same {
 		return Fail(t, fmt.Sprintf(
-			"Expected and actual point to the same object: %p %#[1]v",
-			expected), msgAndArgs...)
+			"Expected and actual point to the same object: %p %s",
+			expected, truncatingFormat("%#v", expected)), msgAndArgs...)
 	}
 	return true
 }
@@ -604,23 +605,24 @@ func samePointers(first, second interface{}) (same bool, ok bool) {
 // to a type conversion in the Go grammar.
 func formatUnequalValues(expected, actual interface{}) (e string, a string) {
 	if reflect.TypeOf(expected) != reflect.TypeOf(actual) {
-		return fmt.Sprintf("%T(%s)", expected, truncatingFormat(expected)),
-			fmt.Sprintf("%T(%s)", actual, truncatingFormat(actual))
+		return fmt.Sprintf("%T(%s)", expected, truncatingFormat("%#v", expected)),
+			fmt.Sprintf("%T(%s)", actual, truncatingFormat("%#v", actual))
 	}
 	switch expected.(type) {
 	case time.Duration:
 		return fmt.Sprintf("%v", expected), fmt.Sprintf("%v", actual)
 	}
-	return truncatingFormat(expected), truncatingFormat(actual)
+	return truncatingFormat("%#v", expected), truncatingFormat("%#v", actual)
 }
 
 // truncatingFormat formats the data and truncates it if it's too long.
 //
 // This helps keep formatted error messages lines from exceeding the
 // bufio.MaxScanTokenSize max line length that the go testing framework imposes.
-func truncatingFormat(data interface{}) string {
-	value := fmt.Sprintf("%#v", data)
-	max := bufio.MaxScanTokenSize - 100 // Give us some space the type info too if needed.
+func truncatingFormat(format string, data interface{}) string {
+	value := fmt.Sprintf(format, data)
+	// Give us space for two truncated objects and the surrounding sentence.
+	max := bufio.MaxScanTokenSize/2 - 100
 	if len(value) > max {
 		value = value[0:max] + "<... truncated>"
 	}
@@ -743,7 +745,7 @@ func Nil(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
-	return Fail(t, fmt.Sprintf("Expected nil, but got: %#v", object), msgAndArgs...)
+	return Fail(t, fmt.Sprintf("Expected nil, but got: %s", truncatingFormat("%#v", object)), msgAndArgs...)
 }
 
 // isEmpty gets whether the specified object is considered empty or not.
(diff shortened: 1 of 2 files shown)


assert/assertions.go
```diff
@@ -604,23 +605,24 @@ func samePointers(first, second interface{}) (same bool, ok bool) {
...
 //
 // This helps keep formatted error messages lines from exceeding the
 // bufio.MaxScanTokenSize max line length that the go testing framework imposes.
-func truncatingFormat(data interface{}) string {
-	value := fmt.Sprintf("%#v", data)
-	max := bufio.MaxScanTokenSize - 100 // Give us some space the type info too if needed.
+func truncatingFormat(format string, data interface{}) string {
+	value := fmt.Sprintf(format, data)
+	// Give us space for two truncated objects and the surrounding sentence.
+	max := bufio.MaxScanTokenSize/2 - 100
```
ccoVeille: Do not shadow built-in `max` function

assert/assertions.go
```diff
@@ -793,7 +795,7 @@ func Empty(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 		if h, ok := t.(tHelper); ok {
 			h.Helper()
 		}
-		Fail(t, fmt.Sprintf("Should be empty, but was %v", object), msgAndArgs...)
+		Fail(t, fmt.Sprintf("Should be empty, but was %s", truncatingFormat("%v", object)), msgAndArgs...)
```
ccoVeille, line 798: Others are like this

```suggestion
		Fail(t, fmt.Sprintf("Should be empty, but was %s", truncatingFormat("%#v", object)), msgAndArgs...)
```

assert/assertions.go
```diff
@@ -836,11 +838,11 @@ func Len(t TestingT, object interface{}, length int, msgAndArgs ...interface{})
 	}
 	l, ok := getLen(object)
 	if !ok {
-		return Fail(t, fmt.Sprintf("\"%v\" could not be applied builtin len()", object), msgAndArgs...)
+		return Fail(t, fmt.Sprintf("%q could not be applied builtin len()", truncatingFormat("%v", object)), msgAndArgs...)
```
ccoVeille, line 841:
```suggestion
		return Fail(t, fmt.Sprintf("%q could not be applied builtin len()", truncatingFormat("%#v", object)), msgAndArgs...)
```

assert/assertions.go
```diff
@@ -836,11 +838,11 @@ func Len(t TestingT, object interface{}, length int, msgAndArgs ...interface{})
...
 	}
 
 	if l != length {
-		return Fail(t, fmt.Sprintf("\"%v\" should have %d item(s), but has %d", object, length, l), msgAndArgs...)
+		return Fail(t, fmt.Sprintf("%q should have %d item(s), but has %d", truncatingFormat("%v", object), length, l), msgAndArgs...)
```
ccoVeille, line 845:
```suggestion
		return Fail(t, fmt.Sprintf("%q should have %d item(s), but has %d", truncatingFormat("%#v", object), length, l), msgAndArgs...)
```

assert/assertions.go
```diff
@@ -1056,7 +1058,7 @@ func Subset(t TestingT, list, subset interface{}, msgAndArgs ...interface{}) (ok
 			return Fail(t, fmt.Sprintf("%#v could not be applied builtin len()", list), msgAndArgs...)
 		}
 		if !found {
-			return Fail(t, fmt.Sprintf("%#v does not contain %#v", list, element), msgAndArgs...)
+			return Fail(t, fmt.Sprintf("%s does not contain %#v", truncatingFormat("%#v", list), element), msgAndArgs...)
```
ccoVeille, line 1061:
```suggestion
			return Fail(t, fmt.Sprintf("%s does not contain %s", truncatingFormat("%#v", list), element), msgAndArgs...)
```

assert/assertions.go
```diff
@@ -1700,7 +1702,7 @@ func ErrorContains(t TestingT, theError error, contains string, msgAndArgs ...in
 
 	actual := theError.Error()
 	if !strings.Contains(actual, contains) {
-		return Fail(t, fmt.Sprintf("Error %#v does not contain %#v", actual, contains), msgAndArgs...)
+		return Fail(t, fmt.Sprintf("Error %s does not contain %#v", truncatingFormat("%#v", actual), contains), msgAndArgs...)
```
ccoVeille, line 1705:
```suggestion
		return Fail(t, fmt.Sprintf("Error %s does not contain %s", truncatingFormat("%#v", actual), contains), msgAndArgs...)
```

assert/assertions.go
```diff
@@ -1031,10 +1033,10 @@ func Subset(t TestingT, list, subset interface{}, msgAndArgs ...interface{}) (ok
 			av := actualMap.MapIndex(k)
 
 			if !av.IsValid() {
-				return Fail(t, fmt.Sprintf("%#v does not contain %#v", list, subset), msgAndArgs...)
+				return Fail(t, fmt.Sprintf("%s does not contain %s", truncatingFormat("%#v", list), truncatingFormat("%#v", subset)), msgAndArgs...)
 			}
 			if !ObjectsAreEqual(ev.Interface(), av.Interface()) {
-				return Fail(t, fmt.Sprintf("%#v does not contain %#v", list, subset), msgAndArgs...)
+				return Fail(t, fmt.Sprintf("%s does not contain %s", truncatingFormat("%#v", list), truncatingFormat("%#v", subset)), msgAndArgs...)
```
ccoVeille, line 1039: I'm a bit worried about these 2

We have 2 list that could be huge that would be truncated here, but displayed on the same line.

Maybe it works because of the max/2-200, but I feel like the line might difficult to read because of the lack of \n that we have for almost all others failures

assert/assertions.go
```diff
@@ -1106,7 +1108,7 @@ func NotSubset(t TestingT, list, subset interface{}, msgAndArgs ...interface{})
 			}
 		}
 
-		return Fail(t, fmt.Sprintf("%q is a subset of %q", subset, list), msgAndArgs...)
+		return Fail(t, fmt.Sprintf("%s is a subset of %s", truncatingFormat("%#v", subset), truncatingFormat("%#v", list)), msgAndArgs...)
```
ccoVeille: See https://github.com/stretchr/testify/pull/1646#discussion_r2353499286

assert/assertions.go
```diff
@@ -1128,7 +1130,7 @@ func NotSubset(t TestingT, list, subset interface{}, msgAndArgs ...interface{})
 		}
 	}
 
-	return Fail(t, fmt.Sprintf("%q is a subset of %q", subset, list), msgAndArgs...)
+	return Fail(t, fmt.Sprintf("%s is a subset of %s", truncatingFormat("%#v", subset), truncatingFormat("%#v", list)), msgAndArgs...)
```
ccoVeille: See https://github.com/stretchr/testify/pull/1646#discussion_r2353499286

assert/assertions_test.go
```diff
@@ -3969,3 +3969,244 @@ func TestNotErrorAs(t *testing.T) {
...
+func TestContainsWithSliceTooLongToPrint(t *testing.T) {
+	t.Parallel()
+	mockT := new(mockTestingT)
+	longSlice := make([]int, 1_000_000)
+	Contains(mockT, longSlice, 1)
+	Contains(t, mockT.errorString(), `
+	Error Trace:	
+	Error:      	[]int{0, 0, 0,`)
+	Contains(t, mockT.errorString(), `<... truncated> does not contain 1`)
+}
```
ccoVeille, line 3993: Can't we define the max value via a global variable used only in tests. Defined to something like 10, and then we check the exact rendering, and not something that check for piece of text

assert/assertions.go
brackendawson: This instance of truncatingFormat must use `%v` because the un-truncated format it replaces does, otherwise it would change the format of the message. This PR only truncates messages that are so long that they cause Go to output no message at all.

assert/assertions.go
brackendawson: This one was `\"%v\"`, which is equivalent to  `\"%s\"` with truncatingFormat(%v), which is equivelant to `%q` with truncatingFormat(%v)

assert/assertions.go
brackendawson: This one was `\"%v\"`, which is equivalent to  `\"%s\"` with truncatingFormat(%v), which is equivelant to `%q` with truncatingFormat(%v)

assert/assertions.go
brackendawson: The second formatted variable (uses `element`) must not change.

assert/assertions.go
brackendawson: I agree about the hard to read part, but that would be a different issue. Again the only aim here is to show strings which previously got swallowed.

assert/assertions.go
brackendawson: I am actually wrong here, it should be:
```suggestion
		return Fail(t, fmt.Sprintf("%s is a subset of %s", truncatingFormat("%q", subset), truncatingFormat("%q", list)), msgAndArgs...)
```
%q quotes the elements of a list o.0

I've raised #1800 to fix the formatting later.

assert/assertions_test.go
brackendawson: We can't because that wouldn't prove that the truncated messages make it through the scanner.

brackendawson: We can't because that wouldn't prove that the truncated message makes it through the reader.

brackendawson:
Thank you @ccoVeille. I've updated the PR to address some comments and responded with a defence in the others.

assert/assertions.go
brackendawson: The second formatted variable (uses `contains`) must not change.

assert/assertions.go
ccoVeille: I understand your point. But mine wasn't about readability.

I just wanted to make sure that comparing two slices that very long won't lead to a line that cannot be displayed.

I feel like that tge point of the max/2-200 but I wanted to make sure it would work.

ccoVeille (approved):

assert/assertions.go
brackendawson: Yes, that doesn't sit 100% with me either. When we follow this up with #1235 and reduce the max length much further I will be very confident.

PR #1653, opened 2024-10-04
Fix Eventually and EventuallyWithT to always run condition

### LINKED ISSUE #1652: assert.Eventually can fail having never run condition
## Description
Witnessed a test flake on one of the go versions (1.21) in CI from the master branch because the assert package took too long: https://github.com/stretchr/testify/actions/runs/11182446108/job/31088748678

## Step To Reproduce
Run CI.

## Expected behavior
CI passes.

## Actual behavior
Tests timed out.

brackendawson:
## Summary
If `Eventually` was called with a `tick` larger than `waitFor`, or a `tick` very slightly smaller than `waitFor`, or very small `tick` and `waitFor` values, then a timing race can occur where the select statement falls into the `waitFor` case first and the `condition` function is never run.

## Changes
* Add a timeout to the unit test so that a deadlock is found  after 1 second rather than 600 seconds.
* Swap the order that the ticker and timer are created so that in the case that `tick` is very slightly smaller than `waitFor`, then the timer will never yield before the ticker.
* Run `condition` once before the loop and initialise `tick` in the for loop to `(<-chan time.Time)(nil)` so that we always run `condition` statically, then check its result, then run `condition` based on ticks.

## Motivation
Having a timing race where it is possible for condition to never be run causes our CI tests to flake, and is not how I think the user would expect Eventually can behave.

## Related issues
Closes #1652

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1932,13 +1932,15 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 
 	ch := make(chan bool, 1)
 
-	timer := time.NewTimer(waitFor)
-	defer timer.Stop()
-
 	ticker := time.NewTicker(tick)
 	defer ticker.Stop()
 
-	for tick := ticker.C; ; {
+	timer := time.NewTimer(waitFor)
+	defer timer.Stop()
+
+	go func() { ch <- condition() }()
+
+	for tick := (<-chan time.Time)(nil); ; {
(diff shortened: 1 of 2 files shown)


brackendawson:
This might be a dupe of #1427

brackendawson:
The same fix is needed in `Never`.

brackendawson:
#1657 is a far better approach I think.

PR #1663, opened 2024-10-16
Fix issue #1662 (comparing infs should fail)

### LINKED ISSUE #1662: assert.InEpsilon has inconsistent results with expected or actual values of Inf
```go
func TestInEpsilon1(t *testing.T) {
	div := 0.0
	require.InEpsilon(t, 1, 1/div, 0.001)
}

func TestInEpsilon2(t *testing.T) {
	div := 0.0
	require.InEpsilon(t, 1/div, 1, 0.001)
}
```

```
> go test -v -run InEps | grep -E '(PASS|FAIL)'
--- FAIL: TestInEpsilon1 (0.00s)
--- PASS: TestInEpsilon2 (0.00s)
```

_Originally posted by @ybrustin-te in https://github.com/stretchr/testify/issues/918#issuecomment-2416889992_

Given the difference between two Inf values cannot be determined, those assertions should both fail.

ybrustin:
## Summary
Comparing Infs should fail as described in related issue

## Related issues
Closes #1662

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1527,6 +1527,9 @@ func InEpsilon(t TestingT, expected, actual interface{}, epsilon float64, msgAnd
 	if err != nil {
 		return Fail(t, err.Error(), msgAndArgs...)
 	}
+	if math.IsNaN(actualEpsilon) {
+		return Fail(t, "relative error is NaN", msgAndArgs...)
+	}
 	if actualEpsilon > epsilon {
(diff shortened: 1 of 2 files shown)


brackendawson (changes requested):
👌 but you can now also delete the NaN check at line 1523.

brackendawson:
Ah, possibly not, we just don't have a test case for that.

brackendawson (approved):
Looks good. Going to allow other maintainers a bit of time to comment.

PR #1664, opened 2024-10-17
NotSame should fail if args are not pointers #1661

### LINKED ISSUE #1661: `NotSame` should fail if args are not pointers
### Discussed in https://github.com/stretchr/testify/discussions/1660

<div type='discussions-op-text'>

<sup>Originally posted by **n4nn31355** October 16, 2024</sup>
It's confusing and error-prone that `NotSame` checks nothing if any of the values is not a pointer.
Is there a reason for this, or may we consider it a bug that needs an issue?
```
a := struct{ byte }{}
require.NotSame(t, a, a)   // Doesn't fail
require.NotSame(t, &a, &a) // Fails
```</div>

### PR DESCRIPTION
**Pull request by sikehish**
## Summary
Reduces the confusion assosciated with NotSame that previously would check nothing if any of the values is not a pointer. The changes made were tested using TestSame, TestNotSame, and Test_samePointers tests, and the changes did clear the tests.

## Changes
1. Modified samePointers to return another bool value(ok) that would determine if the 2 values are of pointer type or not, while the returned "same" bool value would determine if the 2 pointers are same.
2. Modified assert.NotSame to call Fail() if the 2 values are either i)pointers pointing to the same address or ii)either/both of the values are not pointers.
3.  Modified assert.Same to call Fail() if the 2 values are either i)pointers not pointing to the same address or ii)either/both of the values are not pointers.
4. Modified Test_samePointers to handle the new behavior of samePointers by checking both if the inputs are pointers (ok) and if they point to the same object (same).

## Motivation
Ensure NotSame accurately verifies pointer sameness by handling non-pointer inputs explicitly, improving clarity and reducing potential misuse.

## Related issues
Closes #1661

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -502,7 +502,13 @@ func Same(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) b
 		h.Helper()
 	}
 
-	if !samePointers(expected, actual) {
+	same, ok := samePointers(expected, actual)
+	if !ok {
+		return Fail(t, "Both arguments must be pointers", msgAndArgs...)
+	}
+
+	//both are pointers but pointing to different addresses
+	if !same {
 		return Fail(t, fmt.Sprintf("Not same: \n"+
 			"expected: %p %#v\n"+
 			"actual  : %p %#v", expected, expected, actual, actual), msgAndArgs...)
@@ -521,13 +527,19 @@ func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
-
-	if samePointers(expected, actual) {
-		return Fail(t, fmt.Sprintf(
-			"Expected and actual point to the same object: %p %#v",
-			expected, expected), msgAndArgs...)
+	if same, ok := samePointers(expected, actual); ok {
+		//if ok is true, then both arguments are pointers
+		if same {
+			return Fail(t, fmt.Sprintf(
+				"Expected and actual point to the same object: %p %#v",
+				expected, expected), msgAndArgs...) //because Fail return false, but we need to
+		}
+		return true
 	}
-	return true
+
+	//fails when the arguments are not pointers
+	return !(Fail(t, "Both arguments must be pointers", msgAndArgs...))
+	//we return !Fail because Fail return false, but we need to return true if the arguments are not pointers(by default)
 }
 
 // samePointers compares two generic interface objects and returns whether
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -542,10 +542,10 @@ func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
 	return true
 }
 
-// samePointers checks if two generic interface objects are pointers to
-// the same object. It returns two values: a boolean indicating if they
-// point to the same object, and another boolean indicating whether both
-// inputs are valid pointers of the same type.
+// samePointers checks if two generic interface objects are pointers of the same
+// type pointing to the same object. It returns two values: same indicating if
+// they are the same type and point to the same object, and ok indicating that
+// both inputs are pointers.
 func samePointers(first, second interface{}) (same bool, ok bool) {
 	firstPtr, secondPtr := reflect.ValueOf(first), reflect.ValueOf(second)
 	if firstPtr.Kind() != reflect.Ptr || secondPtr.Kind() != reflect.Ptr {
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -527,7 +527,10 @@ func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
-	if same, ok := samePointers(expected, actual); ok {
+
+	same, ok := samePointers(expected, actual)
+
+	if ok {
 		//if ok is true, then both arguments are pointers
 		if same {
 			return Fail(t, fmt.Sprintf(
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -542,8 +542,10 @@ func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
 	//we return !Fail because Fail return false, but we need to return true if the arguments are not pointers(by default)
 }
 
-// samePointers compares two generic interface objects and returns whether
-// they point to the same object
+// samePointers checks if two generic interface objects are pointers to
+// the same object. It returns two values: a boolean indicating if they
+// point to the same object, and another boolean indicating whether both
+// inputs are valid pointers of the same type.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: assert/assertions.go [hunk 1]
```diff
@@ -521,30 +527,36 @@ func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
...
+
+	//fails when the arguments are not pointers
+	return !(Fail(t, "Both arguments must be pointers", msgAndArgs...))
+	//we return !Fail because Fail return false, but we need to return true if the arguments are not pointers(by default)
 }
 
 // samePointers compares two generic interface objects and returns whether
 // they point to the same object
-func samePointers(first, second interface{}) bool {
+func samePointers(first, second interface{}) (same bool, ok bool) {
```
brackendawson, line 547: The doc-string for the internal samePointers needs updating. It wasn't strictly correct before as it doesn't mention the type checking.

assert/assertions.go
```diff
@@ -502,7 +502,13 @@ func Same(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) b
...
 	}
 
-	if !samePointers(expected, actual) {
+	same, ok := samePointers(expected, actual)
+	if !ok {
+		return Fail(t, "Both arguments must be pointers", msgAndArgs...)
+	}
+
+	//both are pointers but pointing to different addresses
+	if !same {
```
brackendawson, line 511: This comment isn't true in that location, it also doesn't mention the types.
```suggestion
	if !same {
		// both are pointers but not the same type & pointing to the same address
```

assert/assertions.go
```diff
@@ -521,30 +527,36 @@ func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
-
-	if samePointers(expected, actual) {
-		return Fail(t, fmt.Sprintf(
-			"Expected and actual point to the same object: %p %#v",
-			expected, expected), msgAndArgs...)
+	if same, ok := samePointers(expected, actual); ok {
```
brackendawson, line 530: Flip the condition on this if to `!ok` to preserve [line of sight](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88).

assert/assertions_test.go
```diff
@@ -637,7 +637,12 @@ func Test_samePointers(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			tt.assertion(t, samePointers(tt.args.first, tt.args.second))
+			same, ok := samePointers(tt.args.first, tt.args.second)
+			if !ok {
+				tt.assertion(t, false) // both arguments must be pointers
```
brackendawson, line 642: You can't use `tt.assertion` for this, it might be expecting `false`.

Rename `assertion` to `same` and add another field called `ok` using the same pattern. Then add tests to the table which pass the three combinations of non-pointers.

brackendawson (changes requested):

assert/assertions.go
sikehish: > Flip the condition on this if to `!ok` to preserve [line of sight](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88).

I've instead split it into 2 different statements to make it more readable and also to preserve line of sight:

```
       same, ok := samePointers(expected, actual)

	if ok {}
```

assert/assertions_test.go
sikehish: Got it! I've made the necessary changes

sikehish:
Hi, @brackendawson.
I've made all the requested amendments. Do let me know if any other changes are to be made :)

assert/assertions.go
```diff
@@ -522,29 +528,40 @@ func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
 		h.Helper()
 	}
 
-	if samePointers(expected, actual) {
-		return Fail(t, fmt.Sprintf(
-			"Expected and actual point to the same object: %p %#v",
-			expected, expected), msgAndArgs...)
+	same, ok := samePointers(expected, actual)
+
+	if ok {
```
brackendawson, line 533: Almost there, if you flip this to `if !ok {` then you can avoid the nested if statements and fully preserve line of sight.

assert/assertions_test.go
```diff
@@ -605,39 +605,47 @@ func Test_samePointers(t *testing.T) {
 		second interface{}
 	}
 	tests := []struct {
-		name      string
-		args      args
-		assertion BoolAssertionFunc
+		name string
+		args args
+		same BoolAssertionFunc
+		ok   BoolAssertionFunc
```
brackendawson, line 611: That's perfect. 👌

Can you add a couple more test cases for when the left and right arguments aren't pointers but the other one is, like:
* `args{first: 1, second: p}`
* `args{first: p, second: 1}`

brackendawson (changes requested):
Sorry for the late re-review, that's looking good, just a couple more bits.

assert/assertions.go
brackendawson: Also we don't really want the blank line first as this is "error handling" for the line above.

assert/assertions_test.go
sikehish: Yup, I've added 2 additional test cases now

assert/assertions.go
sikehish: I've made the changes

sikehish:
Hi @brackendawson. I've made the requested changes. Let me know if any other changes need to be made.

assert/assertions.go
```diff
@@ -522,29 +528,37 @@ func NotSame(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}
...
 }
 
-// samePointers compares two generic interface objects and returns whether
-// they point to the same object
-func samePointers(first, second interface{}) bool {
+// samePointers checks if two generic interface objects are pointers to
+// the same object. It returns two values: a boolean indicating if they
+// point to the same object, and another boolean indicating whether both
+// inputs are valid pointers of the same type.
+func samePointers(first, second interface{}) (same bool, ok bool) {
```
brackendawson, line 549: Sorry, I should have noticed this on the last pass. The type comparison is yielded in `same` and not in `ok`.

```suggestion
// samePointers checks if two generic interface objects are pointers of the same
// type pointing to the same object. It returns two values: same indicating if
// they are the same type and point to the same object, and ok indicating that
// both inputs are pointers.
func samePointers(first, second interface{}) (same bool, ok bool) {
```

brackendawson (changes requested):
Code is spot on 👍 +1:, just found one doc-string inaccuracy.

assert/assertions.go
sikehish: I've made the change :) Let me know if any other changes are to be made

brackendawson (approved):
LGTM, thank you for your contribution.

sikehish:
> LGTM, thank you for your contribution.

Thanks for the opportunity :)

PR #1665, opened 2024-10-22
suite: validate method signatures and continue execution for valid tests

### LINKED ISSUE #1508: test panicked: reflect: Call with too few input arguments

vyas-git:
## Summary
This PR improves method signature validation in Testify's suite execution by ensuring that invalid methods are logged, and valid methods continue to execute.

## Changes
* Added validation of method signatures (checking for arguments and return values) at an earlier stage, after method filtering.
* Replaced `t.Fatalf` with `t.Errorf` to report invalid method signatures while allowing the test suite to continue executing valid methods.
* Introduced additional test cases to cover both valid and invalid method signatures.

## Motivation
* To ensure that invalid method signatures in the suite are reported without stopping execution, allowing valid methods to run and improving overall test suite behavior.

## Related issues
* Closes #1508

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -164,7 +164,7 @@ func Run(t *testing.T, suite TestingSuite) {
 				name: method.Name,
 				run: func(t *testing.T) {
 					t.Errorf(
-						"testify: suite method %q has invalid signature: expected no inputs or return values (has %d inputs, %d outputs)",
+						"testify: suite method %q has invalid signature: expected no input or output parameters, method has %d input parameters and %d output parameters",
 						method.Name, method.Type.NumIn()-1, method.Type.NumOut(),
 					)
 				},
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -149,7 +149,19 @@ func Run(t *testing.T, suite TestingSuite) {
 		if !ok {
 			continue
 		}
-
+		// Check method signature
+		if method.Type.NumIn() > 1 || method.Type.NumOut() > 0 {
+			tests = append(tests, test{
+				name: method.Name,
+				run: func(t *testing.T) {
+					t.Errorf(
+						"testify: suite method %q has invalid signature: expected no inputs or return values (has %d inputs, %d outputs)",
+						method.Name, method.Type.NumIn()-1, method.Type.NumOut(),
+					)
+				},
+			})
+			continue
+		}
 		test := test{
 			name: method.Name,
 			run: func(t *testing.T) {
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -751,3 +751,65 @@ func TestUnInitializedSuites(t *testing.T) {
 		})
 	})
 }
+
+// SuiteSignatureValidationTester tests valid and invalid method signatures.
+type SuiteSignatureValidationTester struct {
+	Suite
+
+	executedTestCount int
+	setUp             bool
+	toreDown          bool
+}
+
+// SetupSuite runs once before any tests.
+func (s *SuiteSignatureValidationTester) SetupSuite() {
+	s.setUp = true
+}
+
+// TearDownSuite runs once after all tests.
+func (s *SuiteSignatureValidationTester) TearDownSuite() {
+	s.toreDown = true
+}
+
+// Valid test method — should run.
+func (s *SuiteSignatureValidationTester) TestValidSignature() {
+	s.executedTestCount++
+}
+
+// Invalid: has return value.
+func (s *SuiteSignatureValidationTester) TestInvalidSignatureReturnValue() interface{} {
+	s.executedTestCount++
+	return nil
+}
+
+// Invalid: has input arg.
+func (s *SuiteSignatureValidationTester) TestInvalidSignatureArg(somearg string) {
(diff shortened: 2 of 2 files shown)


brackendawson (changes requested):
This error shows up when you run a valid test. Previously there is no panic in this scenario.

eg:
```go
package kata_test

// Basic imports
import (
	"testing"

)

// Define the suite, and absorb the built-in basic suite
// functionality from testify - including a T() method which
// returns the current testing context
	suite.Suite
	VariableThatShouldStartAtFive int
}

// Make sure that VariableThatShouldStartAtFive is set to five
// before each test
func (suite *ExampleTestSuite) SetupTest() {
	suite.VariableThatShouldStartAtFive = 5
}

// All methods that begin with "Test" are run as tests within a
// suite.
	assert.Equal(suite.T(), 6, suite.VariableThatShouldStartAtFive)
}

func (suite *ExampleTestSuite) TestSomething(a any) {
	suite.Equal(1, a)
}

// In order for 'go test' to run this suite, we need to create
// a normal test function and pass our suite to suite.Run
}
```

```
% go test -run=TestExampleTestSuite/TestExample
--- FAIL: TestExampleTestSuite (0.00s)
    suite.go:153: testify: suite method 'TestSomething' has 1 input arguments and 0 return values. It should have none.
FAIL
exit status 1
FAIL    github.com/brackendawson/kata   0.280s
```

vyas-git:
@brackendawson Thank you for pointing that out! I understand the concern.

Since `t` is tied to the suite instance, this caused the entire suite to be marked as failed whenever an invalid method was detected—even when running a specific test like `TestExampleTestSuite/TestExample`.

The suite considers all its methods, and `methodFilter` cannot filter down to a specific test.

I considered using `t.Skip` within a `testing.InternalTest` case, but creating extra tests for invalid methods just to call `t.Skip` felt unnecessary. Skipping invalid methods early with a warning message turned out to be a cleaner solution.

To address this, I updated the approach to log a warning message directly and skip invalid methods without using `t.Errorf`. This change ensures:

1. Invalid methods are skipped, but valid methods continue to execute without marking the entire suite as failed.
2. Only the requested test methods are executed, while the suite logs a warning message for any invalid method signatures encountered.

Thank you for your feedback, please let me know if this approach aligns with the intended behavior or if there are further adjustments you’d like to see!

brackendawson:
It's not quite right yet, the subtest with the incorrect signature should fail because that's what happens in v1.10.0. The other tests should also run.

For example, when running this:
```go
package kata_test

import (
	"testing"

)

	suite.Suite
}

func (suite *ExampleTestSuite) TestPass() {
	suite.Assert().True(true)
}

func (suite *ExampleTestSuite) TestFail() {
	suite.Assert().True(false)
}

func (suite *ExampleTestSuite) TestWrong(a any) {
	suite.Assert().True(true)
}

}
```

Should have output like:
```
--- FAIL: TestExampleTestSuite (0.01s)
    --- FAIL: TestExampleTestSuite/TestFail (0.00s)
        /Users/bracken/src/githib.com/brackendawson/kata/kata_test.go:18: 
            	Error Trace:	/Users/bracken/src/githib.com/brackendawson/kata/kata_test.go:18
            	Error:      	Should be true
            	Test:       	TestExampleTestSuite/TestFail
    --- FAIL: TestExampleTestSuite/TestWrong (0.00s)
        /Users/bracken/src/githib.com/brackendawson/kata/value.go:427:
            <SOME USEFUL ERROR MESSAGE>
FAIL
coverage: [no statements]
FAIL	github.com/brackendawson/kata	0.220s
FAIL
```

brackendawson (changes requested):
The log will not be seen if all correctly written tests pass. This is dangerous.

vyas-git:
Now methods with invalid signatures will fail as subtests, and the suite runs all valid tests while clearly reporting the invalid ones.

suite/suite.go
```diff
@@ -158,7 +158,19 @@ func Run(t *testing.T, suite TestingSuite) {
...
 			continue
 		}
-
+		// Check method signature
+		if method.Type.NumIn() > 1 || method.Type.NumOut() > 0 {
+			tests = append(tests, test{
+				name: method.Name,
+				run: func(t *testing.T) {
+					t.Errorf(
+						"testify: suite method %q has invalid signature: expected no inputs or return values (has %d inputs, %d outputs)",
```
brackendawson: return values? outputs? They're input parameters and output parameters according to the reflect documentation.
```suggestion
						"testify: suite method %q has invalid signature: expected no input or output parameters, method has %d input parameters and %d output parameters",
```

brackendawson (changes requested):
I'm satisfied with the new behavior, just a small wording change.

PR #1675, opened 2024-11-02
Update documentation for the Error function in assert or require package

### LINKED ISSUE #1609: Incorrect require.Error() docs

architagr:
## Summary
Incorrect assert.Error() docs

## Changes
Removing the if statement and the Equal call from the example.

## Motivation
Since we already have EqualError, the documentation is trying to use `assert.Equal` to assert the error object's content.

## Related issues
Closes #1609


dolmen (dismissed):
The change now hides that `Error` returns a boolean that can be used for further checks.

I think that we must keep the if statement around in the usage example.

architagr:
> The change now hides that `Error` returns a boolean that can be used for further checks.
> 
> I think that we must keep the if statement around in the usage example.

This change is in accordance with the below conversation by @brackendawson

https://github.com/stretchr/testify/issues/1609#issuecomment-2402331089

brackendawson (approved):
I approve as is.

This was a poor place to document the assert assertions returning a booleans. It's just one assertion out of many. Perhaps one of the examples at the top of the package should be used.

It's important to remove this from the assert.Error doc because it is also used as the require.Error doc, which does not return a boolean! The only alternative is writing some code in the generator to handle it.

I also think the example is a bad example because the inner assert.Equal can stand in place of the entire example.

assert/assertions.go
```diff
@@ -1591,10 +1591,8 @@ func NoError(t TestingT, err error, msgAndArgs ...interface{}) bool {
 
 // Error asserts that a function returned an error (i.e. not `nil`).
 //
-//	  actualObj, err := SomeFunction()
-//	  if assert.Error(t, err) {
-//		   assert.Equal(t, expectedError, err)
-//	  }
+//	actualObj, err := SomeFunction()
+//	assert.Error(t, err)
```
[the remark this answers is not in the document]
brackendawson: This is not good practice IMO, a good test will check the returned actualObj even if the returned error is non-nil.

brackendawson:
@architagr I'd like to get this merged but I can't in good faith dismiss @dolmen's review because this PR still removes the last reference to assert's functions returning a bool. Can you add a paragraph to assert/doc.go which explains that all assertions return a bool indicating if the assertion passed? Then I think we can merge this change.

architagr:
> > @architagr I'd like to get this merged but I can't in good faith dismiss @dolmen's review because this PR still removes the last reference to assert's functions returning a bool. Can you add a paragraph to assert/doc.go which explains that all assertions return a bool indicating if the assertion passed? Then I think we can merge this change.
> 
> Sure I can do that

architagr:
> @architagr I'd like to get this merged but I can't in good faith dismiss @dolmen's review because this PR still removes the last reference to assert's functions returning a bool. Can you add a paragraph to assert/doc.go which explains that all assertions return a bool indicating if the assertion passed? Then I think we can merge this change.

Hi @brackendawson, I have updated the doc.go, can you please have a look if that looks ok.

assert/doc.go
```diff
@@ -1,5 +1,5 @@
 // Package assert provides a set of comprehensive testing tools for use with the normal Go testing system.
-//
+// All functions in this package return a bool value indicating whether the test has passed.
 // # Example Usage
```
ccoVeille, line 3: Did you check with pkgsite if it was still displaying a header for the example?

Because I would expect the comment to be like this

```suggestion
// All functions in this package return a bool value indicating whether the test has passed.
	//
 // # Example Usage
```

Also please consider using this

```suggestion
//
// # Note
// All functions in this package return a bool value indicating whether the test has passed.
	//
 // # Example Usage
```

Please check for difference by using pkgsite

assert/doc.go
```diff
@@ -1,5 +1,8 @@
 // Package assert provides a set of comprehensive testing tools for use with the normal Go testing system.
 //
+// # Note
+// All functions in this package return a bool value indicating whether the test has passed.
+//
 // # Example Usage
 //
 // The following is a complete example using assert in a standard test function:
```
ccoVeille, line 8: The exact syntax should be this

```suggestion
// Package assert provides a set of comprehensive testing tools for use with the normal Go testing system.
//
// # Note
//
// All functions in this package return a bool value indicating whether the test has passed.
//
// # Example Usage
//
// The following is a complete example using assert in a standard test function:
```

Here is the difference of rendering with and without the empty line of comments

image

image

ccoVeille (dismissed):

assert/doc.go
```diff
@@ -1,5 +1,9 @@
 // Package assert provides a set of comprehensive testing tools for use with the normal Go testing system.
 //
+// # Note
+//
+// All functions in this package return a bool value indicating whether the test has passed.
```
brackendawson, line 5:
```suggestion
// All functions in this package return a bool value indicating whether the assertion has passed.
```

We shouldn't say "test" here because Go's testing package defines the test as the test function.

brackendawson (changes requested):
Looks good and renders fine in pkgsite.

brackendawson (approved):
LGTM, thank you for your contribution. And thank you to the other maintainers for your diligent reviews.

PR #1681, opened 2024-11-18
Improve ErrorIs message when error is nil but an error was expected

tsioftas:
## Summary
This improves the UX of users of `ErrorIs()` in case the value `nil` is passed for `err` while an error was expected.

## Changes
- added check if `err == nil && target != nil` -> return a special error-string 
- updated relevant test case

## Motivation
The error message when an error is expected but `err` is `nil` (a common occurence in the world of testing) could be more helpful, e.g.:
```
(before)
some_test.go:20:
                Error Trace:    /[...]/some_test.go:20
                Error:          Target error should be in err chain:
                                expected: "sample error message"
                                in chain:
```
vs
```
(after)
some_test.go:20:
                Error Trace:    /[...]/some_test.go:20
                Error:          Expected error with "sample error message" in chain but got nil.
```

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -3237,9 +3237,9 @@ func TestErrorIs(t *testing.T) {
 				"in chain: \"EOF\"\n",
 		},
 		{
-			err:    nil,
-			target: io.EOF,
-			result: false,
+			err:          nil,
+			target:       io.EOF,
+			result:       false,
 			resultErrMsg: "Expected error \"EOF\" but got nil.\n",
 		},
 		{
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -3240,10 +3240,7 @@ func TestErrorIs(t *testing.T) {
 			err:    nil,
 			target: io.EOF,
 			result: false,
-			resultErrMsg: "" +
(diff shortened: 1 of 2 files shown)


ccoVeille (approved):

assert/assertions_test.go
```diff
@@ -3237,13 +3237,10 @@ func TestErrorIs(t *testing.T) {
...
-			target: io.EOF,
-			result: false,
-			resultErrMsg: "" +
-				"Target error should be in err chain:\n" +
-				"expected: \"EOF\"\n" +
-				"in chain: \n",
+			err:          nil,
+			target:       io.EOF,
+			result:       false,
+			resultErrMsg: "Expected error \"EOF\" but got nil.\n",
```
brackendawson, line 3243: A bit more concise
```suggestion
			resultErrMsg: "Expected error with \"EOF\" in chain but got nil.\n",
```

brackendawson (changes requested):

brackendawson (approved):
LGTM, thank you for your contribution.

ccoVeille:
Thanks for this change.

It would be way clearer.

PR #1690, opened 2024-12-30
Json matching function

akaswenwilk:
## Summary
This enables flexible json matching based on matcher functions in addition to json equality checks.

## Changes
- added new JSONMatchedBy func
- added a jsonmatch package to compare json objects
- added comparison via functions for values of json objects

## Motivation
My colleagues and I from every go based team I've worked with have always had difficulty with json.  Oftentimes we want to check aspects of certain parts of JSON objects without having to do a full equality check.  For instance, determining that a value is a uuid format, or that a datetime is in the past etc etc.  While this can be done in other ways such as unmarshalling an object and checking individual values, giving an overview in the form of a json object can be much more clear. 

an exmple from one of the tests:
```go
func TestJSONMatchesBy_DeeplyNestedMatchers(t *testing.T) {
	mockT := new(testing.T)
	True(t, JSONMatchesBy(
		mockT,
		`["$THREE", {"foo": {"bar": {"baz": "$NOT_EMPTY"}}}, {"nested": { "array": ["$NOT_EMPTY", "foo"]}}]`,
		`[3, {"foo": {"bar": {"baz": "qux"}}}, {"nested": { "array": ["quux", "foo"]}}]`,
		ValueMatchers{
			"$THREE":     func(v interface{}) bool { return v == 3.0 },
			"$NOT_EMPTY": func(v interface{}) bool { return v != "" },
		}))
}
```

The only issue I'd like to still solve in another PR is adjusting the display of the diff to factor in matchers.  I think it would be better to solve that problem separately.


assert/assertion_forward.go
```diff
@@ -3,6 +3,7 @@
 package assert
 
 import (
+	jsonmatch "github.com/stretchr/testify/assert/jsonmatch"
```
ccoVeille, line 6: Fix the import order please

assert/assertion_format.go
```diff
@@ -3,6 +3,7 @@
 package assert
 
 import (
+	jsonmatch "github.com/stretchr/testify/assert/jsonmatch"
```
ccoVeille, line 6: Fix the imports order please

assert/assertions.go
```diff
@@ -1703,7 +1689,6 @@ func NotRegexp(t TestingT, rx interface{}, str interface{}, msgAndArgs ...interf
 	}
 
 	return !match
-
```
ccoVeille, line 1706: These files might need formating for sure, but I would encourage you to revert all these spaces changes. It mess up the history of the files.

If you want to format them, why not, but then please open a PR about it or use a separate commit in your current PR

assert/assertions_test.go
```diff
@@ -842,11 +829,9 @@ func TestNil(t *testing.T) {
 	if Nil(mockT, new(AssertionTesterConformingObject)) {
 		t.Error("Nil should return false: object is not nil")
 	}
-
```
ccoVeille, line 845: Here also your formatter committed so many changes

assert/assertions_test.go
```diff
@@ -2687,7 +2793,7 @@ func TestFailNowWithFullTestingT(t *testing.T) {
 }
 
 func TestBytesEqual(t *testing.T) {
-	var cases = []struct {
+	cases := []struct {
```
ccoVeille, line 2796: I'm pretty sure you are using gofumpt as I do 😁

assert/jsonmatch/matches_by.go
```diff
@@ -0,0 +1,134 @@
...
+		for _, childVal := range v {
+			val = append(val, New(childVal))
+		}
+		n.value = val
+	case string:
+		n.kind = String
+		n.value = v
+	case float64:
+		n.kind = Number
+		n.value = v
```
ccoVeille, line 66: What about json.Number?

require/require.go
```diff
@@ -4,6 +4,7 @@ package require
 
 import (
 	assert "github.com/stretchr/testify/assert"
+	jsonmatch "github.com/stretchr/testify/assert/jsonmatch"
 	http "net/http"
 	url "net/url"
 	time "time"
```
ccoVeille, line 10: Please fix these imports please

Yes existing ones are already messed up

require/require_forward.go
```diff
@@ -4,6 +4,7 @@ package require
 
 import (
 	assert "github.com/stretchr/testify/assert"
+	jsonmatch "github.com/stretchr/testify/assert/jsonmatch"
 	http "net/http"
 	url "net/url"
 	time "time"
```
ccoVeille, line 10: Same here

assert/assertions_test.go
akaswenwilk: hehe, indeed

assert/assertions_test.go
akaswenwilk: oops, i'll adjust this

assert/jsonmatch/matches_by.go
akaswenwilk: instead of float64 for instance?

assert/assertion_format.go
akaswenwilk: what do you mean by the order actually? like alphabetically by package name?

assert/jsonmatch/matches_by.go
akaswenwilk: I'm not opposed to it, but what benefit would it provide?

assert/assertion_forward.go
akaswenwilk: I think this is what is meant? https://github.com/stretchr/testify/pull/1690/commits/bdff01450338f98f2c043d0c08dda4de1ddc9fda

assert/assertion_format.go
akaswenwilk: also here: https://github.com/stretchr/testify/pull/1690/commits/bdff01450338f98f2c043d0c08dda4de1ddc9fda

assert/assertions.go
akaswenwilk: makes sense, undid the auto formatting: https://github.com/stretchr/testify/pull/1690/commits/bdff01450338f98f2c043d0c08dda4de1ddc9fda

assert/assertions_test.go
akaswenwilk: removed: https://github.com/stretchr/testify/pull/1690/commits/bdff01450338f98f2c043d0c08dda4de1ddc9fda

assert/assertions_test.go
akaswenwilk: also here: https://github.com/stretchr/testify/pull/1690/commits/bdff01450338f98f2c043d0c08dda4de1ddc9fda

require/require.go
akaswenwilk: adjusted: https://github.com/stretchr/testify/pull/1690/commits/bdff01450338f98f2c043d0c08dda4de1ddc9fda

require/require_forward.go
akaswenwilk: https://github.com/stretchr/testify/pull/1690/commits/bdff01450338f98f2c043d0c08dda4de1ddc9fda

assert/jsonmatch/matches_by.go
ccoVeille: I'm sorry if I didn't explain it well.

Of course, the float64 case is needed.

My point was about the fact, and maybe I'm wrong, this function could receive data that was, or will be, decoded with UseNumber

https://pkg.go.dev/encoding/json#Decoder.UseNumber

But maybe I'm wrong and it cannot happen

assert/assertion_format.go
```diff
@@ -4,6 +4,7 @@ package assert
 
 import (
 	http "net/http"
+	jsonmatch "github.com/stretchr/testify/assert/jsonmatch"
 	url "net/url"
 	time "time"
 )
```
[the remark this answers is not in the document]
akaswenwilk: ah, i see.  fixed then!

ccoVeille (approved):

assert/jsonmatch/matches_by.go
akaswenwilk: ahh, i see.  I think that would have been caught by this test here: https://github.com/stretchr/testify/pull/1690/files#diff-7190065248695df477329c4cada2d71f323ab7940ecffff2d3d5fa43caf79427R2454

assert/assertion_format.go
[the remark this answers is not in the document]
akaswenwilk: looks like go generate doesn't let me fix the import order: https://github.com/stretchr/testify/actions/runs/12549993792/job/34992001258?pr=1690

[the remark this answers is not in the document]
ccoVeille: Oh indeed then

The files are generated. I assume there is something to improve in the script that generate them. But then it's out of the scope of this PR.

Thank you for trying, and I'm sorry if you had to undo

ccoVeille (approved):

assert/assertion_format.go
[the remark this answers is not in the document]
akaswenwilk: no prob, thanks for the review!

akaswenwilk:
thanks @ccoVeille for your review and suggested changes!  If you have a chance to look over it as well, I'd be grateful to be able to merge it to the main package @boyan-soubachov @dolmen @MovieStoreGuy @arjunmahishi or @brackendawson and also for an idea when/how we might be able to release it.  thanks for the help!

dolmen:
This seems to be an interesting feature, but at this time we do not want to increase the API surface of this framework which is already too huge for the limited maintainers resources.

PR #1692, opened 2025-01-03
CI: make GitHub workflow permissions explicit

cedricvanrompay-datadog:
## Summary

make GitHub workflow permissions explicit

## Changes

Add a top-level `permissions` field in each GitHub workflow file.

Note that this should not lead to any changes in the permissions workflows currently have because the permissions written mirror the default permissions or are overwritten by job-level permissions

## Motivation

It makes workflow permissions easier to read and understand, and it makes some scanner happy like [OSSF Scorecard](https://scorecard.dev/): https://deps.dev/project/github/stretchr%2Ftestify

## Related issues

None.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,10 @@
 name: All builds
 on: [push, pull_request]
 
+permissions:
+  contents: read
+  packages: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,9 @@
 name: Create release from new tag
 
+# setting default permissions to "none"
+# because permissions are set at the job level
+permissions: {}
+
 # this flow will be run only when new tags are pushed that match our pattern
 on:
   push:


dolmen:
Please add more context in the description of this PR, such as links to GitHub documentation about the permissions settings in the workflow file.

.github/workflows/main.yml
```diff
@@ -1,6 +1,10 @@
 name: All builds
 on: [push, pull_request]
 
+permissions:
+  contents: read
+  packages: read
+
```
brackendawson, line 7: This does change the permissions of the workflow. According to the [docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#defining-access-for-the-github_token-permissions) any permission not set in this dict will now be set to `none`, which will be different to the [default](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token).

brackendawson (commented):
I'm all for following the principle of least privilege, why do we need packages: read?

dolmen:
@cedricvanrompay-datadog Please fix this: https://github.com/stretchr/testify/pull/1692#issuecomment-2574767248

dolmen (changes requested):
The `setup-go` action [recommends](https://github.com/actions/setup-go?tab=readme-ov-file#recommended-permissions) `contents: read` permission. No mention of `packages: read`.

The `checkout` action also [recommends](https://github.com/actions/checkout?tab=readme-ov-file#recommended-permissions) `contents: read` permission. No mention of `packages: read`.

So I don't see a reason to grant `packages: read`.

PR #1693, opened 2025-01-03
mock: avoid data races when expecting calls

johanneswuerbach:
Instead of modifying the output as proposed in https://github.com/stretchr/testify/pull/1598, avoid any kind of output when trying to match calls.


peczenyj:
I just got a race condition error on my tests and this patch fixes it

peczenyj:
This fixes #1597

brackendawson (changes requested):
You haven't used the pull request template, so I've had to piece together the approach. Please use the template so that I know I've not missed or misunderstood anything.

Your approach uses the fact that Mock.Called calls Arguments.Diff initially through mock.findExpectedCalls to work out if the call is expected, this doesn't need to format the args at all. Then if the call was unexpected it will call Arguments.Diff again through Mock.findClosestCall which does need to format the arguments. You've provided Arguments.Diff with an operational or no-op formatter to stop it from rendering on the first path.

This;
* prevents data races caused by reads in fmt.Sprintf formatting the actual argument values when the call is expected.
* does not prevent data races caused by reads in fmt.Sprintf formatting the actual argument values when the call is not expected.
* does not prevent data races caused by reads in reflect.DeepEqual comparing the actual argument values to the expected values regardless of whether the call is expected or not.

Basically this only prevents races on values checked against mock.Anything or ignored by a mock.MatchedBy func. This does not close #1597 but is a sufficiently common problem for people to have that I think I'm open to merging this change.

PR #1704, opened 2025-03-12
feat: Use IsZero if types implement it

avorima:
## Summary

Support using `IsZero` instead of checking a types' zero value.

## Changes

Modify the `Zero` and `NotZero` assertions to check if types implement `IsZero() bool` and prefer using it over checking the zero value.

## Motivation

In go 1.24 types may implement `IsZero` to instruct the JSON encoder to omit their values when this function returns `true`.
This library should be able to use the same method.


assert/assertions.go
```diff
@@ -1707,21 +1707,35 @@ func NotRegexp(t TestingT, rx interface{}, str interface{}, msgAndArgs ...interf
...
+		return Fail(t, fmt.Sprintf("Should be zero, but was %v", i), msgAndArgs...)
+	}
 	if i != nil && !reflect.DeepEqual(i, reflect.Zero(reflect.TypeOf(i)).Interface()) {
 		return Fail(t, fmt.Sprintf("Should be zero, but was %v", i), msgAndArgs...)
 	}
 	return true
 }
 
 // NotZero asserts that i is not the zero value for its type.
+// If i implements IsZero it is preferred over checking the zero value.
```
brackendawson, line 1728: What is IsZero? You would need to fully document it as encoding/json does.

brackendawson (changes requested):
IsZero was added to encoding/json package to allow people to control how their types interact with the omitzero tag. Does it extend to uses outside of json marshalling? [reflect.IsZero](https://pkg.go.dev/reflect#Value.IsZero) doesn't use this interface.

ccoVeille:
I opened the following issue as a follow up

- https://github.com/stretchr/testify/issues/1759

assert/assertions.go
dolmen:
```suggestion
// If i has a IsZero() bool method it is preferred over checking the zero value.
```

PR #1706, opened 2025-03-17
suite.Passed: add one more status test report

Ararsa-Derese:
## Summary
This test case verifies that the Passed method returns false when all tests in the suite fail

## Changes
Added a test case to check the scenario where all tests fail.

## Motivation
This test is important to ensure that the Passed method correctly identifies the overall failure state of a test suite when none of the individual tests pass.

 ## Example usage 
This test can be used as a part of the test suite to validate the behavior of the Passed method under failure conditions.

## Related issues
None


brackendawson:
You need to format the code with current gofmt.

PR #1708, opened 2025-03-18
Adopt support of github.com/pmezard/go-difflib

### LINKED ISSUE #1159: Find an alternative for go-difflib

brackendawson:
## Summary
Bring the un-maintained github.com/pmezard/go-difflib library into testify and adopt support of the package.

## Changes
* Move difflib package to internal/difflib so that it cannot be imported by other modules.
* Move internal/difflib into testify, preserving all git history, and preserving the difflib license.
* Remove unused functions from internal/difflib.
* Remove github.com/pmezard/go-difflib from testify's requirements.

## Motivation
Difflib is explicitly unmaintained but depended on by testify.

## Related issues
Closes #1159


brackendawson:
I'm aware, the idea of changing the library used for this gives me some concern. It's not so bad because it's not actually used for any comparisons. But it's likely there will be cases where the change is a negative. If this was on a v2 boundary I'd be less hesitant.

It's also another library which might be considered unsupported by some. I usually don't share the view but a lot of corporations consider projects which don't have recent work on them to be de-facto unsupported. I sometimes consider these to be "stable", but an aim of the current maintenance of testify is to make it so that users with some requirement to use only supported software don't need to re-write all their tests to remove testify.

ccoVeille (approved):

mhussein:
Thanks a lot for handling this. When do you expect to merge this?

internal/difflib/difflib.go
```diff
@@ -0,0 +1,587 @@
...
+//
+// The following class and functions have been ported:
+//
+// - SequenceMatcher
+//
+// - unified_diff
+//
+// Getting unified diffs was the main goal of the port. Keep in mind this code
+// is mostly suitable to output text differences in a human friendly way, there
+// are no guarantees generated diffs are consumable by patch(1).
```
dolmen, line 13: The package documentation must be updated to mention the original source.

dolmen (dismissed):

dolmen:
This is the right thing to do.

internal/difflib/difflib.go
```diff
@@ -11,6 +11,9 @@
 // Getting unified diffs was the main goal of the port. Keep in mind this code
 // is mostly suitable to output text differences in a human friendly way, there
 // are no guarantees generated diffs are consumable by patch(1).
+//
+// This package was adopted from https://github.com/pmezard/go-difflib which is
+// no longer maintained.
```
dolmen, line 16: Add href:
```go
// This package was adopted from [github.com/pmezard/go-difflib] which is
// no longer maintained.
//
// [github.com/pmezard/go-difflib]: https://pkg.go.dev/github.com/pmezard/go-difflib
```

dolmen (changes requested):

internal/difflib/difflib.go
brackendawson: Done and checked the rendered version:
image

dolmen:
@Devourian What is your objection to this change?

internal/difflib/difflib.go
dolmen: Done in a51e3b7be7ff0c19c8ba71865ba2b276f53ed4c4.

However, my suggestion was linking to the documentation of the `difflib` package inside `go-difflib`, as this is what remains after the vendoring. I think that that target is more important to give access to the differences between the API of our vendored version vs the original one.

internal/difflib/difflib.go
dolmen: Done in 6d66080dfb4198a373f96df4fc9476acef8ff378

dolmen:
@brackendawson I'm ok with the current code.

However I would prefer if we could have a clean branch rebased on top of `master` at the time of merge, for a cleaner history.

I've attempted `git rebase -i -r --onto origin/master 7c367bb7bc7dad2d8b17921a604157dc27ebe06f internal-difflib`, but a problem remains as the reference to the external difflib remains in go.mod. We would have to edit the merge commit, but I haven't been able to do it so far.

brackendawson:
I'll make a clean version. Also would appreciate not changing the titles of PRs unless it's necessary, it makes it harder to keep track of things that are in progress.

internal/difflib/LICENSE
ccoVeille, line 1: I feel like having the license is good, bht having a README that somehow explains why it was imported would be sometg to consider.

Its content should be close to what I suggest to move to a doc.go file

See https://github.com/stretchr/testify/pull/1708#discussion_r2135339963

internal/difflib/difflib.go
```diff
@@ -0,0 +1,592 @@
...
+// - unified_diff
+//
+// Getting unified diffs was the main goal of the port. Keep in mind this code
+// is mostly suitable to output text differences in a human friendly way, there
+// are no guarantees generated diffs are consumable by patch(1).
+//
+// This package was adopted from [github.com/pmezard/go-difflib] which
+// is no longer maintained.
+//
+// [github.com/pmezard/go-difflib]: https://github.com/pmezard/go-difflib
```
ccoVeille, line 18: Here you are adding something to code that was copyrighted with a non standard spdx license.

I would move this to a doc.go file

brackendawson: Moving this to doc.go in the same directory would not change its license.

I've made other changes to difflib so that it can sit cleanly in testify and those are licensed under difflib's license. Why do we want this doc snippet to be MIT?

brackendawson:
I don't think the rebase tooling in git supports rebasing across unrelated histories, so the only way to make a version of this branch without another merge commit is to repeat [this procedure](https://stackoverflow.com/a/68536044) from scratch, which I'm not going to do. I think the best way forward is to merge this with a merge commit, as has been done with many other PRs in this repo.

internal/difflib/difflib.go
ccoVeille: I don't want the doc snippet to be under MIT 😅

My point is to provide a file that is a pristine import from the initial repository.

Also, the fact there is a license file in the folder doesn't mean the doc.go would be in under the MIT license.

I'm unsure I'm clearer 😅

brackendawson: The file is already not a pristine import. A lot of the unused content is removed. The purpose of moving it here is so that we now maintain it and that includes making further changes to it if necessary. The history of the file via git is pristine though.

ccoVeille: OK indeed then. It's clearer for me now, so I'm fine with the headef here

brackendawson:
@dolmen or @ccoVeille I've re-based this PR, it's not trivial to do so I'd appreciate it if you could consider re-approval soon.

ccoVeille (approved):

chirag-painter:
Hi, Would it be possible to share an estimated public release date for this change?

brackendawson:
I'd estimate around February.

deepakmokashi:
Hello @brackendawson
Do you have an estimated release date for this change?

PR #1710, opened 2025-03-19
Document consequences of calling t.FailNow()

### LINKED ISSUE #1701: Calling MethodCalled() in a goroutine can result in a call to t.FailNow()
## Description
I'm using this library indirectly, through `vektra/mockery`. I have a mock that I use in a goroutine, and I don't expect any calls on it. However, because of a bug in my code, there sometimes is an unexpected calls. This results in a call to `t.FailNow()`, which causes necessary cleanup to be skipped and the entire test suite to hang.

Looking at `MethodCalled()`, we can see several calls to `m.fail()`:

https://github.com/stretchr/testify/blob/7c367bb7bc7dad2d8b17921a604157dc27ebe06f/mock/mock.go#L509-L524

and `fail()` calls `T.FailNow()`: https://github.com/stretchr/testify/blob/7c367bb7bc7dad2d8b17921a604157dc27ebe06f/mock/mock.go#L352

Looking at [the documentation of `FailNow()`](https://pkg.go.dev/testing#T.FailNow), we can read the following: 

> `FailNow` must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test.

I think both `m.fail()` and `m.MethodCalled()` should come with the same warning.

## Step To Reproduce

create a goroutine that calls a mock, and performs an unexpected call on the mock, then closes a channel that should cause the main thread to hang until it is closed

## Expected behavior

The channel should still get closed.

## Actual behavior

The test hangs forever.

greg0ire:
## Summary

These comments are adapted from `t.FailNow()`'s own documentation.

## Changes

I added warnings on high-level APIs likely to be used by end users, and leading to `t.FailNow()` being called.

## Motivation

Such additional documentation can spare users headache in situations where ignoring the warnings would lead to test suites hanging forever

## Related issues

Closes #1701

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -331,7 +331,8 @@ func (m *Mock) TestData() objx.Map {
 	Setting expectations
 */
 
-// Test sets the test struct variable of the mock object
+// Test sets the [TestingT] on which errors will be reported, otherwise errors
+// will cause a panic.
 // Test should not be called on an object that is going to be used in a
 // goroutine other than the one running the test function.
 func (m *Mock) Test(t TestingT) {
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -332,6 +332,8 @@ func (m *Mock) TestData() objx.Map {
 */
 
 // Test sets the test struct variable of the mock object
+// Test should not be called on an object that is going to be used in a
+// goroutine other than the one running the test function.
 func (m *Mock) Test(t TestingT) {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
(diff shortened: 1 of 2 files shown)


mock/mock.go
```diff
@@ -332,6 +332,8 @@ func (m *Mock) TestData() objx.Map {
 */
 
 // Test sets the test struct variable of the mock object
+// Test should not be called on an object that is going to be used in a
```
brackendawson, line 335: Need to add a `.` on the previous sentence:
image

brackendawson (changes requested):

mock/mock.go
brackendawson: While here, we might improve the existing doc?
``go
// Test sets the [TestingT] on which errors will be reported, otherwise errors will cause a panic.

brackendawson (approved):
LGTM

PR #1711, opened 2025-03-20
Fix kind of Subset

khasanovbi:


brackendawson (changes requested):
Looks like a valid bug :+1:. Can you add a test and fill out the PR template please.

siliconbrain:
Actually, `NotSubset` has the same bug. I fixed both and added tests too in #1729

brackendawson:
Closing in favour of #1729

Issue #1712, opened 2025-03-22
Mock.IsMethodCallable doesn't work with dynamic expected arguments


brackendawson:
Mock.IsMethodCallable implements its own incomplete expected argument checking. So it doesn't always work. It should be using the same methods as Mock.Called to check if it is callable.

It also has no reason to accept a TestingT, it doesn't use it at all, there is no point calling Helper, which also isn't part of TestingT.

I also see no sensible use case for IsMethodCallable.

## Description
If the expected Args of a Call in Mock.ExpectedCalls are dynamic (mock.Anything, mock.MatchedBy, etc...) then Mock.IsMethodCallable will return the wrong answer.

## Step To Reproduce
```go
package kata_test

import (
	"testing"

	"github.com/stretchr/testify/mock"
	"github.com/stretchr/testify/require"
)

type MyMock struct {
	mock.Mock
}

func TestEventuallyFast(t *testing.T) {
	m := &MyMock{}
	m.On("MyMethod", 1).Return()
	m.On("MyMethod", mock.Anything).Return()
	require.True(t, m.IsMethodCallable(t, "MyMethod", 1))
	require.True(t, m.IsMethodCallable(t, "MyMethod", 2))
}
```

## Expected behavior
Testcase passes

## Actual behavior
`m.IsMethodCallable(t, "MyMethod", 2)` incorrectly returns false.

brackendawson:
Also it's amusing that a method which does not exist is "Callable"?

gopherjit:
@brackendawson please assign this issue to me.

PR #1714, opened 2025-03-22
doc: remove ineffective inline code blocks

### LINKED ISSUE #1713: Existing docs use code blocks in many places
Noticed that existing docs use code blocks in many places so tried to follow that even though not supported by godoc.  Updated to use quote 😄 Possibly something to consider over time is updating the other spots in the docs that use code block style.  For example:

https://github.com/techfg/testify/blob/098128fd102520c289cc5aef04afa388f10998a9/assert/assertions.go#L1592-L1593
https://github.com/techfg/testify/blob/098128fd102520c289cc5aef04afa388f10998a9/require/doc.go#L25-L26

_Originally posted by @techfg in https://github.com/stretchr/testify/pull/1709#discussion_r2007570748_

brackendawson:
## Summary
Godoc does not have inline code blocks. Look to the standard library for conventions for inline references to code.

## Changes
* Replace inline code blocks in docs (uses backticks) with formats following the conventions of the standard library.

## Motivation
Go doc does not have inline code blocks and backticks are unusual quotes for human readable documentation.

## Related issues
Closes #1713

### DIFF
--- a/assert/doc.go
+++ b/assert/doc.go
@@ -41,7 +41,7 @@
 // # Assertions
 //
 // Assertions allow you to easily write test code, and are global funcs in the assert package.
-// All assertion functions take, as the first argument, the *testing.T object provided by the
+// All assertion functions take, as the first argument, the [*testing.T] object provided by the
 // testing framework. This allows the assertion funcs to write the failings and other details to
 // the correct place.
 //
--- a/assert/doc.go
+++ b/assert/doc.go
@@ -40,8 +40,8 @@
 //
 // # Assertions
 //
-// Assertions allow you to easily write test code, and are global funcs in the `assert` package.
-// All assertion functions take, as the first argument, the `*testing.T` object provided by the
+// Assertions allow you to easily write test code, and are global funcs in the assert package.
+// All assertion functions take, as the first argument, the *testing.T object provided by the
 // testing framework. This allows the assertion funcs to write the failings and other details to
 // the correct place.
 //
--- a/require/doc.go
+++ b/require/doc.go
@@ -22,7 +22,7 @@
 // # Assertions
 //
 // The require package have same global functions as in the assert package,
-// but instead of returning a boolean result they call "t.FailNow()".
+// but instead of returning a boolean result they call [testing.T.FailNow].
 // A consequence of this is that it must be called from the goroutine running
 // the test function, not from other goroutines created during the test.
 //
--- a/require/doc.go
+++ b/require/doc.go
@@ -1,4 +1,4 @@
-// Package require implements the same assertions as the `assert` package but
+// Package require implements the same assertions as the assert package but
 // stops test execution when a test fails.
 //
 // # Example Usage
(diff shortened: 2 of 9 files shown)


brackendawson:
I discovered #1712 while making this change. I also noticed that suite.Suite.Assert() probably isn't necessary, you can just do:
```go
func (s *MySuite) TestIfy() {
    s.Assertsions.True(true)
}
```

You can't travel far in testify without tripping over strange functions.

brackendawson:
@dolmen I've split apart the formatting changes from the language changes. This PR now only converts the invalid inline code blocks to real code block and makes the code blocks into valid go. The incorrect language in the text of the comment is unchanged. I will follow up this PR with this commit to fix the language: https://github.com/brackendawson/testify/commit/0f9596016b3784e05c75e5d2a6397833adbca459

dolmen:
@brackendawson What about my comments about adding godoc links?

brackendawson:
@dolmen I'm not seeing any reference to godoc links? They wouldn't be relevant to this PR.

assert/doc.go
```diff
@@ -40,8 +40,8 @@
 //
 // # Assertions
 //
-// Assertions allow you to easily write test code, and are global funcs in the `assert` package.
-// All assertion functions take, as the first argument, the `*testing.T` object provided by the
+// Assertions allow you to easily write test code, and are global funcs in the assert package.
+// All assertion functions take, as the first argument, the *testing.T object provided by the
```
ccoVeille: Please use this to add a go doc link


```suggestion
// All assertion functions take, as the first argument, the [*testing.T] object provided by the
```

require/doc.go
```diff
@@ -21,8 +21,8 @@
 //
 // # Assertions
 //
-// The `require` package have same global functions as in the `assert` package,
-// but instead of returning a boolean result they call `t.FailNow()`.
+// The require package have same global functions as in the assert package,
+// but instead of returning a boolean result they call "t.FailNow()".
```
ccoVeille: Maybe this


```suggestion
// but instead of returning a boolean result they call `testing.T.FailNow`.
```

brackendawson: This suggestion is an invalid doc string. The point of this PR is to remove the use of backticks (`) because godoc does not support inline code blocks.

Changing the reference from t.FailNow() to testing.T.FailNow is valid. I'll do it but I have already had to remove a language change for it being out of the scope of this PR. 🤨

assert/doc.go
brackendawson: This is a valid change so I'll make it, but I've already had to remove changes from this PR for being out of scope of of the aim of removing inline code blocks.🤨

ccoVeille: …

I tend to request changes that seems to be obvious when they are part of the code that is updated.

Every people is different, luckily. But I can understand it doesn't help to make a PR validated.

But from my perspectives, based on the huge amont of opened PR and the huge delay in merging the old ones, we shouldn't expect people to open a PR for a small change such as updating a comment because "well, that would be better"

I would only expect this from a very clean repository, with a very reactive team who didn't let PR open for half a decade.

Some may argue it goes against the quality standard. For me, the repository is in such terrible state that arguing about this would be a bad move.

You and I are now maintainers. We can make things better.

I went way further than I expected in this comment, but sometimes things has to be said.

require/doc.go
ccoVeille: See https://github.com/stretchr/testify/pull/1714#discussion_r2361013561

PR #1718, opened 2025-03-25
fix: fix IsMethodCallable

gopherjit:
fixes method IsMethodCallable and respective unit test

## Summary

fixes the method IsMethodCallable and respective unit test

## Changes
calling findExpectedCall to check if the call is expected, if Yes return true, otherwise return true and falsefor the cases if the provided method was called and not called respectively.
fix respective unit test

## Motivation
implementation of IsMethodCallable is not correct also the unit test is not written correctly

## Related issues
Closes https://github.com/stretchr/testify/issues/1712

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -714,28 +714,16 @@ func (m *Mock) AssertNotCalled(t TestingT, methodName string, arguments ...inter
 
 // IsMethodCallable checking that the method can be called
 // If the method was called more than `Repeatability` return false
-func (m *Mock) IsMethodCallable(t TestingT, methodName string, arguments ...interface{}) bool {
-	if h, ok := t.(tHelper); ok {
-		h.Helper()
-	}
+func (m *Mock) IsMethodCallable(methodName string, arguments ...interface{}) bool {
 	m.mutex.Lock()
 	defer m.mutex.Unlock()
-
-	for _, v := range m.ExpectedCalls {
-		if v.Method != methodName {
-			continue
-		}
-		if len(arguments) != len(v.Arguments) {
-			continue
-		}
-		if v.Repeatability < v.totalCalls {
-			continue
-		}
-		if isArgsEqual(v.Arguments, arguments) {
-			return true
-		}
+	index, _ := m.findExpectedCall(methodName, arguments...)
+	if index != -1 {
+		return true
 	}
-	return false
+
+	// return false if call was not expected
+	return m.methodWasCalled(methodName, arguments)
 }
 
 // isArgsEqual compares arguments
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -1712,22 +1712,15 @@ func Test_Mock_AssertNotCalled(t *testing.T) {
 }
 
 func Test_Mock_IsMethodCallable(t *testing.T) {
-	var mockedService = new(TestExampleImplementation)
+	mock := new(TestExampleImplementation)
 
-	arg := []Call{{Repeatability: 1}, {Repeatability: 2}}
-	arg2 := []Call{{Repeatability: 1}, {Repeatability: 1}}
-	arg3 := []Call{{Repeatability: 1}, {Repeatability: 1}}
+	// Case 1: Method is expected and has remaining repeatability
+	mock.Mock.ExpectedCalls = append(mock.Mock.ExpectedCalls, &Call{Method: "TestMethod", Repeatability: 1})
+	assert.True(t, mock.IsMethodCallable("TestMethod"), "TestArg")
 
-	mockedService.On("Test_Mock_IsMethodCallable", arg2).Return(true).Twice()
-
-	assert.False(t, mockedService.IsMethodCallable(t, "Test_Mock_IsMethodCallable", arg))
-	assert.True(t, mockedService.IsMethodCallable(t, "Test_Mock_IsMethodCallable", arg2))
-	assert.True(t, mockedService.IsMethodCallable(t, "Test_Mock_IsMethodCallable", arg3))
-
-	mockedService.MethodCalled("Test_Mock_IsMethodCallable", arg2)
-	mockedService.MethodCalled("Test_Mock_IsMethodCallable", arg2)
-
-	assert.False(t, mockedService.IsMethodCallable(t, "Test_Mock_IsMethodCallable", arg2))
+	// Case 2: Method is expected but has no repeatability left
+	mock.Mock.ExpectedCalls[0].Repeatability = -1
+	assert.False(t, mock.IsMethodCallable("TestMethod"), "TestArg")
 }
 
 func TestIsArgsEqual(t *testing.T) {


gopherjit:
@brackendawson please review

mock/mock.go
```diff
@@ -714,28 +714,16 @@ func (m *Mock) AssertNotCalled(t TestingT, methodName string, arguments ...inter
 
 // IsMethodCallable checking that the method can be called
 // If the method was called more than `Repeatability` return false
-func (m *Mock) IsMethodCallable(t TestingT, methodName string, arguments ...interface{}) bool {
-	if h, ok := t.(tHelper); ok {
-		h.Helper()
-	}
+func (m *Mock) IsMethodCallable(methodName string, arguments ...interface{}) bool {
```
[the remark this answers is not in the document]
gopherjit: Thanks @ccoVeille for looking at the code, I understand this is a breaking change for the API, but t testing.T was redundant for this function, we can discuss if we can accommodate a breaking change in the API with the benefit of little cleaner code, it is always a tradeoff.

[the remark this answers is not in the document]
gopherjit: @brackendawson @ccoVeille I have reintroduced the testing.T type argument.

mock/mock_test.go
```diff
@@ -1712,22 +1712,16 @@ func Test_Mock_AssertNotCalled(t *testing.T) {
...
-	mockedService.MethodCalled("Test_Mock_IsMethodCallable", arg2)
+	// Case 1: Method is expected and has remaining repeatability
+	mock.Mock.ExpectedCalls = append(mock.Mock.ExpectedCalls,
+		&Call{Method: "TestMethod", Repeatability: 1, Arguments: Arguments{"TestArg"}})
+	assert.True(t, mock.IsMethodCallable(t, "TestMethod", "TestArg"))
 
-	assert.False(t, mockedService.IsMethodCallable(t, "Test_Mock_IsMethodCallable", arg2))
+	// Case 2: Method is expected but has no repeatability left
+	mock.Mock.ExpectedCalls[0].Repeatability = -1
+	assert.False(t, mock.IsMethodCallable(t, "TestMethod", "TestArg"))
```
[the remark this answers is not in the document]
gopherjit: the unit test was mocking the function it was intended to test, see this line - https://github.com/stretchr/testify/pull/1718/files#diff-80fc02c2ad6c6b55c9dfd38ea5989af97040aeb73df5d9a5195332953276a96eL1721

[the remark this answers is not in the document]
gopherjit: the tests are now in separate test runners

ccoVeille (approved):
LGTM 👍 in terms of code.

Let's wait for maintainers' feedbacks now

mock/mock.go
```diff
@@ -714,28 +714,15 @@ func (m *Mock) AssertNotCalled(t TestingT, methodName string, arguments ...inter
...
-		if isArgsEqual(v.Arguments, arguments) {
-			return true
-		}
+	index, _ := m.findExpectedCall(methodName, arguments...)
+	if index != -1 {
+		return true
 	}
-	return false
+	// return false if call was not expected
+	return m.methodWasCalled(methodName, arguments)
```
brackendawson, line 725: I don't think we need to call methodWasCalled because findExpectedCall will not return an expected call which has been fully satisfied.

I found this regression by writing the following testcase. I'd like to see something similar added in this PR.
```go
package kata_test

import (
	"testing"

)

type MyMock struct {
	mock.Mock
}

func (m *MyMock) Do(i int) {
	m.Called(i)
}

func TestIfy(t *testing.T) {
	for name, expectedArg := range map[string]any{
		"1":              1,
		"Anything":       mock.Anything,
		"MatchedBy":      mock.MatchedBy(func(i int) bool { return i == 1 }),
		"AnythingOfType": mock.AnythingOfType("int"),
		"IsType":         mock.IsType(0),
	} {
			expectedArg := expectedArg
			t.Parallel()

			m := &MyMock{}

			m.On("Do", expectedArg).Return().Times(2)
			assert.True(t, m.IsMethodCallable(t, "Do", 1))
			m.Do(1)
			assert.True(t, m.IsMethodCallable(t, "Do", 1))
			m.Do(1)
			assert.False(t, m.IsMethodCallable(t, "Do", 1))
		})
	}
}
```

After fixing this I think `isArgsEqual()` and `methodWasCalled()` are unused and should be removed.

mock/mock.go
```diff
@@ -714,28 +714,15 @@ func (m *Mock) AssertNotCalled(t TestingT, methodName string, arguments ...inter
 
 // IsMethodCallable checking that the method can be called
```
brackendawson, line 715: We should also make the docstring make sense, perhaps:
> IsMethodCallable returns true if the methodName and arguments are currently expected.

brackendawson (changes requested):

mock/mock.go
gopherjit: changed the docstring

mock/mock.go
gopherjit: I have removed usage of `methodWasCalled` in "IsMethodCallable"
I have removed `isArgsEqual`, didn't remove `methodWasCalled` as it is used at some other places.
The expected changes in unit tests are not clear to me.

brackendawson: Let me annotate the test with some comments, hopefully it becomes clear:
```go
func TestIfy(t *testing.T) {
	// Test this with each kind of expected argumnet. It works with `1` (raw value)
	// in tstify v1.10.0 but not with any of the other kinds.
	for name, expectedArg := range map[string]any{
		"1":              1,
		"Anything":       mock.Anything,
		"MatchedBy":      mock.MatchedBy(func(i int) bool { return i == 1 }),
		"AnythingOfType": mock.AnythingOfType("int"),
		"IsType":         mock.IsType(0),
	} {
			// Copy expectedArg to local scope before calling Parallel, solves the
			// capturing loop variable issue for go <v1.22.
			expectedArg := expectedArg
			t.Parallel()

			// Make a mock as usual
			m := &MyMock{}

			// Prime the mock as usual
			m.On("Do", expectedArg).Return().Times(2)
			
			// The "Do" call is now expected, it is expected twice.
			assert.True(t, m.IsMethodCallable(t, "Do", 1))
			
			// Call Do once
			m.Do(1)
			
			// The "Do" call is now expected once
			assert.True(t, m.IsMethodCallable(t, "Do", 1))
			
			// Call it again
			m.Do(1)
			
			// The "Do" call is now not expected, so it is not callable.
			assert.False(t, m.IsMethodCallable(t, "Do", 1))
		})
	}
}
```

gopherjit: Thanks @brackendawson, I have added the tests in similar fashion as suggested, please let me know if any other improvements can be made.

gopherjit:
@brackendawson please review.

mock/mock.go
```diff
@@ -712,45 +712,17 @@ func (m *Mock) AssertNotCalled(t TestingT, methodName string, arguments ...inter
...
-			continue
-		}
-		if isArgsEqual(v.Arguments, arguments) {
-			return true
-		}
+	index, _ := m.findExpectedCall(methodName, arguments...)
+	if index != -1 {
+		return true
 	}
 	return false
```
dolmen, line 723:
```suggestion
	return index != -1
```

mock/mock.go
```diff
@@ -712,45 +712,17 @@ func (m *Mock) AssertNotCalled(t TestingT, methodName string, arguments ...inter
 	return true
 }
 
-// IsMethodCallable checking that the method can be called
-// If the method was called more than `Repeatability` return false
-func (m *Mock) IsMethodCallable(t TestingT, methodName string, arguments ...interface{}) bool {
-	if h, ok := t.(tHelper); ok {
-		h.Helper()
-	}
+// IsMethodCallable returns true if the methodName and arguments are currently expected
```
dolmen, line 715:
```suggestion
// IsMethodCallable returns true if the methodName and arguments are currently expected.
```

dolmen (changes requested):
Just a few minor things to fix.

mock/mock_test.go
```diff
@@ -1712,35 +1712,48 @@ func Test_Mock_AssertNotCalled(t *testing.T) {
...
+			expected: false,
+		},
+	}
 
-	// Clone
-	arr := append(([]interface{})(nil), expected...)
-	assert.True(t, isArgsEqual(expected, arr))
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mock.Mock.ExpectedCalls = tt.expectedCalls
```
brackendawson, line 1753: Rather than using mock.Mock.Expected calls, which really shouldn't be exported, I'd like this test to use mock.On as most real users would.

brackendawson (changes requested):
You're still not testing expected calls with all of these argument types:
* `1` - basic types.
* `mock.Anything` - widlcard matcher.
* `mock.MatchedBy(func(i int) bool { return i == 1 })` - functional matcher.
* `mock.AnythingOfType("int")` - legacy type matcher.
* `mock.IsType(0)` - new type matcher.

Issue #1724, opened 2025-04-04
Migrate to a maintained yaml package


xqqp:
Testify uses the yaml package provided by https://github.com/go-yaml/yaml. After no development for a long period, a few days ago it finally got archived. 
Testify should migrate to another yaml package which is actively maintained. There are some forks of go-yaml which could be considered. There is also https://github.com/goccy/go-yaml, which is not a fork but largely API compatible.

dolmen:
Note that the default YAML library is still gopkg.in/yaml.v3, that libraray is already user-pluggable (replaceable) in your own test suite by using build tags. See https://github.com/stretchr/testify/blob/master/assert/yaml/yaml_custom.go and #1579.

dolmen:
At this point I would prefer to deprecate [`assert.YAMLEq`](https://pkg.go.dev/github.com/stretchr/testify/assert#YAMLEq) and [`assert.YAMLEqf`](https://pkg.go.dev/github.com/stretchr/testify/assert#YAMLEqf) than to deal with compatibility issues between two YAML libraries.

I can't imagine that the two libraries work exactly the same and that the replacement will not raise issues. Instead removing one dependency will make maintenance of the core easier.

founderio:
To make that happen, an option would be to remove the dependency completely and make `testify_yaml_fail` or `testify_yaml_custom` the default behaviour.

Tehnically, `testify_yaml_custom` without a configured unmarshaller is not that functionally different from `testify_yaml_fail`, it would just need a default implementation which fails. This would do away with the build tags as well.

For anyone who needs the original behaviour, a code snippet could be provided in the documentation.

brackendawson:
My first priority here as with all testify issues is to not break existing users or require them to do extra work. Deprecating the function without replacement or removing the implementation of the assertion will cause some people to have to do a non-trivial amount of work.

Continuing to use the archived yaml implementation requires some to do no work, but there will be people who can't use archived transient dependencies due to company policies.

[...]

PR #1725, opened 2025-04-04
Replace gopkg.in/yaml.v3 with github.com/goccy/go-yaml

xqqp:
## Summary
Replace `gopkg.in/yaml` with [`github.com/goccy/go-yaml`](https://pkg.go.dev/github.com/goccy/go-yaml). See #1724.
This was previously attempted in #1120, but the PR got closed due to `go-yaml` having a high number of dependencies. This has since been resolved. `go-yaml` does not have any dependencies anymore.

The only downside is that `go-yaml` requires at least golang v1.21


ccoVeille:
Another PR was opened to address this apparently

- https://github.com/stretchr/testify/pull/1727

The kubernetes' yaml version doesn't imply to bump Go minimal version

xqqp:
I also considered the Kubernetes yaml fork, but this one still has the 8 year old `gopkg.in/check.v1` dependency, while `github.com/goccy/go-yaml` does not.

ccoVeille:
The test are failing.

So now, I would say let's wait for a go/no-go from @dolmen or @brackendawson to bump minimal Go version for testify.

I don't feel like there is an urge in fixing tests if your PR is declined because of the minimal Go version bump.

dolmen:
There is a major downside in any replacement: I expect that the new library doesn't have 100% identical behavior.

Instead replacement driven by the end user is already supported and documented: see package [`github.com/stretchr/testify/assert/yaml`](https://pkg.go.dev/github.com/stretchr/testify/assert/yaml).

ccoVeille:
Please note there is also https://github.com/yaml/go-yaml/ in preparation

It's a fork of gopkg.in/yaml made by the official YAML team

brackendawson:
Given the discussion in #1724 I'm quite confident that we won't be moving to github.com/goccy/go-yaml

PR #1726, opened 2025-04-05
Feature/dynamic return values

jlou2u:
## Summary
Adds `RunWithReturn` which is essentially `Run` but adds `Arguments` as the return type and propagates them.

Used this as a starting point https://github.com/stretchr/testify/pull/742

## Changes
- Adds `RunWithReturn`

## Motivation
Helps write simple tests that can dynamically calculate return values.

Example (from test)
```
counter := 0
mockedService.On("TheExampleMethod", Anything, Anything, Anything).
	RunWithReturn(func(args Arguments) Arguments {
		counter++
		a, b, c := args[0].(int), args[1].(int), args[2].(int)
		assert.IsType(t, 1, a)
		assert.IsType(t, 1, b)
		assert.IsType(t, 1, c)
		return Arguments{counter}
	}).
	Twice()

answer, _ := mockedService.TheExampleMethod(2, 4, 5)
assert.Equal(t, 1, answer)

answer, _ = mockedService.TheExampleMethod(44, 4, 5)
assert.Equal(t, 2, answer)
```

## Related issues
https://github.com/stretchr/testify/pull/742


mock/mock.go
```diff
@@ -71,6 +71,10 @@ type Call struct {
 	// decoders.
 	RunFn func(Arguments)
 
+	// RunFnWithReturn adds return arguments to RunFn.  implemented separately
+	// to avoid breaking the API
+	RunFnWithReturn func(Arguments) Arguments
```
brackendawson, line 76: RunFnWithReturn should be unexported. We regret that some of the fields on this struct are exported. It also makes it easy to autocomplete instead of RunWithReturn, which is a paper-cut for the users.

brackendawson (changes requested):
This should override any previous calls to Return but also be overridden by any subsequent call to Return:

```go
package kata_test

import (
	"testing"

)

type myMock struct {
	mock.Mock
}

func (m *myMock) Do() string {
	return m.Called().String(0)
}

func TestIfy(t *testing.T) {
	m := &myMock{}
	m.On("Do").Return("one").Return("two")
	assert.Equal(t, "two", m.Do())

	m = &myMock{}
	m.On("Do").Return("one").RunWithReturn(func(args mock.Arguments) mock.Arguments {
		return mock.Arguments{"two"}
	})
	assert.Equal(t, "two", m.Do())

	m = &myMock{}
	m.On("Do").RunWithReturn(func(args mock.Arguments) mock.Arguments {
		return mock.Arguments{"one"}
	}).Return("two")
	assert.Equal(t, "two", m.Do())
}
```

This is why I suggested calling it ReturnFn, so that folks understand its purpose to to set return values. The purpose of Run is to implement side effects.

mock/mock.go
jlou2u: makes sense, should be resolved now

mock/mock_test.go
```diff
@@ -29,7 +29,7 @@ type TestExampleImplementation struct {
 
 func (i *TestExampleImplementation) TheExampleMethod(a, b, c int) (int, error) {
 	args := i.Called(a, b, c)
-	return args.Int(0), errors.New("Whoops")
```
jlou2u, line 32: this looked wrong -- not sure why this would always return an error?

mock/mock.go
```diff
@@ -187,6 +191,19 @@ func (c *Call) Run(fn func(args Arguments)) *Call {
 	return c
 }
 
+// ReturnFn sets a handler to be called before returning.
+//
+//	Mock.On("MyMethod", arg1, arg2).ReturnFn(func(args Arguments) Arguments {
+//	    return Arguments{args.Get(0) + args.Get(1)}
```
brackendawson, line 197: The example should be valid Go, did you mean Int rather than Get?
```suggestion
//	    return Arguments{args.Int(0) + args.Int(1)}
```

mock/mock.go
```diff
@@ -584,6 +601,14 @@ func (m *Mock) MethodCalled(methodName string, arguments ...interface{}) Argumen
 	returnArgs := call.ReturnArguments
 	m.mutex.Unlock()
 
+	m.mutex.Lock()
+	returnFn := call.returnFn
+	m.mutex.Unlock()
+
+	if returnFn != nil {
```
brackendawson, line 608: Because call.ReturnArguments is exported it should probably be checked by this condition rather than returnFn, eg:
```go
	m = &myMock{}
	m.On("Do").Return("one")
	calls := m.Calls
	m.On("Do").Return("one").Unset()
	m.On("Do").ReturnFn(func(args mock.Arguments) mock.Arguments { return mock.Arguments{"two"} })
	m.Calls = calls
	assert.Equal(t, "one", m.Do())
```

ie. Call.ReturnArguments should always override Call.returnFn

brackendawson (changes requested):
Good work so far. :+1:

This also causes Mock.Unset to misbehave:
```go
	m = &myMock{}
	m.On("Do").ReturnFn(func(args mock.Arguments) mock.Arguments { return mock.Arguments{"one"} })
	m.On("Do").Return("two")
	m.On("Do").Return("two").Unset() // Unset is an awful Method
	assert.Equal(t, "one", m.Do())
```

Mock.Unset is using Arguments.Diff to find matching calls, but because Arguments is empty it implicitly matches. Unset should probably never match any Call when Call.ReturnArguments is empty and Call.returnFn is non-nil. Unset should probably panic when called on a Call when Call.ReturnArguments is empty and Call.returnFn is non-nil, this is because functions are not comparable.

PR #1728, opened 2025-04-22
Issue 1722 suite panic skip setup

FGasper:


FGasper:
@ccoVeille Just a mistake.

We’re making a temporary-use fork until #1728 can be merged & pushed to a new release.

PR #1729, opened 2025-04-23
fix Subset/NotSubset when calling with mixed input types

siliconbrain:
## Summary
`Subset`/`NotSubset` failed with "<subset> has unsupported type map" when called with a `list` of kind `array`/`slice` and `subset` of kind `map` which should work based on the documentation.

## Changes
I added support and test cases for these inputs.

## Motivation
Implementation should conform to documentation.

### DIFF
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -693,12 +693,15 @@ func NotSamef(t TestingT, expected interface{}, actual interface{}, msg string, 
 	return NotSame(t, expected, actual, append([]interface{}{msg}, args...)...)
 }
 
-// NotSubsetf asserts that the specified list(array, slice...) or map does NOT
-// contain all elements given in the specified subset list(array, slice...) or
-// map.
+// NotSubsetf asserts that the list (array, slice, or map) does NOT contain all
+// elements given in the subset (array, slice, or map).
+// Map elements are key-value pairs unless compared with an array or slice where
+// only the map key is evaluated.
 //
 //	assert.NotSubsetf(t, [1, 3, 4], [1, 2], "error message %s", "formatted")
 //	assert.NotSubsetf(t, {"x": 1, "y": 2}, {"z": 3}, "error message %s", "formatted")
+//	assert.NotSubsetf(t, [1, 3, 4], {1: "one", 2: "two"}, "error message %s", "formatted")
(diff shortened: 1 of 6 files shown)


brackendawson (changes requested):
This adds a new behaviour that isn't documented. The choice seems to be that a map's keys can be considered a subset of a slice's elements.

The all maps behaviour requires the "subset" to have key value pairs from the "list". But when comparing a map type list with a slice type subset it's already just the keys that are compared.

The doc is currently:
> Subset asserts that the specified list(array, slice...) or map contains all elements given in the specified subset list(array, slice...) or map.
> 
> ```go
> assert.Subset(t, [1, 2, 3], [1, 2])
> assert.Subset(t, {"x": 1, "y": 2}, {"x": 1})
> ```

Perhaps:
> Subset asserts that the list (array, slice, or map) contains all elements given in the subset (array, slice, or map). Map elements are key-value pairs unless compared with an array or slice where only the map key is evaluated.

siliconbrain:
Since the documentation uses the same phrase ("list(array, slice...) or map") for both parameters, I expected the implementation to be symmetric too and if it worked for the [Not]Subset(map, array/slice) combination (which it does, thanks to the magic in `containsElement`) it should also work for [Not]Subset(array/slice, map).

If you think that the documentation isn't clear enough I'm happy to update it to your suggestion, but I would argue that this behavior is indeed documented even if it's _new_ by way of never being implemented correctly in the first place.

brackendawson:
The part which has never been documented is the behaviour when making asymmetric comparisons between slices and maps.

Previously when comparing a map list with a slice subset; only the keys in the map are considered to be elements, the values in the map are ignored. This was never documented.

You have added the inverse behaviour that when comparing a map list with a slice subset; the only the keys in the map are considered to be elements. the map values are ignored. I am asking you to document this. I think a clean way to document it exists which also documents the previous undocumented behaviour.

brackendawson:
Looks good :+1:. CI is failing because you need to run `go generate ./...` which will update the corresponding docs in the require package for you.

brackendawson (approved):
LGTM, thanks for all your work.

ccoVeille:
I'll have to check whether testifylint might conflicts with this change.

And if nothing is done about it, think about reporting misuage.

cc @mmorel-35

PR #1735, opened 2025-05-08
Improve captureTestingT helper

ccoVeille:
## Summary

A simple change to provide more explicit errors message when testing the code while refactoring asserters

## Changes

This helper is used to capture the testing.TB interface, and compare
the log output with the expected output.

This is useful for testing and refactoring purposes.

This commit improves the helper by displaying:
- the captured and expected outputs on a newline.
- the special characters in the captured output, such as newlines and tabs.

Both help with readability.

Here is an example of the output before and after the change:

Before:

```
    assertions_test.go:3422: Logged Error: Should be in error chain
        expected: *assert.customError
        in chain: "EOF" (*errors.errorString)
    assertions_test.go:3422: Should log Error: Should be in error chain:
        expected: *assert.customError
        in chain: "EOF" (*errors.errorString)
```

After:

```
    assertions_test.go:3486: Actual Logged Error:
        Should be in error chain\n
        expected: *assert.customError\n
        in chain: "EOF" (*errors.errorString)\n
    assertions_test.go:3486: Expected Logged Error:
        Should be in error chain:\n
        expected: *assert.customError\n
        in chain: "EOF" (*errors.errorString)
```

The new format helps to identify the differences:
- the missing colon after "Should be in error chain"
- the extra newline in the captured output

## Motivation

I spent 10 minutes on this change, because I lost 10 minutes in                                                                                                                      
finding the differences between the captured and expected output on a                                                                                                                     
refactoring I was doing.
                               
## Related issues

None

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -3241,10 +3241,71 @@ func (ctt *captureTestingT) checkResultAndErrMsg(t *testing.T, expectedRes, res 
 			if expectedErrMsg == content.content {
 				return
 			}
-			t.Errorf("Logged Error: %v", content.content)
+			t.Errorf("Actual Logged Error: \n%v", escapeSpecialCharacters(content.content))
 		}
 	}
-	t.Errorf("Should log Error: %v", expectedErrMsg)
+	t.Errorf("Expected Logged Error: \n%v", escapeSpecialCharacters(expectedErrMsg))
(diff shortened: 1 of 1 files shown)


ccoVeille:
#1741 was opened to replace this PR opened from the wrong fork

- #1741

PR #1737, opened 2025-05-09
chore: improve tests

ccoVeille:
## Summary
we are now testing the failure messages for the assert package

## Changes

Some tests were also added.

## Motivation

Having no idea about the error message could cause huge issue when refactoring.

I found a bug by doing so, I will report it soon


dolmen:
I won't review a PR that has a high risk of regression just because it deletes the existing testsuite.

Add tests, but don't remove existing ones!

PR #1738, opened 2025-05-09
fix: NotSame was returning true for non-pointers

ccoVeille:
## Summary
The code was different than the comment, and the code behavior was different from other asserters

## Changes
- **fix: NotSame was returning true for non-pointers**
- **chore: improve Same and NotSame tests**

## Motivation

NotSame had a different behavior than the other asserters:
- NotImplements fails and returns false for nil
- NotElementsMatch fails and returns false for non-lists
- NotEqual fails and returns false for invalid arguments
- NotContains fails and returns false if a length cannot be computed
- NotSubset fails and returns false if the objects are not comparable- **fix: NotSame was returning true for non-pointers**

Nothing returns true, and Fail.

Related to #1660, #1661, #1664

- #1660
- #1661
- #1664

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -610,40 +610,176 @@ func ptr(i int) *int {
 }
 
 func TestSame(t *testing.T) {
+	p1 := ptr(1)
+	p1bis := ptr(1)
+	p2 := ptr(2)
 
-	mockT := new(testing.T)
+	tests := []struct {
+		name         string
+		value        interface{}
+		expected     interface{}
+		result       bool
+		resultErrMsg string
+	}{
+		{
+			name:     "Same pointers",
+			value:    p1,
+			expected: p1,
+			result:   true,
+		},
+		{
+			name:     "Different pointers to same variable",
+			value:    p1,
+			expected: p1bis,
+			result:   false,
+			resultErrMsg: "Not same: \n" +
+				fmt.Sprintf("expected: %v %#v\n", p1, p1) +
+				fmt.Sprintf("actual  : %v %#v\n", p1bis, p1bis),
+		},
 
-	if Same(mockT, ptr(1), ptr(1)) {
-		t.Error("Same should return false")
+		{
+			name:     "Pointers to different variables",
+			value:    p1,
+			expected: p2,
+			result:   false,
+			resultErrMsg: "Not same: \n" +
+				fmt.Sprintf("expected: %v %#v\n", p1, p1) +
+				fmt.Sprintf("actual  : %v %#v\n", p2, p2),
+		},
+
+		{
+			name:         "Pointer compared to nil",
+			value:        p1,
+			expected:     nil,
+			result:       false,
+			resultErrMsg: "Both arguments must be pointers\n",
+		},
+
+		{
+			name:         "nil compared to pointer",
+			value:        nil,
+			expected:     p1,
+			result:       false,
+			resultErrMsg: "Both arguments must be pointers\n",
+		},
+
+		{
+			name:         "pointer to value",
+			value:        p1,
+			expected:     *p1,
+			result:       false,
+			resultErrMsg: "Both arguments must be pointers\n",
+		},
+
+		{
+			name:         "pointer to target",
+			value:        p1,
+			expected:     *p1,
+			result:       false,
+			resultErrMsg: "Both arguments must be pointers\n",
+		},
+
+		{
+			name:         "Not pointers",
+			value:        1,
+			expected:     1,
+			result:       false,
+			resultErrMsg: "Both arguments must be pointers\n",
+		},
 	}
-	if Same(mockT, 1, 1) {
-		t.Error("Same should return false")
-	}
-	p := ptr(2)
-	if Same(mockT, p, *p) {
-		t.Error("Same should return false")
-	}
-	if !Same(mockT, p, p) {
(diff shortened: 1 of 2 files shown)


assert/assertions_test.go
```diff
@@ -610,40 +610,176 @@ func ptr(i int) *int {
...
+			result:   true,
+		},
+		{
+			name:     "Different pointers to same variable",
+			value:    p1,
+			expected: p1bis,
+			result:   false,
+			resultErrMsg: "Not same: \n" +
+				fmt.Sprintf("expected: %v %#v\n", p1, p1) +
+				fmt.Sprintf("actual  : %v %#v\n", p1bis, p1bis),
```
dolmen, line 637: - Use a single `fmt.Sprintf` expression for the whole string.
- Use indexed formats like `%#[1]v` to remove parameter duplication (we should do that also in Same/NotSame code, but in a separate PR).
- Use `%p` like in the tested code

assert/assertions_test.go
```diff
@@ -610,40 +610,176 @@ func ptr(i int) *int {
...
-	if Same(mockT, p, *p) {
-		t.Error("Same should return false")
+		{
+			name:     "Pointers to different variables",
+			value:    p1,
+			expected: p2,
+			result:   false,
+			resultErrMsg: "Not same: \n" +
+				fmt.Sprintf("expected: %v %#v\n", p1, p1) +
+				fmt.Sprintf("actual  : %v %#v\n", p2, p2),
```
dolmen, line 647: - Use a single `fmt.Sprintf` expression for the whole string.
- Use indexed formats like `%#[1]v` to remove parameter duplication (we should do that also in Same/NotSame code, but in a separate PR).
- Use `%p` like in the tested code

assert/assertions_test.go
```diff
@@ -610,40 +610,176 @@ func ptr(i int) *int {
...
+		result       bool
+		resultErrMsg string
+	}{
+		{
+			name:     "Same pointers",
+			value:    p1,
+			expected: p1,
+			result:   false,
+			resultErrMsg: "Expected and actual point to the same object: " +
+				fmt.Sprintf("%v %#v\n", p1, p1),
```
dolmen, line 719: - Use a single `fmt.Sprintf` expression for the whole string.
- Use indexed formats like `%#[1]v` to remove parameter duplication (we should do that also in Same/NotSame code, but in a separate PR).
- Use `%p` like in the tested code

dolmen (changes requested):
Do not remove the existing testsuites.

I want to be able to see that with your changes the existing testsuite still pass, and your new tests also.

ccoVeille:
I'm closing this PR and I will split it in two PRs.

This will also allow meto switch to my fork.

ccoVeille:
I feel like I faced the same issue that @olivergondza had with


- https://github.com/stretchr/testify/pull/1400#issuecomment-2930429921

Unless I'm wrong Oliver came to the same conclusion than mine: captureTestingT cannot be used with negative asserters like NotSame and co.

Could you confirm Oliver and share your thoughts about this struct?

Also I'm curious about what you said about other label than errors.

Please note, I'm aware of the code you talked about and the loop on label to look for the one holding `Error`.

I'm just unsure what are the other labels you talked about.

I feel like I should conduct a refactoring of captureTestingT, that could help to handle negative asserter like NotSame, and the other I listed in my previous post above this one.

Thanks

olivergondza:
> I feel like I faced the same issue that @olivergondza had with
> 
> * [assert.PanicsWithError: report error message #1400 (comment)](https://github.com/stretchr/testify/pull/1400#issuecomment-2930429921)
> 
> Unless I'm wrong Oliver came to the same conclusion than mine: captureTestingT cannot be used with negative asserters like NotSame and co.
> 
> Could you confirm Oliver and share your thoughts about this struct?
> 
> Also I'm curious about what you said about other label than errors.
> 
> Please note, I'm aware of the code you talked about and the loop on label to look for the one holding `Error`.
> 
> I'm just unsure what are the other labels you talked about.
> 
> I feel like I should conduct a refactoring of captureTestingT, that could help to handle negative asserter like NotSame, and the other I listed in my previous post above this one.

Glad I am not alone who was fighting with this :)

Long story short, I ended up using explicit assertions on `captureTestingT.msg` over `captureTestingT.checkResultAndErrMsg`.

The reason `captureTestingT.checkResultAndErrMsg` not worked for me all that well is twofold:

- It parses the "mismatch" message into `assertions.labeledContent`, and only check the content of the line labeled with "Error".
- It only assert exact string match of the given portion.

My PR introduced the whole new label for panic, that it does not examine, obviously. And I also needed contains / not contains on the entire unparsed "mismatch" message.

I have spent some time looking into refactoring `checkResultAndErrMsg` to account for all that, but that would force me to sacrifice defect localization - as the messages would become too generic to hardly be helpful. Also, the API would likely bloat somewhat, and for a method that have 4 arguments already, it did not seem like a good idea.

But as long as I can get the entire message, I am fine using `captureTestingT.msg` as demonstrated by the other PR.

ccoVeille:
Thanks for your detailed reply.

I do feel like there are room for improvement yes. cc @brackendawson @dolmen 

You are right, the panic has to be handled differently.

I'm still interested if you could open a PR for a refactoring suggestion of checkResultAndErrMsg

PR #1734, opened 2025-05-08
Improve ErrorAs failure message when error is nil

ccoVeille:
## Summary

The following changes were made to improve the error messages

## Changes

Improve ErrorAs failure message when error is nil

Before:

```
    Should be in error chain:
    expected: *assert.customError
    in chain:
```

After:

```
    An error is expected but got nil.
    expected: *assert.customError
```

The message `An error is expected but got nil.` is the one already reported by `assert.Error`

## Motivation

This follows the changes that were made with #1681

- #1681

## Related issues

Closes #1733

- #1733

PR #1740, opened 2025-05-13
Use the right method to get the expected error type

ccoVeille:
## Summary

The type can be requested directly without getting the value.

## Changes

Use the right method to get the expected error type.

## Motivation

>Use `reflect.TypeOf(target).Elem().String()`

The following feedback was provided after the [previous PR](https://github.com/stretchr/testify/pull/1734) was merged.

_Originally posted by @dolmen in https://github.com/stretchr/testify/pull/1734#discussion_r2086158023_

## Related issues

Related to:

- #1734

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -2202,17 +2202,17 @@ func ErrorAs(t TestingT, err error, target interface{}, msgAndArgs ...interface{
 		return true
 	}
 
-	expectedText := reflect.ValueOf(target).Elem().Type().String()
+	expectedType := reflect.TypeOf(target).Elem().String()
 	if err == nil {
 		return Fail(t, fmt.Sprintf("An error is expected but got nil.\n"+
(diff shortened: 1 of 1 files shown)


ccoVeille:
> It looks we lack coverage by tests. Could you add some checks?

What do you mean?

Here are the TestErrorAs and TestNotErrorAs

https://github.com/stretchr/testify/blob/d2ddb5da5d3a5bfb855d6e9396cbe131cbb9c813/assert/assertions_test.go?#L3570-L3653

It's a PR you suggested to do to simplify the code, there is no changes in the test, because the code changes are stylistic.

Could you explain me more precisely what you would like to see?

I feel like I'm missing something

dolmen (approved):

dolmen:
Please change the commit message. The summary line should be meaningful when looking at the history of the whole repo.

Use the PR title as I have just changed it.

PR #1741, opened 2025-05-13
Improve captureTestingT helper

ccoVeille:
## Summary

A simple change to provide more explicit errors message when testing the code while refactoring asserters

## Changes

This helper is used to capture the testing.TB interface, and compare
the log output with the expected output.

This is useful for testing and refactoring purposes.

This commit improves the helper by displaying:
- the captured and expected outputs on a newline.
- the special characters in the captured output, such as newlines and tabs.

Both help with readability.

Here is an example of the output before and after the change:

Before:

```
    assertions_test.go:3422: Logged Error: Should be in error chain
        expected: *assert.customError
        in chain: "EOF" (*errors.errorString)
    assertions_test.go:3422: Should log Error: Should be in error chain:
        expected: *assert.customError
        in chain: "EOF" (*errors.errorString)
```

After:

```
    assertions_test.go:3486: Logged Error:
        Should be in error chain\n
        expected: *assert.customError\n
        in chain: "EOF" (*errors.errorString)\n
    assertions_test.go:3486: Should log Error: Error:
        Should be in error chain:\n
        expected: *assert.customError\n
        in chain: "EOF" (*errors.errorString)
```

The new format helps to identify the differences:
- the missing colon after "Should be in error chain"
- the extra newline in the captured output

## Motivation

I spent 10 minutes on this change, because I lost 10 minutes in                                                                                                                      
finding the differences between the captured and expected output on a                                                                                                                     
refactoring I was doing.
                               
## Related issues

Replaces: #1735  opened with the wrong fork

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -3241,10 +3241,71 @@ func (ctt *captureTestingT) checkResultAndErrMsg(t *testing.T, expectedRes, res 
 			if expectedErrMsg == content.content {
 				return
 			}
-			t.Errorf("Logged Error: %v", content.content)
+			t.Errorf("Logged Error:\n%v", escapeSpecialCharacters(content.content))
 		}
 	}
-	t.Errorf("Should log Error: %v", expectedErrMsg)
+	t.Errorf("Should log Error:\n%v", escapeSpecialCharacters(expectedErrMsg))
+}
+
+// escapeSpecialCharacters is an helper to spot differences in multiline strings
+func escapeSpecialCharacters(input string) string {
+	replacer := strings.NewReplacer(
+		"\t", "\\t",
+		"\r", "\\r",
+		"\n", "\\n\n", // newline are kept for readability
+	)
+	return replacer.Replace(input)
+}
+
+func Test_escapeSpecialCharacters(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{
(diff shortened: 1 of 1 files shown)


dolmen:
Why not just log the error using `%q`?

ccoVeille:
> Why not just log the error using `%q`?

I had no idea %q was escaping the \n or \t character

ccoVeille:
I can change, but then if becomes a matter of style

but I feel like this is more readable 

```
        assertions_test.go:3409: Logged Error:
            Target error should not be in err chain:\n
            found: ""\n
            in chain: \n
        assertions_test.go:3409: Should log Error:
            Target error should not be in err chain:\n
            found: ""\n
            in chain: 
```

than this

```
        assertions_test.go:3348: Logged Error: "Target error should not be in err chain:\nfound: \"\"\nin chain: \n"
        assertions_test.go:3348: Should log Error: "Target error should not be in err chain:\nfound: \"\"\nin chain: "
```

I can apply the change you request @dolmen, it matches my needs, but I'm not fan of the rendering.
The main issue is with the way the quote are escaped.

ccoVeille:
> Why not just log the error using `%q`?

done

dolmen (approved):

ccoVeille:
@dolmen 

I feel like I could use back tick with CanQuote as you did in a previous PR

What do you think?

PR #1742, opened 2025-05-13
assert.Same/NotSame: improve usage of Sprintf

ccoVeille:
## Summary
Reuse variable indexes instead of duplicating them.

## Changes

Avoid using the same variable multiple times with [fmt.Sprintf](https://pkg.go.dev/fmt#hdr-Explicit_argument_indexes)

## Motivation

> Use indexed formats like `%#[1]v` to remove parameter duplication (we should do that also in Same/NotSame code, but in a separate PR).
> _Originally posted by @dolmen in https://github.com/stretchr/testify/pull/1738#discussion_r2086147685_

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -543,8 +543,9 @@ func Same(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) b
 	if !same {
 		// both are pointers but not the same type & pointing to the same address
 		return Fail(t, fmt.Sprintf("Not same: \n"+
-			"expected: %p %#v\n"+
-			"actual  : %p %#v", expected, expected, actual, actual), msgAndArgs...)
+			"expected: %p %#[1]v\n"+
+			"actual  : %p %#[2]v",
+			expected, actual), msgAndArgs...)
 	}
 
 	return true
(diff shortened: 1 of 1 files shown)


assert/assertions.go
```diff
@@ -528,8 +528,9 @@ func Same(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) b
 	if !same {
 		// both are pointers but not the same type & pointing to the same address
 		return Fail(t, fmt.Sprintf("Not same: \n"+
-			"expected: %p %#v\n"+
-			"actual  : %p %#v", expected, expected, actual, actual), msgAndArgs...)
+			"expected: %#[1]p %#[1]v\n"+
```
dolmen, line 531:
```suggestion
			"expected: %#p %#[1]v\n"+
```

assert/assertions.go
```diff
@@ -528,8 +528,9 @@ func Same(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) b
...
+			"actual  : %#[2]p %#[2]v",
```
dolmen, line 532:
```suggestion
			"actual  : %#p %#[2]v",
```

dolmen (changes requested):

ccoVeille:
@dolmen I have applied your suggestions.

dolmen (approved):

PR #1745, opened 2025-05-17
assert: check test failure message for Empty and NotEmpty 

ccoVeille:
## Summary

Repost of #1737.

check test failure message for Empty and NotEmpty 

## Changes

Only the tests are updated, code is unchanged.
                                                                                                                                                                                               
Previously, the tests were checking only the result of the asserter.                                                                                                                       
Using captureTestingT helper allows to check the error message   

## Motivation

It allows us to make sure the error message reported are OK.

## Related issues

This is a proof of concept PR that follows up #1737 that was closed because:
- it was not using a fork
- there were a lot of changes to review.

- #1737

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -1704,6 +1704,142 @@ func TestEmpty(t *testing.T) {
 	False(t, Empty(mockT, TString("abc")), "non-empty aliased string is empty")
 	False(t, Empty(mockT, xP), "ptr to non-nil value is not empty")
 	False(t, Empty(mockT, [1]int{42}), "array is not state")
+
+	// error messages validation
+	tests := []struct {
+		name           string
+		value          interface{}
+		expectedResult bool
+		expectedErrMsg string
+	}{
+		{
+			name:           "Non Empty string is not empty",
+			value:          "something",
+			expectedResult: false,
+			expectedErrMsg: "Should be empty, but was something\n",
+		},
+		{
+			name:           "Non nil object is not empty",
+			value:          errors.New("something"),
+			expectedResult: false,
+			expectedErrMsg: "Should be empty, but was something\n",
+		},
+		{
+			name:           "Non empty string array is not empty",
+			value:          []string{"something"},
+			expectedResult: false,
+			expectedErrMsg: "Should be empty, but was [something]\n",
+		},
+		{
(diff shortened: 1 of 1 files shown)


assert/assertions_test.go
```diff
@@ -1736,13 +1792,52 @@ func TestNotEmpty(t *testing.T) {
...
-	False(t, NotEmpty(mockT, [1]int{}), "array is state")
+	tests := []struct {
+		name           string
+		value          interface{}
+		expectedErrMsg string
+	}{
+		{
+			name:           "Empty string is empty",
+			value:          "",
+			expectedErrMsg: `Should NOT be empty, but was ` + "\n", // TODO FIX THIS strange error message
```
ccoVeille, line 1803: This PR helps to stop things like that

dolmen (changes requested):
Do not remove existing tests.

ccoVeille:
> Do not remove existing tests.
> 
> Adding tests is fine. Replacing is not. We'll remove redundant tests later... maybe.

Please review

dolmen (approved):

ccoVeille:
> The `require` package also has `TestEmpty` and `TestNotEmpty`, so they should be extended as well.

I'm unsure what you meant, could you explain a bit more ?

Do you want me to check the error messages returned by require.Empty and require.NotEmpty ?
I could, but it's out of the scope I wanted for this PR.

require.Empty and require.NotEmpty are simple wrappers around assert.Empty and assert.NotEmpty, the code is very limited.

I feel like testing the returned values of assert.Empty and assert.NotEmpty is enough.

I don't plan to maintain two sets of tests, one in assert package and one in require one.
I could add a test to check one error message (and so not all of them) is the one expected:
Something like checking the error message returned by Empty with a non-empty string.
But I won't go further.

Unfortunately, I checked code and for now, captureTestingT is not available in require package

I could start a refactoring. captureTestingT could be moved in an internal package to be available for both.
But it would be out of the scope of this PR that I wanted to be limited.

TL;DR; the PR seems good enough for me

Of course, if I misunderstood what you meant, feel free to come back to me.

ccoVeille:
@dolmen could you give me some feedbacks about my last post, and also consider braken one.

dolmen (approved):

PR #1746, opened 2025-05-22
CI: fix shebang in .ci.*.sh scripts

dolmen:
## Summary
Fix shebang in .ci.*.sh scripts to ease running on platforms where bash is not in /bin/bash (ex: MacOS).


ccoVeille (approved):

brackendawson (approved):
ok

PR #1748, opened 2025-05-24
suite: Add support for extra filtering and -testify.r

renzoarreaza:
## Summary
This PR makes 2 changes to the suite package. 
1. Adds a repeat argument `-testify.r` to run tests multiple times without rerunning Setup/TeardownSuite
2. Adds support for users to provide extra filtering

## Changes & Motivation
Change 1: \
The first change allows tests to be repeated multiple times.  Using  go test's `-count` option results in the Setup/TeardownSuite also running multiple times. There are cases where this is not desired since the Setup/TeardownSuite may take a long time.

Change 2: \
The second change allows users of the suite package to "extend" it with their own skip function. 
The tests I added show overly simplified implementation, but this allows you to have test registration, and apply this during the collection phase of the testrun. 
Additionally, the running of the SetupSuite was moved to outside the loop. 
This is to avoid it running unnecessarily when you try to run a specific test that is filtered out but others aren't.
e.g: \
`go test MySuite/MyTest `
`MyTest` is skipped by the skip function, but the iteration over  `methodFinder.NumMethod()` will still go over the other tests of the suite which are possibly not skipped and will run the SetupSuite. However, because you only specified MyTest, they won't run either. Now the SetupSuite is executed unnecessarily.

### DIFF
--- a/.ci.gofmt.sh
+++ b/.ci.gofmt.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 if [ -n "$(gofmt -l .)" ]; then
   echo "Go code is not formatted:"
--- a/.ci.gogenerate.sh
+++ b/.ci.gogenerate.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # If GOMOD is defined we are running with Go Modules enabled, either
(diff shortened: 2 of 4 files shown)


dolmen:
I don't see how the two changes are related.

Please resubmit as 2 independent PRs.

renzoarreaza:
Fair enough. I've created #1749 and #1750 for the two changes.

PR #1749, opened 2025-05-25
suite: add support for user-defined test filtering

renzoarreaza:
## Summary
Adds support for users to provide extra filtering on which tests to run.

## Changes 
A new Exported function was needed in the suite package to support this. 
```
func RunWithSkip(t *testing.T, suite TestingSuite, skip func(string, string) bool)
```
This function is the same as the existing Run function with an extra call to the skip function passed by the user.
I now let the `Run` function call this one. 

Another change was to move the running of the SetupSuite to outside the iteration of the (test) methods.
This is to avoid it running unnecessarily when you try to run a specific test that is filtered out but others aren't.
e.g: \
`go test MySuite/MyTest `
Let say `MyTest` is skipped by the skip function. The iteration over  `methodFinder.NumMethod()` will still go over the other tests of the suite which are possibly not skipped and will run the SetupSuite. However, because you only specified MyTest, they won't run either. Now the SetupSuite (and Teardown) is executed unnecessarily. 

##  Motivation
The main motivation of this change is to allow for test registration. This allows tests to be skipped, as defined by the user, without repeatedly running the Setup/TeardownTest repeatedly for skipped tests.

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -119,6 +119,12 @@ func (suite *Suite) Run(name string, subtest func()) bool {
 // Run takes a testing suite and runs all of the tests attached
 // to it.
 func Run(t *testing.T, suite TestingSuite) {
+	RunWithSkip(t, suite, func(_, _ string) bool { return false })
+}
+
+// RunWithSkip takes a testing suite and a skip function allowing
+// for extra filtering on tests to be run
+func RunWithSkip(t *testing.T, suite TestingSuite, skip func(string, string) bool) {
 	defer recoverAndFailOnPanic(t)
 
 	suite.SetT(t)
@@ -148,16 +154,8 @@ func Run(t *testing.T, suite TestingSuite) {
 			continue
 		}
 
-		if !suiteSetupDone {
-			if stats != nil {
-				stats.Start = time.Now()
-			}
-
-			if setupAllSuite, ok := suite.(SetupAllSuite); ok {
-				setupAllSuite.SetupSuite()
-			}
-
-			suiteSetupDone = true
+		if skip(suiteName, method.Name) {
+			continue
 		}
 
 		test := testing.InternalTest{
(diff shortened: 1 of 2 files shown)


suite/suite.go
```diff
@@ -204,6 +202,19 @@ func Run(t *testing.T, suite TestingSuite) {
...
+	if len(tests) > 0 {
+		if stats != nil {
+			stats.Start = time.Now()
+		}
+
+		if setupAllSuite, ok := suite.(SetupAllSuite); ok {
+			setupAllSuite.SetupSuite()
+		}
+
+		suiteSetupDone = true
```
dolmen, line 215: This `suiteSetupDone` is not necessary anymore.

The `defer` block below can be de-indented.

suite/suite.go
```diff
@@ -204,6 +202,19 @@ func Run(t *testing.T, suite TestingSuite) {
 		}
 		tests = append(tests, test)
 	}
+
+	if len(tests) > 0 {
```
dolmen, line 206: Do not add indent. Instead:
```go
if len(tests) == 0 {
    return
}
```

dolmen (changes requested):
I see some value in the refactoring of `Run` that moves the suite setup out of the loop. It would be worth submitting it as a separate cleanup PR.

However at this point I don't understand the use case. So I didn't take time to review the test suite.

renzoarr:
I have made some changes to the proposal.
The idea is the same, but the implementation has changed. 
Instead of having RunWithSkip and a function that the user has to pass I've added a new interface that the user can optionally implement as part of their suite struct.

### Motivation: 
The goal of this proposal is for the user to have a way to define a filter for which tests should or shouldn't run. 
(you could think of it as an extension of -testify.m)
I use testify for integration tests and SetupSuite and SetupTest are on the slower side.
Depending on how I build the src code, (daily/beta builds, specific feature flags enabled or not etc...) or on cli flags (e.g runLevel express, regular, extreme) I might want to not run some tests. Right now, the only way to achieve this is to have some checks at the beginning of the test and decide if s.T().Skip() needs to be called or not. This however results in SetupTest running unnecessarily. 
These changes allow the user to skip tests at the collection phase and skipping the SetupTest for skipped tests (and even the SetupSuite is skipped if all tests are).

suite/suite.go
renzoarr: Fixed

suite/suite.go
renzoarr: Fixed

dolmen:
@renzoarreaza I have just merged #1769 so a rebase would be welcome.

brackendawson:
I think it would be better to make sure that calling testing.T.Skip in the setup function stops that function and prevents the test and the teardown form being run too. 

Edit: Skip will stop both the rest of BeforeTest, and the test, but not the call to AfterTest. I raised #1781

PR #1750, opened 2025-05-25
suite: add support for -testify.c

renzoarreaza:
## Summary
Add testify's own "count" argument `-testify.c` to run tests multiple times without rerunning Setup/TeardownSuite

## Changes
I've added a new (uint) `flag` to the top of `suite/suite.go`.
And I add the same InternalTest object that is created multiple times to the `tests` slice. 

## Motivation
The first change allows tests to be repeated multiple times.  Using  go test's `-count` option results in the Setup/TeardownSuite also running multiple times. This is especially valuable when you have a SetupSuite that takes a long time.

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -17,6 +17,7 @@ import (
 
 var allTestsFilter = func(_, _ string) (bool, error) { return true, nil }
 var matchMethod = flag.String("testify.m", "", "regular expression to select tests of the testify suite to run")
+var repeatItem = flag.Uint("testify.c", 1, "used to repeat each test multiple times without rerunning Setup/TearDownSuite")
 
 // Suite is a basic testing suite with methods for storing and
 // retrieving the current *testing.T context.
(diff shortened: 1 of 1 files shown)


dolmen:
It doesn't make sense to have this flag named with a `testify.` prefix while the flag is specific to `suite`. I know there is a precedent here with `testify.m` which is really unfortunate.

renzoarr:
What do you think of naming it `suite.c`. We could also add an alias to for `testify.m` to `suite.m`. 
I believe the second change means adding a third-party dependency, but I haven't looked into it yet.

renzoarr:
> What do you think of naming it `suite.c`. We could also add an alias to for `testify.m` to `suite.m`. I believe the second change means adding a third-party dependency, but I haven't looked into it yet.

@dolmen any thoughts on this before I write the code

renzoarr:
I picking this back up again. I've renamed the flags with the `suite` prefix instead of `testify` while keeping support for the existing `testify.m`.

PR #1751, opened 2025-05-27
suite: cleanup use of 'testing' internals at runtime

dolmen:
## Summary

Cleanup runtime use of stdlib's testing internals which was required for older Go versions.

Note: we are still using testing.RunTests in the suite's test suite for now.

## Changes
* drop use of `testing.InternalTest`
* drop use of `testing.RunTests`

## Motivation

Remove use of [`testing`](https://pkg.go.dev/testing) internals which we don't need to depend on anymore.

## Related issues
N/A


ccoVeille (approved):
I had to open the code to suite_test.go to see where the allTestsFilter was used (even with the comment 😅)

But, it seems OK.

A piece code for compatibility purpose with Go 1.6 wow

PR #1753, opened 2025-05-28
assert.Empty: comprehensive doc of "Empty"-ness rules

dolmen:
## Summary

Document the [`assert.Empty`](https://pkg.go.dev/github.com/stretchr/testify/assert#Empty) rules more comprehensively. This exposes our quirks to the user to avoid wrong expectations.

Add many many many more test cases that document edges cases and will allow to catch breaking changes and avoid regressions.

## Changes
* improve doc of `assert.Empty`, `assert.NotEmpty` (and their generated childs in package `require`)
* add many test cases for the internal `isEmpty`. They are added in the current style for now, but I have plans for a future refactoring (using table tests).

## Motivation

[`assert.Empty`](https://pkg.go.dev/github.com/stretchr/testify/assert#Empty) has quirks which desserve to be documented for our users, and for contributors.

The extensive test suite will help to ease future refactorings of Testify internals by catch breaking changes.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -773,10 +773,19 @@ func isEmpty(object interface{}) bool {
 	}
 }
 
-// Empty asserts that the specified object is empty.  I.e. nil, "", false, 0 or either
-// a slice or a channel with len == 0.
+// Empty asserts that the given value is "empty".
+//
+// [Zero values] are "empty".
+//
+// Arrays are "empty" if every element is the zero value of the type (stricter than "empty").
+//
+// Slices, maps and channels with zero length are "empty".
+//
+// Pointer values are "empty" if the pointer is nil or if the pointed value is "empty".
 //
 //	assert.Empty(t, obj)
+//
+// [Zero values]: https://go.dev/ref/spec#The_zero_value
 func Empty(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 	pass := isEmpty(object)
 	if !pass {
@@ -789,8 +798,7 @@ func Empty(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 	return pass
 }
 
-// NotEmpty asserts that the specified object is NOT empty.  I.e. not nil, "", false, 0 or either
-// a slice or a channel with len == 0.
+// NotEmpty asserts that the specified object is NOT [Empty].
 //
 //	if assert.NotEmpty(t, obj) {
 //	  assert.Equal(t, "two", obj[1])
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -1728,24 +1728,119 @@ func Test_isEmpty(t *testing.T) {
 
 	True(t, isEmpty(""))
 	True(t, isEmpty(nil))
+	True(t, isEmpty(error(nil)))
+	True(t, isEmpty((*int)(nil)))
+	True(t, isEmpty((*string)(nil)))
+	True(t, isEmpty(new(string)))
 	True(t, isEmpty([]string{}))
+	True(t, isEmpty([]string(nil)))
+	True(t, isEmpty([]byte(nil)))
+	True(t, isEmpty([]byte{}))
+	True(t, isEmpty([]byte("")))
+	True(t, isEmpty([]bool(nil)))
+	True(t, isEmpty([]bool{}))
+	True(t, isEmpty([]interface{}(nil)))
+	True(t, isEmpty([]interface{}{}))
+	True(t, isEmpty(struct{}{}))
+	True(t, isEmpty(&struct{}{}))
+	True(t, isEmpty(struct{ A int }{A: 0}))
+	True(t, isEmpty(struct{ a int }{a: 0}))
+	True(t, isEmpty(struct {
+		a int
+		B int
+	}{a: 0, B: 0}))
 	True(t, isEmpty(0))
+	True(t, isEmpty(int(0)))
+	True(t, isEmpty(int8(0)))
+	True(t, isEmpty(int16(0)))
+	True(t, isEmpty(uint16(0)))
 	True(t, isEmpty(int32(0)))
+	True(t, isEmpty(uint32(0)))
 	True(t, isEmpty(int64(0)))
+	True(t, isEmpty(uint64(0)))
+	True(t, isEmpty('\u0000')) // rune => int32
+	True(t, isEmpty(float32(0)))
+	True(t, isEmpty(float64(0)))
+	True(t, isEmpty(0i))   // complex
+	True(t, isEmpty(0.0i)) // complex
 	True(t, isEmpty(false))
+	True(t, isEmpty(new(bool)))
 	True(t, isEmpty(map[string]string{}))
+	True(t, isEmpty(map[string]string(nil)))
 	True(t, isEmpty(new(time.Time)))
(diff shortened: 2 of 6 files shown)


dolmen:
Cc: @PeterEFinch could you review this?

assert/assertions_test.go
```diff
@@ -1728,24 +1728,83 @@ func Test_isEmpty(t *testing.T) {
...
+	True(t, isEmpty(uint32(0)))
 	True(t, isEmpty(int64(0)))
+	True(t, isEmpty(uint64(0)))
+	True(t, isEmpty(float32(0)))
+	True(t, isEmpty(float64(0)))
+	True(t, isEmpty(0i))   // complex
+	True(t, isEmpty(0.0i)) // complex
 	True(t, isEmpty(false))
+	True(t, isEmpty(new(bool)))
 	True(t, isEmpty(map[string]string{}))
```
ccoVeille, line 1762: I think these ones could be interesting also

```go
isEmpty(map[string]string{"":""}))
isEmpty([]byte{})
isEmpty([]byte(""))
isEmpty([]byte{0})
isEmpty([]byte{0,0})
isEmpty([]string{""})
isEmpty([]string{"",""})
isEmpty([]bool{false})
isEmpty([]bool{false,false})
```

assert/assertions_test.go
```diff
@@ -1728,24 +1728,83 @@ func Test_isEmpty(t *testing.T) {
...
-	True(t, isEmpty([1]int{}))
+	True(t, isEmpty(chan struct{}(nil)))
+	True(t, isEmpty(chan<- struct{}(nil)))
+	True(t, isEmpty(make(chan struct{})))
+	True(t, isEmpty(make(chan<- struct{})))
+	True(t, isEmpty(make(chan struct{}, 1)))
+	True(t, isEmpty(make(chan<- struct{}, 1)))
+	True(t, isEmpty([1]int{0}))
+	True(t, isEmpty([2]int{0, 0}))
+	True(t, isEmpty(errors.New(""))) // BEWARE
```
ccoVeille, line 1775: What about adding

```go
isEmpty([]error{errors.New("")})
isEmpty([]error{errors.New(""), nil})
```

ccoVeille:
I learned things BTW

assert/assertions_test.go
```diff
@@ -1728,24 +1728,83 @@ func Test_isEmpty(t *testing.T) {
...
 	True(t, isEmpty(new(time.Time)))
 	True(t, isEmpty(time.Time{}))
 	True(t, isEmpty(make(chan struct{})))
-	True(t, isEmpty([1]int{}))
+	True(t, isEmpty(chan struct{}(nil)))
+	True(t, isEmpty(chan<- struct{}(nil)))
+	True(t, isEmpty(make(chan struct{})))
+	True(t, isEmpty(make(chan<- struct{})))
+	True(t, isEmpty(make(chan struct{}, 1)))
+	True(t, isEmpty(make(chan<- struct{}, 1)))
```
ccoVeille, line 1772:
```go
	isEmpty(make([]int, 8)))
```

assert/assertions_test.go
```diff
@@ -1728,24 +1728,83 @@ func Test_isEmpty(t *testing.T) {
 
 	True(t, isEmpty(""))
 	True(t, isEmpty(nil))
+	True(t, isEmpty(error(nil)))
+	True(t, isEmpty((*int)(nil)))
+	True(t, isEmpty((*string)(nil)))
+	True(t, isEmpty(new(string)))
```
PeterEFinch, line 1734: I think your list is pretty extensive.

You could test some interfaces with nil values but have a types associated with them but that might be a bit overkill. For example, you can have err being nil according to the assertions but is not equal to nil
```go
type MyError struct{}

func (e *MyError) Error() string {
	return "my error"
}

func TestMyError_isNil(t *testing.T) {
	var e *MyError
	err := error(e)
	assert.False(t, err == nil)
	assert.Empty(t, err)
	assert.Nil(t, err)
	assert.Zero(t, err)
}
```

assert/assertions_test.go
```diff
@@ -1728,24 +1728,83 @@ func Test_isEmpty(t *testing.T) {
...
+	True(t, isEmpty([2]int{0, 0}))
+	True(t, isEmpty(errors.New(""))) // BEWARE
+	True(t, isEmpty(&[1]int{0}))
+	True(t, isEmpty(&[2]int{0, 0}))
 	False(t, isEmpty("something"))
 	False(t, isEmpty(errors.New("something")))
 	False(t, isEmpty([]string{"something"}))
 	False(t, isEmpty(1))
+	False(t, isEmpty(uint(1)))
+	False(t, isEmpty(byte(1)))
```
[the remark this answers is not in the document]
brackendawson: rune is an alias for int32. I see we test True for int32 but not False.

assert/assertions_test.go
brackendawson: Because testify works at runtime it does not have visibility of your code's interface types, they get always converted into `interface{}`, eg:

```go
var e *MyError // type: *MyError, value: nil
assert.Empty(t, e)
...
func Empty(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
	object // type: interface{}, value: (*MyError)(nil)
...
err := error(e) // type: error, value: (*MyError)(nil)
assert.Empty(t, err)
...
func Empty(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
	object // type: interface{}, value: (*MyError)(nil)
```

Put simply, testify cannot tell that you passed an interface.

assert/assertions.go
```diff
@@ -772,8 +772,14 @@ func isEmpty(object interface{}) bool {
 	}
 }
 
-// Empty asserts that the specified object is empty.  I.e. nil, "", false, 0 or either
-// a slice or a channel with len == 0.
+// Empty asserts that the given value is "empty".
+//
+// [Zero values](https://go.dev/ref/spec#The_zero_value) are "Empty".
+// Arrays are "Empty" if every element is the zero value of the type (stricter than "Empty").
```
brackendawson, line 778: Malformed Zero values link and missing line break between Zero values and Arrays:
image

brackendawson (changes requested):

assert/assertions.go
```diff
@@ -772,8 +772,14 @@ func isEmpty(object interface{}) bool {
...
+//
+// Slices, maps and channels with zero length are "Empty".
+//
+// Pointer values are "Empty" if the pointer is nil or if the pointed value is "Empty".
```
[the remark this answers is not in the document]
dolmen: `reflect.Value.IsZero()` did not exist when `assert.Empty` has been added.

It is a good idea for a future refactoring. That's why we first need a strong test suite.

assert/assertions_test.go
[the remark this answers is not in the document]
dolmen: Added.

About rune, added check for `'\u0000'`.

Thanks.

assert/assertions_test.go
dolmen: Make for slices are redundant with slice literals. There would be no runtime difference.

I've put `make` with channels because that's the only way to make a non-nil channel.

ccoVeille (approved):

Issue #1752, opened 2025-05-28
Naive question: do we need objx dependency?


ccoVeille:
This question follows this PR

- https://github.com/stretchr/testify/pull/1567

- https://github.com/stretchr/objx/pull/140

I opened to solve the issue on objx side.

- https://github.com/stretchr/objx/pull/154

But maybe another approach is to remove this dependency?

cc @dolmen @brackendawson

brackendawson:
I don't think we ever needed it. It's used to embed a scratchpad in struts which embed mock.Mock. But anyone that needs a scratchpad could trivially add one themselves.

We can't remove it because it's a breaking change.

dolmen:
We could start with a deprecation of [`Mock.TestData`](https://pkg.go.dev/github.com/stretchr/mock#Mock.TestData).

What do you think?

ccoVeille:
Go for it, it doesn't sounds much used in public repositories

https://github.com/search?q=language%3Ago+testify+%22mock.testdata%22+-is%3Afork+-is%3Aarchived+-org%3Astretchr+-path%3A%2Ftestify%2F&type=code

dolmen:
This code search returns more relevant results (but confirms the rare use):

https://github.com/search?q=language%3AGo+%22ock.TestData%28%29%22&type=code

dolmen:
Here is a use of TestData (in project grafana/tanka) that can just be replaced by storing test data directly in the mock object.
https://github.com/grafana/tanka/pull/1467/files

brackendawson:
We would still need to bump the major version number to remove this else risk breaking innocent project's builds.

Deprecating it isn't so bad as there is an easy alternative of simply embedding the objx.Map in the struct next to mock.Mock. But we will still have the dependency, so I don't think there is actually any plus side to justify the (relatively minor) pain.

dolmen:
While we can't just remove the `objx` dependency in v1 because  of `TestData()` API breakage, we can still make it easier for downstream users to not dependend on it. I already did related things for YAML in #1579.

Here is a plan:
1. Move objx-dependent code in their own source files. That will ease the maintenance of downstream forks of Testify that would remove `objx`
2. Add compile tags allowing to compile `testify/mock` without `TestData`
3. Deprecate `TestData` with clear documentation about the reason of the deprecation and with workaround examples.

dolmen:
Step 1 is implemented in #1757.

emilien-puget:
[...]

PR #1757, opened 2025-05-30
mock: isolate objx-dependent code in their own source files

dolmen:
## Summary

In order to ease maintenance of downstream forks of Testify that would remove dependency on github.com/stretchr/objx we move all uses of that module in isolated source files. A fork without that dependency could just remove those files.
See https://github.com/stretchr/testify/issues/1752#issuecomment-2921545929

The use of objx is quite contained: it is only used in Mock.TestData(). Note that we can't just remove that method or change the return value because that would be a breaking change.

## Changes

* Move `Mock.TestData()` in a new file `mock/mock_objx.go`
* Move tests of TestData in a new file `mock/mock_objx_test.go`

## Motivation

See https://github.com/stretchr/testify/issues/1752#issuecomment-2921545929

With this change it will also be possible to add a `//go:build !testify_mock_no_objx` to `mock/mock_objx.go` and `mock/mock_objx_test.go` to allow downstream users to use testify/mock without the objx dependency.

## Related issues

* #1752

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,6 +19,7 @@ jobs:
       - run: ./.ci.gofmt.sh
       - run: ./.ci.govet.sh
       - run: go test -v -race ./...
+      - run: go test -tags testify_no_objx ./...
 
   test-old-go:
     runs-on: ubuntu-latest
@@ -40,6 +41,7 @@ jobs:
         with:
           go-version: ${{ matrix.go_version }}
       - run: go test -v -race ./...
+      - run: go test -tags testify_no_objx ./...
 
   check-actions-hashes:
     runs-on: ubuntu-latest
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -11,8 +11,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/stretchr/objx"
-
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/internal/difflib"
 	"github.com/stretchr/testify/internal/spew"
@@ -311,7 +309,7 @@ type Mock struct {
 
 	// TestData holds any data that might be useful for testing.  Testify ignores
 	// this data completely allowing you to do whatever you like with it.
-	testData objx.Map
+	testData testData
 
 	mutex sync.Mutex
 }
@@ -322,16 +320,6 @@ type Mock struct {
 // without acquiring the mutex, which is detected by go test -race.
 func (m *Mock) String() string {
 	return fmt.Sprintf("%[1]T<%[1]p>", m)
-}
-
-// TestData holds any data that might be useful for testing.  Testify ignores
-// this data completely allowing you to do whatever you like with it.
-func (m *Mock) TestData() objx.Map {
-	if m.testData == nil {
-		m.testData = make(objx.Map)
-	}
-
-	return m.testData
 }
 
 /*
--- a/mock/mock_no_objx.go
+++ b/mock/mock_no_objx.go
@@ -1,0 +1,5 @@
+//go:build testify_no_objx || testify_no_deps
+
+package mock
+
+type testData = struct{}
--- a/mock/mock_objx.go
+++ b/mock/mock_objx.go
@@ -1,0 +1,19 @@
+// This source file isolates the uses of the objx module to ease
+// maintenance of downstream forks that remove that dependency.
+// See https://github.com/stretchr/testify/issues/1752
+
+package mock
+
+import "github.com/stretchr/objx"
+
+type testData = objx.Map
+
+// TestData holds any data that might be useful for testing.  Testify ignores
+// this data completely allowing you to do whatever you like with it.
+func (m *Mock) TestData() objx.Map {
+	if m.testData == nil {
+		m.testData = make(objx.Map)
+	}
+
+	return m.testData
+}
--- a/mock/mock_objx.go
+++ b/mock/mock_objx.go
@@ -2,6 +2,8 @@
 // maintenance of downstream forks that remove that dependency.
 // See https://github.com/stretchr/testify/issues/1752
 
+//go:build !testify_no_objx && !testify_no_deps
+
 package mock
 
 import "github.com/stretchr/objx"
--- a/mock/mock_objx_test.go
+++ b/mock/mock_objx_test.go
@@ -1,3 +1,5 @@
+//go:build !testify_no_objx && !testify_no_deps
+
 package mock
 
 import (
--- a/mock/mock_objx_test.go
+++ b/mock/mock_objx_test.go
@@ -1,0 +1,19 @@
+package mock
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_Mock_TestData(t *testing.T) {
+	t.Parallel()
+
+	var mockedService = new(TestExampleImplementation)
+
+	if assert.NotNil(t, mockedService.TestData()) {
+
+		mockedService.TestData().Set("something", 123)
+		assert.Equal(t, 123, mockedService.TestData().Get("something").Data())
+	}
+}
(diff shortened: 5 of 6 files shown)


brackendawson (commented):
In my opinion I don't think this is a worthwhile change, removing my review without blocking.

This change also does the opposite of what it intends: If a fork of testify has removed objx then this change will be a merge conflict when they pull from upstream. If we touch nothing in the objx referenced code then it will not be a merge conflict for them.

brackendawson (changes requested):
This PR achieves the opposite of its intention. The stated aim is to make it easier for forks of testify to remove the objx dependency by moving code containing references to objx to new files.

Every time we touch the code referencing objx we cause a merge conflict for such forks and that includes this change. We should not make any change to make it easier for forks unless we also have to make another change to lines which reference objx.

> With this change it will be possible to add a //go:build !testify_mock_no_objx to mock/mock_objx.go and mock/mock_objx_test.go.

This will still not allow such a fork to remove objx from go.mod.

Please reference a request from the maintainer of a fork for this functionality.

ccoVeille (approved):

ccoVeille (changes requested):
I don't think it's worth trying this

dolmen:
@brackendawson You wrote

> Please reference a request from the maintainer of a fork for this functionality.

Are you aware of a such fork already existing?

brackendawson:
I'm not aware of any. I think this feature should not be done unless it delivers good value to a customer who exists.

ccoVeille:
> > This will still not allow such a fork to remove objx from go.mod.
> 
> An automated fork that tracks our master would be easy to do. Just a single patch that would be rebased. I think that resolving conflicts which involve just removing 2 files and `go mod tidy` would be easy to automate. 
> 
> > Please reference a request from the maintainer of a fork for this functionality.
> 
> Of course no such fork exists yet.
> 
> But I think that:
> 1. it would be easy to create
> 2. it would be easy to automate once this change is accepted. But I need this change to be accepted to demonstrate it.
> 3. we could provide that fork ourself on a branch in this repo `master-no-objx` automatically updated by CI

I have read everything again (this PR but also PR that were done in objx and testify itself), and I changed my mind. I feel like we should try.

At least we should add a deprecation warning as soon as possible

ccoVeille (dismissed):

fredbi:
@ccoVeille  @brackendawson 

I am discovering this piece of work.

From recent experiments at managing dependencies, I can tell for a fact that using build tags to guard dependencies does not achieve anything: consumers of the lib still get the dependencies in their go.mod.

the only viable way i found to make dependencies optional (like stuff used only in the specific use-casz of mocks) is to make an independent module.

this is doable but managing a mono repo with multiple modules requires a bit of extra attention.

i could transform the idea in that direction if you like it

brackendawson:
@fredbi in the case where assert.YAMLEq has to work without an initialisation by the users; does a separate module even remove the dependency from the primary module? We would have to import it, no?

fredbi:
Yes you would have to import it _explicitly_

Initialization may be explicit (e.g call « module.RegisterSomeFeature()”) or may be implicit (eg blank import calls init).

this would work for the yaml part but also for any “niche” dependency caused by some specific part of the framework (e.g mocks, testsuites and other added chrome - like colorized output and what not)

If you’re interested i can showcase the concept with a recent successful deep transformation of a similar “large bandwidth” library that i’ve sliced in a dozen highly focused modules.

ccoVeille:
I like your ideas for YAMLEq.

Here objx is very limited I think. It's about having a struct if I remembered well.

Can't we simply use a replace directive in our project, to force to use a local lib?

Would it work? Coukd it help to solve the dependency issue?

Yes, I agree with you the build tags things looks like an idea that brought nothing except a dead end

fredbi:
No replace directives only act locally not for the consumers of your package

fredbi:
@ccoVeille since you know the place, take a look at github.com/go-openapi/swag. The replace directives exist only for local development and to be able to cut releases (since at this very moment you put tags in the go.mod’s that don’t exist yet)

fredbi:
Also notice that go.work (go1.18+) is by no means essential. It just provides a slightly improved developer experience (as it was designed for). Replace directives are essential for the multi-module with inter-dependent modules, but not resolved by the consumers of your modules

fredbi:
The main issue in testify is to support optional dependency without bringing a breaking change. And that part is hard.

I don’t it is reasonable to expect absolutely no breaking change (apart by reimplementing these deps with the std library only). There the testify linter could help, with an autofix to add the desired import wherever we call something that needs an extra import. Could be YAMLEq, but also the stuff mentioned above related to the mock.

on a given code base, only a tiny fraction would actually use YAMLEq so i’d expect a limited impact

ccoVeille:
> No replace directives only act locally not for the consumers of your package

Thanks for the confirmation. I assumed it was something like that.

> @ccoVeille since you know the place, take a look at github.com/go-openapi/swag. The replace directives exist only for local development and to be able to cut releases (since at this very moment you put tags in the go.mod’s that don’t exist yet)

👍

ccoVeille:
> The main issue in testify is to support optional dependency without bringing a breaking change. And that part is hard.
> 
> I don’t it is reasonable to expect absolutely no breaking change (apart by reimplementing these deps with the std library only). 

Yes, it's an option I thought about.

Import the minimal code needed to check a YAML is equivalent to another, and remove the dependency. Here it's possible because we are not exporting yaml type out of testify.

objx is different as it exposes a type from the dependency.

YAMLEq would require to copy a lot of thing, but it could be a better option as the scope is very limited.

> There the testify linter could help, with an autofix to add the desired import wherever we call something that needs an extra import. Could be YAMLEq, but also the stuff mentioned above related to the mock.
> 
> on a given code base, only a tiny fraction would actually use YAMLEq so i’d expect a limited impact 

This would be a breaking change

fredbi:
An intermediate position could be to indulge into “blessed” dependencies only: objx is part of the stretchr org (or could be a module inside testify) and that’s fine by me. When consumers of my packages see in its go.mod all the imports “stretchr/testify/xxx”  they instantly get that this is test dependencies and moves on.

my auto-merging bots can recognize those as blessed just as well.

but the yaml stuff, the spew lib and the go-difflib we should IMHO isolate somehow

fredbi:
Here is yet another possible strategy (i am planning to apply that one on another lib, to wit girhub.com/go-openapi/runtime, another large-bandwidth stuff we have).

There i want to cut the dependency to open telemetry. But i can't do that without a breaking change.

so what i am planning is that i’ll freeze the dependency, publish the deprecation and encourage users to use the newly published module which is maintained with regular updates.

this is a bit similar to your idea of copying go-yaml once to ensure it works, but would relieve the maintainers from the burden of further maintaining that copy.

dolmen:
@brackendawson
Your review is blocking the merge. And as proposed in #1852 I've based #1870 on this one.
Could you unblock the merge so we could then merge it, and then rebase #1870?

dolmen:
@brackendawson With #1852 I want to drop the dependency on `objx` and move on. By having this (#1757) applied first that gives a clean fork point for downstream users (if they exist...) that still rely on `objx` features of `Mock.TestData()`.

brackendawson:
My review is blocking because I think this is a change which we should not make. The same feature for the yaml dependency caused forward comparability issues. It also does not deliver tangible benefit to anyone because it does not remove the dependency from the go.mod file, it only produces a different philosophical stance that the objx dependency is un-used, which users already have.

dolmen:
Rebased after merge of #1823 (upgrade of objx to v0.5.3).

dolmen:
Rebased.

I still want this to be merged.

@brackendawson 
> My review is blocking because I think this is a change which we should not make. The same feature for the yaml dependency caused forward comparability issues.

Which comparability issues?

brackendawson:
> Which comparability issues?

See: https://github.com/stretchr/testify/issues/1281#issuecomment-3278613428 

In short it inappropriately exposed the internal implementaiton of the yaml assertions and prevented us from fixing bugs. I will not approve any more such build modes which pretend to remove a dependency.

Edit: I did mean compatibility issues

PR #1758, opened 2025-05-30
Dolmen/suite faster method filtering

dolmen:
## Summary

In package `suite`, speed-up the filtering of methods (`-testify.m` flag) by compiling the regexp just once instead of when checking each method.

## Changes
* faster filtering of suite methods which are tests by using `strings.HasPrefix` instead of a `^Test` regexp compiled for every checked method
* faster filtering when `-testify.m` is given: compile the regexp once

## Motivation

Runtime speed!

## Related issues

N/A

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -140,16 +140,24 @@ func Run(t *testing.T, suite TestingSuite) {
 	methodFinder := reflect.TypeOf(suite)
 	suiteName := methodFinder.Elem().Name()
 
-	for i := 0; i < methodFinder.NumMethod(); i++ {
-		method := methodFinder.Method(i)
-
-		ok, err := methodFilter(method.Name)
+	var matchMethodRE *regexp.Regexp
+	if *matchMethod != "" {
+		var err error
+		matchMethodRE, err = regexp.Compile(*matchMethod)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "testify: invalid regexp for -m: %s\n", err)
 			os.Exit(1)
 		}
+	}
 
-		if !ok {
+	for i := 0; i < methodFinder.NumMethod(); i++ {
+		method := methodFinder.Method(i)
+
+		if !strings.HasPrefix(method.Name, "Test") {
+			continue
+		}
+		// Apply -testify.m filter
+		if matchMethodRE != nil && !matchMethodRE.MatchString(method.Name) {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


suite/suite.go
```diff
@@ -139,16 +140,24 @@ func Run(t *testing.T, suite TestingSuite) {
...
 			fmt.Fprintf(os.Stderr, "testify: invalid regexp for -m: %s\n", err)
 			os.Exit(1)
 		}
+	}
+
+	for i := 0; i < methodFinder.NumMethod(); i++ {
+		method := methodFinder.Method(i)
 
-		if !ok {
+		if !strings.HasPrefix(method.Name, "Test") {
```
ccoVeille, line 156: What about Benchmark Example and Fuzz?

dolmen: Using a regexp to match a prefix is just 300x slower than `strings.HasPrefix`.

```go
package main_test

import (
	"regexp"
	"strings"
	"testing"
)

func Benchmark(b *testing.B) {
	input := "TestFirst"
	var result bool

	b.Run("strings.HasPrefix", func(b *testing.B) {
		for b.Loop() {
			result = strings.HasPrefix(input, "Test")
		}
	})
	b.Run("regexp.MatchString", func(b *testing.B) {
		for b.Loop() {
			result, _ = regexp.MatchString("^Test", input)
		}
	})
	rePrefix := regexp.MustCompile("^Test")
	b.Run("rePrefix.MatchString", func(b *testing.B) {
		for b.Loop() {
			result = rePrefix.MatchString(input)
		}
	})

	_ = result
}
```

Results:
```console
$ go test -bench B
goos: darwin
goarch: arm64
pkg: bench.hasprefix
cpu: Apple M2
Benchmark/strings.HasPrefix-8         	343131853	        3.472 ns/op
Benchmark/regexp.MatchString-8        	 998090	     1164 ns/op
Benchmark/rePrefix.MatchString-8      	42221060	       28.82 ns/op
PASS
ok  	bench.hasprefix	3.760s
```

ccoVeille: My point was about the ability to filter tests on `ExampleFoo`, `FuzzFoo` and `BenchmarkFoo`, where `Foo` is an identifier.

I'm sorry if it leads you to benchmark obvious

So here my remark was more about having something like this:

```go
		switch {
			case strings.HasPrefix(method.Name, "Test"),
				strings.HasPrefix(method.Name, "Example"),
				strings.HasPrefix(method.Name, "Fuzz"),
				strings.HasPrefix(method.Name, "Benchmark"):
			default:
				continue
		}
```

(naive implementation)

dolmen: This is irrelevant. This PR is just a refactoring. Don't push ideas for functional changes in the review of such PR. Stay focused and talk about your ideas in separate issues.

By the way the scope of `suite` has always been about testing, never benchmarking or fuzzing.

ccoVeille (approved):

PR #1760, opened 2025-06-02
doc: improve godoc of top level 'testify' package

dolmen:
## Summary

Improve package doc of [`github.com/stretchr/testify`](https://pkg.go.dev/github.com/stretchr/testify)

## Motivation

Add godoc links to other packages `assert`, `require`, `mock`, `suite`.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1,15 +1,17 @@
-// Package testify is a set of packages that provide many tools for testifying that your code will behave as you intend.
+// Module testify is a set of packages that provide many tools for testifying that your code will behave as you intend.
 //
-// testify contains the following packages:
+// Testify contains the following packages:
 //
-// The assert package provides a comprehensive set of assertion functions that tie in to the Go testing system.
+// The [github.com/stretchr/testify/assert] package provides a comprehensive set of assertion functions that tie in to [the Go testing system].
+// The [github.com/stretchr/testify/require] package provides the same assertions but as fatal checks.
 //
-// The mock package provides a system by which it is possible to mock your objects and verify calls are happening as expected.
+// The [github.com/stretchr/testify/mock] package provides a system by which it is possible to mock your objects and verify calls are happening as expected.
 //
-// The suite package provides a basic structure for using structs as testing suites, and methods on those structs as tests.  It includes setup/teardown functionality in the way of interfaces.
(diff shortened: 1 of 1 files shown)


doc.go
```diff
@@ -1,15 +1,17 @@
-// Package testify is a set of packages that provide many tools for testifying that your code will behave as you intend.
+// Module testify is a set of packages that provide many tools for testifying that your code will behave as you intend.
```
ccoVeille, line 1: I don't think this is the norm.

> As can be seen in this example, Go doc comments use complete sentences. For a package comment, that means the first sentence begins with “Package ”.
> Source: https://go.dev/doc/comment#package

I would roll back this.

doc.go
```diff
@@ -1,15 +1,17 @@
...
 //
-// testify contains the following packages:
+// Testify contains the following packages:
 //
-// The assert package provides a comprehensive set of assertion functions that tie in to the Go testing system.
+// The [github.com/stretchr/testify/assert] package provides a comprehensive set of assertion functions that tie in to [the Go testing system].
+// The [github.com/stretchr/testify/require] package provides the same assertions but as fatal checks.
```
ccoVeille, line 6: I feel like this is enough


```suggestion
// The [assert] package provides a comprehensive set of assertion functions that tie in to [the Go testing system].
// The [require] package provides the same assertions but as fatal checks.
```

Please check with pkgsite

cc @brackendawson 

I report it here, but other links should be replaced

doc.go
dolmen: "package [...] is a set of packages" doesn't make sense.

ccoVeille: I know, but here it's about respecting Go specs.

So the solution is to change something else

What about this?

// Package testify consists in a set of tools, organized in packages, for testifying that your code will behave as you intend.

doc.go
dolmen: I tried hard but that doesn't work.
Those links would point to stdlib packages named assert/require/mock/suite.

The only way to be able to use short names for packages in links is to have them in import. But if they aren't used in the code, this is a vet error.

Using the full import path is the only solution.

ccoVeille: OK thanks for trying and providing such detailed feedbacks

doc.go
dolmen: *Module* is the right word to use in this context.

Starting with *Package* is a best practice, but not mandatory (not a "spec"), and in this case that goes against the goal which is to provide clear documentation to readers.

PR #1761, opened 2025-06-03
assert: faster and simpler isEmpty using reflect.Value.IsZero

dolmen:
## Summary

Refactor `isEmpty` which is the backend of `assert.Empty` and `assert.NotEmpty`.

## Changes
* `add Benchmark_isEmpty` that shows that some cases of `isEmpty` allocate memory
* split `isEmpty` into `isEmpty` and `isEmptyValue` for more efficient recursion when checking pointer values
* simplify `isEmptyValue` by using `reflect.Value.IsZero` (available since Go 1.13, so after the initial implementation of `assert.IsEmpty`).

## Motivation

* simpler implementation
* faster:
   * check for zero value using `reflect.Value.IsZero` which doesn't allocate memory.

## Related issues

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -752,8 +752,11 @@ func isEmpty(object interface{}) bool {
 		return true
 	}
 
-	objValue := reflect.ValueOf(object)
+	return isEmptyValue(reflect.ValueOf(object))
+}
 
+// isEmptyValue gets whether the specified reflect.Value is considered empty or not.
+func isEmptyValue(objValue reflect.Value) bool {
 	switch objValue.Kind() {
 	// collection types are empty when they have no element
 	case reflect.Chan, reflect.Map, reflect.Slice:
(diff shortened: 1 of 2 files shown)


assert/assertions.go
```diff
@@ -752,25 +752,25 @@ func isEmpty(object interface{}) bool {
...
-			return true
-		}
-		deref := objValue.Elem().Interface()
-		return isEmpty(deref)
-	// for all other types, compare against the zero value
-	// array types are empty when they match their zero-initialized state
-	default:
-		zero := reflect.Zero(objValue.Type())
-		return reflect.DeepEqual(object, zero.Interface())
+		return isEmptyValue(objValue.Elem())
```
ccoVeille, line 771: Here is the code I'm reviewing

```go
func isEmptyValue(objValue reflect.Value) bool {
	if objValue.IsZero() {
		return true
	}
	// Special cases of non-zero values that we consider empty
	switch objValue.Kind() {
	// collection types are empty when they have no element
	// Note: array types are empty when they match their zero-initialized state.
	case reflect.Chan, reflect.Map, reflect.Slice:
		return objValue.Len() == 0
	// non-nil pointers are empty if the value they point to is empty
	case reflect.Ptr:
		return isEmptyValue(objValue.Elem())
	}
	return false
}
```

I'm copy-pasting it here as the diff is uneasy to read.

One question, the code now uses isEmptyValue for Pointer, previously it was using isEmpty

can't this cause a problem with a pointer to a nil value ?

I feel like there is a risk, so I prefer to raise the point.

Note: I didn't fetch the code or check the coverage in the tests.

dolmen: If the diff is uneasy to read, you might instead review commit by commit. Also, each commit has significant messages.

dolmen: > One question, the code now uses isEmptyValue for Pointer, previously it was using isEmpty
> can't this cause a problem with a pointer to a nil value ?

The `nil` case is a case of zero value, so it is now handled by the `objValue.IsZero()` check.

PR #1765, opened 2025-06-07
Feature/assert match method

### LINKED ISSUE #1762: assert: enhanced array/slice diff
## Problem

When `assert.Equal()` fails on arrays/slices, it's hard to find the exact difference:

```
Not equal:
expected: []struct { ID int; Status string; Value int }{struct { ID int; Status string; Value int }{ID:1, Status:"active", Value:100}, struct { ID int; Status string; Value int }{ID:3, Status:"active", Value:150}, struct { ID int; Status string; Value int }{ID:5, Status:"active", Value:300}}
actual  : []struct { ID int; Status string; Value int }{struct { ID int; Status string; Value int }{ID:1, Status:"active", Value:100}, struct { ID int; Status string; Value int }{ID:3, Status:"active", Value:140}, struct { ID int; Status string; Value int }{ID:5, Status:"active", Value:250}}
```

Finding the differences requires manually comparing two long lines of text.

## Solution

Add detailed diff that shows **which index** differs:

```go
assert.Match(t, expected, actual) // or similar name
```

**Output:**
```
[index 1] Not equal:
expected: [{ID:1, Status:"active", Value:100}, {ID:3, Status:"active", Value:150}, {ID:5, Status:"active", Value:300}]
actual  : [{ID:1, Status:"active", Value:100}, {ID:3, Status:"active", Value:140}, {ID:5, Status:"active", Value:250}]
Field differences:
  └─ [1].Value: 150 ≠ 140
  └─ [2].Value: 300 ≠ 250
```

**For nested structs:**
```
[index 0] Not equal:
expected: {ID: 1, Profile: {Bio: "Engineer", Country: "Brasil"}}
actual  : {ID: 1, Profile: {Bio: "Developer", Country: "Brazil"}}
Field differences:
  └─ Profile.Bio: "Engineer" ≠ "Developer"  
  └─ Profile.Country: "Brasil" ≠ "Brazil"
```

## Benefits

- **Instant debugging** - no more hunting through arrays
- **Works with existing patterns** - same familiar API
- **Zero breaking changes** - completely additive

## Implementation

I can implement this. Would follow testify's existing conventions.

**Would the maintainers be open to this enhancement? Happy to submit a PR if so.**

Andrei-hub11:
## Summary

Add Match() for precise slice/array element comparison with field-by-field difference reporting

## Changes

- Implement Match() function for comparing arrays/slices with the same element order
- Add detailed field-level difference reporting with path information
- Document behavior differences from ElementsMatch (order matters in Match)
- Add comprehensive test suite for the new functionality

## Motivation

When comparing complex slices/arrays, it's often important to not only know IF they differ, but also WHERE and HOW they differ. The new Match() function provides field-by-field comparison with precise path reporting for easier debugging.

## Example output

```
Error: Differences found:
  MATCH: [index 0] Not equal:
  expected: {Name: "Alice", Age: 30, Department: {Manager: "Alice", Budget: 100000.5}}
  actual  : {Name: "Alice", Age: 31, Department: {Manager: nil, Budget: 95000}}
  Field differences:
    └─ Age: 30 ≠ 31
    └─ Department.Budget: 100000.5 ≠ 95000
    └─ Department.Manager: "Alice" ≠ nil

  MATCH: [index 1] Not equal:
  expected: {Name: "Bob", Age: 25, Department: {Manager: "Bob", Budget: 75000.0}}
  actual  : {Name: "Bob", Age: 25, Department: {Manager: "Alice", Budget: 72000.0}}
  Field differences:
    └─ Department.Manager: "Bob" ≠ "Alice"
    └─ Department.Budget: 75000.0 ≠ 72000.0

  MATCH: [index 2] Not equal:
  expected: {Name: "Charlie", Age: 35, Department: {Manager: "Dave", Budget: 120000.0}}
  actual  : {Name: "Charlie", Age: 42, Department: {Manager: "Dave", Budget: 150000.0}}
  Field differences:
    └─ Age: 35 ≠ 42
    └─ Department.Budget: 120000.0 ≠ 150000.0
```

## Related issues

Closes #1762

@dolmen, let me know if there's anything I should update in the PR.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -56,6 +56,15 @@ type PanicAssertionFunc = func(t TestingT, f PanicTestFunc, msgAndArgs ...interf
 // Comparison is a custom function that returns true on success and false on failure
 type Comparison func() (success bool)
 
+// FieldDiff represents a single difference between two compared values.
+// It stores the path to the differing field, along with the expected and actual values.
+// This is used by the Match function to provide detailed information about differences.
+type FieldDiff struct {
+	Path     string      // The path to the field, using dot notation for nested fields
+	Expected interface{} // The expected value at this path
+	Actual   interface{} // The actual value at this path
+}
+
 /*
 	Helper functions
 */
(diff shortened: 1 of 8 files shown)


assert/assertions.go
```diff
@@ -2292,3 +2301,357 @@ func buildErrorChainString(err error, withType bool) string {
...
+	if actualLen != expectedLen {
+		Fail(t, fmt.Sprintf("Lengths not equal:\nexpected: %d\nactual  : %d", expectedLen, actualLen), msgAndArgs...)
+		return false
+	}
+
+	for i := 0; i < minLen; i++ {
+		expectedItem := expectedValue.Index(i).Interface()
+		actualItem := actualValue.Index(i).Interface()
+
+		if !ObjectsAreEqual(expectedItem, actualItem) {
```
ccoVeille: You can reduce the nesting level by doing this I think

```go
		if ObjectsAreEqual(expectedItem, actualItem) {
			continue
		}
```

assert/assertions.go
```diff
@@ -2292,3 +2301,357 @@ func buildErrorChainString(err error, withType bool) string {
...
+		}
+	}
+
+	if len(differences) > 0 {
+		diffMessage := strings.Join(differences, "\n")
+		Fail(t, fmt.Sprintf("Differences found:\n%s", diffMessage), msgAndArgs...)
+		return false
+	}
+
+	return true
```
ccoVeille: This might improve the readability

```suggestion
	if len(differences) == 0 {
		return true
	}

	diffMessage := strings.Join(differences, "\n")
	return Fail(t, fmt.Sprintf("Differences found:\n%s", diffMessage), msgAndArgs...)
```

assert/assertions.go
```diff
@@ -2292,3 +2301,357 @@ func buildErrorChainString(err error, withType bool) string {
...
+			fieldValue := v.Field(i)
+			parts = append(parts, fmt.Sprintf("%s: %s", field.Name, formatValueComparison(fieldValue)))
+		}
+		return fmt.Sprintf("{%s}", strings.Join(parts, ", "))
+
+	case reflect.String:
+		return fmt.Sprintf(`"%s"`, v.String())
+
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return fmt.Sprintf("%d", v.Int())
```
ccoVeille: I feel like most of the Sprintf could be this, because they are almost equivalent to sprintf("%v", whatever) so sprint(whatever)

```suggestion
		return fmt.Sprint(v.Int())
```

Please double check.

You can use perfsprint linter if you are curious

assert/assertions.go
```diff
@@ -2292,3 +2301,357 @@ func buildErrorChainString(err error, withType bool) string {
...
+	default:
+		return fmt.Sprintf("%v", v.Interface())
+	}
+}
+
+// formatDiffValue formats a value specifically for showing differences.
+// It handles basic types differently than complex types for better readability.
+func formatDiffValue(value interface{}) string {
+	if value == nil {
+		return "<nil>"
```
ccoVeille: Here I see `<nil>` the previous were `nil`

Any reason?

assert/assertions.go
```diff
@@ -2292,3 +2301,357 @@ func buildErrorChainString(err error, withType bool) string {
...
+	case string:
+		return fmt.Sprintf("%q", v)
+	case bool:
+		return fmt.Sprintf("%t", v)
+	case int, int8, int16, int32, int64:
+		return fmt.Sprintf("%v", v)
+	case uint, uint8, uint16, uint32, uint64:
+		return fmt.Sprintf("%v", v)
+	case float32, float64:
+		return fmt.Sprintf("%v", v)
```
ccoVeille:
```suggestion
	case bool, int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64, float32, float64:
		return fmt.Sprint(v)
```

ccoVeille (commented):
Minor stylistic suggestions, when not nitpicking 😅

assert/assertions.go
Andrei-hub11: An output like the following:

```
  Error:          Differences found:
                                MATCH: [index 0] Not equal:
                                expected: nil
                                actual  : "not nil"
                                Field differences:
                                  └─ : <nil> ≠ "not nil"
```

ccoVeille: It sounds legit, indeed.

It also used in other asserter also

assert/assertions.go
```diff
@@ -56,6 +56,15 @@ type PanicAssertionFunc = func(t TestingT, f PanicTestFunc, msgAndArgs ...interf
 // Comparison is a custom function that returns true on success and false on failure
 type Comparison func() (success bool)
 
+// FieldDiff represents a single difference between two compared values.
+// It stores the path to the differing field, along with the expected and actual values.
+// This is used by the Match function to provide detailed information about differences.
+type FieldDiff struct {
```
ccoVeille: Open question.

Is there a need for this struct to be exported?

assert/assertions.go
```diff
@@ -2292,3 +2301,364 @@ func buildErrorChainString(err error, withType bool) string {
...
+	minLen := expectedLen
+	if actualLen < minLen {
+		minLen = actualLen
+	}
+
+	var differences []string
+
+	if actualLen != expectedLen {
+		Fail(t, fmt.Sprintf("Lengths not equal:\nexpected: %d\nactual  : %d", expectedLen, actualLen), msgAndArgs...)
+		return false
```
ccoVeille: This pattern is more common in the other asserters, I think

```suggestion
		return Fail(t, fmt.Sprintf("Lengths not equal:\nexpected: %d\nactual  : %d", expectedLen, actualLen), msgAndArgs...)
```

assert/assertions.go
```diff
@@ -2292,3 +2301,364 @@ func buildErrorChainString(err error, withType bool) string {
...
+			if !field.IsExported() {
+				continue
+			}
+			fieldValue := v.Field(i)
+			parts = append(parts, fmt.Sprintf("%s: %s", field.Name, formatValueComparison(fieldValue)))
+		}
+		return fmt.Sprintf("{%s}", strings.Join(parts, ", "))
+
+	case reflect.String:
+		return fmt.Sprintf(`"%s"`, v.String())
```
ccoVeille, line 2419: Any reason to do not use %q ?


```suggestion
		return fmt.Sprintf(`%q`, v.String())
```

assert/assertions.go
```diff
@@ -2292,3 +2301,364 @@ func buildErrorChainString(err error, withType bool) string {
...
+			value := v.MapIndex(key)
+			pairs = append(pairs, fmt.Sprintf("%s: %s", formatValueComparison(key), formatValueComparison(value)))
+		}
+		return fmt.Sprintf("map[%s]", strings.Join(pairs, ", "))
+
+	default:
+		if v.CanInterface() {
+			return fmt.Sprintf("%v", v.Interface())
+		}
+		return fmt.Sprintf("%v", v)
```
ccoVeille: Try avoiding call to formatting method when there is no need.

```suggestion
		if v.CanInterface() {
			return fmt.Sprint(v.Interface())
		}
		return fmt.Sprint(v)
```

assert/assertions.go
```diff
@@ -2292,3 +2301,364 @@ func buildErrorChainString(err error, withType bool) string {
...
+func findDifferences(expected, actual interface{}) []FieldDiff {
+	var diffs []FieldDiff
+	compareExpectedActual(expected, actual, "", &diffs)
+	return diffs
+}
+
+// compareExpectedActual compares two values recursively and records any differences in the provided diffs slice.
+// It handles complex structures like structs, maps, slices, and arrays.
+func compareExpectedActual(expected, actual interface{}, path string, diffs *[]FieldDiff) {
+	expectedValue := reflect.ValueOf(expected)
```
ccoVeille: Is there a reason to use a pointer here?


```suggestion
func findDifferences(expected, actual interface{}) []FieldDiff {
	return compareExpectedActual(expected, actual, "")
}

// compareExpectedActual compares two values recursively and records any differences in the provided diffs slice.
// It handles complex structures like structs, maps, slices, and arrays.
func compareExpectedActual(expected, actual interface{}, path string) (diffs []FieldDiff) {
	expectedValue := reflect.ValueOf(expected)
```

Then you don't have to juggle with *diffs when using append

assert/assertions.go
Andrei-hub11: Yeah. It would look weird

```
 Error:          
Differences found:
                   MATCH: [index 0] Not equal:
                   expected: {Name: ""Alice"", Age: 30, Department: {Manager: ""Alice"", Budget: 100000.5}}
                   actual  : {Name: ""Alice"", Age: 31, Department: {Manager: nil, Budget: 95000}}
                   Field differences:
                     └─ Age: 30 ≠ 31
                     └─ Department.Manager: ""Alice"" ≠ nil
                     └─ Department.Budget: 100000.5 ≠ 95000
                Test:           TestMatchExample
```

assert/assertions.go
```diff
@@ -2292,3 +2301,375 @@ func buildErrorChainString(err error, withType bool) string {
...
+	case reflect.Map:
+		if v.IsNil() {
+			return "nil"
+		}
+		var pairs []string
+		for _, key := range v.MapKeys() {
+			value := v.MapIndex(key)
+			pairs = append(pairs, fmt.Sprintf("%s: %s", formatValueComparison(key), formatValueComparison(value)))
+		}
+		return fmt.Sprintf("map[%s]", strings.Join(pairs, ", "))
```
ccoVeille, line 2452: There is something that intrigues me here

maps are not ordered. 

MapKeys neither.

https://pkg.go.dev/reflect#Value.MapKeys

I feel like the keys should be ordered when you compare

Otherwise `map[a: 1, b: 2]` string won't be equal to `map[b: 2, a: 1]` one

ccoVeille: I feel like you simply need to call sort.Slices before the strings.Join(pairs, ", ")

Andrei-hub11: You're right that MapKeys() is non-deterministic, but this doesn't affect the correctness of the comparison. The logic ensures that:

1. Every key in expected is checked against actual

2. Every key in actual is checked against expected

The final result is always correct — only the order of the error messages may vary between runs.

Additionally, I've added another test specifically to document this behavior.

ccoVeille: Exactly, but if we consider the rendering output should be tested. The fact the result varies will cause an issue.

I would like to suggest you adding example of rendering with captureTestingT mock.

It would ensure the way things are rendered

assert/find_differences_test.go
```diff
@@ -0,0 +1,1021 @@
...
+			actual: map[int]string{
+				1: "one",
+				2: "deux",
+				4: "four",
+			},
+			want: []fieldDiff{
+				{
+					Path:     "[2]",
+					Expected: "two",
+					Actual:   "deux",
```
ccoVeille, line 687: See here this test would be flacky, no?

The "position 2" in the map is not guaranteed?

Or am I commenting a different piece of code that my previous comment?

assert/assertions.go
Andrei-hub11: Here are two examples of the output — let me know if they’re enough.

Captura de Tela (348)
Captura de Tela (349)

assert/find_differences_test.go
```diff
@@ -0,0 +1,1021 @@
...
+		if !ok {
+			return false
+		}
+
+		// For simplicity, we'll just compare expected and actual as strings
+		// This isn't perfect but works for basic testing
+		if fmt.Sprintf("%v", diffA.Expected) != fmt.Sprintf("%v", diffB.Expected) ||
+			fmt.Sprintf("%v", diffA.Actual) != fmt.Sprintf("%v", diffB.Actual) {
+			return false
+		}
```
ccoVeille, line 1017: This could be a switch, it would improve readability

Also, you can use sprint without formatting here

assert/assertions.go
ccoVeille: My point was about using captureTestingT, instead of the code you have that uses Contain multiple times

ccoVeille: But let's say that for now you applied a lot of changes I suggested.

Let's wait for another maintainer feedbacks

dolmen:
So far I don't see the need to add a new `Match` function that does the same assertion than `Equal`. But let's discuss that on #1762.

PR #1768, opened 2025-06-16
assert.CallerInfo: micro cleanup

dolmen:
## Summary

Move the stackFrameBufferSize const which was in package scope but used only by CallerInfo, into CallerInfo body.

## Changes

* Move the `stackFrameBufferSize` const into `CallerInfo` body
* Small refactor use of stack frame handling to remove one reference to that const

## Motivation

Cleanup: the `stackFrameBufferSize` constant doesn't pollute the package scope anymore.

That constant wasn't even defined near the `CallerInfo` body, but at the top of the file, which made it harder to lookup.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -24,8 +24,6 @@ import (
 	"github.com/stretchr/testify/assert/yaml"
 )
 
-const stackFrameBufferSize = 10
-
 //go:generate sh -c "cd ../_codegen && go build && cd - && ../_codegen/_codegen -output-package=assert -template=assertion_format.go.tmpl"
 
 // TestingT is an interface wrapper around *testing.T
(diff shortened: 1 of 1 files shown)


assert/assertions.go
```diff
@@ -217,8 +215,10 @@ func CallerInfo() []string {
 	var line int
 	var name string
 
-	callers := []string{}
+	const stackFrameBufferSize = 10
```
ccoVeille, line 218: Maybe a comment could be added mentioning why 10 was chosen.

ccoVeille (approved):

assert/assertions.go
dolmen: 10 is the value in the example from the [`runtime.Frames` documentation](https://pkg.go.dev/runtime#Frames).

PR #1770, opened 2025-06-25
chore(deps): replace gopkg.in/yaml.v3 with github.com/goccy/go-yaml

janisz:
## Summary
`gopkg.in/yaml.v3 ` is not maintained let's use maintained alternative.

## Changes
Basically search and replace and run `go mod tidy`

## Motivation
The current lib is archived and not maintained. The alternative will be to use k8s fork but it's heavier (in terms of dependencies).

## Related issues
- https://github.com/kubernetes-sigs/yaml

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -20,7 +20,7 @@ import (
 	"github.com/davecgh/go-spew/spew"
 	"github.com/pmezard/go-difflib/difflib"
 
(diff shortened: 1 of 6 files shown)


janisz:
Looks like there are some problems as this lib requires Go 1.21
Do we need to test against 1.17 and 1.18 as they are quite old?

Issue #1695, opened 2025-01-15
Feature: support user implementations of `Match` when diffing mock arguments


connorszczepaniak-wk:
## Description
I've found it to be common that I want to set up a mock expectation for a call that takes in a struct type, and I want the matcher to match only some of the struct fields. Usually it's because there's a field with something like metrics or a timestamp that I don't consider to be relevant to my test and it's fine if those don't match, but the rest should.

The existing solution to this is to write a `mock.MatchedBy` and write a function that only checks the fields I care about. This works well enough, but there are types where we want to define globally that only certain fields should be matched when a value of that type is passed into a `mock.On`.

## Proposed solution
A solution to this would be to use an interface within `Arguments.Diff` -- currently it type-asserts on `argumentMatcher` which handles the `mock.MatchedBy` case, but if there instead was an interface with a `Matches` function that that type assertion uses, it would allow users to implement `Matches` on their own types to define custom matching behavior.

The relevant branch is here:
https://github.com/stretchr/testify/blob/master/mock/mock.go#L976-L991

The interface could look like this:
```go
type Matcher interface {
	Matches(argument interface{}) bool
}
```

And then the type assertion would simply change to the following:
```go
if matcher, ok := expected.(Matcher); ok {
```

This would be backwards compatible with `mock.MatchedBy` because it already implements the new interface.

## Use case
See above, but briefly: I want to define custom matching behavior for a given type that the `mock` package uses if a type implements the interface.

connorszczepaniak-wk:
Another thought: maybe if we went down this path, we should use a name other than `Matches` for the interface function because that seems likely to collide with other methods that may prefer to be called `Matches`. Maybe we could do something like this:

```go
type ArgumentMatcher interface {
	MockArgumentMatches(argument interface{}) bool
}
```

And then we'd just implement `MockArgumentMatches` on the current `mock.MatchedBy` implementation.

iwittkau:
This would also be nice for expecters: instead of relying on `RunAndReturn` you could pass the expected argument directly with the matcher.

connorszczepaniak-wk:
I'd like to bump this! @brackendawson you reacted with 👎 above, can you please elaborate?

Issue #1234, opened 2022-07-20
Diff on mocking library


golden-protocol:
I have a structure with a field that is a map[string]interface{}, that is matching here is the diff. 

```
map[string]interface {}=map[KEY:30]) != (map[string]interface {}=map[KEY:30]
```

However you can see they match.... am i missing something?

brackendawson:
I don't know what you're doing here, can your provide a minimal testcase which shows this issue?

brackendawson:
Thanks for the updated description. Sounds like we need to change the way the values are printed here.

Issue #1002, opened 2020-09-29
Export mock.argumentMatcher


rzajac:
I would really benefit from mock.argumentMatcher to be exported. I know there was already a pull request #333 but it was closed.

The rationale is the same as in the pull request:

> This allows the creation of predicate-generating functions, which can make code a lot cleaner. To give a very simple example, this is not possible with the current non-exported argumentMatcher:

Issue #817, opened 2019-09-10
[feature] mock.MatchedByRegexp(regexp string)


KrzysztofMadejski:
A method  that tests string argument if it matches regexp would be helpful.

Currently I'm using `MatchedBy(func)`:
```go
mock.MatchedBy(func(path string) bool {
	return regexp.MustCompile(pathRegexp).MatchString(path)
})
```

But the error (panic btw) is hard to analyze easily:
```
mock: Unexpected Method Call
-----------------------------

Create(string)
		0: "/Home/user/development/20190910170936437.csv"

The closest call I have is: 

Create(mock.argumentMatcher)
		0: mock.argumentMatcher{fn:reflect.Value{typ:(*reflect.rtype)(0x9be540), ptr:(unsafe.Pointer)(0xc000021260), flag:0x13}}


Diff: 0: FAIL:  (string=/Home/user/development/20190910170936437.csv) not matched by func(string) bool
```

It would be a lot nicer to have standard error: 
```
Expected: "^....$" (regexp)
Got: ""
```

dolmen:
Please add a full example of your issue that runs on go.dev/play. Here is a template: https://go.dev/play/p/AsRNcxCRdWx

dolmen:
@KrzysztofMadejski Here is a simple workaround: add this function to your test file:

```go
func mockMatchedByRegexp(pattern string) {
    mock.MatchedBy(func(path string) bool {
        ok, err := regexp.MatchString(pattern, path)
        if err != nil {
            panic(err)
        }
        return ok
    }
}
```

dolmen:
@KrzysztofMadejski Adding a fully runnable example would be helpful to understand your pain. Here is a template: https://go.dev/play/p/AsRNcxCRdWx

PR #1773, opened 2025-07-23
feat: add support for go-cmp argument matching

IlyasYOY:
## Summary

Added support for Google's `go-cmp` library in the mock package for customizable deep comparisons during test assertions: 

- Updated `go.mod` to include `go-cmp` dependency
- Introduced `goCmpArgument` type in `mock/mock.go` to encapsulate comparison values and options
- Implemented `GoCmp()` helper function to create comparison arguments
- Use `cmp.Diff()` for precise comparison output
- Allow customization via `cmp.Options`

## Changes

Adds new assertion behavior (not a bug fix):
- Introduces `GoCmp()` as an alternative to `mock.MatchedBy()`
- Enables field-level comparison customization via `cmp.Option`
- Adds diff output formatting in error messages

## Motivation

Support more precise comparisons than `reflect.DeepEqual` (e.g., ignoring specific fields)

## Example usage

```go
// Basic usage with a map
mock.On("Method", mock.GoCmp(map[string]bool{"test": true}))

// With options for unexported fields
mock.On("Method", mock.GoCmp(&ExampleType{ran: false}, cmp.AllowUnexported(ExampleType{})))
```

## Related issues

Might relate to some issues looking for a better way to match arguments: 

- https://github.com/stretchr/testify/issues/1695
- https://github.com/stretchr/testify/issues/1234
- https://github.com/stretchr/testify/issues/1002
- https://github.com/stretchr/testify/issues/817

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ go 1.17
 
 require (
 	github.com/davecgh/go-spew v1.1.1
+	github.com/google/go-cmp v0.6.0
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/stretchr/objx v0.5.2 // To avoid a cycle the version of testify used by objx should be excluded below
 	gopkg.in/yaml.v3 v3.0.1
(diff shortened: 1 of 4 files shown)


IlyasYOY:
This PR looks like very much like https://github.com/stretchr/testify/pull/579 (but for the `mock` module).

brackendawson:
I'm confident in saying we're not open to switching out matching libraries via different function names in testify's packages, or adding more external dependencies. I'm sorry but this is a not planned and close.

PR #1779, opened 2025-08-14
Fix: a data race condition after a timeout on an `assert.Eventually` call.

kcross-ctoken:
## Summary
Fixes a data race when using assert.Eventually so the variables can be safely used after a timeout

This was found in my test code when the eventually timed out and I was using something set in the closure. It required the `--race`  `go test` parameter set.

## Changes
added a wait group to wait for the end of the go routine call in assert.Eventually

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,8 @@ _testmain.go
 
 .DS_Store
 
+.idea
+
 # Output of "go test -c"
 /assert/assert.test
 /require/require.test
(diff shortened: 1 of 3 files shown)


brackendawson (changes requested):
I think you've found a valid bug but your fix changes the behavior of Eventually. Right now it will return and fail the test as soon as the timeout is hit but with this change it will now wait until the last condition function returns. While I actually think waiting is the better behavior it's not something we can just go changing.

This also doesn't fix the issue for other Eventually/Consistently/Never assertions.

brackendawson:
#1657 would resolve the issue here I believe.

PR #1780, opened 2025-08-20
fix(require): correct invalid examples in doc comments #issue-1776

### LINKED ISSUE #1776: require: invalid examples from doc comments as functions don't return bool
## Description

The doc comments for multiple functions in the `require` package have code examples that are invalid. Example:

https://github.com/stretchr/testify/blob/a53be35c3b0cfcd5189cffcfd75df60ea581104c/require/require.go#L1391

The example reads:
```go
	  if require.NoError(t, err) {
		   require.Equal(t, expectedObj, actualObj)
	  }
```
but the `NoError()` function does not return any value, so this is invalid. 

There are similar invalid examples in the doc comments for `NoErrorf`, `NotEmpty` and `NotEmptyf`.

PCloud63514:
## Summary
Correct invalid examples in require package doc comments.
require does not return a boolean value; its responsibility is to fail immediately.

## Changes
- Updated doc comments where examples incorrectly used require inside an if condition.
```go
// NoError asserts that a function returned no error (i.e. `nil`).
//
//	  actualObj, err := SomeFunction()
//	  require.NoError(t, err)
//	  require.Equal(t, expectedObj, actualObj)
func NoError(t TestingT, err error, msgAndArgs ...interface{})
```

## Motivation
Unlike assert, require functions do not return a boolean value.
The previous examples suggested patterns such as if require.NoError(t, err) { ... }, which are invalid.
This PR corrects the documentation to properly reflect the intended usage of require.

## Related issues

Closes #1776

### DIFF
--- a/require/require.go
+++ b/require/require.go
@@ -1388,9 +1388,8 @@ func NoDirExistsf(t TestingT, path string, msg string, args ...interface{}) {
 // NoError asserts that a function returned no error (i.e. `nil`).
 //
 //	  actualObj, err := SomeFunction()
-//	  if require.NoError(t, err) {
-//		   require.Equal(t, expectedObj, actualObj)
-//	  }
+//	  require.NoError(t, err)
+//	  require.Equal(t, expectedObj, actualObj)
 func NoError(t TestingT, err error, msgAndArgs ...interface{}) {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
@@ -1404,9 +1403,8 @@ func NoError(t TestingT, err error, msgAndArgs ...interface{}) {
 // NoErrorf asserts that a function returned no error (i.e. `nil`).
 //
 //	  actualObj, err := SomeFunction()
-//	  if require.NoErrorf(t, err, "error message %s", "formatted") {
-//		   require.Equal(t, expectedObj, actualObj)
-//	  }
+//	  require.NoErrorf(t, err, "error message %s", "formatted")
+//	  require.Equal(t, expectedObj, actualObj)
 func NoErrorf(t TestingT, err error, msg string, args ...interface{}) {
 	if h, ok := t.(tHelper); ok {
 		h.Helper()


_codegen/main.go
```diff
@@ -298,6 +299,61 @@ func (f *testFunc) CommentWithoutT(receiver string) string {
 	return strings.Replace(f.Comment(), search, replace, -1)
 }
 
+func requireCommentParseIf(s string) string {
```
ccoVeille, line 318: This method should be tested

_codegen/main.go
```diff
@@ -298,6 +299,61 @@ func (f *testFunc) CommentWithoutT(receiver string) string {
...
+				body := strings.TrimLeft(line[slash+2:], " \t")
+				out = append(out, commentPrefix+body)
+			} else {
+				out = append(out, line)
+			}
+		}
+	}
+
+	return strings.Join(out, "\n")
+}
```
ccoVeille: I'm worried about a badly indented comment.

I'm unsure how to handle this correctly.

PCloud63514:
@ccoVeille Thanks, I also felt the code was a bit messy, so I tried to simplify it further.

Regarding your comment about tests: since this is only used for code generation,
I'm not sure if it makes sense to add dedicated tests here.
Do you have any suggestion for a good approach?

_codegen/main.go
dolmen: This is codegen. Tests are not strictly necessary: checking the generated output is enough.

However, tests would still be helpful to document what the function expects and does. Tests will also be useful for future refactorings.

Even more helpful would be some godoc for the function because the `require` prefix is misleading: on a first read I got it as verb while it is about the `require` package.

brackendawson: An issue is that it's unclear from the function name or the code; what is this function is supposed to do. A test or a good comment before the function would help.

_codegen/main.go
```diff
@@ -298,6 +299,50 @@ func (f *testFunc) CommentWithoutT(receiver string) string {
...
+	rePrefix := regexp.MustCompile(`//[ \t]*`)
+	ifDepth := 0
+	existsIfRequire := false
+	SomeFunctionLineIdx := 0
+
+	for i, line := range lines {
+		commentPrefix := rePrefix.FindString(line)
+		comment := strings.TrimSpace(line[2:])
+
+		if strings.HasSuffix(comment, "SomeFunction()") {
```
dolmen: I don't like the explicit reference to `SomeFunction()`. This should instead look for the first line of an example block by looking at indent. But maybe there is a reason why this wasn't enough?

dolmen (changes requested):

_codegen/main.go
ccoVeille: That was my point, indeed

_codegen/main.go
PCloud63514: Hi @dolmen 

I’ve removed the explicit `SomeFunction()` reference.
That logic was originally added considering the `type S struct {` comment,
but now the handling is simplified to focus only on if require. cases.

_codegen/main.go
PCloud63514: https://github.com/stretchr/testify/pull/1780#issuecomment-3210206675
I wasn’t familiar with the reply feature, so I didn’t realize multiple conversations were happening. 😅
Considering tests, would it be better to separate this method and handle it individually?

dolmen: @PCloud63514 Please start by adding a comment block to document the function.

PCloud63514: Added a doc comment for `requireCommentParseIf` to clarify its purpose.requireCommentParseIf

PCloud63514:
Is there anything more I should do, or anything I might be misunderstanding?

PR #1782, opened 2025-08-25
_codegen: copy dependency github.com/ernesto-jimenez/gogen/imports

dolmen:
## Summary

Remove external dependency on module `github.com/ernesto-jimenez/gogen` by copying its package [`github.com/ernesto-jimenez/gogen/imports`](https://pkg.go.dev/github.com/ernesto-jimenez/gogen/imports) (MIT license) as `_codegen/internal/imports`. 

## Changes
* copy `github.com/ernesto-jimenez/gogen/imports/imports.go` as `_codegen/internal/imports/imports.go` and add a license block to the file
* update imports in `_codegen/main.go` and `_codegen/go.mod`.

## Motivation
The removal of this external dependency (by duplicating its code) improves the control of our source code. It will also allow more refactoring of the codegen steps (in particular this now allow to remove the separate go module).

Cc: @ernesto-jimenez

### DIFF
--- a/_codegen/go.mod
+++ b/_codegen/go.mod
@@ -1,5 +1,3 @@
 module github.com/stretchr/testify/_codegen
 
 go 1.11
-
-require github.com/ernesto-jimenez/gogen v0.0.0-20180125220232-d7d4131e6607
--- a/_codegen/go.sum
+++ b/_codegen/go.sum
@@ -1,2 +1,0 @@
-github.com/ernesto-jimenez/gogen v0.0.0-20180125220232-d7d4131e6607 h1:cTavhURetDkezJCvxFggiyLeP40Mrk/TtVg2+ycw1Es=
-github.com/ernesto-jimenez/gogen v0.0.0-20180125220232-d7d4131e6607/go.mod h1:Cg4fM0vhYWOZdgM7RIOSTRNIc8/VT7CXClC3Ni86lu4=
--- a/_codegen/main.go
+++ b/_codegen/main.go
@@ -23,7 +23,7 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/ernesto-jimenez/gogen/imports"
+	"github.com/stretchr/testify/_codegen/internal/imports"
 )
 
 var (


_codegen/internal/imports/imports.go
ccoVeille, line 1: I would add a README file aside this one mentioning where the code come from (its URL) and why it was imported (the reason)

Also I would add a LICENCE file instead of this header.

brackendawson:
This doesn't bring the commit history from source repo, just the files. Should we bring in a subtree instead, or do you think this is a better approach? I know the subtree has proven extremely difficult to re-base.

ccoVeille (approved):
LGTM 👍

A few remarks anyway

_codegen/internal/imports/imports.go
brackendawson: The imported code and testify are both MIT licenced, the MIT license of testify applying to the now embedded codegen package is quite correct.

ccoVeille: Indeed, what about the README file then?

dolmen:
@brackendawson wrote:
> This doesn't bring the commit history from source repo, just the files.

I don't think we need the history for the file. My goal is to get rid of the code we don't need in that package, and may be ultimately move the few things we use into `_codegen/main.go`.

Also we don't need a test suite for that package: we just need the code generation to work with the minimum maintainable code, so the code generation output is the test suite.

_codegen/internal/imports/imports.go
dolmen: My aim is to get rid of the package and to replace the code from that package with code in `_codegen/main.go`.
The source URL is in the commit message and in this PR should be enough to document history.

I've assigned @ernesto-jimenez as a reviewer as he wrote that code and was also a maintainer of Testify, so his opinion is what matters the most to me.

ernesto-jimenez: No objections from my side :)

brackendawson:
Code is always more maintainable when it has tests. The code you've copied in verbatim had one test which you have not vendored. Why are we not bringing that too?

PR #1787, opened 2025-08-27
assert: IsIncreasing et al can return false w/out failing

### LINKED ISSUE #1419: assert.IsIncreasing can return false without failing the test

brackendawson:
## Summary
If you passed a non-collection to IsIncreasing or any of its compatriots then the assertion would return false without failing the test.

## Changes
* Make `isOrdered()` return `Fail()` rather than `false` in validation so that the test will fail.
* Test the above.

## Motivation
The following invalid usage would previously pass:
```go
var nonList int
assert.IsIncreasing(t, nonList)
```

## Related issues
Closes #1419

### DIFF
--- a/assert/assertion_order.go
+++ b/assert/assertion_order.go
@@ -9,7 +9,7 @@ import (
 func isOrdered(t TestingT, object interface{}, allowedComparesResults []CompareType, failMessage string, msgAndArgs ...interface{}) bool {
 	objKind := reflect.TypeOf(object).Kind()
 	if objKind != reflect.Slice && objKind != reflect.Array {
-		return false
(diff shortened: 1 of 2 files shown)


brackendawson:
Thank you @ccoVeille, I've re-based the PR and resolved the conflicts.

PR #1791, opened 2025-09-11
assert.YAMLEq: Document mutlidoc behavior

### LINKED ISSUE #1281: Asserting nested yaml documents are working as expected

brackendawson:
## Summary
Document the limitation that YAMLEq only compares the first document.

## Changes
* Add tests to cover more of the current YAMLEq behavior.
* Document that YAMLEq only compares the first document.

## Motivation
It would be preferable to fix YAMLEq to compare all documents because that's what it outwardly claims to do and is what one would expect it to do. #1284 would do that, but #1579 as made this practically impossible. So in testify v1 we must document the behavior as is.

## Related issues
Addresses but does not resolve #1281

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1880,7 +1880,7 @@ func JSONEq(t TestingT, expected string, actual string, msgAndArgs ...interface{
 	return Equal(t, expectedJSONAsInterface, actualJSONAsInterface, msgAndArgs...)
 }
 
-// YAMLEq asserts that the first document in the two YAML strings is equivalent.
+// YAMLEq asserts that the first documents in the two YAML strings are equivalent.
 //
 //	expected := `---
 //	key: value
--- a/assert/assertion_format.go
+++ b/assert/assertion_format.go
@@ -849,7 +849,7 @@ func WithinRangef(t TestingT, actual time.Time, start time.Time, end time.Time, 
 	return WithinRange(t, actual, start, end, append([]interface{}{msg}, args...)...)
 }
 
-// YAMLEqf asserts that the first document in the two YAML strings is equivalent.
+// YAMLEqf asserts that the first documents in the two YAML strings are equivalent.
 //
 //	expected := `---
 //	key: value
--- a/assert/assertion_forward.go
+++ b/assert/assertion_forward.go
@@ -1690,7 +1690,7 @@ func (a *Assertions) WithinRangef(actual time.Time, start time.Time, end time.Ti
 	return WithinRangef(a.t, actual, start, end, msg, args...)
 }
 
-// YAMLEq asserts that the first document in the two YAML strings is equivalent.
+// YAMLEq asserts that the first documents in the two YAML strings are equivalent.
 //
 //	expected := `---
 //	key: value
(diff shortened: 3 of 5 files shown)


assert/assertions.go
```diff
@@ -1880,7 +1880,7 @@ func JSONEq(t TestingT, expected string, actual string, msgAndArgs ...interface{
 	return Equal(t, expectedJSONAsInterface, actualJSONAsInterface, msgAndArgs...)
 }
 
-// YAMLEq asserts that two YAML strings are equivalent.
+// YAMLEq asserts that the first document in the two YAML strings is equivalent.
```
ccoVeille:
```suggestion
// YAMLEq asserts that the first document in the YAML documents is equivalent to the expected one.
```

This would be better.  First and two didn't sound good

ccoVeille: I don't think it's enough.

There should be an example close yo what you have in the unit test.

brackendawson: Its not accurate because it also only considers the first document in the expected yaml string.

ccoVeille: Oh indeed.

ccoVeille: Thanks it's way clearer

assert/assertions.go
```diff
@@ -1880,7 +1880,19 @@ func JSONEq(t TestingT, expected string, actual string, msgAndArgs ...interface{
...
```
ccoVeille: The current proposal sounds grammatically incorrect to me.

What about this?

```suggestion
// YAMLEq asserts that the first documents in the two YAML strings are equivalent.
```

Or this maybe


```suggestion
// YAMLEq asserts that the first YAML document in each string is equivalent.
```

Or this

```suggestion
// YAMLEq asserts that the first YAML document in one string is equivalent to the first YAML document in the other string.
```

Wording is a pain, grammar is worst.

assert/assertions.go
```diff
@@ -1880,7 +1880,19 @@ func JSONEq(t TestingT, expected string, actual string, msgAndArgs ...interface{
 	return Equal(t, expectedJSONAsInterface, actualJSONAsInterface, msgAndArgs...)
 }
 
-// YAMLEq asserts that two YAML strings are equivalent.
+// YAMLEq asserts that the first documents in the two YAML strings are equivalent.
```
ccoVeille, line 1883: I'm sorry, I thought about it again.

My problem now us the fact "first documents are the same" could imply that it's about checking the first nth documents

What about these?

```suggestion
// YAMLEq asserts that the first YAML document of each string is equivalent.
```

```suggestion
// YAMLEq asserts that the first document of each YAML string is equivalent.
```
I prefer the first one

brackendawson: > YAMLEq asserts that the first YAML document of each string is equivalent.

This implies that the strings contain documents other than YAML.

> YAMLEq asserts that the first document of each YAML string is equivalent.

The first document of each YAML string is equivalent to what?

Your suggestion which I accepted is the most correct.

PR #1796, opened 2025-09-16
CI: run tests separately from build jobs

dolmen:
## Summary
Reorganize GitHub Actions workflows to run tests the same way on all Go versions, and so remove running tests from the 'build' jobs (which run job only on 'oldstable' and 'stable').

## Changes
* Refactor .github/workflows/main.yml
* Add comments on build matrices to explain how the order of Go versions is defined

## Motivation
This makes the build matrix of the 'test' job more consistent.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,6 +7,8 @@ jobs:
     strategy:
       matrix:
         go_version:
+          # Our developer environment (especially the code generator)
+          # must work with all the supported Go releases.
           - stable
           - oldstable
     steps:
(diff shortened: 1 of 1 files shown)


brackendawson (changes requested):
The `build:` job need only run for stable now.

I'm not 100% happy with it being called "Build" but I can't think of a better name.

dolmen:
I think that the 'build' tasks must always be runnable with both `stable` and `oldstable`. This is important after a new Go release, when not every contributor has switched to the new one.

So I prefer to keep `oldstable`.

.github/workflows/main.yml
alexandear, line 43: For keeping in sorted order:
```
          - "1.17"
          - "1.18"
          - "1.19"
          - "1.20"
          - "1.21"
          - "1.22"
          - "1.23"
          - "1.24"
          - oldstable
          - stable
```

oldstable - 1.25, stable - 1.26

alexandear (approved):

.github/workflows/main.yml
dolmen: The tests are executed in the order of the array.
It is important to get the results for `stable` and `oldstable` first, especially failures.

So `stable` and `oldstable` must always stay at the top.

ccoVeille: Maybe we could invert all the list then

stable
oldatable
1.24
1.23
...
1.17

ccoVeille (dismissed):

.github/workflows/main.yml
alexandear: > So stable and oldstable must always stay at the top.

Should we extract them to a separate job?

dolmen: I have now added comments in the file itself to explain the logic to any future contributor.

ccoVeille (approved):

alexandear (approved):

PR #1797, opened 2025-09-16
_codegen: add license and merge package 'imports'

dolmen:
## Summary
Merge code from package internal/imports (recently copied from package github.com/ernesto-jimenez/gogen/imports) into the codegen program itself. Also copy the license attached to it into the _codegen/main.go source, as _codegen/main.go was originally by the same author (@ernesto-jimenez) as the imported code and matches the license terms of the global Testify project.
Also: @ernesto-jimenez who was queried about this project, didn't block it. https://github.com/stretchr/testify/pull/1782#discussion_r2322798422

## Changes
* add license from _codegen/internal/imports/imports.go into _codegen/main.go, and add also other _codegen contributor (from Git history)
* copy type `_codegen/internal/imports.imp` into _codegen/main.go as `imports`.
* remove package `_codegen/internal/imports` that isn't need anymore

## Motivation
Reduce code in _codegen

## Related issues
This a follow-up to #1782.


_codegen/main.go
```diff
@@ -1,3 +1,31 @@
+/*
+The MIT License (MIT)
+
+Copyright (c) 2015,2016 Ernesto Jiménez
+Copyright (c) 2019 Leigh McCulloch
+Copyright (c) 2020 Matt Gorzka
+Copyright (c) 2024 Simon Schulte
+Copyright (c) 2023,2025 Olivier Mengué
```
brackendawson: This is unusual and would require careful attention to keep up to date. Follow the same pattern as testify and allow VCS to be the source of the names: https://github.com/stretchr/testify/blob/6c30310a8d0c34a6aa840c28898363f2d434147c/LICENSE#L3

```suggestion
Copyright (c) 2015-2025 Ernesto Jiménez and contributors.
```

brackendawson (changes requested):

_codegen/main.go
dolmen: I was tempted to do that, but to stay on the safe side, I chose to mention all contributors I found in the Git history.

As this is not really engineering, but legal stuff, I would be more confident if I you could give me some references to back the choice of a particular style of copyright mentions.

brackendawson: I can't give you the original decision to claim on behalf of the contributors because it's not written down, but in addition to what is already done in our own source code I found this in my research: https://opensource.stackexchange.com/questions/4960/can-a-team-be-the-copyright-holder-mit

We should be consistent with ourselves.

dolmen: Fixed, and updated to 2026 as I did the change today.

ccoVeille (approved):

PR #1799, opened 2025-09-17
mock.AnythingOfType fix panic when matching nil type

### LINKED ISSUE #1209: Mock panics when argument of nil type checked against AnythingOfType

brackendawson:
Fix a panic in the AnythingOfType matcher when matching nil type.

It's reported that IsType had the same panic but a refactor between then and now appears to have fixed this.

## Changes
* Handle the nil type case before calling reflect.Type.String.
* Because (<nil>=<nil>) looks confusing, change the failure message value format to the more idiomatic (%T)(%v).
* In order to better test the failure message, change the mockTestingT to capture the messages passed to it.

## Motivation
AnythingOfType would panic if the actual arg value was of type nil. 

## Related issues
* Closes #1209
* Supersedes #1219

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -808,7 +808,8 @@ type AnythingOfTypeArgument = anythingOfTypeArgument
 type anythingOfTypeArgument string
 
 // AnythingOfType returns a special value containing the
-// name of the type to check for. The type name will be matched against the type name returned by [reflect.Type.String].
+// name of the type to check for. The type name will be matched against the type
+// name returned by [reflect.Type.String], or against "<nil>" for nil type .
 //
 // Used in Diff and Assert.
 //
@@ -1007,10 +1008,16 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			switch expected := expected.(type) {
 			case anythingOfTypeArgument:
 				// type checking
-				if reflect.TypeOf(actual).Name() != string(expected) && reflect.TypeOf(actual).String() != string(expected) {
+				actualTypeName := "<nil>"
+				actualTypeString := "<nil>"
+				if actual != nil {
+					actualTypeName = reflect.TypeOf(actual).Name()
+					actualTypeString = reflect.TypeOf(actual).String()
+				}
+				if actualTypeName != string(expected) && actualTypeString != string(expected) {
 					// not match
 					differences++
-					output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type %s - %s\n", output, i, expected, reflect.TypeOf(actual).Name(), actualFmt)
+					output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type %s - %s\n", output, i, expected, actualTypeName, actualFmt)
 				}
 			case *IsTypeArgument:
 				actualT := reflect.TypeOf(actual)
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -131,7 +131,8 @@ func (i *TestExampleImplementation) TheExampleMethodFuncType(fn ExampleFuncType)
 
 // MockTestingT mocks a test struct
 type MockTestingT struct {
-	logfCount, errorfCount, failNowCount int
+	logfMessages, errorfMessages []string
+	failNowCount                 int
 }
 
 // Helper is like [testing.T.Helper] but does nothing.
@@ -139,12 +140,12 @@ func (MockTestingT) Helper() {}
 
 const mockTestingTFailNowCalled = "FailNow was called"
 
-func (m *MockTestingT) Logf(string, ...interface{}) {
-	m.logfCount++
+func (m *MockTestingT) Logf(format string, a ...interface{}) {
+	m.logfMessages = append(m.logfMessages, fmt.Sprintf(format, a...))
 }
 
-func (m *MockTestingT) Errorf(string, ...interface{}) {
-	m.errorfCount++
+func (m *MockTestingT) Errorf(format string, a ...interface{}) {
+	m.errorfMessages = append(m.errorfMessages, fmt.Sprintf(format, a...))
 }
 
 // FailNow mocks the FailNow call.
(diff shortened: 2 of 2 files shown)


mock/mock_test.go
```diff
@@ -131,20 +131,21 @@ func (i *TestExampleImplementation) TheExampleMethodFuncType(fn ExampleFuncType)
...
 
 // Helper is like [testing.T.Helper] but does nothing.
 func (MockTestingT) Helper() {}
 
 const mockTestingTFailNowCalled = "FailNow was called"
 
-func (m *MockTestingT) Logf(string, ...interface{}) {
-	m.logfCount++
+func (m *MockTestingT) Logf(format string, a ...interface{}) {
+	m.logfMessages = append(m.logfMessages, fmt.Sprintf(format, a...))
```
ccoVeille, line 144: 🎨 nitpicking here

They are no longer logfmessages if you used sprintf before adding string to them

mock/mock_test.go
```diff
@@ -131,20 +131,21 @@ func (i *TestExampleImplementation) TheExampleMethodFuncType(fn ExampleFuncType)
...
 }
 
-func (m *MockTestingT) Errorf(string, ...interface{}) {
-	m.errorfCount++
+func (m *MockTestingT) Errorf(format string, a ...interface{}) {
+	m.errorfMessages = append(m.errorfMessages, fmt.Sprintf(format, a...))
```
ccoVeille, line 148: Same remark

https://github.com/stretchr/testify/pull/1799#discussion_r2354563701

mock/mock.go
```diff
@@ -1006,6 +1007,13 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			switch expected := expected.(type) {
 			case anythingOfTypeArgument:
 				// type checking
+				if actual == nil {
+					if expected != "<nil>" {
+						differences++
+						output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type <nil> - %s\n", output, i, expected, actualFmt)
+					}
+					continue
```
ccoVeille: I don't get the comparison on `"<nil>"` here.

Can't we check on a real value?

I'm paranoid but won't the code react to a string being `"<nil>"`

mock/mock.go
```diff
@@ -1006,6 +1007,13 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			switch expected := expected.(type) {
 			case anythingOfTypeArgument:
 				// type checking
+				if actual == nil {
+					if expected != "<nil>" {
+						differences++
+						output = fmt.Sprintf("%s\t%d: FAIL:  type %s != type <nil> - %s\n", output, i, expected, actualFmt)
```
ccoVeille: <nil> is not a type itself


```suggestion
						output = fmt.Sprintf("%s\t%d: FAIL:  type %s != <nil> - %s\n", output, i, expected, actualFmt)
```

mock/mock_test.go
```diff
@@ -1914,8 +1915,8 @@ func Test_Arguments_Diff(t *testing.T) {
 	diff, count = args.Diff([]interface{}{"Hello World", 456, "false"})
 
 	assert.Equal(t, 2, count)
-	assert.Contains(t, diff, `(int=456) != (int=123)`)
-	assert.Contains(t, diff, `(string=false) != (bool=true)`)
+	assert.Contains(t, diff, `(int)(456) != (int)(123)`)
+	assert.Contains(t, diff, `(string)(false) != (bool)(true)`)
```
ccoVeille: This format seems awkward to me.

I feel like we are using something like this everywhere else


```suggestion
	assert.Contains(t, diff, `456(int) != 123(int)`)
	assert.Contains(t, diff, `false(string) != true(bool)`)
```

ccoVeille: Think about something storing "bool" in a string. The result is confusing

mock/mock_test.go
brackendawson: Messages logged via Logf, logfMessages.

mock/mock.go
brackendawson: In AnythingOfType argument, expected is a named type of string.

We can't compare to a type because we're supposed to be comparing strings. I have re-factored this to set actual to `<nil>` do the comparison once.

mock/mock.go
brackendawson: This was to match the format of the other failure case, it should be kept to disambiguate type <nil> from value <nil>. I have now deleted this line, though.

mock/mock_test.go
brackendawson: We've been moving to (%T)(%s) for these types of statement because it's the style fmt's %v uses to convey a type and a value. I think it's the most clear in the face of all type, especially pointers:

```
(*uint16)(9)
```

brackendawson: `(string)(false)` is a bit confusing. However `*int(1)` is also confusing because it's invalid in go. I'm going to restore the original format.

ccoVeille: I'm a bit lost there

I was suggesting to use "false (string)"

I felt like it was better than "(false)(string)"

So the example of "*int(1)" is incorrect. I was asking to use "1 (*int)" (even if I'm unsure this format exists as you said)

My point was to use something like "%v (%T)" that we are already using. And fix the strange format that exists in code for this method.

brackendawson: I mis-read your suggestion, `%v(%T)` is also quite weird to me because I was hoping to find something with a grounding in Go syntax.

This probably warrants de-coupling into a separate issue where we can consider the format we want testify to move to as a whole. As I said for now this PR just fixes the panic and leaves the format alone.

PR #1809, opened 2025-10-14
Do not hang `Eventually` on failed condition

ubunatic:
## Summary
Do not hang `Eventually` when condition does not return normally and fail fast instead. \
Do not hang `Never` when condition does not return normally to prevent false positives.

## Code Changes
* Unblock blocking `ch <- condition()` with a `defer` func in `Eventually`
* Unblock blocking `ch <- condition()` with a `defer` func in `Never`
* React on all `runtime.Goexit()` variants to fail accordingly in `Eventually`, `Never`, and `EventuallyWithT`
* Report any unclean condition goroutine exit as `"Condition exited unexpectedly"`

## Details/Effect 
* Previously hanging tests caused by unclean condition exits will fail fast
* Entirely failed tests will not wait for the full timeout anymore but fail fast
* `Eventually` and `Never` will behave more unsurprisingly and fail on exited conditions
* `Never` will not report false positives caused by unclean exits
* `EventuallyWithT` can distinguish between `collect` failure and outer failure and will fail on an exit if it was not recorded in the `collect`. The condition func must call `collect.FailNow()` (also indirectly) or return normally to not fail
* The sum of changes allow for safely calling `require.*` functions also on the outer/parent `t` inside conditions to fail the test immediately 

## Doc/Internal Changes
* Updated and linked code and repo documentation on the behavior
* Extended code examples to explain usage of `Eventually`, `EventuallyWithT`, and `Never` with details on best practices
  * general: Use atomic operations!
  * new: It's ok to call `require.*`
* Updated `_codegen` to respect source code links like in `// Also see [Eventually] ...` and to reuse the last string parameter as error message (updated a few code examples to comply)
* Changed `MockT.Failed` to `MockT.Failed()` to better comply with `testing` interfaces and for allowing to reuse the `CollectT` implementation for testing tests.

## Motivation
* Currently when calling `require.Fail`, or `t.FailNow()`, or a similar function inside the `Eventually` condition, or when the condition does not return due to any indirect call to `runtime.Goexit`, the `ch <- condition` will block and cause the tick loop to hang until the timeout is reached.
* There was no way to detect the failed condition and hanging loop.

## Related issues
* https://github.com/stretchr/testify/issues/1810

## Tests
* added many test to test the new fast exit behavior
* added "demo" tests to show that panic still exit the runner (run with env: `TestPanic=1`)

## Discussion
* I propose to use a new message `"Condition exited unexpectedly"`
* Alternatively we can use the same error message `"Condition never satisfied"` as used for the timeout case, to not break existing external test analysis approaches.
* (In the future) we could also detect panics and report a `"Condition panicked"` to educate developers what actually happened. Out of scope for this change, could be a `v2` feature.

## Impact
1. (Our) Go developers tend to write `require.NoError` or similar on the outer `t` inside the `Eventually` condition, causing the tests to hang. This regularly creates confusion. With this change the test behavior becomes more explicit. A `require.Fail(t)` or similar inside the `Eventually` condition will just behave as if called outside: make the test fail, but without hanging the test. "As a developer" this is how I would expect my test tests to fail.

2. Just looking at our internal codebase, I can see hundreds of tests that will fail fast with this change. Globally, I expect to save thousands of hours of wasted compute caused by the hanging tests.

3. The more explicit error will better educate developers what is wrong in their code. The fixed `Never` behavior will reveal hidden issues. And the safe use of `require.*` in the condition functions will make 1000s of existing test no longer violate the best practices.

## Risk Assessment
* **Panics stay as is**. The behavior is the same -> **low risk**
* **`Eventually` exits faster**. This may cause some timing issues, but the core logic stays the same -> **low risk**
* **`EventuallyWithT` detects external exits**. Logic on the `collect` stays as is. External exits through the outer `t` will just fail fast. Before, `EventuallyWithT` would succeed (wrongly) and only the outer test would fail. Now both will fail. -> **low risk**
* **`Never` reports more failures.** It exits with an error now and does not block anymore. Before, it would block and never check the condition anymore, which may be satisfied in the meantime, and then Never would wrongly report a success. This will reveal hidden satisfy bugs -> **low risk**
* **`MockT.Failed()`**. MockT is not used outside of this repo ([GH search](https://github.com/search?q=require.MockT&type=code) + [Sourcegraph search](https://sourcegraph.com/search?q=context:global+%22MockT.Failed%22+OR+%22require.MockT%22&patternType=keyword&case=yes&sm=0)) -> **low risk**

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,11 @@
+# This file is used to configure Git attributes for the repository.
+# It is particularly useful for managing how generated code is treated in pull requests.
+# For more information, see:
+# https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github
+
+# Code for assertions and requirements is generated automatically by _codegen.
+# This also updates code comments and code examples in the documentation.
+# These comments and examples must be reviewed when they change,
+# since updates to the generated code comments may not always be correct.
+assert/*.go  linguist-generated=false
+require/*.go linguist-generated=false


assert/assertions.go
```diff
@@ -2029,13 +2044,24 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
...
+			go checkCond() // Schedule the next check.
+		case v := <-resultCh:
+			switch v {
+			case failed:
+				// Condition panicked or test failed and finished.
+				// Cannot determine correct result.
+				// Cannot decide if we should continue gracefully or not.
+				// We can stop here and now, and mark test as failed with
+				// the same error message as the timeout case.
+				return Fail(t, "Condition never satisfied", msgAndArgs...)
```
ubunatic: Maybe we should use `FailNow` here, since this unexpected behavior.
Also see [here](https://github.com/ubunatic/go-contrib-testify/pull/1) for a more advanced variant.

assert/assertions.go
```diff
@@ -2010,15 +2010,30 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
...
-	checkCond := func() { ch <- condition() }
+	const failed = 0
+	const stop = 1
+	const noStop = 2
+
+	resultCh := make(chan int, 1)
+	checkCond := func() {
+		result := failed
+		defer func() {
+			resultCh <- result
```
ubunatic, line 2075: For cleaner test results, we could also catch panics here and make the two failure modes `failed` and `panic` explicit with separate error messages. See my other variant [here](https://github.com/ubunatic/go-contrib-testify/blob/b9276033b772155ac653fa542e31ea89ff72d52c/assert/assertions.go#L2032).

assert/assertions.go
ubunatic: On the other hand, the `t` must be failed here. Panics should not lead to this code path, since we do not catch them.
Only `runtime.Goexit()`, i.e., `testing.T` failure will lead here.

- `Fail(t, ...)` should be safe here and is formally closer to the original logic.
- `FailNow(t, ...)` would just be a more defensive variant.

assert/assertions.go
```diff
@@ -2010,15 +2010,30 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 		h.Helper()
 	}
 
-	ch := make(chan bool, 1)
-	checkCond := func() { ch <- condition() }
+	const failed = 0
+	const stop = 1
+	const noStop = 2
```
brackendawson:
```suggestion
	const (
		failed = iota
		stop
		noStop
	)
```

Though better names might be "conditionExited", "conditionPassed", and "conditionFailed".

assert/assertions.go
```diff
@@ -2029,13 +2044,24 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
...
-			tickC = nil
-			go checkCond()
-		case v := <-ch:
-			if v {
+			tickC = nil    // Do not check again until we get a result.
+			go checkCond() // Schedule the next check.
+		case v := <-resultCh:
+			switch v {
+			case failed:
+				// Condition panicked or test failed and finished.
```
brackendawson: We don't get here is the condition function panicked.

assert/assertions.go
brackendawson: Don't FailNow from the assertion variant.

assert/assertions.go
brackendawson: We should probably not change this to capture panics in the condition function, getting very close to a breaking change in the behavior.

assert/assertions.go
brackendawson: We should exit with a different error message which makes it clear that the assertion failed early.

Imagine asking for the condition to be true within 10 seconds and being told it was never true after only 0 seconds. More confusing then the case we are attempting to fix I think.

brackendawson (changes requested):
This must be tested, you could do so like this:
```go
func TestEventuallyFailsFast(t *testing.T) {
	mockT := &mockTestingT{}
	wait := make(chan struct{})
	go func() {

		Eventually(mockT, func() bool {
			// Emulate a call to testing.T.FailNow
			runtime.Goexit()
			return true
		}, time.Hour, time.Millisecond)
		close(wait)
	}()
	select {
	case <-wait:
	case <-time.After(time.Second):
		t.Fatal("Eventually did not fail fast")
	}
	// assert that eventually adds an error to mockT
	True(t, mockT.Failed())
}
```

The documentation of Eventually must be changed to explain the new failure behaviour when runtime.Goexit is called from the condition function.

The same change should be made in the other eventually/never functions where it makes sense.

The above being a review of the code, I'm hesitent about making a change like this. This relies on unsupported behavior in the standard library, namely the behavior of calling [t.FailNow](https://pkg.go.dev/testing#B.FailNow) from a goroutine other than the one running the test. This behavior isn't guaranteed to remain the same, though with our particular case being synchronous I have no good reason to doubt it would change.

This is also inconsistent when compared with the T versions of this assertion, where a failing require function fails the condition and not the entire assertion. However that is already what happens, just more slowly.

assert/assertions.go
ubunatic: I agree, this is what I also do on my extended examples.
There are multiple exit scenarios:
1. timeout (as before) -> `Condition never satisfied`
2. runtime.Goexit() called manually -> `Condition exited` (or maybe more context)
3. runtime.Goexit() called via t.FailNow() -> `Condition failed` 
4. panic (not catched in this PRs code) -> `Condition panicked`

The code above does not distinguish between 2 and 3 yet.

assert/assertions.go
ubunatic: I agree.

ubunatic: Would it be a breaking change if we forward the panic?

1. DO catch panic
2. but not handle it as regular Condition error 
3. stop the loop (to not hang the test)
4. panic

This would be closest to exiting behavior but without the blocking through the  dead channel.

I could try to add a test for checking that panics stay unhandled and take the correct code path, either in the current **blocking** variant or the **forwarding** variant.

ubunatic:
@brackendawson I tried to address the issue overall, incl. `WithT`  and `Never`.
- The `mockTestingT` and also `CollectT` work well for tests. Thanks for the hint.
- I could test most of the behavior, I think
- The new error message would be `"Condition exited unexpectedly"`. Happy to change that as needed.
- Also added docs and code comments.

Please have a look again at the proposed changes. Thank you!

ubunatic:
I pushed a few smaller improvements on the tests docs and updated the PR and GH Issue description. I think The PR now covers `Eventually`, `Never`, and `EventuallyWithT` comprehensively and holistically.

ubunatic:
While writing more tests I found my initial `CollectT` result check not to be consistent.
I extended `CollectT` with a `finished` state (like in `testing.T`). This way we can clearly distinguish if `FailNow` was called on `collect` or on the parent `t` and safely fail fast.

I updated the docs and tests for this, please have a look. @brackendawson 

Note that I changed `MockT`'s `Failed` to `Failed()` and reused an embedded `CollectT` there, since it provides all what is needed to collect errors for tests. I don't think `MockT` is used outside of this repo (GH search did not show anything, also did a sourcegraph [code search](https://sourcegraph.com/search?q=context:global+%22MockT.Failed%22+OR+%22require.MockT%22&patternType=keyword&case=yes&sm=0)). Impact should be low, except for quite a few updated tests. If needed, I can make this a separate PR.

assert/http_assertions.go
```diff
@@ -127,7 +127,8 @@ func HTTPBody(handler http.HandlerFunc, method, url string, values url.Values) s
 // HTTPBodyContains asserts that a specified handler returns a
 // body that contains a string.
 //
-//	assert.HTTPBodyContains(t, myHandler, "GET", "www.google.com", nil, "I'm Feeling Lucky")
+//	expectVal := "I'm Feeling Lucky"
```
ubunatic, line 130: made this a var to better reflect real world usage and help the codegen to not detect the last string param as error message

brackendawson: The http package has long been deprecated.

.gitignore
```diff
@@ -23,6 +23,9 @@ _testmain.go
 
 .DS_Store
 
+# AI helpers
+*-discussion.md
+
```
brackendawson, line 28: I expect this to become tech debt almost immediately as AI helpers improve their context management.

I'll accept AI generated code into testify so long as I can't tell it's AI. It should be as if a human made it.

I will not accept these files into testify.

README.md
```diff
@@ -2,7 +2,7 @@ Testify - Thou Shalt Write Tests
 ================================
 
 > [!NOTE]
-> Testify is being maintained at v1, no breaking changes will be accepted in this repo.  
+> Testify is being maintained at v1, no breaking changes will be accepted in this repo.
```
brackendawson, line 5: Remove this unrelated change from the PR.

Two trailing spaces in Markdown is equivalent to a line break, they may be necessary.

README.md
```diff
@@ -31,6 +31,17 @@ The `assert` package provides some helpful methods that allow you to write bette
...
+
+> [!Note]
+> See [EVENTUALLY.md](EVENTUALLY.md) for details about timing, exit behavior, and panic handling,
+> and read the source code and source code comments carefully.
+>
+> The `Eventually` functions behavior got some recent bug fixes and behavior changes for longstanding issues.
+>
+> 👉️ Please **read** the [document](EVENTUALLY.md) and the **source code comments**! \
+> 👉️ Please **adapt** your code if you were relying on the buggy behavior!
+
```
brackendawson, line 44: This is too much detail for the main README, and it is too sloppy. Emoji are not bullet points even if your LLM thinks they are.

Remove this change.

assert/http_assertions.go
brackendawson, line 130: Why are we changing this file? Do not change this file.

mock/mock_test.go
```diff
@@ -1342,7 +1342,11 @@ func Test_Mock_Called(t *testing.T) {
 }
 
 func asyncCall(m *Mock, ch chan Arguments) {
-	ch <- m.Called(1, 2, 3)
+	var args Arguments
```
brackendawson, line 1345: Why are we changing mock? This is unrelated to mock. Do not change this file.

require/requirements_testing_test.go
```diff
@@ -0,0 +1,63 @@
+//go:build !novet
+// +build !novet
```
brackendawson, line 2: // why no vet?

require/requirements_testing_test.go
```diff
@@ -0,0 +1,63 @@
...
+	done := make(chan struct{})
+
+	// Run Eventually in a separate goroutine to capture the Goexit behavior.
+	// This avoid test panics from the unssupported Goexit call in the main test goroutine.
+	// Note that this will trigger linter warnings about goroutines in tests. (SA2002)
+	go func(tt *testing.T) {
+		defer close(done)
+		defer func() {
+			r := recover() // [runtime.Goexit] does not trigger a panic
+			// If we see a panic here, the condition function misbehaved
```
brackendawson, line 47: Why are we dealing with panics now? As I said, will not try to handle panics in this PR.

.ci.govet.sh
```diff
@@ -2,4 +2,14 @@
...
+# trigger vet warnings, use the 'novet' build tag. This is used in the
+# following tests:
+#
+# require/requirements_testing_test.go:
+#
+#    The 'testing' tests test testify behavior 😜 against a real testing.T,
+#    running tests in goroutines to capture Goexit behavior.
+#    Such usage triggers would normally trigger vet warnings (SA2002).
+
+go vet -tags novet ./...
```
brackendawson, line 15: Why is this necessary, you've not documented the precise behavior which triggers SA2002 either here or in the file which causes it.

I'd prefer to preserve the fully vetted status of testify. Ignoring entire files is certainly too coarse.

ubunatic:
Since this got out of hand, I will restart from [73ecd57](https://github.com/stretchr/testify/pull/1809/commits/73ecd5785cd8ba7100b9a1b11cd018ea58b63e18) as proposed and make new PR, as small as possible, touching as few files as possible. Thanks for taking the time to review.

----
#### Edit
I will still address all the comments to not leave them open.

If there is value in any of the extensive docs in this PR, we may cherry pick them later; without emojis, as I now understand. The obscurity of `Eventually` and frequent discussions with other Gophers brought me here. I though adding more elaborate docs and better examples would help other Gophers using the lib.

> IMHO: With better tooling across all OSes, code + emojis is easy to do and provides better UX; if not used too wildly.

Other than that, changing scripts and unrelated files was really a mistake. Again, sorry for that, and for wasting your time.

#### Next
I now opened a new [PR](https://github.com/stretchr/testify/pull/1819). There, I still fix the `runtime.Goexit` issue holistically for `Eventually`, `Never`, and `EventuallyWithT` to provide consistent behavior. To me this is not out of scope. For `EventuallyWithT` this required a small internal change to `CollectT`.

README.md
ubunatic: This was hand-written and I use more emojis in docs these days because they are easy to type (not just on MacOS) and add value, IMHO. Anyway, removed it so I can leave this PR in a clean way according to your standards and not mine. Sorry.

.ci.govet.sh
ubunatic: The idea was to add tests on real `testing.T`. Calling FailNow outside the main goroutine caused vet errors.

Out of scope for the original issue, and now removed from the branch.

.gitignore
ubunatic: removed

require/requirements_testing_test.go
ubunatic: To allow calling FailNow on testing.T.
Out of scope + removed now.

assert/http_assertions.go
ubunatic: Understood. The change was made to support the improved codegen. I have now reset everything on the branch.

require/requirements_testing_test.go
ubunatic: This was supposed to catch the test panic: `panic("unreachable")`, but it would actually now work. The code was useless.

I removed the whole file now.

PR #1814, opened 2025-10-18
mock: add OnF method

### LINKED ISSUE #425: Use function pointers to set up mock expectations
testify provides the On() function to set up mock expectations, which uses a string to identify the mocked function. However, Go provides an API [runtime.FuncForPC()](https://golang.org/pkg/runtime/#FuncForPC) which allows users to obtain a function pointer's name.

```go
type Sap struct {
}

func (i *Sap) Magic() {
	fmt.Println("Magic")
}

func dump(a interface{}) {
	name := runtime.FuncForPC(reflect.ValueOf(a).Pointer()).Name()
	fmt.Println(name)
}

func main() {
	dump((*Sap).Magic) // "main.(*Sap).Magic"
	s := &Sap{}
	dump(s.Magic) // "main.(*Sap).Magic-fm"
	sa := Sap{}
	dump(sa.Magic) // "main.(*Sap).Magic-fm"
}
```

https://play.golang.org/p/RAB0zOkB0L

What are your thoughts on providing a new API similar to On() which accepts function pointers in the form of interface{} instead of string? I think this could really reduce pains in refactoring code - such as method renaming - because tools can now pick up these tokens.

burik666:
## Summary
Adds method `OnFn` to `mock.Mock` and `mock.Call` to allow defining mock expectations using function references instead of string method names.

## Changes
* Added `OnFn` method to [`mock.Mock`](https://pkg.go.dev/github.com/stretchr/testify#Mock) and [`mock.Call`](https://pkg.go.dev/github.com/stretchr/testify#Call).
* OnFn extracts the function name using reflection and runtime information ([`runtime.FuncForPC`](https://pkg.go.dev/runtime#FuncForPC)).
* Provides a more type-safe, refactor-friendly alternative to `On`.

## Motivation
Currently, mock expectations are defined using string literals for method names, which can be error-prone and break easily during refactoring.
OnFn allows developers to use actual function references, improving type safety and maintainability.

## Related issues
* Close #425

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -470,7 +470,7 @@ func callString(method string, arguments Arguments, includeArgumentValues bool) 
 
 // Called tells the mock object that a method has been called, and gets an array
 // of arguments to return.  Panics if the call is unexpected (i.e. not preceded by
-// appropriate .On .Return() calls)
+// appropriate [Mock.On] calls)
 // If Call.WaitFor is set, blocks until the channel is closed or receives a message.
 func (m *Mock) Called(arguments ...interface{}) Arguments {
 	// get the calling function's name
@@ -493,7 +493,7 @@ func (m *Mock) Called(arguments ...interface{}) Arguments {
 
 // MethodCalled tells the mock object that the given method has been called, and gets
 // an array of arguments to return. Panics if the call is unexpected (i.e. not preceded
-// by appropriate .On .Return() calls)
+// by appropriate [Mock.On] calls)
 // If Call.WaitFor is set, blocks until the channel is closed or receives a message.
 func (m *Mock) MethodCalled(methodName string, arguments ...interface{}) Arguments {
 	m.mutex.Lock()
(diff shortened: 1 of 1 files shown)


burik666:
bump.
what do you think about this?

dolmen:
I like the implementation, but I don't like the method name.

mock/mock.go
dolmen, line 205: Add link to OnF with a [godoc link](https://go.dev/doc/comment#doc-links) to make the new method more discoverable (especially by agents).

dolmen (changes requested):

mock/mock.go
burik666: Added a short godoc cross-reference to `OnF` from `On` to improve discoverability.

burik666:
@dolmen 
> I don't like the method name.

Renamed to `OnFn`.
Feels consistent with the existing `On` method while being a bit clearer.

burik666:
@dolmen, 
Is this better? What do you think needs to be improved?

Issue #1813, opened 2025-10-17
assert.Equal panic on diff unexported map with array as key


olitez:
## Description
assert.Equal could panic in some cases:
1. values must be same type but not equal
2. type must be struct with unexported field
3. field must be map with array as keys
4. map must have at least 2 values

## Step To Reproduce
```go
type test struct {
	m map[[1]byte]*struct{}
}

a := test{
	map[[1]byte]*struct{}{
		{1}: nil,
		{2}: nil,
	},
}
b := test{}

assert.Equal(t, a, b)
```
https://go.dev/play/p/1tziFuRiCw4

## Expected behavior
show diff

## Actual behavior
```
--- FAIL: TestTest (0.00s)
panic: reflect.Value.Interface: cannot return value obtained from unexported field or method [recovered]
	panic: reflect.Value.Interface: cannot return value obtained from unexported field or method
```

It fails on diff when trying to obtain dump from [davecgh/go-spew](https://github.com/davecgh/go-spew), it's long time no updates nor fixes so probably worth to switch on something like [yassinebenaid/godump](https://github.com/yassinebenaid/godump)

dolmen:
👎 for replacing `go-spew` by another dependency.

Testify has a major place in the Go ecosystem, so we have strong responsibilities regarding our dependency chain. Adding a new dependency would have a significant impact on dependents.

olitez:
Hi @dolmen 

Thanks for your opinion! I agree that adding a dependency is not best scenario (not only for testify but mostly everywhere), but having existed dependency that panic on normal case is not very good for ecosystem imho.
If testify is really relying on go-spew maybe it worth to fork it and continue maintain under testify maintainers? This exact bug is just about single line fix add reflect.CanInterface before trying to reflect.Interface

HaraldNordgren:
@olitez @dolmen Potential fix, although maybe we want to generalize it more? It there are broaded trend of failure cases?

https://github.com/stretchr/testify/pull/1816

olitez:
@HaraldNordgren 

thanks for fix, not showing actual diff but at least no panic
 ```
        	Error:      	Not equal:
        	            	expected: query.test{m:map[[1]uint8]string{[1]uint8{0x1}:"one", [1]uint8{0x2}:"two"}}
        	            	actual  : query.test{m:map[[1]uint8]string(nil)}
        	            	
        	            	Diff:
        	Test:       	TestXxx
```

[...]

PR #1816, opened 2025-10-25
internal/spew: don't panic on unexported struct key in map

HaraldNordgren:
Fixes https://github.com/stretchr/testify/issues/1813.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1972,8 +1972,8 @@ func diff(expected interface{}, actual interface{}) string {
 		e = spewConfigStringerEnabled.Sdump(expected)
 		a = spewConfigStringerEnabled.Sdump(actual)
 	case isStructWithUnexportedMapField(et):
-		e = reflect.ValueOf(expected).String()
-		a = reflect.ValueOf(actual).String()
+		e = fmt.Sprintf("%+v", expected)
+		a = fmt.Sprintf("%+v", actual)
 	default:
 		e = spewConfig.Sdump(expected)
 		a = spewConfig.Sdump(actual)


olitez:
hi all,

not showing actual diff, but no panic so it's looks good for me
```
        	Error:      	Not equal:
        	            	expected: query.test{m:map[[1]uint8]string{[1]uint8{0x1}:"one", [1]uint8{0x2}:"two"}}
        	            	actual  : query.test{m:map[[1]uint8]string(nil)}
        	            	
        	            	Diff:
        	Test:       	TestXxx
```

HaraldNordgren:
Ping @dolmen @brackendawson @ccoVeille

ccoVeille:
I feel like the test you could cover more use cases.

Let me try yo understand and rephrase what this PR is trying to fix:

If we refers to the issue, the code was panicking when the struct had no exported fields, there is a map with at least 2 values.

So here we replace the go-spew diffs, by something else. But as reported this "something" is not a diff, but something else.

the fact go-spew could panic is random. Here you identify an issue but there could be others.

I wonder whether having a defer recover with error management to go-spew couldn't help.

The fact the diff is broken shouldn't break the assertion lib. It should report the test is falling, and here you could try to render things with the solution you found.

I'm not saying your way is wrong, not my solution is a good one. I'm asking for feedbacks.

I'm a bit afraid we are replacing a diff that works in many cases except a few random ones. When one of them was clearly identified, while others may stay hidden in the darkness able to panic.

Also, I'm a bit afraid, the code added could lead to a placeholder reporting that cannot compete with the diffs available in go-spew, when it might not fail in some edge cases we didn't identified yet.

Again, I'm not asking for code changes, but for an open discussion.

I have no problem in being wrong 😁

HaraldNordgren:
It's a valid question for sure and you have a good point that this is just one cause of potentially many others.

But using `recover` as a strategy feels wrong. Maybe it's just my feeling.

ccoVeille:
@brackendawson do you remember the discussion we had about importing the code of go-spew, the same way we imported difflib ?

The conclusion was not yet, because we don't know yet the "unforseen consequences" of importing the code of go-difflib, so we don't have enough information and experience on this.

Maybe it's time to reconsider it.

Importing the code of go-spew diff and and fixing the bug in the code could be considered.

ccoVeille:
Based on go-spew code

This is what's panic is the call to [reflect.Value.Interface](https://pkg.go.dev/reflect#Value.Interface)

https://cs.opensource.google/go/go/+/refs/tags/go1.25.4:src/reflect/value.go;drc=f1e6ae2f6f5424f9f5c6dc915866d4f457c1483e;l=1484

So the fix would be about calling [reflect.Value.CanInterface](https://pkg.go.dev/reflect#Value.CanInterface)

https://cs.opensource.google/go/go/+/refs/tags/go1.25.4:src/reflect/value.go;drc=f1e6ae2f6f5424f9f5c6dc915866d4f457c1483e;l=1457

Especially if we consider the comment

> CanInterface reports whether Value.Interface can be used without panicking. 

So the fix would be about adding calls to reflect.Value.CanInterface in this for loop before calling reflect.Value.Interface

https://github.com/davecgh/go-spew/blob/8991bc29aa16c548c550c7ff78260e27b9ab7c73/spew/common.go#L315

assert/assertions.go
```diff
@@ -1928,6 +1928,23 @@ func typeAndKind(v interface{}) (reflect.Type, reflect.Kind) {
...
+		field := t.Field(i)
+		if !field.IsExported() &&
+			field.Type.Kind() == reflect.Map &&
+			field.Type.Key().Kind() == reflect.Array {
+			return true
+		}
+	}
+
+	return false
+}
```
ccoVeille, line 1945: Based on the code the discussion that happened here


And based on the code of go-spew

https://github.com/stretchr/testify/pull/1816#issuecomment-3493594441

assert/assertions.go
```diff
@@ -1948,13 +1965,16 @@ func diff(expected interface{}, actual interface{}) string {
...
+	case et == reflect.TypeOf(""):
 		e = reflect.ValueOf(expected).String()
 		a = reflect.ValueOf(actual).String()
-	case reflect.TypeOf(time.Time{}):
+	case et == reflect.TypeOf(time.Time{}):
 		e = spewConfigStringerEnabled.Sdump(expected)
 		a = spewConfigStringerEnabled.Sdump(actual)
+	case isStructWithUnexportedMapField(et):
+		e = reflect.ValueOf(expected).String()
+		a = reflect.ValueOf(actual).String()
```
ccoVeille: I would return "" then here.

Maybe directly.

```suggestion
	case isStructWithUnexportedMapField(et):
		return ""
```

The diff being convert as string would call stringer, and so could report invalid things

If there is a need for keeping this diff item, I feel like, this could be considered

```suggestion
	case isStructWithUnexportedMapField(et):
		e = reflect.ValueOf(expected).String()
		a = reflect.ValueOf(actual).String()

		if a == b {
			// only the difference makes senses in term of diff
			return ""
		}
```

ccoVeille: Another solution could be to replicate the code of fmt.Sprintf with %+v

https://github.com/golang/go/blob/8cf7a0b4c956aad5a8b98efde8ea6b7cde173902/src/fmt/print.go#L834-L855

And compare them.

But I'm unsure how much it differs from reflect.Value.String

ccoVeille: There is no need to duplicate the code of sprintf.

We could do this

```suggestion
	case isStructWithUnexportedMapField(et):
		e = fmt.Sprintf("%+v", expanded)
		a = fmt.Sprintf("%+v", actual)
```


Here is the result

```
Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-{m:map[[1]:<nil> [2]:<nil>]}
        	            	+{m:map[]}
```

https://go.dev/play/p/x6c5ZEyilNO

HaraldNordgren: Yeah, that's nice!

assert/assertions.go
ccoVeille: I feel like started writting something, but I posted before putting all together.

So here, I was spotting the code here limited to unexported field that are maps with 2 elements, but go-spew code would panic when dealing with Interface and here, maybe a better fix would be to use CanInterface and check nothing else.

I'm unsure what other code examples could panic, but maybe it could be considered as a better approach than the "recover" I suggested.

ccoVeille:
> It's a valid question for sure and you have a good point that this is just one cause of potentially many others.
> 
> But using `recover` as a strategy feels wrong. Maybe it's just my feeling.

You are right, considering catching the panic with recover and then try to apply the patch is not a good solution.

I provided you some feedback and suggestion to your PR, please take a look at them.

But I feel like  the recover idea should be considered anyway.

The fact the diff can panic is bad.
Your idea to add condition to avoid the panic that was identified is good.

But maybe we could also add a recover to invalidate the diff from error reporting when it panic, instead if breaking everything.

HaraldNordgren:
@ccoVeille Thanks for the ideas, but I'm not exactly sure which of these ideas you want me to implement. All of them or only a few?

I'm not exactly sure how to properly use `CanInterface`, I tried a few things and it panicked 🤗

ccoVeille:
> @ccoVeille Thanks for the ideas, but I'm not exactly sure which of these ideas you want me to implement. All of them or only a few?

I'm sorry about that. I should have been clearer.

Many of the point I raised are open for debate with everyone, including you, and also other maintainers.

We can discuss about them, some of them or all the ideas I raised might be rejected. I just shared my thoughts about the possible resolutions. If we decide to go in that direction new PR will be opened I think.

Right now, I'm good with your changes. That's why I approve your PR.

> I'm not exactly sure how to properly use `CanInterface`, I tried a few things at it panicked 🤗

The CanInterface ideas were for fixing the bug in go-spew code.

HaraldNordgren:
@ccoVeille Got it!

Are there any plans to fix go-spew? I guess it would need to be by forking (since @davecgh seems to have abandoned it), which then in practice introduces a new transitive dependency for all testify users.

HaraldNordgren:
@ccoVeille Let's merge this?

ccoVeille:
I was waiting reaction from other maintainers

assert/assertions.go
ccoVeille: I feel this might not be enough.

I checked at go-spew code while working on https://github.com/stretchr/testify/pull/1824/files

I feel like a code like this won't be caught by your fix

```go
type Bar struct {
	  m map[[1]byte]*struct{}
}

type Foo struct {
  Foo Bar
}
```

When using a Foo, the problem is wider that simply recursively parsing the fields, as you need to also go through the whole stack, check for the pointer, the slices, maps, and arrays

So have something that goes as deep as go-spew.

brackendawson:
This only fixes a narrow set of cases where spew may try to derive a value from an un-exported field, right? I think we should either avoid calling spew in all such cases or we should look at fixing 'spew. Fixing spew might mean moving it into this repository.

Not really a fan of just merging this for now, I believe it's an incomplete fix and is not urgent, this bug has been there for a long time.

Issue #1794, opened 2025-09-15
assert.Regexp panics with invalid rx expression


brackendawson:
## Description
Because assert.Regexp uses [regexp.MustCompile](https://pkg.go.dev/regexp#MustCompile) which is for "initialization of global variables holding compiled regular expressions" a runtime panic is possible. We should use Compile and handle the error.

## Step To Reproduce
```go
package kata_test

import (
	"testing"

	"github.com/stretchr/testify/require"
)

func TestIfy(t *testing.T) {
	require.Regexp(t, "\\C", "I don't like this at all")
}
```

## Expected behavior
The test fails with a useful error.

## Actual behavior
A panic:
```
Running tool: /Users/brackendawson/sdk/go1.24.5/bin/go test -timeout 30s -coverprofile=/var/folders/8c/f9y61rb14cn8c6_6lnz10q300000gn/T/vscode-gohVso1s/go-code-cover github.com/brackendawson/kata -race

--- FAIL: TestIfy (0.00s)
panic: regexp: Compile(`\C`): error parsing regexp: invalid escape sequence: `\C` [recovered]
	panic: regexp: Compile(`\C`): error parsing regexp: invalid escape sequence: `\C`

goroutine 35 [running]:
testing.tRunner.func1.2({0x1025cc440, 0xc0000924f0})
	/Users/brackendawson/sdk/go1.24.5/src/testing/testing.go:1734 +0x2bc
testing.tRunner.func1()
	/Users/brackendawson/sdk/go1.24.5/src/testing/testing.go:1737 +0x47c
panic({0x1025cc440?, 0xc0000924f0?})
	/Users/brackendawson/sdk/go1.24.5/src/runtime/panic.go:792 +0x124
regexp.MustCompile({0xc000090d4e, 0x2})
	/Users/brackendawson/sdk/go1.24.5/src/regexp/regexp.go:313 +0xcc
github.com/stretchr/testify/assert.matchRegexp({0x1025cc440, 0x1026239f0}, {0x1025cc440, 0x102623a00})
	/Users/brackendawson/src/github.com/stretchr/testify/assert/assertions.go:1715 +0x88
github.com/stretchr/testify/assert.Regexp({0x103cb8170, 0xc000082c40}, {0x1025cc440, 0x1026239f0}, {0x1025cc440, 0x102623a00}, {0x0, 0x0, 0x0})
	/Users/brackendawson/src/github.com/stretchr/testify/assert/assertions.go:1737 +0x90
[...]

PR #1818, opened 2025-10-25
fix(assert): avoid panic on invalid regex in Regexp/NotRegexp (#1794)

kdt523:
Summary
Make assert.Regexp / assert.NotRegexp fail the assertion instead of panicking when given an invalid regex pattern.

Changes
Use regexp.Compile (rather than regexp.MustCompile) when compiling user-supplied patterns in assert.
Propagate compile errors so Regexp/NotRegexp report a clear failure message like:
invalid regular expression "<pattern>": <error>
Add assert/regexp_invalid_test.go with tests that ensure invalid patterns do not cause a panic.
Files touched:
assert/assertions.go
assert/regexp_invalid_test.go
Motivation
Passing a malformed regex used to trigger a runtime panic in tests. That made debugging harder and could abort the whole test run. Reporting a failed assertion with a helpful error message is a safer, more useful behavior for test authors.

Related issues
 #1794

### DIFF
--- a/assert/regexp_invalid_test.go
+++ b/assert/regexp_invalid_test.go
@@ -5,15 +5,15 @@ import "testing"
 // Verifies that invalid patterns no longer cause a panic when using Regexp/NotRegexp.
 // Instead, the assertion should fail and return false.
 func TestRegexp_InvalidPattern_NoPanic(t *testing.T) {
-    NotPanics(t, func() {
-        mockT := new(testing.T)
-        False(t, Regexp(mockT, "\\C", "whatever"))
-    })
+	NotPanics(t, func() {
+		mockT := new(testing.T)
+		False(t, Regexp(mockT, "\\C", "whatever"))
+	})
 }
 
 func TestNotRegexp_InvalidPattern_NoPanic(t *testing.T) {
-    NotPanics(t, func() {
-        mockT := new(testing.T)
-        False(t, NotRegexp(mockT, "\\C", "whatever"))
-    })
+	NotPanics(t, func() {
+		mockT := new(testing.T)
+		False(t, NotRegexp(mockT, "\\C", "whatever"))
+	})
 }


assert/assertions.go
```diff
@@ -1753,7 +1763,11 @@ func NotRegexp(t TestingT, rx interface{}, str interface{}, msgAndArgs ...interf
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
-	match := matchRegexp(rx, str)
+	match, err := matchRegexp(rx, str)
+	if err != nil {
+		Fail(t, fmt.Sprintf("invalid regular expression %q: %v", fmt.Sprint(rx), err), msgAndArgs...)
```
ccoVeille, line 1768: This would be better, no?

```suggestion
		Fail(t, fmt.Sprintf("invalid regular expression %q: %v",  rx, err), msgAndArgs...)
```

assert/assertions.go
```diff
@@ -1736,7 +1742,11 @@ func Regexp(t TestingT, rx interface{}, str interface{}, msgAndArgs ...interface
 		h.Helper()
 	}
 
-	match := matchRegexp(rx, str)
+	match, err := matchRegexp(rx, str)
+	if err != nil {
+		Fail(t, fmt.Sprintf("invalid regular expression %q: %v", fmt.Sprint(rx), err), msgAndArgs...)
```
ccoVeille, line 1747: This would be better, no?

```suggestion
		Fail(t, fmt.Sprintf("invalid regular expression %q: %v", rx, err), msgAndArgs...)
```

kdt523: Hi @ccoVeille — thanks for the suggestion, very helpful to question that line.

Quick reason I didn’t pass rx directly to %q: %q expects a string (or string-like) value. In our code rx is an interface{} and callers sometimes pass a compiled *regexp.Regexp. If you do fmt.Sprintf("%q", rx) with a *regexp.Regexp, Go prints an unhelpful %!q(...) type dump instead of the pattern. That makes the error message confusing.
Small examples:
rx := "start" → fmt.Sprintf("%q", rx) => "start" (good)
rx := regexp.MustCompile("start") → fmt.Sprintf("%q", rx) => %!q(*regexp.Regexp=...) (bad)
Safe alternatives

1.Keep the quoted style by converting rx to a string first:
fmt.Sprintf("invalid regular expression %q: %v", fmt.Sprint(rx), err)
Pros: preserves a quoted pattern; simple.

2.Don’t quote, just use %v:
fmt.Sprintf("invalid regular expression %v: %v", rx, err)
Pros: always works for any type; no weird %!q output. Cons: loses the quoted look.

3.(Preferred) Be explicit: if rx is a *regexp.Regexp, call .String(), otherwise fall back to fmt.Sprint:
var rxStr string
switch v := rx.(type) {
case *regexp.Regexp:
    rxStr = v.String()
default:
    rxStr = fmt.Sprint(rx)
}
Fail(t, fmt.Sprintf("invalid regular expression %q: %v", rxStr, err), msgAndArgs...)

Pros: most robust and readable; shows the actual pattern for compiled regexps and handles any other types cleanly.
If you’re OK with that, I can apply option 3 (tiny, explicit change) to the fix/assert-regexp-compile-error branch and push it. If you prefer a simpler tweak, I can apply option 1 instead — tell me which and I’ll open a small commit/PR.

ccoVeille: I thought %q would behave the same way %s with quote.

I checked https://cs.opensource.google/go/go/+/master:src/fmt/format.go

I'm a bit surprised it would differ.

Your reply sounds like it would, I will have to check further.

But for now, let's say the code can stay like you did.

ccoVeille: I'm still doubtful

https://go.dev/play/p/WBwS5ClJU0i

kdt523: You’re right that for *regexp.Regexp, %q and %s both output the pattern, with %q just adding quotes. In most cases, this isn’t harmful.

My main reason for switching to %v (or [fmt.Sprint]) was to avoid unnecessary quoting and ensure consistent, clear error messages—especially if the input isn’t a string or is a different type. This way, the formatting is robust for all cases, not just regex patterns.

If you find any edge cases where %q would be preferable, I’m happy to revisit! For now, I think the current approach is a bit safer and more general. Thanks again for your thoughtful feedback!

dolmen:
> Passing a malformed regex used to trigger a runtime panic in tests. That made debugging harder and could abort the whole test run.

This is on purpose. There is a bug in the testsuite code, not in the tested data. This is a good reason for not reporting the issue in the same way.

PR #1821, opened 2025-11-08
docs(require): correct example usage to use assert.CollectT (require.CollectT does not exist)

### LINKED ISSUE #1843: `require.CollectT` in documentation for EventuallyWithT{,t} does not exist
There is no `require.CollectT` but in code comment it is mentioned several times as below.

https://github.com/stretchr/testify/blob/65697cefca1d61fba40e001c90d2166be5c8df7f/require/require.go#L432

https://github.com/stretchr/testify/blob/65697cefca1d61fba40e001c90d2166be5c8df7f/require/require.go#L460

a2not:
## Summary

Update documentation comments in `require.go` to use `assert.CollectT` instead of non-existing `require.CollectT` in `EventuallyWithT` and `EventuallyWithTf` examples. This clarifies the intended usage and prevents confusion for users.

## Changes

- Created wrapper function for strings.ReplaceAll which preserve certain identifiers. So far we only need to preserve `assert.CollectT`.

## Motivation

The documentation was confusing since it uses non-existent type.

## Related issues

Closes #1843

### DIFF
--- a/_codegen/main.go
+++ b/_codegen/main.go
@@ -106,9 +106,7 @@ func parseTemplates() (*template.Template, *template.Template, error) {
 		}
 		funcTemplate = string(f)
 	}
-	tmpl, err := template.New("function").Funcs(template.FuncMap{
-		"replace": strings.ReplaceAll,
-	}).Parse(funcTemplate)
+	tmpl, err := template.New("function").Parse(funcTemplate)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -331,6 +329,8 @@ func requireComment(comment string) string {
 
 func (f *testFunc) CommentRequire() string {
 	comment := strings.ReplaceAll(f.DocInfo.Doc, "assert.", "require.")
+	// Preserve assert.CollectT, even in package 'require'
+	comment = strings.ReplaceAll(comment, "require.CollectT", "assert.CollectT")
 	return requireComment(comment)
 }
 
--- a/require/require.go
+++ b/require/require.go
@@ -429,7 +429,7 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 //		time.Sleep(8*time.Second)
 //		externalValue = true
 //	}()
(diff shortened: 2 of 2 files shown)


a2not:
Oops. I didn't know that it was in generated file.
I'll think of another approach then.

_codegen/main.go
```diff
@@ -298,6 +296,31 @@ func (f *testFunc) CommentWithoutT(receiver string) string {
 	return strings.Replace(f.Comment(), search, replace, -1)
 }
 
+func (f *testFunc) Replace(comment, search, replace string) string {
+	// replace strings, while preserving some identifiers
+	identifiersToBePreserved := []string{
+		"*assert.CollectT",
+	}
```
a2not: This might be over-generalizing things, since so far we only want to preserve one keyword.
Let me know if I need to simplify this so that it only cares about `assert.CollectT` like https://github.com/stretchr/testify/pull/1821/commits/a39aa2d6d38af3f5cfd7c4150bd9df364029d9cf#diff-4b4cb0af6b5935c8adcb9f00b426367349e989296edf13eee95831c4bb93c831R299-R310.

dolmen: This is in fact over-engineered.

What about this:
```go
comment = strings.ReplaceAll(comment, search, replace)
// Preserve assert.CollectT, even in package 'require'
comment = strings.ReplaceAll(comment, "require.CollectT", "assert.CollectT")
```

a2not: What a neat solution! I'm sold.
9b474d6a94c18c7a58ad8fd74f06b7cdbacfe67d

dolmen:
@a2not Notice that I have just merged #1850 which also changes code generation!

a2not:
@dolmen No worries I got you! Resolved conflicts with #1850 .

dolmen:
@a2not Thanks!

a2not:
@dolmen Thank you for your review 👍

PR #1822, opened 2025-11-15
Ensure deterministic order when diffing maps

robbiemcmichael:
## Summary

Ensure deterministic order when diffing maps to make the diffs readable.

## Changes

Enable the [`SpewKeys` setting in `spew.ConfigState`](https://pkg.go.dev/github.com/davecgh/go-spew/spew#ConfigState).

## Motivation

When generating the diff for two maps where the map key is not a basic type, the order is non-deterministic which makes the diffs unreadable.

By enabling the `SpewKeys` setting, the map keys will be spewed to strings for the purpose of sorting them. This results in a deterministic order and makes the diffs readable.

For the test case added in this PR, the old config that has `SpewKeys` disabled would look like this:

```diff
--- Expected
+++ Actual
@@ -1,5 +1,2 @@
 (map[assert.Key]int) (len=4) {
- (assert.Key) {
-  x: (int) 4
- }: (int) 4,
  (assert.Key) {
@@ -12,3 +9,6 @@
   x: (int) 3
- }: (int) 3
+ }: (int) 3,
+ (assert.Key) {
+  x: (int) 4
+ }: (int) 999
 }
```

After enabling `SpewKeys`, we get the simpler diff that we would expect:

```diff
--- Expected
+++ Actual
@@ -12,3 +12,3 @@
   x: (int) 4
- }: (int) 4
+ }: (int) 999
 }
```

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -2989,6 +2989,28 @@ Diff:
 Diff:
 --- Expected
 +++ Actual
+@@ -12,3 +12,3 @@
+   x: (int) 4
+- }: (int) 4
++ }: (int) 999
+ }
+`
+
+	type Key struct {
+		x int
+	}
+
+	actual = diff(
+		map[Key]int{Key{1}: 1, Key{2}: 2, Key{3}: 3, Key{4}: 4},
+		map[Key]int{Key{1}: 1, Key{2}: 2, Key{3}: 3, Key{4}: 999},
+	)
+	Equal(t, expected, actual)
+
+	expected = `
+
+Diff:
+--- Expected
++++ Actual
 @@ -1,3 +1,3 @@
  (*errors.errorString)({
 - s: (string) (len=19) "some expected error"
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1985,6 +1985,7 @@ var spewConfig = spew.ConfigState{
 	DisablePointerAddresses: true,
 	DisableCapacities:       true,
 	SortKeys:                true,
+	SpewKeys:                true,
 	DisableMethods:          true,
 	MaxDepth:                10,
 }
@@ -1994,6 +1995,7 @@ var spewConfigStringerEnabled = spew.ConfigState{
 	DisablePointerAddresses: true,
 	DisableCapacities:       true,
 	SortKeys:                true,
+	SpewKeys:                true,
 	MaxDepth:                10,
 }
 


assert/assertions_test.go
```diff
@@ -2986,6 +2986,28 @@ Diff:
...
+
+	type Key struct {
+		x int
+	}
+
+	actual = diff(
+		map[Key]int{Key{1}: 1, Key{2}: 2, Key{3}: 3, Key{4}: 4},
+		map[Key]int{Key{1}: 1, Key{2}: 2, Key{3}: 3, Key{4}: 999},
+	)
+	Equal(t, expected, actual)
```
ccoVeille, line 3007: I would like you to change a few things

- The value should differ from the key in the map. Otherwise it's unclear how it's ordered.

- your example should also display wgat happen when a value is added and removed

- there is no need for a local exported type

Please consider something like this

```suggestion
	type key struct {
		x int
	}

	actual = diff(
		map[key]int{key{1}: 789, key{2}: 678, key{3}: 567, key{4}: 456},
		map[key]int{key{2}: 678, key{3}: -1, key{4}: 456, key{5}: 345},
	)
	Equal(t, expected, actual)
```

robbiemcmichael: @ccoVeille: I think that's a different kind of test case that is trying to assert a *particular* order for the map, while I would argue that we only need a *deterministic* order. The fact that the current non-deterministic order produces unreadable diffs should be considered a bug (otherwise `SortKeys` wouldn't be enabled in the first place).

So while I could remove a key, add a key and replace a value in this test case, the problem is that it makes the diff much more confusing:

```diff
--- Expected
+++ Actual
@@ -1,5 +1,2 @@
 (map[assert.key]int) (len=4) {
- (assert.key) {
-  x: (int) 1
- }: (int) 789,
  (assert.key) {
@@ -9,6 +6,9 @@
   x: (int) 3
- }: (int) 567,
+ }: (int) -1,
  (assert.key) {
   x: (int) 4
- }: (int) 456
+ }: (int) 456,
+ (assert.key) {
+  x: (int) 5
+ }: (int) 345
 }
```

It's no longer obvious what we're testing and the diff is about as unreadable as the original broken diff. If someone were to break this test case in the future, there's a good chance they might just decide to replace the diff with whatever it spat out or delete the test case entirely.

I think my test case is about as simple as it can be while asserting that the map order is deterministic. It would be obvious that something has gone wrong if someone were to break this in the future because the diff would become unreadable again.

If we do want to assert a particular order for maps then I can add a separate test case for that, but I'm not sure if it's very valuable. The actual order will be not be particularly intuitive because it's just spew emitting a string representation of the struct and sorting that, so you would see orders like `Key{1}, Key{11}, Key{2}`.

ccoVeille: Indeed

ccoVeille:
@robbiemcmichael I feel like what you report is a bug in go-spew. Did you find any issue related to this ordering in map issue?

I worked on #1827 to vendor go-spew, maybe the way would be to fix gospew behavior.

I'm asking because I'm now unsure about the consequence of enabling this setting in term of rendering/ordering.

I analyzed a lot the code of gospew lately, and I'm a bit worried enabling a new config flag could lead us to face yet-to-disvover gospew bug/panic hidden in the darkness

robbiemcmichael:
@ccoVeille: Looking at the docs for [`SortKeys`](https://pkg.go.dev/github.com/davecgh/go-spew/spew#ConfigState) setting:

> 	// SortKeys specifies map keys should be sorted before being printed. Use
	// this to have a more deterministic, diffable output.  Note that only
	// native types (bool, int, uint, floats, uintptr and string) and types
	// that support the error or Stringer interfaces (if methods are
	// enabled) are supported, with other types sorted according to the
	// reflect.Value.String() output which guarantees display stability.

It does seem as though it should be using [`reflect.Value.String()`](https://pkg.go.dev/reflect#Value.String)      to try to turn the value into a string for sorting, but that seems to be a misunderstanding of what that function does. It only prints a string containing `<T Value>` where `T` is the type, so it doesn't vary at all for values of the same type.

I think this could be considered a bug in go-spew, though the fix would need us to turn the value into a string which will probably end up being similar to using the setting I've enabled in this PR. I think the current approach should be safe as we must already be spewing the value to a string for the sake of printing it, so if it panics when trying to turn it into a string for sorting then it should already panic when it turns it into a string for printing.

robbiemcmichael:
@ccoVeille: Any preference on how to proceed with this? I can't see many other sensible ways to turn the key into a string that can be sorted other than by spewing it to a string, which is exactly what the `SpewKeys` setting is meant to do. These keys already get spewed to strings when they get printed for the diff, so I don't think this is likely to introduce any new panics.

Here's the code that shows what the `SpewKeys` setting does:

https://github.com/stretchr/testify/blob/65697cefca1d61fba40e001c90d2166be5c8df7f/internal/spew/common.go#L244-L249

ccoVeille:
Sorry for the delay, Christmas period is uneasy.

Thanks for pointing out the code.

I agree with you it sounds OK. But I see one issue we may have to address.

reflect.Value.Interface() can panic in some condition. I'm unsure if it makes sense to have a map with a struct with unexported method used as a key. But something as awkward as this could fail.

Maybe we could find go-spew issue about this strange use case. But I didn't find some.

Maybe we could take a look at [utter](https://github.com/kortschak/utter) the go-spew fork that rewrote almost everything. Especially [this](https://github.com/kortschak/utter/blob/b8cb875e83b36b3190dbe140385ae1510a20444e/dump.go#L506) 

But anyway, _the fix is easy I think_ ™️ 😅: adding a check on CanInterface() before and fallback on reflect.Value.String() (that never panic, but we know to be imperfect) or on fmt.SPrint of the value could be something to do.

Maybe I'm simply wrong and there is no problem to activate SpewKeys

Do you think you could add some tests and then update the "internalized" go-spew lib ?

Thanks

I'm confident the path to achieve is OK.

PR #1825, opened 2025-11-15
Fix panic when using EqualValues with uncomparable types

### LINKED ISSUE #1699: EqualValues is not panic-safe
### Prerequisite

```go
	// ConvertibleTo reports whether a value of the type is convertible to type u.
	// Even if ConvertibleTo returns true, the conversion may still panic.
	// For example, a slice of type []T is convertible to *[N]T,
	// but the conversion will panic if its length is less than N.
	ConvertibleTo(u Type) bool
```

### Example

```go
package main

import (
	"testing"

)

func TestEqualValues(t *testing.T) {
	a := []int{1, 2}
	b := (*[3]int)(nil)
	assert.EqualValues(t, a, b)
}
```

### Result

```
=== RUN   TestEqualValues
--- FAIL: TestEqualValues (0.00s)
panic: reflect: cannot convert slice with length 2 to pointer to array with length 3 [recovered]
	panic: reflect: cannot convert slice with length 2 to pointer to array with length 3
```

P.S. Inspired by https://github.com/Antonboom/testifylint/issues/88

ccoVeille:
## Summary

Fix panic when using EqualValues with uncomparable types

## Changes

Add calls to CanConvert and CanInterface to avoid panics

## Motivation

Avoid panic

## Related issues
Closes #1699

- #1699

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -177,10 +177,25 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
 		return false
 	}
 
+	// Attempt conversion of expected to actual type
+	// This handles more cases than just the ConvertibleTo check above
+	if !expectedValue.CanConvert(actualType) {
+		// Types are not convertible, so they cannot be equal
+		// This prevents panics when calling [reflect.Value.Convert]
+		return false
+	}
+
+	expectedConverted := expectedValue.Convert(actualType)
+	if !expectedConverted.CanInterface() {
+		// Cannot interface after conversion, so cannot be equal
+		// This prevents panics when calling [reflect.Value.Interface]
+		return false
+	}
+
 	if !isNumericType(expectedType) || !isNumericType(actualType) {
 		// Attempt comparison after type conversion
 		return reflect.DeepEqual(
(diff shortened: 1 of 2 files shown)


assert/assertions.go
```diff
@@ -177,21 +177,49 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
 		return false
 	}
 
+	// Attempt conversion of expected to actual type
+	// This handles more cases than just the ConvertibleTo check above
+	if !expectedValue.CanConvert(actualType) {
+		// Types are not convertible, so they cannot be equal
+		// This prevents panics when calling [reflect.Value.Convert]
+		return false
+	}
```
dolmen, line 186: Convert should probably happen _after_ check of numeric types.

```console
$ goeval 'fmt.Println(reflect.ValueOf(1.0e50).CanConvert(reflect.TypeFor[byte]()))'
true
$ goeval 'fmt.Println(reflect.ValueOf(1.0e50).Convert(reflect.TypeFor[byte]()))' 
255
```

`CanConvert` doesn't protect against overflow.

dolmen (changes requested):

PR #1827, opened 2025-11-18
chore: vendor go-spew from https://github.com/davecgh/go-spew

### LINKED ISSUE #1826: go-spew vendoring
go-spew can panic when trying to print certain types of values, there are open issues about this on their GitHub repository.

go-spew is unfortunately unmaintained, we cannot expect a fix any time soon.

vendoring go-spew, the same way we vendored go-difflib would help us to remove bugs, but also keep the current pretty print lib without moving to something else.

Note: we have already vendored the go-difflib we are using to compare the go-spew results.

Testify is widely used in Go community. We cannot guess the unpredictable effects it could lead to change the pretty print lib.

See #1061

## Related issues

- #480

- #1824
- #1816
- #1813
- maybe others

It would also help to address old issues because we were depending on go-spew, without a chance so see changes being merged upstream

- time.Time rendering

- #1079
- #1078
- #895

- map

- #288

ccoVeille:
## Summary

vendor [go-spew](https://github.com/davecgh/go-spew) 

## Changes

Import unmaintained library

## Motivation

go-spew is unfortunately unmaintained, we cannot expect a fix any time soon.

vendoring go-spew, the same way we vendored go-difflib would help us to remove bugs, but also keep the current pretty print lib without moving to something else.

Note: we have already vendored the go-difflib we are using to compare the go-spew results.

## Related issues

Closes #1826

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -17,11 +17,10 @@ import (
 	"unicode"
 	"unicode/utf8"
 
-	"github.com/davecgh/go-spew/spew"
-	"github.com/stretchr/testify/internal/difflib"
-
 	// Wrapper around gopkg.in/yaml.v3
 	"github.com/stretchr/testify/assert/yaml"
+	"github.com/stretchr/testify/internal/difflib"
+	"github.com/stretchr/testify/internal/spew"
 )
 
 //go:generate sh -c "cd ../_codegen && go build && cd - && ../_codegen/_codegen -output-package=assert -template=assertion_format.go.tmpl"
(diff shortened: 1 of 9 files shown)


brackendawson (approved):
The license is handled properly and is compatible with what we're doing. I have checked and the import is a faithful verbatim copy. In this case we didn't keep the history as we did with difflib, but we decided that's not mandatory in the vendoring of ernesto-jimenez/gogen. The modifications to fit it into the internal directory are also fine.

Big!

PR #1842, opened 2026-01-08
docs: Fix EventuallyWithTf documentation with proper placement of formatting arguments

### LINKED ISSUE #1833: EventuallyWithTf function help is incorrect, misleading callback signature
Documentation shows that the callback function format is `func(c *assert.CollectT, "error message %s", "formatted")` which is incorrect. 

Formatting arguments are only valid for `EventuallyWithTf` itself, while the callback is just `func(c *assert.CollectT)`.

The issue appears for:
- `assert.EventuallyWithTf()`: https://github.com/stretchr/testify/blob/429ee0b8f4efde767b69e2b6421ec6dfbd75fd22/assert/assertion_format.go#L193
- `require.EventuallyWithTf()`: https://github.com/stretchr/testify/blob/429ee0b8f4efde767b69e2b6421ec6dfbd75fd22/require/require.go#L460
- As a method on`*Assertion` receiver in `assert` package: https://github.com/stretchr/testify/blob/429ee0b8f4efde767b69e2b6421ec6dfbd75fd22/assert/assertion_forward.go#L375
- As a method on`*Assertion` receiver in `require` package: https://github.com/stretchr/testify/blob/429ee0b8f4efde767b69e2b6421ec6dfbd75fd22/require/require_forward.go#L376

a2not:
## Summary

Fixed the regexp for `CommentFormat()` code-generation function such that it properly captures the outer most closing parenthesis of a function call in code comment.
https://github.com/stretchr/testify/blob/65697cefca1d61fba40e001c90d2166be5c8df7f/_codegen/main.go#L291

## Changes

- Changed regexp. It got much more of complexity but regexp is hard so what can I say...
- Minor refactoring such as removing a formating of a string which in reality does nothing.
- The original `EventuallyWithT` has `msgAndArgs` as the last argument, which needs to be replaced by two separate `msg` and `args` in `EventuallyWithTf`. I've implemented with very naive approach which might need to be improved.

## Motivation

Misleading documentation

## Related issues

Closes #1833

### DIFF
--- a/_codegen/main.go
+++ b/_codegen/main.go
@@ -283,10 +283,15 @@ func (f *testFunc) Comment() string {
 }
 
 func (f *testFunc) CommentFormat() string {
-	search := fmt.Sprintf("%s", f.DocInfo.Name)
+	search := f.DocInfo.Name
 	replace := fmt.Sprintf("%sf", f.DocInfo.Name)
-	comment := strings.Replace(f.Comment(), search, replace, -1)
-	exp := regexp.MustCompile(replace + `\(((\(\)|[^\n])+)\)`)
+	comment := strings.ReplaceAll(f.Comment(), search, replace)
+
+	// NOTE: Some functions have msgAndArgs at the end. It needs to be omitted. (currently only EventuallyWithT)
+	// Change here if the original comment changed.
+	comment = strings.Replace(comment, `, "external state has not changed to 'true'; still false"`, "", 1)
+
+	exp := regexp.MustCompile(replace + `\((([^()]*|\([^()]*\))*)\)`)
 	return exp.ReplaceAllString(comment, replace+`($1, "error message %s", "formatted")`)
 }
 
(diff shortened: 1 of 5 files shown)


_codegen/main.go
```diff
@@ -285,10 +285,15 @@ func (f *testFunc) Comment() string {
 }
 
 func (f *testFunc) CommentFormat() string {
-	search := fmt.Sprintf("%s", f.DocInfo.Name)
+	search := f.DocInfo.Name
 	replace := fmt.Sprintf("%sf", f.DocInfo.Name)
-	comment := strings.Replace(f.Comment(), search, replace, -1)
-	exp := regexp.MustCompile(replace + `\(((\(\)|[^\n])+)\)`)
+	comment := strings.ReplaceAll(f.Comment(), search, replace)
+	if strings.Contains(search, "EventuallyWithT") {
```
dolmen: What about looking for the `..."external state...` string instead?

a2not: This suggestion simplified the implementation. Thanks!

PR #1845, opened 2026-01-26
Fix regression introduced in 1.11.x

cicoyle:
## Summary
Revert the check early behavior introduced in [this PR](https://github.com/stretchr/testify/pull/1427/changes). Add tests to ensure there are no panics due to nil derefs. Restores v1.10 behavior for [Never()](https://github.com/stretchr/testify/blob/v1.10.0/assert/assertions.go#L2074), [Eventually()](https://github.com/stretchr/testify/blob/v1.10.0/assert/assertions.go#L1943), [EventuallyWithT()](https://github.com/stretchr/testify/blob/v1.10.0/assert/assertions.go#L1943)

## Motivation
We hit nil deref panics in our dapr/dapr unit tests after upgrading to v1.11.x here in [this PR](https://github.com/dapr/dapr/pull/9345) ([see panic here](https://github.com/dapr/dapr/actions/runs/21289530220/job/61312418538#step:4:846)) because `EventuallyWithT` now runs the condition immediately on entry. You can see where I pinned back to 1.10 to resolve the issue in [this PR as a test to confirm the regression](https://github.com/dapr/dapr/pull/9352) and after confirming merged the pinned version [in this PR](https://github.com/dapr/dapr/pull/9345). The test from the linked CI failure is [here](https://github.com/dapr/dapr/blob/master/pkg/scheduler/server/internal/pool/loops/stream/stream_test.go#L240) for easy reference where we use atomic ptr, hence why I added a very similar test to your test suite.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -2018,12 +2018,7 @@ func Eventually(t TestingT, condition func() bool, waitFor time.Duration, tick t
 	ticker := time.NewTicker(tick)
 	defer ticker.Stop()
 
-	var tickC <-chan time.Time
-
-	// Check the condition once first on the initial call.
-	go checkCond()
-
-	for {
+	for tickC := ticker.C; ; {
 		select {
 		case <-timer.C:
 			return Fail(t, "Condition never satisfied", msgAndArgs...)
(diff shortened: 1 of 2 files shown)


cicoyle:
@brackendawson, @ccoVeille, @dolmen  - mind reviewing this one? 🙏🏻

ccoVeille:
I feel like @brackendawson is busy lately.

I am also. Let's wait for a bit

cc @fredbi who might be interested by the issue on his own fork

fredbi:
Thanks @ccoVeille I’ll look into it asap. I am trying to keep up with upstream

PR #1848, opened 2026-02-10
fix: use %#v instead of %q in Subset/NotSubset error messages

### LINKED ISSUE #1800: assert.NotSubset incorrecly formats with %q
## Description
NotSubset (but not Subset) seems to be using %q to format error output, that doesn't work for many collection element types.

## Step To Reproduce
```go
package kata_test

import (
	"testing"

)

type User struct {
	ID, Name string
	Age      int
}

func TestIfy(t *testing.T) {
	assert.NotSubset(t, []bool{true}, []bool{true})
}
```

## Expected behavior
Values of arguments to be formatted well.

## Actual behavior
```
--- FAIL: TestIfy (0.00s)
    /Users/brackendawson/src/github.com/brackendawson/kata/kata_test.go:15: 
        	Error Trace:	/Users/brackendawson/src/github.com/brackendawson/kata/kata_test.go:15
        	Error:      	[%!q(bool=true)] is a subset of [%!q(bool=true)]
        	Test:       	TestIfy
FAIL
coverage: 0.0% of statements
FAIL	github.com/brackendawson/kata	0.202s
FAIL
```

veeceey:
## Summary

- Fixes `NotSubset` (and `Subset`) error messages that use `%q` format verb, which produces broken output for non-string types (e.g., `%!q(bool=true)` for booleans, `'\x01'` for integers)
- Changes to `%#v` (Go-syntax representation) to match how `Subset` already formats its "does not contain" failure messages
- Updates corresponding test expectations to reflect the corrected output format

**Before (broken):**
```
Error: [%!q(bool=true)] is a subset of [%!q(bool=true)]
```

**After (fixed):**
```
Error: []bool{true} is a subset of []bool{true}
```

Fixes #1800

## Test plan

- [x] All existing `TestSubsetNotSubset` tests pass with updated expectations
- [x] `TestNotSubsetNil` passes unchanged
- [x] `TestNotSubsetWithSliceTooLongToPrint` passes with updated expectations
- [x] `TestNotSubsetWithMapTooLongToPrint` passes with updated expectations
- [x] `TestSubsetWithSliceTooLongToPrint` and `TestSubsetWithMapTooLongToPrint` pass unchanged
- [x] Full test suite (`go test ./...`) passes with no regressions

### DIFF
--- a/assert/assertions_test.go
+++ b/assert/assertions_test.go
@@ -1292,6 +1292,7 @@ func TestNotSubsetNil(t *testing.T) {
 // readable error messages using Go-syntax formatting (%#v) for various data types.
 // This catches regressions where %q was used, which produces broken output like
 // %!q(bool=true) for non-string types.
+// Related: https://github.com/stretchr/testify/issues/1800
 func TestSubsetNotSubsetErrorMessages(t *testing.T) {
 	t.Parallel()
 
@@ -1300,8 +1301,6 @@ func TestSubsetNotSubsetErrorMessages(t *testing.T) {
 		NotSubset(mockT, []bool{true, false}, []bool{true})
 		msg := mockT.errorString()
 		Contains(t, msg, "[]bool{true} is a subset of []bool{true, false}")
-		// Ensure no broken %q formatting like %!q(bool=true)
-		NotContains(t, msg, "%!q")
 	})
 
 	t.Run("NotSubset with float64 slices", func(t *testing.T) {
@@ -1309,7 +1308,6 @@ func TestSubsetNotSubsetErrorMessages(t *testing.T) {
 		NotSubset(mockT, []float64{1.1, 2.2, 3.3}, []float64{1.1, 2.2})
 		msg := mockT.errorString()
 		Contains(t, msg, "[]float64{1.1, 2.2} is a subset of []float64{1.1, 2.2, 3.3}")
-		NotContains(t, msg, "%!q")
 	})
 
 	t.Run("NotSubset with uint slices", func(t *testing.T) {
@@ -1317,7 +1315,6 @@ func TestSubsetNotSubsetErrorMessages(t *testing.T) {
 		NotSubset(mockT, []uint{10, 20, 30}, []uint{10, 20})
 		msg := mockT.errorString()
 		Contains(t, msg, "[]uint{0xa, 0x14} is a subset of []uint{0xa, 0x14, 0x1e}")
-		NotContains(t, msg, "%!q")
 	})
 
 	t.Run("NotSubset with map of int to bool", func(t *testing.T) {
@@ -1328,9 +1325,7 @@ func TestSubsetNotSubsetErrorMessages(t *testing.T) {
 		)
 		msg := mockT.errorString()
 		Contains(t, msg, "is a subset of")
-		// Verify %#v formatting is used (type info present, no %!q artifacts)
 		Contains(t, msg, "map[int]bool{")
-		NotContains(t, msg, "%!q")
 	})
 
 	t.Run("Subset failure with bool slices", func(t *testing.T) {
@@ -1338,7 +1333,6 @@ func TestSubsetNotSubsetErrorMessages(t *testing.T) {
 		Subset(mockT, []bool{true}, []bool{true, false})
 		msg := mockT.errorString()
 		Contains(t, msg, "[]bool{true} does not contain false")
-		NotContains(t, msg, "%!q")
 	})
 
 	t.Run("Subset failure with float64 slices", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

ccoVeille:
This sounds reasonable, but it's a known thing our test coverage on edge cases is low. So I wonder how this little change could affect existing code. But as it's only about the reporting message. I feel we are good.

Maybe there is a need to limit the changes to one that were failing. So a method could be added to return the old %q for things that can be reported with it, and %#v for others.

I might be wrong, and we are good, but I prefer to raise the point

veeceey:
Thanks for the review @ccoVeille! Good call on scoping the changes — I can limit it to just the methods that were actually failing with the bad message. Want me to narrow it down?

veeceey:
Narrowed the scope to only the methods that were producing confusing output. Good call on keeping the blast radius small.

ccoVeille:
>Thanks for the review. Good call on scoping the changes — I can limit it to just the methods that were actually failing with the bad message.

> Narrowed the scope to only the methods that were producing confusing output. Good call on keeping the blast radius small.

Thanks for this change.

Here again the commit shows off the lack of tests we have in testify.

You were able to change this without having to change unit tests.

Could you please add unit tests that are missing ? I assume you have a good knowledge of the types this method handles.

veeceey:
Absolutely, I can add unit tests for the affected methods. You're right that changing the formatting without touching tests is a red flag for coverage gaps. I'll add tests that exercise the different data types and edge cases these methods handle to make sure the new formatting works as expected.

veeceey:
Added more unit tests to expand coverage for the error message formatting across different types:

- **String slices** -- verifies `%#v` still produces readable output for the one type `%q` handled well (e.g. `[]string{"hello"}`)
- **Map-to-map with int keys** -- exercises the map-to-map comparison path in `NotSubset` with non-string keys
- **Subset failure with map of int keys** -- exercises the map "does not contain" path in `Subset`
- **Subset failure with uint slices** -- covers the slice "does not contain" path with unsigned integers
- **NotSubset with int32 slices** -- covers a narrower integer type to verify no `%!q` artifacts

All tests pass, including the full suite (`go test ./...`). The test function `TestSubsetNotSubsetErrorMessages` now has 14 subtests covering bool, float64, uint, int32, byte, string, struct, and map types for both `Subset` and `NotSubset` error paths.

ccoVeille:
@veeceey could you check at Copilot feedbacks ? Maybe there is Jo need, but just in case.

Please add tests if you changes some code

assert/assertions_test.go
```diff
@@ -1288,6 +1288,198 @@ func TestNotSubsetNil(t *testing.T) {
 	}
 }
 
+// TestSubsetNotSubsetErrorMessages verifies that Subset and NotSubset produce
+// readable error messages using Go-syntax formatting (%#v) for various data types.
+// This catches regressions where %q was used, which produces broken output like
+// %!q(bool=true) for non-string types.
```
ccoVeille, line 1294: You should mention the related issues too

assert/assertions_test.go
```diff
@@ -1288,6 +1288,198 @@ func TestNotSubsetNil(t *testing.T) {
...
+func TestSubsetNotSubsetErrorMessages(t *testing.T) {
+	t.Parallel()
+
+	t.Run("NotSubset with bool slices", func(t *testing.T) {
+		mockT := new(mockTestingT)
+		NotSubset(mockT, []bool{true, false}, []bool{true})
+		msg := mockT.errorString()
+		Contains(t, msg, "[]bool{true} is a subset of []bool{true, false}")
+		// Ensure no broken %q formatting like %!q(bool=true)
+		NotContains(t, msg, "%!q")
```
ccoVeille: I think this is enough

```suggestion
		Contains(t, msg, "[]bool{true} is a subset of []bool{true, false}")
```

The code may divert and this refers to something you remove, and that no longer exists once this PR will be merged.

For me checking, Contains is enough.

This comment apply to everything you added to this PR

ccoVeille: If we want to add a detection of bad formatting looking for "%!" would be OK if we consider it will detect:
- %!(MISSING)
- %!(EXTRA string=foo) and variation
- %!d(string=hello), %!q(bool=true)

But then I feel it should be part of testify test suite, and not limited to this PR, and the refactoring would be way larger than the scope of this PR and related issue

veeceey:
Hey @ccoVeille, took a look at the Copilot feedback and your review comments. Good points across the board — I'll simplify the test assertions to just use Contains like you suggested, and add the related issue references. Will push those changes shortly.

ccoVeille (approved):

veeceey:
Pushed the changes — simplified all assertions to just use Contains (removed the NotContains checks), and added issue references in the test comments. All tests passing.

veeceey:
hey @ccoVeille, all the review feedback has been addressed and the changes are pushed. anything else needed to get this merged?

ccoVeille:
I would expect other maintainers to take a look

 @dolmen @brackendawson

dolmen:
@veeceey I'm sorry, I hadn't noticed your PR. I have just merged #1888 which is about fixing the same issue.

Please rebase (in case there is something missing in #1888) or close.

PR #1854, opened 2026-02-24
fix(assert): handle invalid regexp gracefully instead of panicking

### LINKED ISSUE #1794: assert.Regexp panics with invalid rx expression

bk-simon:
## Summary

Fixes #1794

`assert.Regexp` and `assert.NotRegexp` used `regexp.MustCompile` internally (via `matchRegexp`) when the `rx` argument was a string or non-`*regexp.Regexp` value. `MustCompile` panics on an invalid regexp pattern, which is documented as intended only for package-level initialization of global variables.

This causes a runtime panic rather than a test failure when a user accidentally passes an invalid regexp:

```go
    require.Regexp(t, `\C`, "some string") // panics!
}
```

## What changed

- Changed `matchRegexp` signature from `func matchRegexp(rx, str interface{}) bool` to `func matchRegexp(rx, str interface{}) (bool, error)` 
- Replaced `regexp.MustCompile` with `regexp.Compile` — on error the function returns the compile error
- `Regexp` and `NotRegexp` now call `Fail` with a descriptive `"Invalid regexp: ..."` message when the pattern fails to compile, so the test fails gracefully rather than panicking
- Added `TestRegexpInvalidExpression` to guard against regression

## Test plan

- [x] `go test ./assert/ -run TestRegexp` — all pass
- [x] `go test ./...` — all packages pass
- [x] `TestRegexpInvalidExpression` specifically verifies invalid patterns fail gracefully (not panic)


assert/assertions_test.go
```diff
@@ -2557,6 +2557,23 @@ func TestRegexp(t *testing.T) {
...
 
+func TestRegexpInvalidExpression(t *testing.T) {
+	t.Parallel()
+
+	// Verify that an invalid regexp expression fails the test with a useful
+	// error message instead of panicking (regression test for issue #1794).
+	mockT := new(testing.T)
+
+	// \C is an invalid escape sequence - previously this caused a panic via
+	// regexp.MustCompile; now it should fail gracefully.
```
ccoVeille: I dislike the idea of having an inline comment that explain what the code was before.

It's a valuable information in the commit message and the PR description, but not with an inline comment.

Inline comments are here to explain what the code is about, and here this should be enough

```suggestion
	// \C is an invalid escape sequence, it should fail gracefully.
```

My motivation here is that once merged there is no need to talk about what was, and it should be clear what this test is about

assert/assertions_test.go
```diff
@@ -2557,6 +2557,23 @@ func TestRegexp(t *testing.T) {
 	}
 }
 
+func TestRegexpInvalidExpression(t *testing.T) {
+	t.Parallel()
+
+	// Verify that an invalid regexp expression fails the test with a useful
+	// error message instead of panicking (regression test for issue #1794).
```
ccoVeille: Right now, this test is not validating that the error message is the one we expect, "one that is useful" as you mention

You can use mockTestingT struct for that

assert/assertions_test.go
bk-simon: Good call — simplified to just `// \C is an invalid escape sequence, it should fail gracefully.` Historical context belongs in the commit message, not inline.

assert/assertions_test.go
bk-simon: Switched to `mockTestingT` and added `Contains(t, mockT.errorString(), "Invalid regexp")` for both `Regexp` and `NotRegexp`. Now the test actually verifies the error message is informative, not just that it failed.

ccoVeille (approved):

bk-simon:
Closing — this was from an exploratory session and I won't be able to shepherd it through. The fix is valid (collaborator-approved) so feel free to cherry-pick or reopen if you'd like to merge it.

PR #1861, opened 2026-03-13
add ErrorAsType and related functions

### LINKED ISSUE #1860: Add require.ErrorAsType for Go 1.26+
## Description
Go 1.26 introduced [`errors.AsType`](https://pkg.go.dev/errors#AsType).  It would be nice to have a Testify equivalent.

## Use case

Previously, I might have:
```go
var syntaxErr *json.SyntaxError
require.ErrorAs(t, err, &syntaxErr)
```

Now, with Go 1.26, I _could_ write this:
```go
_, ok := errors.AsType[*json.SyntaxError](err)
require.True(t, ok)
```
That removes the allocation, but loses details on failure.

## Proposed solution

I'd like to be able to write:
```go
require.ErrorAsType[*json.SyntaxError](t, err)
```

mattjohnsonpint:
## Summary
Adds Testify functions that mirror the new `errors.AsType[E]` in Go 1.26.

## Changes

For Go 1.26+ consumers, adds the following:

```go
assert.ErrorAsType[E error](t TestingT, err error, msgAndArgs ...any) (E, bool)
assert.ErrorAsTypef[E error](t TestingT, err error, msg string, args ...any) (E, bool)
assert.NotErrorAsType[E error](t TestingT, err error, msgAndArgs ...any) bool
assert.NotErrorAsTypef[E error](t TestingT, err error, msg string, args ...any) bool

require.ErrorAsType[E error](t TestingT, err error, msgAndArgs ...any) E
require.ErrorAsTypef[E error](t TestingT, err error, msg string, args ...any) E
require.NotErrorAsType[E error](t TestingT, err error, msgAndArgs ...any)
require.NotErrorAsTypef[E error](t TestingT, err error, msg string, args ...any)
```

Note that `E` is returned where appropriate, to allow the caller to follow with additional requirements without further casting.

## Motivation

Instead of:
```go
var syntaxErr *json.SyntaxError
require.ErrorAs(t, err, &syntaxErr)
```

With this change, and if I'm using Go 1.26.0 or later, I can now write:
```go
require.ErrorAsType[*json.SyntaxError](t, err)
```

Other functions are similar in usage to their existing `*ErrorAs*` counterparts.

## Related issues
Closes #1860

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,6 +34,8 @@ jobs:
           - "1.22"
           - "1.23"
           - "1.24"
+          - "1.25"
+          - "1.26"
     steps:
       - uses: actions/checkout@v5
       - name: Setup Go
--- a/_codegen/main.go
+++ b/_codegen/main.go
@@ -163,6 +163,12 @@ func analyzeCode(scope *types.Scope, docs *doc.Package) (imports.Importer, []tes
 			continue
 		}
 
+		// Skip generic functions (type parameters present) — they cannot be
+		// reproduced by codegen and are maintained by hand.
+		if sig.TypeParams() != nil && sig.TypeParams().Len() > 0 {
+			continue
+		}
+
 		funcs = append(funcs, testFunc{*outputPkg, fdocs, fn})
 		importer.AddImportsFrom(sig.Params())
 	}
--- a/assert/assertions_go1.26.go
+++ b/assert/assertions_go1.26.go
@@ -1,0 +1,79 @@
+//go:build go1.26
+
+package assert
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+)
+
+// ErrorAsType asserts that at least one of the errors in err's tree matches
+// type E, using errors.AsType. On success it returns the matched error value.
+// This is a Go 1.26+ generic alternative to ErrorAs that avoids the need for
+// a pre-declared target variable.
+//
+//	assert.ErrorAsType[*json.SyntaxError](t, err)
+func ErrorAsType[E error](t TestingT, err error, msgAndArgs ...any) (E, bool) {
+	if h, ok := t.(tHelper); ok {
+		h.Helper()
+	}
+
+	if target, ok := errors.AsType[E](err); ok {
+		return target, true
+	}
+
+	expectedType := reflect.TypeFor[E]().String()
+	if err == nil {
+		Fail(t, fmt.Sprintf("An error is expected but got nil.\n"+
+			"expected: %s", expectedType), msgAndArgs...)
+		var zero E
+		return zero, false
+	}
+
+	chain := buildErrorChainString(err, true)
+	Fail(t, fmt.Sprintf("Should be in error chain:\n"+
+		"expected: %s\n"+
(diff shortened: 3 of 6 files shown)


_codegen/main.go
```diff
@@ -163,6 +163,12 @@ func analyzeCode(scope *types.Scope, docs *doc.Package) (imports.Importer, []tes
 			continue
 		}
 
+		// Skip generic functions (type parameters present) — they cannot be
+		// reproduced by codegen and are maintained by hand.
+		if sig.TypeParams() != nil && sig.TypeParams().Len() > 0 {
+			continue
+		}
+
```
ccoVeille, line 171: I'm unsure what this could mean and what would be the problems if you are not adding this to the PR

.github/workflows/main.yml
```diff
@@ -34,6 +34,8 @@ jobs:
           - "1.22"
           - "1.23"
           - "1.24"
+          - "1.25"
+          - "1.26"
```
dolmen, line 38: This must be a separate PR.

ccoVeille (commented):
This sounds reasonable. I didn't check by pulling the branch, but the first look seems OK .

I was afraid you might break old Go versions by using a recent feature, but you used gu:build tag 👍

My only concern now is to check if the new files you add are auto generated or not.

_codegen/main.go
```diff
@@ -163,6 +163,12 @@ func analyzeCode(scope *types.Scope, docs *doc.Package) (imports.Importer, []tes
 			continue
 		}
 
+		// Skip generic functions (type parameters present) — they cannot be
+		// reproduced by codegen and are maintained by hand.
```
dolmen, line 167: This is not the way we must take.

To introduce generics in Testify, someone has to do the job of porting the code generator to support generics.

dolmen (changes requested):
To introduce generic functions in `assert` and `require` the code generator has to be improved.

Introducing generic functions without codegen will be too heavy to maintain.

dolmen:
In Testify `assert`, functions are also exposed as methods of the [`Assertions`](https://pkg.go.dev/github.com/stretchr/testify/assert#Assertions) type.

Go doesn't yet supports methods with type parameters. While this is coming (see golang/go#77273), we aren't there yet. We need changes in the codegen that will be ready to handle methods with type parameters once they are there.

mattjohnsonpint:
Thanks for the feedback.  I'm away for the weekend but will revisit this soon.

fredbi:
Yes interesting. The assertion expression is definitely lighter.

mattjohnsonpint:
Revised substantially for Go 1.27 in #1937

fredbi:
@mattjohnsonpint an implementation of this feature already exists at my fork github.com/go-openapi/testify/v2 - it might provide you some extra inspiration, although it is essentially in the same spirit as your implementation. The fork is planned to support the generic "forward" methods in a few weeks (waiting for go1.27 to release).

Super happy that the code generator at stretchr/testify gets some uplifting! thanks.

Fred

mattjohnsonpint:
@fredbi - yeah, that's what my #1937 PR does.  Go 1.27 was released today BTW.

PR #1863, opened 2026-03-17
assert.EqualValues: fix float32 and float64 comparisions

### LINKED ISSUE #1576: v1.9.0 breaks float comparisons with EqualValues
## Description
Migrating from 1.8.4 to 1.9.0 seems to have broken some of our unit tests that relied on comparing floats of varying bit depths with `EqualValues(...)`.

I tried fixing it myself but given I'm not very familiar with the `reflect`, I didn't get much further than discovering the issue stemmed from comparing the output of `reflect.ValueOf(aFloat32).Convert(float64Type).Interface()` against the actual `float64` - `Convert` seems to misinterpret the underlying bits and the interface of the former operand comes out to eg. `10.100000381469727`

## Step To Reproduce
Run the following suite:

```go
package a_test

import (
	"testing"

)

type FailingMinimalSuite struct {
	suite.Suite
}

func (s *FailingMinimalSuite) TestFloatEqualValues() {
	s.EqualValues(float32(10.1), float64(10.1))
}

func TestRunFailingMinimalSuite(t *testing.T) {
	suite.Run(t, &FailingMinimalSuite{})
}
```

Alternatively, inject the following case into `TestObjectsAreEqualValues`:
```go
{float32(10.1), float64(10.1), true},
```

## Expected behavior
I'd expect the `EqualValues(float32(10.1), float64(10.1))` assertion to pass.

## Actual behavior
The abovementioned assertion fails.

snopan:
## Summary
This PR attempts to provide a solution to #1576 where converting `float32` to `float64` for comparison will fail due to differences in binary representation. Where 
```golang
fmt.Println(float32(10.1))          // 10.1
fmt.Println(float64(10.1))          // 10.1
fmt.Println(float64(float32(10.1))) //10.100000381469727
```
This solution only allows comparison when `float64` value has just as many decimal digits defined as `float32`. For example
```golang
fmt.Println(float32(10.123456))      // 10.123456
fmt.Println(float64(10.12345600000)) // 10.123456
```

But for `1.0 / 3.0`, these are very different values for both `float32` and `float64`.
```golang
fmt.Println(float32(1.0 / 3.0)) // 0.33333334
fmt.Println(float64(1.0 / 3.0)) // 0.3333333333333333
```
Because the `float64` value has more decimal places defined than `float32`, they are not equal anyway.

[playground](https://go.dev/play/p/iyq8V_1o2Qb)
## Changes
* Added rounding logic for converted `float32` value to remove the trailing non zero decimals after the 6th decimal place. 
* Added some variables to simplify logic which variable to convert and round

## Motivation
When converting to `float64`, zeros are added the binary representation losing the accuracy of the original value. When we are comparing the `float32` we only care about the digits that are accurate, which is why we remove the trailing digits. The choice to round to 6 digits was aligning to what I found in golang playground, `float32` seems to round to 6 digits.
```golang
fmt.Println(float32(10.123456))  // 10.123456
fmt.Println(float32(10.1234567)) // 10.123457
```
[playground](https://go.dev/play/p/GUt2b2hqnwD)

## Related issues
Closes #1576

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -187,11 +187,24 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
 	// If BOTH values are numeric, there are chances of false positives due
 	// to overflow or underflow. So, we need to make sure to always convert
 	// the smaller type to a larger type before comparing.
-	if expectedType.Size() >= actualType.Size() {
-		return actualValue.Convert(expectedType).Interface() == expected
+	fromType := actualType
+	toType := expectedType
+	fromValue := actualValue
+	toValue := expectedValue
+	if expectedType.Size() < actualType.Size() {
+		fromType = expectedType
+		toType = actualType
+		fromValue = expectedValue
+		toValue = actualValue
 	}
 
-	return expectedValue.Convert(actualType).Interface() == actual
+	newValue := fromValue.Convert(toType).Interface()
+	if fromType.Kind() == reflect.Float32 && toType.Kind() == reflect.Float64 {
+		scale := math.Pow(10, 6)
+		newValue = math.Round(newValue.(float64)*scale) / scale
+	}
+
+	return newValue == toValue.Interface()
 }
 
 // isNumericType returns true if the type is one of:
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -198,6 +198,11 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
 		toValue = actualValue
 	}
 
+	// If we are converting from float32 to float64, the converted value will
+	// have trailing non zero decimals due to binary representation differences
+	// For example: float64(float32(10.1)) = 10.100000381469727
+	// To remove the trailing decimals we can round the 64-bit value to
+	// expected precision of 32-bit which is 6 decimal places
 	newValue := fromValue.Convert(toType).Interface()
 	if fromType.Kind() == reflect.Float32 && toType.Kind() == reflect.Float64 {
 		scale := math.Pow(10, 6)
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -177,39 +177,53 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
 		return false
 	}
 
+	convertedExpectedValue := expectedValue.Convert(actualType).Interface()
+
 	if !isNumericType(expectedType) || !isNumericType(actualType) {
 		// Attempt comparison after type conversion
-		return reflect.DeepEqual(
-			expectedValue.Convert(actualType).Interface(), actual,
-		)
+		return reflect.DeepEqual(convertedExpectedValue, actual)
 	}
 
 	// If BOTH values are numeric, there are chances of false positives due
 	// to overflow or underflow. So, we need to make sure to always convert
 	// the smaller type to a larger type before comparing.
-	fromType := actualType
-	toType := expectedType
-	fromValue := actualValue
-	toValue := expectedValue
-	if expectedType.Size() < actualType.Size() {
-		fromType = expectedType
-		toType = actualType
-		fromValue = expectedValue
-		toValue = actualValue
+	// Assume smaller is expected value and larger is actual value
+	smallerTypeValue, largerTypeValue := expectedValue, actualValue
+	smallerValueCmp, largerValueCmp := convertedExpectedValue, actual
+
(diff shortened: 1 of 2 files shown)


assert/assertions.go
```diff
@@ -187,11 +187,29 @@ func ObjectsAreEqualValues(expected, actual interface{}) bool {
...
+	// For example: float64(float32(10.1)) = 10.100000381469727
+	// To remove the trailing decimals we can round the 64-bit value to
+	// expected precision of 32-bit which is 6 decimal places
+	newValue := fromValue.Convert(toType).Interface()
+	if fromType.Kind() == reflect.Float32 && toType.Kind() == reflect.Float64 {
+		scale := math.Pow(10, 6)
+		newValue = math.Round(newValue.(float64)*scale) / scale
+	}
+
+	return newValue == toValue.Interface()
```
ccoVeille: Does this approach work for comparing very large or very small number?

I mean the 10^6 sounds fit the exact use case of 10.1

Shouldn't simply use assert.InEpsilon with an hard coded value here?

ccoVeille: While I understand why you do it, I'm a bit surprised by the fact the from and to are not clear.

I will suggest something

assert/assertions_test.go
```diff
@@ -161,6 +161,11 @@ func TestObjectsAreEqualValues(t *testing.T) {
 		{3.14, complex128(1e+100 + 1e+100i), false},
 		{complex128(1e+10 + 1e+10i), complex64(1e+10 + 1e+10i), true},
 		{complex64(1e+10 + 1e+10i), complex128(1e+10 + 1e+10i), true},
+		{float32(10.1), float64(10.1), true},
+		{float64(10.1), float32(10.1), true},
+		{float32(10.123456), float64(10.12345600), true},
+		{float32(10.123456), float64(10.12345678), false},
+		{float32(1.0 / 3.0), float64(1.0 / 3.0), false},
```
ccoVeille, line 172: This doesn't make sense to me, comparing 1.0/3 and 1.0/3 should say they are equal, no ?

assert/assertions.go
ccoVeille: Oh I get it now, but so you are comparing a 1e-6 factor, but this won't work with a number close to math.Inf, because if you multiply by 1e6, you will explode math.Inf

snopan: Yeah you're right, I think it might explode if it's close to `math.Inf `

I could try using `assert.Epsilon ` like you suggested, it looks like it would similar comparison to mine and hopefully doesn't have the  `math.Inf` problem. Will have a look into it

assert/assertions_test.go
snopan: I should probably make this clear in the PR aswell, basically this solution is to allow comparing `float32` and `float64` when the `float64` value is the exact same (with only six decimal digits) as the `float32` value. For example
```golang
fmt.Println(float32(10.123456))      // 10.123456
fmt.Println(float64(10.12345600000)) // 10.123456
```

But for `1.0 / 3.0`, these are very different values for both `float32` and `float64`.
```golang
fmt.Println(float32(1.0 / 3.0)) // 0.33333334
fmt.Println(float64(1.0 / 3.0)) // 0.3333333333333333
```
Because the `float64` value has more decimal places defined than `float32`, they are not equal anyway.

assert/assertions.go
snopan: Sure thing, happy to rework this to something that's clearer

ccoVeille: I have made the changes, but I cannot push my changes to your branch even if you use "allow maintainer edits" the reason is that your `master` branch is protected in your fork (which is a normal)

You shouldn't have worked in your master branch :smirk: 

I can open a new PR to testify, or a PR to your fork

I will choose the PR to your fork, and we will discuss there

snopan: Ahh okay, sort of my first few times contributing to open source. Should I create a local branch on my fork then contribute by pointing that non main branch to stretchrs?

ccoVeille: There are many way, but yes having a dedicated branch in your fork is good habit.

assert/assertions_test.go
ccoVeille: You are right, I opened https://github.com/snopan/testify/pull/2 to only back up your code with more tests

assert/assertions.go
ccoVeille: You are right, I opened https://github.com/snopan/testify/pull/2 to only back up your code with more tests

PR #1866, opened 2026-04-03
mock: reduce data races in Arguments.Diff for pointer-like arguments

### LINKED ISSUE #1597: Mock code is prone to data races

pavelzeman:
## Summary

Fixes `Arguments.Diff` causing data races when mock arguments (pointers, maps, slices) are concurrently modified by other goroutines.

## Problem

`Arguments.Diff` uses `fmt.Sprintf("%v")` to format arguments for diff output. This traverses the underlying data structure, which races with concurrent writers:

- **Maps**: triggers `fatal error: concurrent map iteration and map write` — crashes the entire test process (non-recoverable by `recover()`)
- **Pointers/slices**: causes data races detectable by `go test -race`

This is a real-world issue. For example, when testing code that passes `*sql.DB` to mocked methods, the database connection cleaner goroutine modifies internal fields while testify formats the struct for diff output.

## Solution

Introduces `safeFormatArg()` which uses `%p` (address only) for reference types (`Ptr`, `Map`, `Slice`, `Chan`) instead of `%v`, avoiding traversal of the underlying data structure entirely. All other types (structs, primitives, strings) use the full `%v` representation since they are value types and cannot race.

### Why not deep-copy?

Copying the arguments before formatting seems like it would preserve the full `%v` output, but:
- **Maps**: even iterating to copy triggers the fatal `concurrent map iteration and map write` error
- **Pointers**: copying the pointed-to value still triggers the race detector under `-race` during the copy itself

### Why not recover()?

- **Maps**: `concurrent map iteration and map write` is a runtime throw, not a panic — `recover()` cannot catch it
- **Pointers/slices**: while panics are recoverable, the `-race` detector flags the concurrent access before any panic occurs, causing test failures

## Tests

Three new concurrent-modification tests:

1. **`Test_Arguments_Diff_ConcurrentPointerModification`** — pointer struct modified by another goroutine (reproducer from #1597)
2. **`Test_Arguments_Diff_ConcurrentMapModification`** — map modified by another goroutine
3. **`Test_Arguments_Diff_ConcurrentSliceModification`** — slice modified by another goroutine

All three crash or trigger race detector failures without the fix and pass cleanly with it.

## Tradeoff

Reference-type arguments now show addresses instead of values in diff output (e.g., `([]int=0xc0000b4030)` instead of `([]int=[1 2 3])`). The detailed structural diff (from `ObjectsAreEqual`) still appears above the summary line when arguments don't match, so the diagnostic value loss is minimal.

Fixes #1597

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -977,7 +977,7 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			actualFmt = "(Missing)"
 		} else {
 			actual = objects[i]
-			actualFmt = fmt.Sprintf("(%[1]T=%[1]v)", actual)
+			actualFmt = safeFormatArg(actual)
 		}
 
 		if len(args) <= i {
@@ -985,7 +985,7 @@ func (args Arguments) Diff(objects []interface{}) (string, int) {
 			expectedFmt = "(Missing)"
 		} else {
 			expected = args[i]
-			expectedFmt = fmt.Sprintf("(%[1]T=%[1]v)", expected)
+			expectedFmt = safeFormatArg(expected)
 		}
 
 		if matcher, ok := expected.(argumentMatcher); ok {
@@ -1309,4 +1309,28 @@ func isFuncSame(f1, f2 *runtime.Func) bool {
 	f2File, f2Loc := f2.FileLine(f2.Entry())
 
 	return f1File == f2File && f1Loc == f2Loc
+}
+
+// safeFormatArg formats an argument for display in diff output, handling
+// concurrent modification safely. For types that contain inherent references
+// (pointers, maps, slices, channels), fmt.Sprintf("%v") would traverse the
+// underlying data structure, which races with concurrent writers. For maps
+// this is a non-recoverable fatal error; for pointers and slices it triggers
+// data races detectable by -race.
+//
+// To avoid this, reference types are formatted with %p (address only) instead
+// of %v. All other types (structs, primitives, strings) are value types that
+// cannot race, so they use the full %v representation.
+//
+// See https://github.com/stretchr/testify/issues/1597
+func safeFormatArg(v interface{}) string {
+	if v == nil {
+		return fmt.Sprintf("(%[1]T=%[1]v)", v)
+	}
+	switch reflect.TypeOf(v).Kind() {
+	case reflect.Ptr, reflect.Map, reflect.Slice, reflect.Chan:
+		return fmt.Sprintf("(%T=%p)", v, v)
+	default:
+		return fmt.Sprintf("(%[1]T=%[1]v)", v)
+	}
 }
--- a/mock/mock_test.go
+++ b/mock/mock_test.go
@@ -2044,6 +2044,81 @@ func Test_Arguments_Diff_WithArgMatcher(t *testing.T) {
 	assert.Contains(t, diff, `No differences.`)
 }
 
+// Test_Arguments_Diff_ConcurrentPointerModification verifies that
+// Arguments.Diff does not race when a pointer argument is concurrently
+// modified. This is a regression test for https://github.com/stretchr/testify/issues/1597.
+// Adapted from https://github.com/stretchr/testify/pull/1598.
+func Test_Arguments_Diff_ConcurrentPointerModification(t *testing.T) {
+	t.Parallel()
+
+	type data struct {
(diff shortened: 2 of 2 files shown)


dolmen:
Related commit in project Mattermost's testsuite: https://github.com/mattermost/mattermost/commit/4b8a4ae2b35381a46b891f831518a731d6f9e313

Previous closed PR: #1865

mock/mock.go
```diff
@@ -1310,3 +1310,27 @@ func isFuncSame(f1, f2 *runtime.Func) bool {
...
+// underlying data structure, which races with concurrent writers. For maps
+// this is a non-recoverable fatal error; for pointers and slices it triggers
+// data races detectable by -race.
+//
+// To avoid this, reference types are formatted with %p (address only) instead
+// of %v. All other types (structs, primitives, strings) are value types that
+// cannot race, so they use the full %v representation.
+//
+// See https://github.com/stretchr/testify/issues/1597
+func safeFormatArg(v interface{}) string {
```
dolmen, line 1326: We need a testsuite for this new function.

Also, in its current implementation it is safer but no yet "safe". So the name is misleading.

mock/mock.go
```diff
@@ -1310,3 +1310,27 @@ func isFuncSame(f1, f2 *runtime.Func) bool {
...
+	if v == nil {
+		return fmt.Sprintf("(%[1]T=%[1]v)", v)
+	}
+	switch reflect.TypeOf(v).Kind() {
+	case reflect.Ptr, reflect.Map, reflect.Slice, reflect.Chan:
```
dolmen, line 1331: Replace `reflect.Ptr` (deprecated) with `reflect.Pointer`.

Other types must also be handled. In particular `reflect.Array`, `reflect.Interface`. We definitely need a comprehensive testsuite.

mock/mock.go
```diff
@@ -1310,3 +1310,27 @@ func isFuncSame(f1, f2 *runtime.Func) bool {
...
+		return fmt.Sprintf("(%T=%p)", v, v)
```
dolmen, line 1332: Have you checked the output of `safeFormatArg([]int(nil))`?
We need a comprehensive test suite.

Also use `[1]` like in other paths.

mock/mock_test.go
```diff
@@ -2044,6 +2044,81 @@ func Test_Arguments_Diff_WithArgMatcher(t *testing.T) {
 	assert.Contains(t, diff, `No differences.`)
 }
 
+// Test_Arguments_Diff_ConcurrentPointerModification verifies that
+// Arguments.Diff does not race when a pointer argument is concurrently
```
dolmen, line 2048: The 3 tests have much common code. Please refactor to extract a `func(t *testing.T, arg interface{}, accessArg func())`.

Also group them as a top level test with 3 subtests.

dolmen (changes requested):
The submission claims "Fixes #1594", but this isn't the case: some cases of races are avoided in formatting the output (so this PR would still be an improvement), but races could also happen in the comparison itself.

So the claims in the commit message and the description of this PR must be lowered.

Some test cases are added that show the issue when the diff output is formatted, but they cover just Arguments.Diff, and more tests that cover the full call from Mock would be welcome.

Once my specific issues are fixed (ex: missing testsuite for safeFormatArg, renamming the function...) we will probably be able to merge this, but #1594 will still not be fixed.

PR #1873, opened 2026-04-19
suite: skip AfterTest when SetupTest or BeforeTest skips (#1781)

### LINKED ISSUE #1781: suite: Skipping a test in BeforeTest prevents test execution but not AfterTest
## Description
Related to #473 and #1722. If testing.T.Skip is called in Before test, we should skip that test and its teardowns. If it's called in the suite setup, we should skip the entire suite.

## Step To Reproduce
```go
package kata_test

import (
	"testing"

)

type My struct {
	suite.Suite
}

func (m *My) BeforeTest(s, t string) {
	m.T().Skip()
	panic("before")
}

func (m *My) AfterTest(s, t string) {
	panic("after")
}

func (m *My) TestA() {
	panic("during")
}

func TestIfy(t *testing.T) {
	suite.Run(t, &My{})
}
```

## Expected behavior
No panics

## Actual behavior
panic: "after"

barry3406:
## Summary
Skip the `AfterTest` hook when a `SetupTest` or `BeforeTest` call caused the test to be skipped, so `AfterTest` always has a matching `BeforeTest`.

## Changes
- `suite/suite.go`: track whether `BeforeTest` finished via a local `beforeTestFinished` flag; only invoke `AfterTest` in the deferred cleanup when that flag is set.
- `suite/suite_test.go`: add tests covering skip-in-`BeforeTest`, skip-in-`SetupTest`, and skip-in-test-body (regression guard).

`TearDownTest` is intentionally left unchanged: it still runs in all cases so users can clean up partially initialized state set up before the skip.

## Motivation
Per #1781, calling `t.Skip()` from `BeforeTest` correctly prevents the test body from running, but `AfterTest` is still invoked — leaving `AfterTest` without the paired `BeforeTest` it expects. Scenario 3 in the reporter's example causes a `panic("after")` today even though the test should be skipped cleanly. `SetupTest` has the same inconsistency (scenario 2). Gating `AfterTest` on `BeforeTest` having finished resolves both without changing the test-body-skip path.

## Related issues
Closes #1781

### DIFF
--- a/suite/suite.go
+++ b/suite/suite.go
@@ -181,6 +181,7 @@ func Run(t *testing.T, suite TestingSuite) {
 				parentT := suite.T()
 				suite.SetT(t)
 				defer recoverAndFailOnPanic(t)
+				var beforeTestFinished bool
 				defer func() {
 					t.Helper()
 
@@ -189,7 +190,12 @@ func Run(t *testing.T, suite TestingSuite) {
 					stats.end(method.Name, !t.Failed() && r == nil)
 
 					if afterTestSuite, ok := suite.(AfterTest); ok {
-						afterTestSuite.AfterTest(suiteName, method.Name)
+						// Only run AfterTest if BeforeTest finished without
+						// calling Skip. This mirrors the guarantee that
+						// AfterTest has a matching BeforeTest.
+						if beforeTestFinished {
+							afterTestSuite.AfterTest(suiteName, method.Name)
+						}
 					}
 
 					if tearDownTestSuite, ok := suite.(TearDownTestSuite); ok {
@@ -206,6 +212,7 @@ func Run(t *testing.T, suite TestingSuite) {
 				if beforeTestSuite, ok := suite.(BeforeTest); ok {
 					beforeTestSuite.BeforeTest(methodFinder.Elem().Name(), method.Name)
 				}
+				beforeTestFinished = true
 
 				stats.start(method.Name)
 
--- a/suite/suite_test.go
+++ b/suite/suite_test.go
@@ -813,3 +813,138 @@ func TestSuiteSignatureValidation(t *testing.T) {
 	assert.True(t, suiteTester.setUp, "SetupSuite should have been executed")
 	assert.True(t, suiteTester.toreDown, "TearDownSuite should have been executed")
 }
+
+// skipInBeforeTestSuite is used to verify that AfterTest is not called when
+// BeforeTest skips the test (#1781).
+type skipInBeforeTestSuite struct {
+	Suite
+	beforeTestCount int
+	afterTestCount  int
+	testRunCount    int
+}
+
+func (s *skipInBeforeTestSuite) BeforeTest(_, _ string) {
+	s.beforeTestCount++
+	s.T().Skip()
+}
+
+func (s *skipInBeforeTestSuite) AfterTest(_, _ string) {
+	s.afterTestCount++
+}
+
+func (s *skipInBeforeTestSuite) TestA() {
+	s.testRunCount++
+}
+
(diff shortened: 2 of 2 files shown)


suite/suite.go
```diff
@@ -189,7 +190,12 @@ func Run(t *testing.T, suite TestingSuite) {
 					stats.end(method.Name, !t.Failed() && r == nil)
 
 					if afterTestSuite, ok := suite.(AfterTest); ok {
-						afterTestSuite.AfterTest(suiteName, method.Name)
+						// Only run AfterTest if BeforeTest finished without
+						// calling Skip. This mirrors the guarantee that
+						// AfterTest has a matching BeforeTest.
+						if beforeTestFinished {
+							afterTestSuite.AfterTest(suiteName, method.Name)
+						}
```
dolmen, line 198: Move the `if beforeTestFinished` around the `if suite.(AfterTest)` block: I prefer to keep the cast really close to the method call.

suite/suite_test.go
```diff
@@ -813,3 +813,138 @@ func TestSuiteSignatureValidation(t *testing.T) {
 	assert.True(t, suiteTester.setUp, "SetupSuite should have been executed")
 	assert.True(t, suiteTester.toreDown, "TearDownSuite should have been executed")
 }
+
+// skipInBeforeTestSuite is used to verify that AfterTest is not called when
+// BeforeTest skips the test (#1781).
+type skipInBeforeTestSuite struct {
```
dolmen, line 819: Wrap SuiteTester that already handles counting.
You'll have to override only the method where Skip() happens.

With this the case will be much shorter and so clearer.

```go
type skipInBeforeTestSuite struct {
	SuiteTester
}

func (s *skipInBeforeTestSuite) BeforeTest(suiteName, testName string) {
	s.SuiteTester.BeforeTest(suiteName, testName)
	if testName == "TestA" {
		s.T().Skip()
	}
}
```

Also add a TestB and check that its Before/After are correctly called

suite/suite_test.go
```diff
@@ -813,3 +813,138 @@ func TestSuiteSignatureValidation(t *testing.T) {
...
+	assert.True(t, ok, "suite should not fail when BeforeTest skips")
+	assert.Equal(t, 1, s.beforeTestCount, "BeforeTest should run once")
+	assert.Equal(t, 0, s.testRunCount, "test body should not run when BeforeTest skips")
+	assert.Equal(t, 0, s.afterTestCount, "AfterTest should not run when BeforeTest skips")
+}
+
+// skipInSetupTestSuite is used to verify that AfterTest is not called when
+// SetupTest skips the test (#1781).
+type skipInSetupTestSuite struct {
+	Suite
```
dolmen, line 861: Wrap SuiteTester also here.

suite/suite_test.go
```diff
@@ -813,3 +813,138 @@ func TestSuiteSignatureValidation(t *testing.T) {
...
+	assert.Equal(t, 0, s.beforeTestCount, "BeforeTest should not run when SetupTest skips")
+	assert.Equal(t, 0, s.testRunCount, "test body should not run when SetupTest skips")
+	assert.Equal(t, 0, s.afterTestCount, "AfterTest should not run when SetupTest skips")
+	assert.Equal(t, 1, s.tearDownTestCount, "TearDownTest should still run when SetupTest skips")
+}
+
+// skipInTestBodySuite is used to verify that AfterTest is still called when
+// the test body itself calls Skip (regression guard for existing behavior).
+type skipInTestBodySuite struct {
+	Suite
```
dolmen, line 916: Wrap SuiteTester also here.

dolmen (changes requested):

PR #1875, opened 2026-04-22
quote string values in empty assertion failures

Bojun-Vvibe:
**Repo:** stretchr/testify (⭐ 23000)
**Type:** bugfix
**Files changed:** 2
**Lines:** +20/-22

## What
This change updates `assert.Empty` and `assert.NotEmpty` to quote string values in failure messages, including aliased string types. The related `assert` package tests were updated so whitespace-only, newline-containing, and empty-string cases now assert against explicit, readable output like `"\\n"` and `""` instead of visually ambiguous raw text.

## Why
Current failure output for string values uses `%v`, which makes empty strings and whitespace-heavy strings hard to distinguish from formatting noise. That weakens assertion diagnostics in exactly the cases where users need clarity. Quoting string values keeps behavior unchanged while making failures immediately understandable and resolves the existing TODO-marked test expectations around these messages.

## Testing
Ran `GOCACHE=$(pwd)/.gocache go test ./assert -run 'Test(Empty|NotEmpty)$'` from the repo root. The targeted tests passed.

## Risk
Low / Message formatting only changes for string values in `Empty` and `NotEmpty`; assertion semantics are unchanged.

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -757,6 +757,15 @@ func isEmpty(object interface{}) bool {
 	return isEmptyValue(reflect.ValueOf(object))
 }
 
+func formatEmptyMessageValue(object interface{}) string {
+	value := reflect.ValueOf(object)
+	if value.IsValid() && value.Kind() == reflect.String {
+		return truncatingFormat("%q", object)
+	}
+
+	return truncatingFormat("%v", object)
+}
+
 // isEmptyValue gets whether the specified reflect.Value is considered empty or not.
 func isEmptyValue(objValue reflect.Value) bool {
 	if objValue.IsZero() {
@@ -794,7 +803,7 @@ func Empty(t TestingT, object interface{}, msgAndArgs ...interface{}) bool {
 		if h, ok := t.(tHelper); ok {
 			h.Helper()
 		}
-		Fail(t, fmt.Sprintf("Should be empty, but was %s", truncatingFormat("%v", object)), msgAndArgs...)
+		Fail(t, fmt.Sprintf("Should be empty, but was %s", formatEmptyMessageValue(object)), msgAndArgs...)
 	}
 
 	return pass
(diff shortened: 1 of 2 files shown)


dolmen (changes requested):
Changes should happen in truncatingFormat.

Bojun-Vvibe:
Thanks for the review @dolmen. To make sure I move this in the right direction — did you mean: (a) have `truncatingFormat` itself detect string values and switch the verb to `%q` automatically (so callers don't need to choose), or (b) keep the per-call-site verb but drop the `formatEmptyMessageValue` helper and just inline a `truncatingFormat("%q", obj)` call in `Empty`/`NotEmpty`? I lean toward (b) since `truncatingFormat` is also called with `%v` and `%#v` in many places where auto-quoting strings would be a behavior change, but happy to do (a) if you prefer the centralization. Will push the update once you confirm.

Bojun-Vvibe:
@dolmen thanks for the review. To make sure I take this in the right direction: `truncatingFormat` is a shared helper used by other assertion paths too (e.g. `Equal`/`NotEqual`-adjacent diagnostics), so making it always emit `%q` for strings would change error-message output for callers beyond `Empty`/`NotEmpty`. Are you suggesting:

1. `truncatingFormat` always quotes strings (changes behavior for all callers), or
2. a new `truncatingFormat`-style helper (e.g. `truncatingFormatValue`) that picks `%q` vs `%v` by kind, and have `Empty`/`NotEmpty` use that, or
3. give `truncatingFormat` a format-string-aware path so callers opt in via the existing `%q` token?

Happy to refactor along any of these — just want to avoid silently changing other assertions' output. Inclined toward (2) since it keeps the existing `truncatingFormat` callsites byte-stable, but defer to your call.

dolmen:
I won't review this shit.

PR #1884, opened 2026-05-07
Modernize Go to 1.18: interface{} -> any

SuperQ:
## Summary
Run the Go modernize tool against the codebase.

## Changes
* Update minimum Go to 1.18.
* Modernize with [`modernize`][0] tool.

## Motivation
Updates syntax to current Go style and use.

## Related issues

[0]: https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,6 @@ jobs:
     strategy:
       matrix:
         go_version:
-          - "1.17"
           - "1.18"
           - "1.19"
           - "1.20"
--- a/README.md
+++ b/README.md
@@ -343,7 +343,7 @@ To update Testify to the latest version, use `go get -u github.com/stretchr/test
 Supported go versions
 ==================
 
-We currently support the most recent major Go versions from 1.19 onward.
+We currently support the most recent major Go versions from 1.18 onward.
 
 ------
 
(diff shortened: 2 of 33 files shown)


.github/workflows/main.yml
```diff
@@ -26,14 +26,15 @@ jobs:
...
-          - "1.17"
           - "1.18"
           - "1.19"
           - "1.20"
           - "1.21"
           - "1.22"
           - "1.23"
           - "1.24"
+          - "1.25"
+          - "1.26"
```
dolmen: We already have `stable` and `oldstable` in the `build` job.

We don't need them here.

dolmen (changes requested):
1. Reviewing the changes submitted by an external contributor are more work than running the modernizer tool myself and checking the changes it made.

2. The change `interface{}` -> `any` is cosmetic, but not a fundamental change that brings significant features to justify upgrading the minimum Go version

SuperQ:
Well, if you don't want external contributions, just say that.

I was planning to contribute more, but if the attitude of the maintainers here is that external contributions are more work than they're worth, I'm happy to just not.

As a maintainer of a large downstream ecosystem (`github.com/prometheus`) that uses testify, I would prefer it be in good working condition.

But it's also just an option for us to stop using testify.

SuperQ:
I was thinking about additional modernization to include new stuff like `range` and the `slices` stdlib package. (1.21.x+)

PR #1883, opened 2026-05-07
CI: upgrade GitHub Actions and pin hashes

SuperQ:
## Summary
It's [recommended][0] to pin GitHub actions to their full SHA version for supply chain security. Dependabot will take care of opening PRs to bump these version hashes.

[0]: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions
## Changes
* Update actions to latest version.
* Pin with the SHA.

PR #1885, opened 2026-05-07
CI: add check of GitHub Action pinned hashes against tag

dolmen:
## Summary
Add `.ci.ghactions.sh` script to validate hashes of GitHub Actions against the declared tag.

## Changes
* add `.ci.ghactions.sh` script.
* use `.ci.ghactions.sh` script from new GH Actions job `check-actions-hashes`.

## Motivation
Supply chain security.

## Related issues
#1883

Cc: @SuperQ

### DIFF
--- a/.ci.ghactions.sh
+++ b/.ci.ghactions.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+
+# MIT License
+#
+# Copyright (c) 2026 Olivier Mengué and contributors.
+# 
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+#
+# Verify that hashes of GitHub actions match the declared tag in attached comment.
+#
+
+set -euo pipefail
+
+declare -A seen
+status=0
+
+for w in .github/workflows/*.yml
+do
+	sed -n -e '/uses: / s!^ *-\{0,1\} uses: \([^@]*\)@\([0-9a-f][0-9a-f]*\) *# *\(v.*\)$!\1 \2 \3!p' "$w" | while read -r action hash tag
+	do
+		if (( ${seen["$action-$hash-$tag"]:-0} )); then
+			printf "\e[1;32m%s: %s@%s == %s\e[m\n" "$w" "$action" "$tag" "$hash"
+			continue
+		fi
+		seen["$action-$hash-$tag"]=1
+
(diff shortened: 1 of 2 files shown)


SuperQ:
If it helps, this is the script that I use to validate tags:

```bash
$ cat gh_fix_action_hash.sh 
#!/usr/bin/env bash
#
# Descriptoin: Pin GitHub Actions to a specific tag.

set -u -o pipefail

if [[ $# -le 1 ]]; then
  echo "usage: $(basename $0) <action> <file...>"
  exit 1
fi

action="$1"
shift 1

repo="$(echo "${action}" | cut -f1-2 -d'/')"

get_latest_tag() {
  local repo="$1"
  gh api "repos/${repo}/releases" \
    | jq -r '.[] | select(.prerelease == false) | .tag_name' \
    | sort -V \
    | tail -n 1
}

latest_tag="$(get_latest_tag "${repo}" 2> /dev/null)"
if [[ "${latest_tag}" == "null" ]] ; then
  echo "ERROR: Unable to get latest release tag"
  exit 1
fi

latest_hash="$(gh api "repos/${repo}/git/ref/tags/${latest_tag}" 2> /dev/null | jq -r '.object.sha')"
if [[ "${latest_hash}" == "null" ]] ; then
  echo "ERROR: Unable to get latest SHA hash"
  exit 1
fi

new_uses="uses: ${action}@${latest_hash} # ${latest_tag}"

echo "Updating: ${new_uses}"
```

It uses the github CLI and jq for parsing. I'm pretty sure those are both available in the github actions standard runtime.

.ci.ghactions.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+# Verify that hashes of GitHub actions match the declared tag in attached comment
+#
+# Author: Olivier Mengué
```
ccoVeille: Why not adding an information about the license here

The sane way you did with #1889

https://github.com/stretchr/testify/blob/dolmen%2Freadme-gofmt-pure-go/_readme-gofmt%2Fmain.go#L4

ccoVeille (dismissed):

.github/workflows/main.yml
```diff
@@ -41,3 +42,9 @@ jobs:
         with:
           go-version: ${{ matrix.go_version }}
       - run: go test -v -race ./...
+
+  check-actions-hashes:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - run: ./.ci.ghactions.sh
```
alexandear, line 48: Why not use [zizmor](https://docs.zizmor.sh/integrations/#github-actions)?

zizmor is a static analysis tool for GitHub Actions. It can find and fix many common security issues in typical GitHub Actions CI/CD setups.


```suggestion
      - uses: zizmorcore/zizmor-action@b1d7e1fb5de872772f31590499237e7cce841e8e # v0.5.3
        with:
          advanced-security: false
          annotations: true
```

See the real world example at https://github.com/golangci/golangci-lint/blob/0fffb1b46b3adbf5fd8e76b8e59a77e3d2724cb9/.github/workflows/pr-checks.yml#L32

.ci.ghactions.sh
dolmen: Full license text added.

dolmen:
@ccoVeille While modifying the script to add the license I noticed a few things to clean (such as shellcheck warnings) which I have fixed, so it seems that another full review is required.

ccoVeille (approved):
Sounds good even the changes

Did you try launching shellcheck on it ?

.github/workflows/main.yml
ccoVeille: @dolmen what are your thoughts here?

It's a good thing to rely on an external dependency that is community reviewed.

Here the code is not tested a future PR could introduce a small change that could break the feature of checking tags

zizmor is tested and maintained.

Also, we could use cooldown feature in dependabot also, trying to mitigate attacks

~~https://github.com/samber/cc-skills-golang/blob/main/skills%2Fgolang-modernize%2FSKILL.md~~

https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-

dolmen: There is a balance between adding external dependencies (which are supply chain risk) and what they bring.

At this point I prefer a simple shell script that anyone can audit (and run on its own dev environment) to trusting another external dependency as a Github Action. Even [made](https://github.com/zizmorcore/zizmor/graphs/contributors?from=07%2F02%2F2026) by a [security researcher](https://yossarian.net/resume.pdf) who now works at OpenAI.

dolmen: This is also the reason why I haven't proposed to enable SonarQube or Codecov like I have on some of my [other](https://github.com/dolmen-go/goeval) [projects](https://github.com/dolmen-go/sqlfunc): they give valuable insight, but also bring supply chain risks. And with 25k ⭐ , Testify is a high profile target.

PR #1858, opened 2026-03-09
chore: use mdsf action and stable schema

alexandear:
## Summary

Use `mdsf` action instead of installing via `npm`.
Use the latest available schema for mdsf (see https://github.com/hougesen/mdsf/issues/1524).

## Changes

- Replace `npm install -g mdsf-cli` with `hougesen/mdsf@v0.12.0`.
- Update `$schema` in `mdsf.json`.

## Motivation

Because of supply chain security concerns.

This allows referencing a stable URL schema without needing to update it every time a new version is released.

## Related issues

#1687

PR #1889, opened 2026-05-11
CI: add _readme-gofmt tool to reformat Go code in README

dolmen:
## Summary

Add `_readme-gofmt` internal tool to replace the use of mdsf, a JS tool.

## Changes
* `_readme-gofmt` applies 'gofmt -s' to each Go block in README.md. The output is a diff, if any change has been detected.
* CI scripts/jobs are updated to drop mdsf and use `_readme-gofmt`

## Motivation
Drop use of mdsf, which is installed via npm. So [many](https://tanstack.com/blog/npm-supply-chain-compromise-postmortem) supply chain attacks via npm these days...

This reduces supply chain concerns with internal tooling in pure Go using only stdlib.

## Related changes
* #1687
* #1858

Cc: @ alexandear


.ci.readme.fmt.sh
alexandear, line 1: Please also remove [.mdsf.json](https://github.com/stretchr/testify/blob/master/.mdsf.json). 

FYI: mdsf is written in Rust, not in JS

_readme-gofmt/main.go
```diff
@@ -0,0 +1,93 @@
+//go:build go1.18
```
alexandear: The `slices` package is available only in Go 1.21+. 

Anyway, I think this build tag is redundant and can be removed.

_readme-gofmt/main.go
```diff
@@ -0,0 +1,93 @@
...
+		cmd.Stderr = os.Stderr
+		err := cmd.Run()
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		outputBytes := bytes.ReplaceAll(output.Bytes(), []byte("\r"), []byte("")) // CRLF -> LF
+		if !bytes.Equal(outputBytes, block) {
+			changes++
+			// fmt.Printf("-----\n%s+++++\n%s-----\n", block, outputBytes)
```
alexandear: Let's remove commented code.

_readme-gofmt/main.go
dolmen: I prefer to make it explicit as we still have go1.17 in go.mod.

On the other hand, when editing we don't get gopls helper. :(

alexandear (approved):

ccoVeille (approved):

_readme-gofmt/main.go
```diff
@@ -0,0 +1,91 @@
+//go:build go1.21
+
+/*
+MIT License
+
+Copyright (c) 2026 Olivier Mengué and contributors.
```
ccoVeille, line 6: Why not SPDX identifiers instead of full text ?

dolmen: Well, the text of the license doesn't speak about SPDX identifiers.

Instead it says:
> The above copyright notice and this permission notice shall be included in all
> copies or substantial portions of the Software.

PR #1828, opened 2025-11-18
fix: panic in spew when dealing with unexported fields

ccoVeille:
## Summary

Fix panic that happens when spew tries to sort unexported fields

## Changes

Fix the panic raised by [Interface()](https://pkg.go.dev/reflect#Value.Interface) on unexported fields

[CanInterface()](https://pkg.go.dev/reflect#Value.CanInterface) is the solution

## Motivation

Now, go-spew is vendored, we can fix old issues we had
## Related issues

- Related to #1826
- Closes #480
- Closes #1813
- Closes #1816
- Supersedes #1824

PR #1894, opened 2026-05-18
assert: don't crash diff() when spew panics on awkward inputs

### LINKED ISSUE #1813: assert.Equal panic on diff unexported map with array as key

c-tonneslan:
From #1813. A struct with an unexported map field whose keys are arrays panics inside `internal/spew/common.go:sortValues`, which calls `.Interface()` on array elements to dedupe map keys. Unexported `reflect.Value`s don't allow `Interface()`, so the assertion takes the test run down instead of reporting the mismatch.

Repro from the issue:

```go
type test struct {
    m map[[1]byte]*struct{}
}
a := test{map[[1]byte]*struct{}{{1}: nil, {2}: nil}}
b := test{}
assert.Equal(t, a, b)
```

The diff string is decoration on the failure message, the assertion itself is what matters. Recovering inside `diff()` and returning empty preserves the failure report and just drops the diff for inputs spew can't safely walk. That keeps the fix contained to assert without touching the spew fork.

Added a regression test that exercises exactly the repro.

Fixes #1813

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1929,7 +1929,7 @@ func typeAndKind(v interface{}) (reflect.Type, reflect.Kind) {
 
 // diff returns a diff of both values as long as both are of the same type and
 // are a struct, map, slice, array or string. Otherwise it returns an empty string.
-func diff(expected interface{}, actual interface{}) string {
+func diff(expected interface{}, actual interface{}) (result string) {
 	if expected == nil || actual == nil {
 		return ""
 	}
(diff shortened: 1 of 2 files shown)


ccoVeille:
I would prefer https://github.com/stretchr/testify/pull/1828 to be merged

c-tonneslan:
Closing in favor of #1828 — that fixes the root cause in spew's sortValues with CanInterface() rather than swallowing the panic in diff(). Thanks @ccoVeille.

PR #1896, opened 2026-05-23
assert.InDeltaSlice: fix lack of length mismatch handling

### LINKED ISSUE #1263: InDeltaSlice doesn't check all expected values present
For example
```
assert.InDeltaSlice(t, []float64{1, 2}, []float64{1}, 0.001)
```
passes because all elements of `actual` have a successful match in `expected`, but the opposite is not tested. Is this intentional?

puneetdixit200:
Fixes #1263.

`InDeltaSlice` already compares values by index, but it did not first verify that the two slices had the same length. That meant extra expected values were ignored, while extra actual values could panic by indexing past the expected slice.

This adds an explicit length check before the element-wise comparison. I also added regression coverage for both mismatch directions.

Verification:
- `go test ./assert -run TestInDeltaSlice -count=1` failed before the fix: the shorter actual slice passed and the longer actual slice panicked with `reflect: slice index out of range`
- `go test ./assert -run TestInDeltaSlice -count=1`
- `go test ./assert -count=1`
- `go test ./... -count=1`
- `go vet ./...`
- `gofmt -w assert/assertions.go assert/assertions_test.go`
- `git diff --check`

### DIFF
--- a/assert/assertions.go
+++ b/assert/assertions.go
@@ -1488,23 +1488,25 @@ func InDeltaSlice(t TestingT, expected, actual interface{}, delta float64, msgAn
 	if h, ok := t.(tHelper); ok {
 		h.Helper()
 	}
-	if expected == nil || actual == nil ||
-		reflect.TypeOf(actual).Kind() != reflect.Slice ||
-		reflect.TypeOf(expected).Kind() != reflect.Slice {
+	if expected == nil || actual == nil {
 		return Fail(t, "Parameters must be slice", msgAndArgs...)
 	}
 
-	actualSlice := reflect.ValueOf(actual)
 	expectedSlice := reflect.ValueOf(expected)
+	actualSlice := reflect.ValueOf(actual)
 
-	if actualSlice.Len() != expectedSlice.Len() {
-		return Fail(t, "Arguments must have the same number of elements", msgAndArgs...)
+	if expectedSlice.Type().Kind() != reflect.Slice {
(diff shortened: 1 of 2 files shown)


assert/assertions.go
```diff
@@ -1497,6 +1497,10 @@ func InDeltaSlice(t TestingT, expected, actual interface{}, delta float64, msgAn
 	actualSlice := reflect.ValueOf(actual)
 	expectedSlice := reflect.ValueOf(expected)
 
+	if actualSlice.Len() != expectedSlice.Len() {
```
dolmen: Use `assert.Len` for the check.

More generally, the code of `InDeltaSlice` should much look like `InEpsilonSlice` by also using `IsType`.

dolmen (changes requested):

puneetdixit200:
Updated in f402eee. `InDeltaSlice` now follows the `InEpsilonSlice` validation pattern more closely: it checks the expected value is a slice, uses `IsType`, uses `Len` for length mismatches, and reports element failures with the index. Added coverage for mismatched slice types. Verified with `go test ./...`.

assert/assertions.go
puneetdixit200: Addressed on the current head `f402eee`: `InDeltaSlice` now matches `InEpsilonSlice` by using `IsType` and `Len`, and the test covers both length mismatch directions plus slice type mismatch. Focused verification: `go test ./assert -run TestInDeltaSlice -count=1`.

PR #1897, opened 2026-05-24
assert.Regexp: document allowed rx argument types for Regexp/NotRegexp

### LINKED ISSUE #1793: assert.Regexp: Document allowed rx argument values #268

SAY-5:
## Summary
Document that the rx argument of Regexp/NotRegexp accepts a *regexp.Regexp or any other value.

## Changes
* Add a note to the assert.Regexp and assert.NotRegexp doc comments explaining that rx may be a *regexp.Regexp (used directly) or any other value (stringified with fmt.Sprint and compiled with regexp.MustCompile, which panics on an invalid pattern).
* Regenerate the require, format and forward variants with `go generate ./...`.

## Motivation
Requested in #1793 by @brackendawson: "The documentation for these functions should be changed to explain that the value should be a *regexp.Regexp, but that if any other type is provided" it is compiled at runtime. The current doc comments give examples for both forms but never state the accepted types or the panic on an invalid expression.

This is a documentation-only change; the behavior is unchanged.

## Related issues
Closes #1793


dolmen (changes requested):
I think that it is a bug that we accept anything that can be stringified: this leaves to much space to misuse of the API. We should only accept `*regexp.Regexp` and strict `string`.

See also https://github.com/Antonboom/testifylint/pull/292

So the last thing I want is to document this as a _feature_.

We must only document that we accept `*regexp.Regexp` and `string` (but unfortunately leave the implementation as is for compatibility).

PR #1898, opened 2026-05-24
assert: include map key in InDeltaMapValues error message

### LINKED ISSUE #1254: Prepend key name to assert.InDeltaMapValues error message
currently the message isn't too helpful, it shows the values that are wrong but does not say which key that value belongs to. Currently test code

```go
assert.InDeltaMapValues(t, map[string]int{"a": 3, "b": 1}, map[string]int{"a": 4, "b": 1}, 0.01)
```
will just produce
```
    t_test.go:7: 
        	Error Trace:	/tmp/a/t_test.go:7
        	Error:      	Max difference between 3 and 4 allowed is 0.01, but difference was -1
        	Test:       	TestT

```
with no indication which value is wrong

I did *try* to write a patch for it and it does *work*

https://github.com/efigence/testify/tree/key-name-map-value

```
--- FAIL: TestT (0.00s)
    t_test.go:7: 
        	Error Trace:	/tmp/a/t_test.go:7
        	Error:      	key["a"]: Max difference between 3 and 4 allowed is 0.01, but difference was -1
        	Test:       	TestT
```

but it is hacky *at best*, I wanted to avoid duplicating `InDelta` but I haven't found a good way to do it

https://github.com/efigence/testify/tree/key-name-map-value

gaurav0107:
## Summary

Closes #1254. `assert.InDeltaMapValues` now prefixes per-key delta failures with `key[<k>]:` so users can immediately see which entry diverged.

## Why

Without the key, a failing `InDeltaMapValues` against a multi-entry map gives no hint about which key broke. Previously:

```
Error: Max difference between 3 and 4 allowed is 0.01, but difference was -1
```

After this change:

```
Error: key[a]: Max difference between 3 and 4 allowed is 0.01, but difference was -1
```

The reporter showed the same before/after on the issue and even prototyped the fix; this PR finishes that work.

## Changes

- `assert/assertions.go`: inline the relevant `InDelta` math inside `InDeltaMapValues` so the failure message can include the map key. The numeric/NaN/delta branches mirror `InDelta` exactly, with the only addition being a `key[<k>]:` prefix.
- `assert/assertions_test.go`: add `TestInDeltaMapValues_ErrorIncludesKey` covering the new message format.
- `assert/assertion_format.go`, `assert/assertion_forward.go`, `require/require.go`, `require/require_forward.go`: regenerated by `go generate ./...` to propagate the updated doc comment.

## Testing

- `go test -v -race ./...` — all packages pass
- `./.ci.gofmt.sh`, `./.ci.govet.sh`, `./.ci.gogenerate.sh` — pass
- New test `TestInDeltaMapValues_ErrorIncludesKey` asserts `key[a]:` appears in the captured failure output.

## Backward compatibility

Behaviour for valid input is unchanged — only the failure-message text is enriched. `True`/`False` semantics are preserved. No tests in the repo pattern-match the previous message text.


assert/assertions.go
```diff
@@ -1537,14 +1539,23 @@ func InDeltaMapValues(t TestingT, expected, actual interface{}, delta float64, m
...
+		}
+		if math.IsNaN(af) {
+			return Fail(t, fmt.Sprintf("key[%v]: Expected must not be NaN", k), msgAndArgs...)
+		}
+		if math.IsNaN(bf) {
+			return Fail(t, fmt.Sprintf("key[%v]: Expected %v with delta %v, but was NaN", k, ev.Interface(), delta), msgAndArgs...)
+		}
+		dt := af - bf
+		if dt < -delta || dt > delta {
+			return Fail(t, fmt.Sprintf("key[%v]: Max difference between %v and %v allowed is %v, but difference was %v", k, ev.Interface(), av.Interface(), delta, dt), msgAndArgs...)
```
HaraldNordgren, line 1558: This seems like a quite a big change for a minor copy fix, is it possible to fix it inside of `InDelta` without re-writing everything, maybe pass the `key[%v]:` prefix with msgAndArgs?

PR #1899, opened 2026-05-25
fix(assert): pass custom message through InEpsilonSlice

### LINKED ISSUE #763: InEpsilonSlice doesn't print the message
`InEpsilonSlice` is implemented as a series of [calls to `InEpsilon`](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1173) (which is not unexpected), but when it calls `InEpsilon`, it does not pass the `msgAndArgs ...interface{}`. As a result, the failures are not very helpful, because the default output of `InEpsilon` only tells the relative error, not the expected & actual values.

It isn't immediately obvious to me how `InEpsilonSlice` should deal with the messages when calling `InEpsilon`-- especially if there are multiple mismatches, would you want to print the same message each time?

wyf027:
## Summary

- `InEpsilonSlice` now forwards `msgAndArgs` to each per-element `InEpsilon` call
- Keeps the existing `"at index %d"` hint and appends caller-provided context
- Adds a regression test asserting custom messages appear in failure output

Fixes #763

## Problem

`InEpsilonSlice` loops over elements and calls `InEpsilon`, but only passed an index hint and omitted `msgAndArgs`. Failures therefore lacked caller-provided context that other assertions include under `Messages:`.

`InDeltaSlice` already forwarded `msgAndArgs`; this aligns `InEpsilonSlice` with that behavior.

## Test plan

- [x] `go test ./assert/ -run TestInEpsilonSlice -v`
- [x] Added `TestInEpsilonSliceMessage` verifying custom message appears in failure output


assert/assertions_test.go
```diff
@@ -2516,6 +2516,19 @@ func TestInEpsilonSlice(t *testing.T) {
...
+	t.Parallel()
+
+	mockT := new(mockTestingT)
+	InEpsilonSlice(mockT,
+		[]float64{2.2, 2.0},
+		[]float64{2.1, 2.1},
+		0.04,
+		"custom context",
+	)
+	Contains(t, mockT.errorString(), "custom context")
```
HaraldNordgren, line 2529: Would it be possible to assert the full error here?

PR #1905, opened 2026-06-03
mock: Mock.Return does not exist anymore

Kentzo:
## Summary
Remove mentions of removed function the docs and comment.

## Changes
Removed references to the Return

## Motivation
I'm a new user and was getting familiar with the API. Noticed that docs for some of the Mock functions mentioned the `Return` function which didn't exist.


brackendawson (approved):
Mock.Return never existed, it has always been Call.Return. Probably the original comment intended to say On().Return().

This is a good improvement and renders fine in pkgsite.

Kentzo:
> Mock.Return never existed

It did in the [initial commit](https://github.com/stretchr/testify/blob/2930d903bf969b4e153bb2f2d05a65455d214949/mock/mock.go#L69-L72) :)

PR #1909, opened 2026-06-11
Add Go fuzz tests for OSS-Fuzz integration

canolgun:
Adds Go native fuzz targets covering critical code paths. Part of OSS-Fuzz integration (google/oss-fuzz#15667). Verified: go test -fuzz=. -fuzztime=30s


canolgun:
Recreating with go1.18 build tag fix + gofmt compliance

PR #1920, opened 2026-07-03
fix: NotNil fails on non-nillable types

### LINKED ISSUE #570: assert: NotNil accepts non-nillable types and doesn't fail

happysnaker:
## Summary

Fixes #570

Currently `NotNil(t, 0)`, `NotNil(t, "")`, `NotNil(t, mystruct{})` all pass without error because `isNil` returns `false` for non-nillable types, making `NotNil` always succeed for values like int, string, and struct.

## Changes

1. Added `isNillableKind` helper that checks if a `reflect.Kind` can hold a nil value
2. Updated `NotNil` to fail when called with a non-nillable type
3. Refactored `isNil` to use the new `isNillableKind` helper

## Before

```go
assert.NotNil(t, 0)     // passes (incorrectly)
assert.NotNil(t, "")   // passes (incorrectly)
assert.NotNil(t, MyStruct{}) // passes (incorrectly)
```

## After

```go
assert.NotNil(t, 0)     // fails: Expected value not to be nil, but got non-nillable type int
assert.NotNil(t, "")   // fails: Expected value not to be nil, but got non-nillable type string
```


happysnaker:
Tests fail because existing tests use NotNil with non-nillable types like true, 0, "true". This is a breaking change that requires updating those tests. Closing for now — will re-approach with a less breaking fix.

PR #1916, opened 2026-06-23
Fix EqualExportedValues on recursive values

cyphercodes:
## Summary
Fix `EqualExportedValues` so copying exported fields terminates for recursive values.

## Changes
* Track visited pointer/slice/map values while copying exported fields.
* Reuse in-progress copies when a recursive value is encountered.
* Add a regression test for self-referential struct pointers with differing unexported fields.

## Motivation
`copyExportedFields` recursed indefinitely on pointer cycles, causing `EqualExportedValues` to time out or overflow the stack for recursive data structures.

Exact test commands run:
* `go test ./assert -run TestEqualExportedValuesRecursiveStruct -count=1 -timeout=3s` (failed before the fix; timed out reproducing the bug)
* `go test ./assert -run 'TestEqualExportedValuesRecursiveStruct|TestEqualExportedValues|TestObjectsExportedFieldsAreEqual|TestCopyExportedFields' -count=1`
* `go test ./assert -count=1`
* `go test ./... -count=1`
* `git diff --check`

## Related issues
Fixes #1915

PR #1922, opened 2026-07-12
fix: detect pointer cycles in copyExportedFields to prevent stack overflow

### LINKED ISSUE #1915: `EqualExportedValues` overflow the stack on recursive data structures
## Description

`assert.EqualExportedValues` recurses through a value's exported fields via `copyExportedFields`. When the value contains a pointer cycle (a struct that, directly or indirectly, points back to itself), `copyExportedFields` follows the cycle forever and never terminates. The result is a `fatal error: stack overflow` that aborts the whole test binary. Tested with testify v1.11.1 and Go 1.26.4.

## Step To Reproduce

Run this test file:

```go
package repro

import (
	"testing"

)

type Node struct {
	Self *Node
}

func TestRecursive(t *testing.T) {
	a := &Node{}
	a.Self = a
	b := &Node{}
	b.Self = b
	assert.EqualExportedValues(t, a, b)
}
```

## Expected behavior

The assertion should detect the cycle and report the two values as equal.

`assert.EqualExportedValues` already compares by structural equality, recursive structures are no different:

<details>

```go
func TestChain(t *testing.T) {
	a := &Node{Self: &Node{Self: &Node{}}}
	b := &Node{Self: &Node{Self: &Node{}}}
	assert.EqualExportedValues(t, a, b)
}
```

</details>

## Actual behavior

`copyExportedFields` recurses on the self-referential pointer until the goroutine stack is exhausted, killing the entire test run:

```
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0x... stack=[...]
fatal error: stack overflow

goroutine 5 [running]:
internal/abi.Name.Name(...)
	.../internal/abi/type.go:659 +0xc0 ...
reflect.(*structType).Field(...)
	.../reflect/type.go:1189 +0x90 ...
reflect.(*rtype).Field(...)
	.../reflect/type.go:777 +0x50 ...
github.com/stretchr/testify/assert.copyExportedFields(...)
	.../testify@v1.11.1/assert/assertions.go:99 +0x42c ...
github.com/stretchr/testify/assert.copyExportedFields(...)
	.../testify@v1.11.1/assert/assertions.go:114 +0x310 ...
github.com/stretchr/testify/assert.copyExportedFields(...)
	.../testify@v1.11.1/assert/assertions.go:106 +0x538 ...
[copyExportedFields repeats until the stack is exhausted]
```

---

I'd be happy to take a stab at addressing this if your willing to review.

tamish560:
## What

`copyExportedFields` now tracks visited pointer addresses to detect and break pointer cycles, preventing stack overflow on recursive data structures.

## Why

`EqualExportedValues` recurses through a value's exported fields via `copyExportedFields`. When the value contains a pointer cycle (a struct that directly or indirectly points back to itself), the function follows the cycle indefinitely and exhausts the goroutine stack, killing the entire test binary.

Closes #1915

## How

1. Extracted the recursion into `copyExportedFieldsWithVisited` which accepts a `map[uintptr]bool` of visited pointer addresses.
2. For pointer-typed values, the pointer address is checked against the visited map. If already seen, `reflect.Zero` of the pointer type is returned, breaking the cycle.
3. `copyExportedFields` calls `copyExportedFieldsWithVisited` with a fresh map.
4. All existing call sites remain unchanged.

The returned zero value for cyclic pointers means both `a` and `b` in a cycle like `a.Self = a` / `b.Self = b` get `Self = nil` in their copies, so the comparison reports them as equal.

## Testing

Added `TestEqualExportedValuesRecursiveStruct` which constructs a self-referential struct and verifies `EqualExportedValues` completes without stack overflow. All existing tests pass unchanged.


tamish560:
Closing this as a duplicate of #1916 by @cyphercodes, which was opened 19 days earlier and addresses the same issue (#1915). Should have checked for existing PRs before opening this one.

PR #1929, opened 2026-07-23
feat: Comparer interface for custom Equal assertions

### LINKED ISSUE #1204: Custom assert equal comparator interface
Is there a way, or a v2 plan to provide more user friendly ways of diffing values for large structs?

Two types I make use of a lot are times and money-friendly decimals, and this combined with needing to compare large structs results in...

image

image

Which is unreadable.

For simple cases I compare the `.String()` output like `a.Equal(want.String(), got.String())` which is a little more typing but it's fine for single values.

The problem is when asserting structs that have many fields. I don't want to waste time pulling out all these fields into tons of assertions just so I can add `.string()` to each item.

Ideally, I'd like to satisfy some comparator interface for our custom money type so diffs and comparisons are done against the rendered value not the underlying data. One other detail of decimal values is different combinations of bases and exponents can result in the same actual number. Assertions fail on this even if the actual resulting number is the same.

```go

type MonetaryValue struct {
	Num decimal.Decimal `json:"num"` // The underlying value: 1234.69
	Cur BaseCurrency    `json:"cur"` // the currency identifier: USD
	Fmt string          `json:"fmt"` // the formatted value: $1,234.69
}

func (v MonetaryValue) GoString() string {
	return fmt.Sprintf(`"%s" (%s, %f)`, v.Fmt, v.Cur, v.Num.InexactFloat64())
}

func (v MonetaryValue) String() string {
	return v.Fmt
}
```

I've already got this, which gets me slightly nicer diffs but the comparison is still conducted against the underlying struct data, which isn't helpful for this particular case.

What would be ideal is a simple test interface:

```go
func (v MonetaryValue) TestCompare(against MonetaryValue) bool {
  return v.Fmt == against.Fmt
}
```

Or maybe a way to surface better diff strings too.

go-cmp _kind of_ solves this but it's not across all tests via an interface, it's just for a single assert - which is cumbersome to add to every test when you have hundreds.

```go
gt.DeepEqual(t, want, got, cmp.Transformer("MonetaryValue", func(in currency.MonetaryValue) string {
		return in.Fmt
}))
```

sonnemusk:
## Summary

Adds `assert.Comparer` so domain types can define equality for `Equal` / `NotEqual` (and anything else using `ObjectsAreEqual`) without `reflect.DeepEqual` field-by-field noise (issue #1204).

```go
type Money struct {
    Cents int
    Label string // display only
}

func (m Money) Equal(other interface{}) bool {
    o, ok := other.(Money)
    return ok && m.Cents == o.Cents
}

assert.Equal(t, Money{100, "$1.00"}, Money{100, "1 dollar"}) // true
```

### Behavior

- If **expected** implements `Comparer`, use `expected.Equal(actual)`
- Else if **actual** implements `Comparer`, use `actual.Equal(expected)`
- Else existing `[]byte` / `reflect.DeepEqual` path

No codegen changes; only `ObjectsAreEqual` grows a short interface check.

Fixes #1204


assert/assertions.go
```diff
@@ -60,11 +60,38 @@ type Comparison func() (success bool)
...
+	// Equal reports whether this value equals other for test purposes.
+	Equal(other interface{}) bool
+}
+
 func ObjectsAreEqual(expected, actual interface{}) bool {
 	if expected == nil || actual == nil {
 		return expected == actual
 	}
 
+	if c, ok := expected.(Comparer); ok {
```
brackendawson, line 88: This approach is not workable for the reasons I mentioned in #1204

This change would break existing tests which compare time.Time instances and expect to assert the time zone too.

brackendawson (changes requested):

sonnemusk:
fair enough — using Equal via an interface would stomp on types that already define Equal with different semantics (time.Time being the obvious one). closing this.

if something like an opt-in registry or an explicit EqualValues helper ever lands for #1204 I'll revisit from that angle.

PR #1934, opened 2026-08-10
assert: compare equivalent YAML numeric values safely

### LINKED ISSUE #1933: assert.YAMLEq fails when comparing equivalent numbers with different formats
### Description
The `assert.YAMLEq` assertion compares two YAML documents by unmarshaling both to `interface{}` and comparing them with `assert.Equal`.

However, the underlying YAML parser (`gopkg.in/yaml.v3`) deserializes integers as `int` (or `int64`) and floating-point numbers as `float64`. This causes `assert.YAMLEq` to fail when comparing conceptually equivalent numbers if their formats differ (such as `80` and `80.0`). Standard `assert.Equal` performs strict type checking and treats `int` and `float64` as different.

This behavior is inconsistent with `assert.JSONEq`. In `JSONEq`, the standard `json.Unmarshal` parses all numbers as `float64` under `interface{}`, which ensures stable and format agnostic number comparison.

### Root Cause Analysis
In `assertions.go` inside `YAMLEq`:

```go
	if err := yaml.Unmarshal([]byte(expected), &expectedYAMLAsInterface); err != nil {
		return Fail(t, fmt.Sprintf("Expected value ('%s') is not valid yaml.\nYAML parsing error: '%s'", expected, err.Error()), msgAndArgs...)
	}

	// ...

	if err := yaml.Unmarshal([]byte(actual), &actualYAMLAsInterface); err != nil {
		return Fail(t, fmt.Sprintf("Input ('%s') needs to be valid yaml.\nYAML error: '%s'", actual, err.Error()), msgAndArgs...)
	}

	return Equal(t, expectedYAMLAsInterface, actualYAMLAsInterface, msgAndArgs...)
```

When comparing `port: 80` (parsed as `int`) with `port: 80.0` (parsed as `float64`), `Equal` performs strict type matching and reports that they are not equal, even though the YAML documents are semantically equivalent.

### Reproduction Code

```go
package main

import (
	"fmt"
)

type mockTestingT struct{}

func (m mockTestingT) Errorf(format string, args ...interface{}) {
	fmt.Printf(format, args...)
	fmt.Println()
}

func main() {
	t := mockTestingT{}
	
	// Compare "port: 80" and "port: 80.0"
	ok := assert.YAMLEq(t, "port: 80", "port: 80.0")
	fmt.Println("YAMLEq result:", ok)
}
```

#### Output:
```text
	Error Trace:	repro_yamleq.go:19
	Error:      	Not equal: 
	            	expected: map[string]interface {}{"port":80}
	            	actual  : map[string]interface {}{"port":80}
	            	
	            	Diff:
	            	--- Expected
	            	+++ Actual
	            	@@ -1,3 +1,3 @@
	            	 (map[string]interface {}) (len=1) {
	            	- (string) (len=4) "port": (int) 80
	            	+ (string) (len=4) "port": (float64) 80
	            	 }

YAMLEq result: false
```

### Proposed Fix
Implement a recursive normalization function for the deserialized YAML objects before calling `Equal`. This function should convert all numeric types (integers, unsigned integers and floats) to `float64`, matching how `JSONEq` compares numbers.

Boulea7:
## Summary

Make `assert.YAMLEq` treat exactly equivalent YAML numbers as equal across the Go types produced by `yaml.v3`, without converting integers to `float64` and losing precision.

## Changes

- Compare decoded `int`, `int64`, `uint64`, and `float64` values through a canonical numeric representation that preserves signed, unsigned, fractional, infinity, NaN, and integer-boundary behavior.
- Recursively compare YAML maps and slices while leaving nonnumeric scalar comparison unchanged.
- Group numeric map keys by canonical numeric value. When YAML decoding preserves multiple numeric Go key types in one equivalence class, compare the associated values as a bounded unordered group.
- Keep nonnumeric map keys on native Go map lookup semantics, including named key types, pointer identity, and comparable struct values.
- Preserve the existing `Equal` fallback so failed assertions retain Testify's normal diff output.
- Add symmetric regression coverage for nested values, numeric map keys, mixed key classes, nonnumeric keys, NaN/infinity/negative zero, signed and unsigned boundaries, float precision boundaries, allocations, and 32-bit behavior.

For a map of width `n`, numeric grouping remains O(n). Unordered matching is limited to the decoded numeric key types (`int`, `int64`, `uint64`, and `float64`), so each equivalence group contains at most four entries.

## Motivation

`YAMLEq` currently reports documents such as `port: 80` and `port: 80.0` as unequal because `yaml.v3` decodes them to different Go numeric types. Normalizing every number to `float64`, as initially suggested, would make common cases pass but would introduce false equality around `int64`/`uint64` and IEEE-754 precision boundaries. This change compares numeric values exactly instead.

## Related issues

Closes #1933

## Validation

- `./.ci.gofmt.sh`
- `env -u GOROOT go test ./assert -run 'TestYAML' -count=20`
- `env -u GOROOT go test ./assert`
- `env -u GOROOT go test ./...`
- `env -u GOROOT go test -race ./...`
- `env -u GOROOT ./.ci.govet.sh`
- `env -u GOROOT golangci-lint run --new-from-rev origin/master` (0 issues)
- `env -u GOROOT GOARCH=386 go test ./assert`
- `env -u GOROOT GOOS=linux GOARCH=arm64 go test -exec /bin/true ./assert` (compile-only)

An early batch that launched several validations concurrently saw intermittent failures in existing timing-sensitive `Eventually` tests. Isolated repeated checks and subsequent serial all-package and race runs passed.

## AI assistance

Codex assisted with implementation, tests, and review. I reviewed the resulting changes and validation results before submission.


brackendawson:
See #1933 for why we aren't accepting this change

Boulea7:
Thanks for the clarification. I understand that YAMLEq intentionally preserves YAML numeric type semantics, so I’ll leave this closed.

PR #1937, opened 2026-08-20
feat: add ErrorAsType assertions for Go 1.27

### LINKED ISSUE #1860: Add require.ErrorAsType for Go 1.26+

mattjohnsonpint:
## Summary

Add generic `ErrorAsType` and `NotErrorAsType` assertions for Go 1.27 consumers.

The underlying `errors.AsType` function arrived in Go 1.26, but Testify could not expose an equivalent API consistently because Go did not yet support generic methods. Go 1.27 removes that restriction, allowing these assertions to be generated as both package functions and methods on `Assertions`.

These assertions return the matched error with its concrete type while preserving Testify's failure diagnostics.

Closes #1860.

## Motivation

Previously, checking and then using a specific error type required a separate target variable:

```go
var syntaxErr *json.SyntaxError
require.ErrorAs(t, err, &syntaxErr)
```

With this change:

```go
syntaxErr := require.ErrorAsType[*json.SyntaxError](t, err)
```

The `assert` variant also returns whether the assertion succeeded:

```go
syntaxErr, ok := assert.ErrorAsType[*json.SyntaxError](t, err)
```

This avoids the target variable and subsequent cast without losing Testify's error-chain diagnostics.

## Added API

For Go 1.27 and later, package `assert` adds:

```go
assert.ErrorAsType[E error](t, err, msgAndArgs...) (E, bool)
assert.ErrorAsTypef[E error](t, err, msg, args...) (E, bool)
assert.NotErrorAsType[E error](t, err, msgAndArgs...) bool
assert.NotErrorAsTypef[E error](t, err, msg, args...) bool
```

The same assertions are available as generic methods:

```go
assert.New(t).ErrorAsType[E](err, msgAndArgs...)
assert.New(t).ErrorAsTypef[E](err, msg, args...)
assert.New(t).NotErrorAsType[E](err, msgAndArgs...)
assert.New(t).NotErrorAsTypef[E](err, msg, args...)
```

Package `require` adds corresponding functions:

```go
require.ErrorAsType[E error](t, err, msgAndArgs...) E
require.ErrorAsTypef[E error](t, err, msg, args...) E
require.NotErrorAsType[E error](t, err, msgAndArgs...)
require.NotErrorAsTypef[E error](t, err, msg, args...)
```

These are also available as methods on `require.Assertions`:

```go
require.New(t).ErrorAsType[E](err, msgAndArgs...)
require.New(t).ErrorAsTypef[E](err, msg, args...)
require.New(t).NotErrorAsType[E](err, msgAndArgs...)
require.New(t).NotErrorAsTypef[E](err, msg, args...)
```

The `require.ErrorAsType` variants return the matched error directly after terminating the test on failure.

## Code Generation

The code generator now supports:

- Generic type parameters and constraints.
- Generic method generation.
- Assertions that return typed values followed by a success `bool`.
- Corresponding `require` functions that return typed values.
- Imports referenced by generic constraints, arguments, and results.
- Version-specific generated output inferred from the source file's build constraint.

This keeps the new functions and methods generated consistently with Testify's existing API rather than introducing handwritten exceptions.

The `_codegen` module now requires Go 1.18 because generic signature inspection uses `go/types` APIs introduced in that release. This does not change Testify's Go 1.17 minimum. `_codegen` is a separate module used only during generation.

## Compatibility

All new API and generated files use:

```go
//go:build go1.27
```

The main module remains compatible with Go 1.17. Existing generated output remains stable when generation is run under Go 1.26.

The CI matrix also adds explicit Go 1.25 coverage. Go 1.26 and Go 1.27 are covered by `oldstable` and `stable`.

## Testing

The tests cover:

- Successful matches and returned typed errors.
- Errors that do not contain the requested type.
- Nil errors.
- Wrapped error chains and failure diagnostics.
- Formatted variants.
- Package functions and generic `Assertions` methods.
- `assert` and `require` behavior.
- Generation under Go 1.26.
- Package compatibility with Go 1.17.

---

## Potential Future Work

The following APIs are suggestions for separate future changes. They are not included in this PR.

The generic code-generation support introduced here could enable a small number of additional assertions whose implementations genuinely rely on generic operations.

### `AsType`

```go
assert.AsType[T any](t, value, ...) (T, bool)
require.AsType[T any](t, value, ...) T
```

This could perform a type assertion against a type parameter:

```go
target, ok := value.(T)
```

It could validate and return either a concrete type or an interface implementation.

### `PanicsAsType`

```go
assert.PanicsAsType[T any](t, f, ...) (T, bool)
require.PanicsAsType[T any](t, f, ...) T
```

This could apply a generic type assertion to the recovered panic value and return the typed payload for further inspection.

### Typed channel assertions

Arbitrary typed channels cannot be converted to `<-chan any`, so assertions that accept and return their element type could also use generics directly:

```go
assert.EventuallyReceives[T any](t, ch <-chan T, waitFor, ...) (T, bool)
require.EventuallyReceives[T any](t, ch <-chan T, waitFor, ...) T

assert.NeverReceives[T any](t, ch <-chan T, waitFor, ...) bool
require.NeverReceives[T any](t, ch <-chan T, waitFor, ...)
```

This PR intentionally remains focused on `ErrorAsType`; these APIs are listed only to show where the new generator support could be useful.

### DIFF
--- a/.ci.ghactions.sh
+++ b/.ci.ghactions.sh
@@ -28,28 +28,40 @@
 
 set -euo pipefail
 
-declare -A seen
+seen=("")
 status=0
 
 for w in .github/workflows/*.yml
 do
-	sed -n -e '/uses: / s!^ *-\{0,1\} uses: \([^@]*\)@\([0-9a-f][0-9a-f]*\) *# *\(v.*\)$!\1 \2 \3!p' "$w" | while read -r action hash tag
+	actions="$(sed -n -e '/uses: / s!^ *-\{0,1\} uses: \([^@]*\)@\([0-9a-f][0-9a-f]*\) *# *\(v.*\)$!\1 \2 \3!p' "$w")"
+	while read -r action hash tag
 	do
-		if (( ${seen["$action-$hash-$tag"]:-0} )); then
-			printf "\e[1;32m%s: %s@%s == %s\e[m\n" "$w" "$action" "$tag" "$hash"
+		if [[ -z "$action" ]]; then
 			continue
 		fi
-		seen["$action-$hash-$tag"]=1
+		key="$action-$hash-$tag"
+		duplicate=0
+		for seen_key in "${seen[@]}"
+		do
+			if [[ "$seen_key" == "$key" ]]; then
+				duplicate=1
+				break
+			fi
+		done
+		if (( duplicate )); then
+			continue
+		fi
+		seen+=("$key")
 
-		if eval "$( curl -s -H "Accept: application/vnd.github+json" \
-			"https://api.github.com/repos/$action/commits/$tag" | jq -r '.sha == "'"$hash"'"' )"
+		if curl --fail --silent --show-error -H "Accept: application/vnd.github+json" \
+			"https://api.github.com/repos/$action/commits/$tag" | jq -e --arg hash "$hash" '.sha == $hash' >/dev/null
 		then
 			printf "\e[1;32m%s: %s@%s == %s\e[m\n" "$w" "$action" "$tag" "$hash"
 		else
 			printf "\e[1;31m%s: %s@%s != %s\e[m\n" "$w" "$action" "$tag" "$hash"
 			status=1
 		fi
-	done
+	done <<< "$actions"
 done
 
 exit $status
--- a/.ci.gofmt.sh
+++ b/.ci.gofmt.sh
@@ -2,7 +2,8 @@
 
 set -euo pipefail
 
-if [ -n "$(gofmt -l .)" ]; then
+unformatted_files="$(gofmt -l .)"
+if [ -n "$unformatted_files" ]; then
   echo "Go code is not formatted:"
   gofmt -d .
   exit 1
@@ -11,7 +12,8 @@ fi
 go run ./_readme-gofmt/main.go
 
 go generate ./...
-if [ -n "$(git status -s -uno)" ]; then
+repository_status="$(git status --short)"
+if [ -n "$repository_status" ]; then
   echo "Go generate output does not match commit."
   echo "Did you forget to run go generate ./... ?"
   exit 1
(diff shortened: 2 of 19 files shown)


mattjohnsonpint:
While reviewing the CI run, I noticed that the Go 1.26 `gofmt` step emitted errors for the Go 1.27 generic methods but still reported success. Build constraints do not prevent `gofmt` from parsing those files, and the script’s conditional command substitution masked its nonzero exit status.

I updated the CI changes to:

- Run formatting validation only with the stable Go toolchain.
- Ensure formatting and generation command failures propagate correctly.
- Detect newly generated, untracked files.
- Fix similar failure-masking behavior in the GitHub Actions hash checker, including errors hidden by a pipeline subshell or failed API request.

The updated checks now fail when their underlying validation fails, rather than producing error annotations within an otherwise successful job.

dolmen:
Please move the CI changes to a separate PR.

mattjohnsonpint:
> Please move the CI changes to a separate PR.

Done in #1941.  I'll rebase this PR after it's merged.  Thanks.

ccoVeille:
The issue I see here, is the fact I feel like it break the logic of

assert.Whatever returns true/false
require.Whatever doesn't return a value.

PR #1938, opened 2026-08-20
mock: do not let a surplus matcher match a missing argument

### LINKED ISSUE #587: mock does not enforce arity if parameter is Anything

### LINKED ISSUE #1211: Mock: nothing does match mock.Anything
Having more arguments in a mock expectation than the interface has does not fail if those expectations match the string `(Missing)`. This is only really confusing if the matcher is fairly generic, like `mock.Anything`.

Perhaps the value used here should be a private type?

Testify: v1.7.4
Go: go version go1.18 darwin/amd64

```go
package kata

import (
	"testing"

)

type MyMock struct {
	mock.Mock
}

func (t *MyMock) Do() {
	t.Called()
}

func TestFoo(t *testing.T) {
	m := &MyMock{}
	m.Test(t)
	m.On("Do", mock.Anything, mock.AnythingOfType("string"), "(Missing)", mock.MatchedBy(func(s string) bool {
		t.Errorf("%q", s)
		return true
	})).Return()
	m.Do()
	m.AssertExpectations(t)
}
```

```
Running tool: /usr/local/go/bin/go test -timeout 30s -coverprofile=/var/folders/06/jjq25lws0cbcxv37ns0hb_mm0000gn/T/vscode-gohOmAs5/go-code-cover github.com/brackendawson/kata -race

--- FAIL: TestFoo (0.00s)
    /Users/alandaws/src/github.com/brackendawson/kata/kata_test.go:21: "(Missing)"
    /Users/alandaws/src/github.com/brackendawson/kata/kata_test.go:21: "(Missing)"
    /Users/alandaws/src/github.com/brackendawson/kata/kata_test.go:25: PASS:	Do(string,mock.AnythingOfTypeArgument,string,mock.argumentMatcher)
FAIL
coverage: 0.0% of statements
FAIL	github.com/brackendawson/kata	0.195s
FAIL
```

dylanpulver:
## Summary

A matcher in a surplus position could match an argument that was never passed.

## Changes

* `Arguments.Diff` used the string `"(Missing)"` to stand in for an absent argument. That is now an unexported `missingArgument` type.
* `Diff` records the length mismatch before it consults any matcher.
* `MatchedBy` is no longer called with the sentinel.
* Printed output is unchanged. `missingArgument.String()` returns `(Missing)`.

## Motivation

With a plain string as the sentinel, `mock.Anything`, `mock.AnythingOfType("string")`, `mock.IsType("")` and the literal `"(Missing)"` all matched an argument that did not exist, so an `On` call listing too many arguments passed instead of failing. `mock.FunctionalOptions()` panicked on it.

## Related issues

Closes #1211
Closes #587

### DIFF
--- a/mock/mock.go
+++ b/mock/mock.go
@@ -952,6 +952,18 @@ func (args Arguments) Is(objects ...interface{}) bool {
 	return true
 }
 
+// missingArgument is the sentinel [Arguments.Diff] uses for an argument that
+// only one side has, either because the call passed fewer arguments than the
+// expectation declares or because it passed more. The type is unexported and
+// has no constructor, so an expectation can never hold a value equal to it.
+type missingArgument struct{}
+
+// String returns the text used for a missing argument in failure messages.
+func (missingArgument) String() string { return "(Missing)" }
+
+// missing is the only value of type [missingArgument].
+var missing missingArgument
+
 // Diff gets a string describing the differences between the arguments
 // and the specified objects.
 //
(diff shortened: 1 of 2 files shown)


brackendawson:
Can *you* write a PR summary using our [template](https://github.com/stretchr/testify/blob/master/.github/pull_request_template.md) please.

brackendawson (approved):
Lovely, thank you. 👍

PR #1941, opened 2026-08-21
CI: keep version coverage and checks reliable

mattjohnsonpint:
As stable and oldstable advance to Go 1.27 and Go 1.26, retain explicit Go 1.25 coverage so every supported release remains tested. Run language-level formatting validation with the stable toolchain because Go 1.26 cannot parse generic methods, even in files excluded by build constraints.

Several CI helpers could previously report success when their underlying commands, generation checks, or action-pin validation failed. Ensure those failures are propagated and newly generated files are detected so a green build means the checks actually completed successfully.

### DIFF
--- a/.ci.gofmt.sh
+++ b/.ci.gofmt.sh
@@ -2,7 +2,8 @@
 
 set -euo pipefail
 
-if [ -n "$(gofmt -l .)" ]; then
+unformatted_files="$(gofmt -l .)"
+if [ -n "$unformatted_files" ]; then
   echo "Go code is not formatted:"
   gofmt -d .
   exit 1
@@ -11,7 +12,8 @@ fi
 go run ./_readme-gofmt/main.go
 
 go generate ./...
-if [ -n "$(git status -s -uno)" ]; then
+repository_status="$(git status --short)"
+if [ -n "$repository_status" ]; then
   echo "Go generate output does not match commit."
   echo "Did you forget to run go generate ./... ?"
   exit 1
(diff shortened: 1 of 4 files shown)


dolmen (changes requested):
Too many unrelated changes in a single commit.

Please split in multiple commits (or even multiple PRs as I don't think that changing the workflow is related to the changes in the scripts). Each commit must explain the intent of the change.

.ci.gofmt.sh
```diff
@@ -2,7 +2,8 @@
 
 set -euo pipefail
 
-if [ -n "$(gofmt -l .)" ]; then
+unformatted_files="$(gofmt -l .)"
+if [ -n "$unformatted_files" ]; then
```
dolmen, line 6: This change is an example of making the review harder than necessary:
- this is not a functional change: dropping that change would have no impact, but the reviewer has to check that
- as this is mixed with plenty of other changes, the reviewer has to check how those changes interact. If that single change was in a single commit that tells the intent, it would be clearer that the change is contained
- the refactor is just to clarify the action done by this test. But just adding a comment before the code (instead of changing the code) would have been easier to review as there would be no question about interaction with other changes

mattjohnsonpint: Got it.  I'll revise, simplify and split up as necessary for clarity.

mattjohnsonpint:
Closing this PR. I'll create a few new ones.  Thanks.

mattjohnsonpint:
Replaced by:
- #1950
- #1951
- #1952
- #1953
- #1954

PR #1942, opened 2026-08-22
Fix: Add ErrorNotContains assertion

### LINKED ISSUE #1940: Add ErrorNotContains assertion
I'd like to propose adding the counterpart of "ErrorContains".

Today we can write:

require.ErrorContains(t, err, "/my/new/path")

instead of:

require.Contains(t, err.Error(), "/my/new/path")

However, there is no equivalent for the negative case, so we still have to write:

require.Error(t, err)
require.NotContains(t, err.Error(), "a/b/c")

Would you consider adding something like "ErrorNotContains" (or "NotErrorContains")?

The idea would be to assert that:

- "err" is not "nil"
- "err.Error()" does not contain the given substring

It would make the API more symmetric with "ErrorContains" and avoid having to call "err.Error()" directly.

It would also allow tools such as testifylint to safely simplify the pattern above without losing the err != nil assertion.

TheRodzz:
Fixes #1940. Adds ErrorNotContains and its corresponding formatted/wrapper variants to complement ErrorContains, along with comprehensive test coverage.


ccoVeille:
I feel like you didn't generate the files, or did it the wrong way.

The CI fails.

Could you please check?

TheRodzz:
@ccoVeille Addressed CI failures

assert/assertion_forward.go
```diff
@@ -311,6 +311,30 @@ func (a *Assertions) ErrorIsf(err error, target error, msg string, args ...inter
...
+		h.Helper()
+	}
+	return ErrorNotContains(a.t, theError, contains, msgAndArgs...)
+}
+
+// ErrorNotContainsf asserts that a function returned a non-nil error (i.e. an
+// error) and that the error does not contain the specified substring.
+//
+//	actualObj, err := SomeFunction()
+//	a.ErrorNotContainsf(err,  expectedErrorSubString, "error message %s", "formatted")
```
ccoVeille, line 330:
```suggestion
//	a.ErrorNotContainsf(err, expectedErrorSubString, "error message %s", "formatted")
```

Extra space here.

Reported here, but valid for everywhere.

assert/assertion_forward.go
```diff
@@ -311,6 +311,30 @@ func (a *Assertions) ErrorIsf(err error, target error, msg string, args ...inter
...
+func (a *Assertions) ErrorNotContainsf(theError error, contains string, msg string, args ...interface{}) bool {
```
ccoVeille, line 331: Use err not theError, to be consistent with other methods and Go specification


```suggestion
func (a *Assertions) ErrorNotContainsf(err error, contains string, msg string, args ...interface{}) bool {
```

Reported here, but valid for everything

ccoVeille (commented):
I'm reportimg things, but I also notice the comment of existing ErrorContains is invalid.

I'm unsure what is the best:
- fix everything including ErrorContains
- fix only NotErrorContains in that PR

What do you think @brackendawson @dolmen

PR #1946, opened 2026-08-27
assert: detect cycles in EqualExportedValues

### LINKED ISSUE #1915: `EqualExportedValues` overflow the stack on recursive data structures

22elix3r:
## Summary

`EqualExportedValues` copies exported fields via `copyExportedFields`, which recursed forever on pointer, slice, and map cycles and aborted the test binary with `fatal error: stack overflow`. This change detects those cycles (same identity keys as `encoding/json`) and reuses the in-progress copy so cyclic equal structures compare equal, matching `reflect.DeepEqual`.

This is a bug fix. Example (the report from #1915):

```go
type Node struct {
    Self *Node
}

func TestRecursive(t *testing.T) {
    a := &Node{}
    a.Self = a
    b := &Node{}
    b.Self = b
    assert.EqualExportedValues(t, a, b) // now succeeds instead of overflowing
}
```

Cyclic structures that differ in exported fields still compare unequal. Cyclic vs non-cyclic values (self-pointer vs nil) also remain unequal, because the copy preserves the cycle instead of zeroing it.

## Related Issue

Fixes #1915

## Changes Made

- Track visited pointer, slice, and map identities in `copyExportedFields` using a visit key in the same shape as `encoding/json` (type + pointer, plus length for slices).
- Check that map once at the start of each recursive call; store the new copy before walking children so a cycle reuses the in-progress value.
- Format `EqualExportedValues` failure messages with spew (which already handles cycles and is used for diffs) instead of `fmt` `%#v`, so unequal cyclic values do not overflow while printing.
- Add regression tests for self-referential structs, mutual pointers, map cycles, slice cycles, arrays of cyclic pointers, finite chains, unequal exported fields, and cyclic vs non-cyclic.

## Testing

- `gofmt` clean
- `go vet ./assert`
- `go test ./assert -count=1 -timeout=30s -run 'TestCopyExportedFields|TestEqualExportedValues|TestObjectsExportedFieldsAreEqual'`
- `go test ./assert -count=1`
- `go test -race ./assert -count=1`
- `go test ./... -count=1`

## Notes

- Zeroing cyclic pointers would make a self-referential value compare equal to one with a nil pointer, which is wrong. Reusing the in-progress copy preserves the cycle so `reflect.DeepEqual` can tell those cases apart.
- Arrays are values and cannot be cyclic by themselves; cycles through arrays are detected via the pointers they hold.
- #1916 is an earlier cycle-tracking patch. A maintainer asked there for map/slice/array tests and a single visit check; this PR includes those and also covers the failure-message overflow (`fmt` `%#v` does not handle cycles). Happy to close this if #1916 is updated instead.


22elix3r:
Ready for review whenever a maintainer is available. I cannot request a reviewer from this account.

22elix3r:
Withdrawing: I missed that this issue already had earlier open PRs. Sorry for the noise — leaving the field to those patches.

22elix3r:
Withdrawing this PR so the earlier open work on the issue can proceed.

PR #1953, opened 2026-09-01
CI: fail incomplete action pin validation

mattjohnsonpint:
## Summary
Ensure the GitHub Actions pin validation check fails when validation cannot be completed successfully.

## Changes
Update `.ci.ghactions.sh`:

- Keep the validation loop in the current shell with Bash `lastpipe`.
- Require successful GitHub API requests with `curl --fail`.
- Use `jq -e` to validate the returned commit SHA without `eval`.

## Motivation
The validation loop previously ran in a pipeline subshell, so setting its failure status did not affect the script’s exit status. Empty or invalid GitHub API responses could also produce an empty `eval` command that succeeded. Together, these issues allowed action pin validation to report success without validating every pin.

### DIFF
--- a/.ci.ghactions.sh
+++ b/.ci.ghactions.sh
@@ -28,6 +28,9 @@
 
 set -euo pipefail
 
+# Keep the pipeline's while loop in this shell so status updates survive.
+shopt -s lastpipe
+
 declare -A seen
 status=0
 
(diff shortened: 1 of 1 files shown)


.ci.ghactions.sh
```diff
@@ -41,8 +44,8 @@ do
 		fi
 		seen["$action-$hash-$tag"]=1
 
-		if eval "$( curl -s -H "Accept: application/vnd.github+json" \
-			"https://api.github.com/repos/$action/commits/$tag" | jq -r '.sha == "'"$hash"'"' )"
+		if curl --fail --silent --show-error -H "Accept: application/vnd.github+json" \
+			"https://api.github.com/repos/$action/commits/$tag" | jq -e --arg hash "$hash" '.sha == $hash' >/dev/null
```
mattjohnsonpint, line 48: An alternative would be to avoid the GitHub API and jq by checking the remote tag with Git directly.

```suggestion
		if git ls-remote --exit-code "https://github.com/$action.git" \
			"refs/tags/$tag" "refs/tags/$tag^{}" |
			grep -q "^$hash[[:space:]]"
```

Let me know which approach you prefer.

dolmen (approved):

PR #1956, opened 2026-09-04
assert: Empty looks through a pointer to an interface

### LINKED ISSUE #1955: assert.Empty regression in v1.11.0: a pointer to an interface is never empty
## Description

`isEmpty` was rewritten in #1761 (`assert: faster and simpler isEmpty using reflect.Value.IsZero`, released in v1.11.0). The rewrite changed behavior for one shape that the old implementation handled: **a pointer to an interface**.

Since v1.11.0, `isEmpty` returns `false` for any non-nil pointer to a non-nil interface, regardless of what the interface holds. So `Empty` can never pass for a `*SomeInterface`, and `NotEmpty` always passes for one, even when the pointee is unambiguously empty.

Root cause, comparing the two implementations:

v1.10.0 (`assert/assertions.go:717`):

```go
case reflect.Ptr:
    if objValue.IsNil() {
        return true
    }
    deref := objValue.Elem().Interface()   // re-boxes through interface{}
    return isEmpty(deref)
```

v1.12.1 (`assert/assertions.go:761`):

```go
func isEmptyValue(objValue reflect.Value) bool {
    if objValue.IsZero() {
        return true
    }
    switch objValue.Kind() {
    case reflect.Chan, reflect.Map, reflect.Slice:
        return objValue.Len() == 0
    case reflect.Ptr:
        return isEmptyValue(objValue.Elem())
    }
    return false
}
```

The old code's `objValue.Elem().Interface()` round-trips through `interface{}`, and `reflect.ValueOf` on the result collapses the interface layer, so recursion continued into the concrete pointee. The new code walks `reflect.Value` directly, so `objValue.Elem()` on a `*I` yields a value with `Kind() == reflect.Interface`. That kind matches no case in the switch, so it falls through to `return false`.

This looks unintentional: #1761 was framed as a performance and simplification change, and the extensive `Empty`-ness cases added just before it in #1753 do not cover a pointer to an interface.

## Step To Reproduce

Self-contained, no dependencies beyond testify:

```go
package repro

import (
	"testing"

)

type Shape interface{ Area() int }

type Square struct{ Side int }

func (s *Square) Area() int { return s.Side * s.Side }

func TestEmpty_PointerToInterface(t *testing.T) {
	var s Shape = &Square{}

	// Unwrapped forms are Empty in every version.
	assert.Empty(t, Square{})  // ok
	assert.Empty(t, &Square{}) // ok
	assert.Empty(t, s)         // ok

	// v1.10.0: passes.  v1.11.0+: fails.
	assert.Empty(t, &s)
}

func TestNotEmpty_PointerToInterface(t *testing.T) {
	var empty any = ""

	// v1.10.0: fails, correctly.  v1.11.0+: passes.
	assert.NotEmpty(t, &empty)
}
```

Under v1.11.0+ the first test fails and the second passes; under v1.10.0 it is exactly the reverse. The same holds for `*any` holding `[]int{}` or `0`.

## Expected behavior

`Empty` should look through the interface layer, the way it looks through a pointer, so that `Empty(&s)` agrees with `Empty(s)`. Concretely, `isEmptyValue` should recurse on `reflect.Interface` the same way it recurses on `reflect.Ptr`:

```go
    case reflect.Ptr, reflect.Interface:
        return isEmptyValue(objValue.Elem())
```

That restores v1.10.0 behavior for this shape and keeps the `IsZero` fast path. (`Elem()` on a nil interface is the zero `Value`, but that case is already caught by the `objValue.IsZero()` check above.)

I applied that one-line change to a local v1.12.1 checkout: the repro above then behaves identically to v1.10.0, and `go test ./...` on testify itself is unchanged (only `TestDirExists` / `TestNoDirExists` fail, and they fail the same way on an unpatched v1.12.1 in my environment, so they are unrelated). Happy to send it as a PR with test cases if you agree with the direction.

Alternatively, if the new behavior is intended, it is worth calling out in the `Empty` doc comment, which currently says only that "pointer values are 'empty' if the pointer is nil or if the pointed value is 'empty'". That sentence reads as though `Empty(&s) == Empty(s)`.

## Actual behavior

```
Error:      	Should be empty, but was 0xc000010030
Test:       	TestEmpty_PointerToInterface
```

and `NotEmpty` on a pointer to an interface holding `""` passes.

## Version

Reproduced on v1.11.0, v1.11.1 and v1.12.1. Last good version is v1.10.0. Go 1.26.3, linux/amd64.

justinhwang:
## Summary
isEmpty was rewritten in #1761 to walk reflect.Value directly instead of re-boxing each dereference through interface{}. That round-trip used to collapse the interface layer, so recursion continued into the concrete pointee. Walking reflect.Value directly, Elem() on a *I yields a value whose Kind is reflect.Interface, which matched no case in the switch and fell through to false.

The result, since v1.11.0, is that isEmpty reports false for every non-nil pointer to a non-nil interface, whatever the interface holds: Empty can never pass for a *SomeInterface, and NotEmpty always passes for one. That was an unintended side effect of a change framed as performance and simplification.

## Changes
Recurse on reflect.Interface the same way we recurse on reflect.Ptr, so that Empty(&i) agrees with Empty(i) again as it did in v1.10.0. Elem() on a nil interface returns the zero Value, but IsZero already catches that above, and the IsZero fast path is otherwise untouched.

## Motivation
Fix regression in v1.11.0+

## Related issues
Fixes #1955


ccoVeille:
Thanks for working on this.

Your logic seems good.

Let's wait for other maintainers feedback

cc @fredbi for its curiosity

justinhwang:
Quick bump on this, hoping to get this landed and a patch cut

justinhwang:
@ccoVeille @colmen @fredbi

PR #1961, opened 2026-09-10
fix(assert): check slice lengths in InDeltaSlice and fix argument ordering

aoright:
In InDeltaSlice, if expected and actual had different lengths:
- If expected had more elements than actual, the loop only iterated up to actualSlice.Len(), returning true even though slices differed.
- If actual had more elements than expected, indexing expectedSlice.Index(i) panicked with reflect: slice index out of range.
Additionally, InDelta was invoked with inverted arguments (actual was passed as expected).

This patch:
1. Adds an explicit length check before element comparison so length mismatches fail cleanly with an informative message.
2. Fixes the argument ordering passed to InDelta (expected then actual).
3. Adds test cases covering mismatched slice lengths.


ccoVeille:
Related to #1263 #1059

- #1263
- #1059

Duplicate of #1896 #1060

- #1896
- #1060

ccoVeille:
Maybe it's time to fix all these 😅

